PR #48, opened 2014-12-05
Added use of Atomics for expired and lastActivityTime

mondain:
The use of Atomics to replace synchronized blocks prevents thread contention and possible deadlocking. Atomics use CAS and are guaranteed to be atomic. In addition this style is also known to speed things up in a lot of instances.
http://www.ibm.com/developerworks/library/j-jtp11234/


bgrozev:
This code has been refactored quite a lot now

PR #55, opened 2015-02-14
Adds comments to the config file that the debian installer generates.

gpolitis:


gpolitis:
hey @turint @damencho. This is an extremely simple patch. Could you please review and merge it, if you think it's OK? Thanks!

PR #74, opened 2015-06-12
add handling in the rest api for header extensions field

brianh5:
REST api is missing a few features compared to the xmpp api, not sure if there's a bigger update in the works but I added support for header extensions here.

### DIFF
--- a/src/org/jitsi/videobridge/rest/JSONDeserializer.java
+++ b/src/org/jitsi/videobridge/rest/JSONDeserializer.java
@@ -520,7 +520,8 @@ final class JSONDeserializer
 
     public static RTPHdrExtPacketExtension deserializeHeaderExtension(
             JSONObject headerExtension,
-            ColibriConferenceIQ.Channel channelIQ) {
+            ColibriConferenceIQ.Channel channelIQ)
+    {
         RTPHdrExtPacketExtension headerExtensionIQ;
         if (headerExtension == null)
         {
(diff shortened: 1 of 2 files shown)


brianh5:
Oops, realized this included more than I had planned.  I'll try and get it cleaned up.

bgrozev:
Hi Brian,

Thank you for your contribution. The RTP header extensions support for REST is something that we'd like to merge. Before we can do that we need you to sign our contributor agreement as either an [individual](https://jitsi.org/icla) or a [corporation](https://jitsi.org/ccla).
Also, there are a couple of minor changes necessary for the code, so that it follows our style guide: opening braces should be on new lines, a Logger instance should be used instead of System.err for logging errors.

Thanks again,
Boris

brianh5:
Cool! Thanks Boris.  I'll work on getting this cleaned up.

brianh5:
Hey Boris, still figuring out the contract piece, but I've tried to make the appropriate style changes here to be consistent with the rest of the file.

brianh5:
@bgrozev just got the contributor agreement signed and submitted

brianh5:
@bgrozev let me know if there's anything else you need from me on this.

bgrozev:
Sorry, Brian, it's been a busy couple of days. I'll review and merge when I get a chance.

jitsi-developers:
Sure, no problem!  Wanted to make sure everything with the contributor
agreement was right.

markreg:
@brianh5 are you able to specify rtcp-fb: ccm fir, nack pli, and goog-remb via the rest api?

brianh5:
@markreg yes, we're able to set that.  i do notice, however, that when you access the conference state via the rest api, that information will not show up.  it looks like RtpChannel::describe (which serializes the bridge data into json to respond to rest requests) only does the ssrc--so if you're worried your rtcp-fb information isn't showing up in rest, i think that's why.

bgrozev:
Videobridge currently doesn't use any of the rtcp-fb fields. It doesn't even store the information, which is why it doesn't show up in replies (either XMPP or REST)

markreg:
So, is it not used because its features are implicit in Videobridge or are the rtcp-fb features seen as unnecessary?

markreg:
Would you accept a pull request with rtcp-fb support?

bgrozev:
Do you mean rtcp-fb support in COLIBRI and the REST interface, or somehow handling the specific rtcp-fb extensions set on a channel? I think we would accept the first, since we can use it to detect FIR, NACK, REMB support for endpoints.
Do you have a specific use-case where you need this?

PR #80, opened 2015-08-13
upgrade influxdb to 0.9.2

joshuawarner32:


src/main/java/org/jitsi/videobridge/influxdb/LoggingHandler.java
```diff
@@ -86,27 +86,27 @@
             "conference_id",
             "endpoint_id",
 
-            "stats.local_ip",
-            "stats.local_port",
-            "stats.remote_ip",
-            "stats.remote_port",
+            "stats_local_ip",
```
joshuawarner32: 0.9.2 has a bug where it doesn't accept column names that start with reserved words. (https://github.com/influxdb/influxdb/issues/3547)

src/main/java/org/jitsi/videobridge/influxdb/LoggingHandler.java
```diff
@@ -306,71 +309,59 @@ public LoggingHandler(ConfigurationService cfg)
...
-        if (values[0] instanceof Object[])
-        {
-            multipoint = true;
-            pointCount = values.length;
-        }
-
-        if (multipoint)
-        {
-            for (int i = 0; i < pointCount; i++)
+        /* The following is a sample JSON message in the format used by InfluxDB v0.9.0
```
joshuawarner32: Technically, this JSON format is deprecated as of 0.9.2, so next time we upgrade, we'll likely have to switch to their new custom text-based format for data submissions.

PR #81, opened 2015-08-20
Fix: race condition between addChannel() and close() when assigning SctpConnection

innerverse:
This request fixes a race condition caused by close and addChannel mutating the reference to SctpConnection and channelForDtls.  This can result in a stack trace causing a cast exception.  The result is the participant in the conference receives no media.

The fix uses an AtomicReference to keep track of the SctpConnection reference as a guard;  there was still a race condition with with ChannelForDtls, however.  Since the flow was a bit more complex, I introduced a simple synchronized block with a lock.  I looked for any chances of dead-locks by carefully analyzing any locks obtained within the new synchronize block and didn't find any.

Although this uses a couple of different concurrent solutions, I think the compareAndSet approach allows us to have the smallest synchronized() block that we can and be lock free in the guard case.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
+++ b/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
@@ -19,6 +19,7 @@ import java.beans.*;
 import java.io.*;
 import java.net.*;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


emcho:
was this closed by accident?

PR #82, opened 2015-08-21
Fix: race condition between addChannel() and close() when assigning sctpConnection AND channelForDtls

innerverse:
This request fixes a race condition caused by close and addChannel mutating the reference to SctpConnection and channelForDtls. This can result in a stack trace causing a cast exception. The result is the participant in the conference receives no media.

The fix is a simple synchronize block, since there is a tight connection between the value of the sctpConnection and channelForDtls and two areas where these values can be mutated - addChannel and close. I looked for any chances of dead-locks by carefully analyzing any locks obtained within the new synchronize block and didn't find any;  I also ensured all areas that read and write these variables are done while the lock is held, which turned out to be the same areas, allowing for this simpler approach.

Load testing this fix reveals no race conditions are occurring now.  The previous pull request made the situation much better, but didn't cover all cases.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
+++ b/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
@@ -189,6 +189,11 @@ public class IceUdpTransportManager
     private Channel channelForDtls = null;
 
     /**
+     *  Used to synchronize access to {@link #channelForDtls}.
+     */
+    private final Object channelForDtlsLock = new Object();
+
+    /**
      * Whether this <tt>TransportManager</tt> has been closed.
      */
     private boolean closed = false;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -391,65 +396,67 @@ public boolean addChannel(Channel channel)
...
+            if (channel instanceof SctpConnection
+                    && sctpConnection != null
+                    && sctpConnection != channel)
             {
-                /*
-                 * channelForDtls is necessarily an RtpChannel, because we don't
-                 * add more than one SctpConnection. The SctpConnection socket
-                 * will automatically accept DTLS.
-                 */
-                RtpChannel rtpChannelForDtls = (RtpChannel) channelForDtls;
```
innerverse: This can still cause a cast exception.  I can submit another pull request with the fix.  The comment is misleading as this can change types during a close.

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -391,65 +396,67 @@ public boolean addChannel(Channel channel)
...
-        {
-            logd("Not adding a second SctpConnection to TransportManager.");
-            return false;
-        }
-
-        if (!super.addChannel(channel))
-            return false;
-
-        if (channel instanceof SctpConnection)
+        synchronized (channelForDtlsLock)
```
innerverse: The diff here is actually pretty small - the only thing that changes inside the synchronize block in this commit is whitespace.

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -682,42 +689,45 @@ public boolean close(Channel channel)
 
         if (removed)
         {
-            if (channel == sctpConnection)
-            {
-                sctpConnection = null;
-            }
-
-            if (channel == channelForDtls)
+            synchronized (channelForDtlsLock)
```
innerverse: Again - the diff here looks bigger on GitHub's tool than it may appear - this is mainly whitespace due to being inside a synchronized block.

bgrozev:
Thank you for the contribution! This look reasonable to me, but I haven't looked at it in detail. I don't know when someone will be able to review it, but before we can merge it we would need you to sign our contributor agreement (for an [individual](https://jitsi.org/icla) or a [corporation](https://jitsi.org/ccla))

innerverse:
Corporate CLA signed;  I added a commit to check the data type as well.

bgrozev:
Closing as it's a stale request for v1.0

PR #83, opened 2015-08-22
Pass java flags in from init script

spditner:
Allows setting java runtime values such as sip-communicator flags by setting
JAVA_FLAGS in /etc/jitsi/videobridge/config, rather supplying them in
/usr/share/jitsi-videobridge/.sip-communicator/sip-commnicator.properties

For example, enabling statistics in /etc/jitsi/videobridge/config becomes:

   JAVA_FLAGS="-Dorg.jitsi.videobridge.ENABLE_STATISTICS=true"

### DIFF
--- a/resources/install/debian/init.d
+++ b/resources/install/debian/init.d
@@ -67,7 +67,7 @@ start() {
         exit 1
     fi
     echo -n "Starting $DESC: "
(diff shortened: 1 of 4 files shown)


damencho:
A similar property was added JAVA_SYS_PROPS and is part of commit 9de9773, moving config files in /etc/.

PR #85, opened 2015-08-27
fix(Close Stream Race): Race condition closing a stream resulting in NullPointer

innerverse:
This fixes the NPE when two threads attempt to close the channel at the same time:

```
java.lang.NullPointerException
    at org.jitsi.videobridge.RtpChannel.closeStream(RtpChannel.java:599)
    at org.jitsi.videobridge.Channel.expire(Channel.java:339)
    at org.jitsi.videobridge.VideobridgeExpireThread.expire(VideobridgeExpireThread.java:141)
    at org.jitsi.videobridge.VideobridgeExpireThread.run(VideobridgeExpireThread.java:216)
```

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -174,7 +174,7 @@ public class RtpChannel
     /**
      * Whether {@link #stream} has been closed.
      */
-    private boolean streamClosed = false;
+    private final AtomicBoolean streamClosed = new AtomicBoolean(false);
 
     /**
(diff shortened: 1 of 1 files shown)


innerverse:
Corporate CLA signed;  this account should be a listed account now.

Stefan1oo:
+1

bgrozev:
ACK on the CLA. And thank you for the contribution!

The patch looks good to me, but I am trying to understand under what conditions the problem occurs. Unless I am missing something RtpChannel#closeStream is only called from Channel#expire, which makes sure it executes only once.

PR #91, opened 2015-10-09
Add a generic 'endpoint' colibri message that allows clients to commu…

brianh5:
…nicate an arbitrary payload to one another/an entire conference.

I thought I'd run this idea by the jitsi team.  I wanted to add a generic colibri message that applications can use to communicate application-specific information over the data channel.  Our use cases centered around endpoints being able to send information in either '1:1' (endpoint sends a message specifically to another endpoint) or 'broadcast' (endpoint sends a message to everyone else in the conference) modes.  This way applications can send their own messages over the existing datachannel when connecting via the bridge.

The way I've got it formatted now is this:
message {
  colibriClass: "EndpointMessage",
  to: "broadcast" | <user_id [the id used by the bridge]>,
  msgPayload: <arbitrary message payload, jitsi doesn't care, up to application developer>
}
but I'm open to suggestions, this was just what I threw together to match our use case.  Let me know if this is a feature that you think would make sense to be included upstream.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -227,8 +227,6 @@ public class Conference
      *  see broadcastMessageOnDataChannels below
      */
     public void sendMessageOnDataChannels(String msg, List<Endpoint> endpoints) {
-        System.out.println("broadcasting message to " + endpoints.size() + " endpoints");
-        System.out.println(msg);
         for (Endpoint endpoint : endpoints) {
             try
             {
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -221,6 +221,28 @@ public class Conference
     }
 
     /**
+     * Used to send a message to a subset of endpoints in the call, primary use case
+     *  being a message that has originated from an endpoint (as opposed to a message
+     *  originating from the bridge and being sent to all endpoints in the call, for that
+     *  see broadcastMessageOnDataChannels below
+     */
+    public void sendMessageOnDataChannels(String msg, List<Endpoint> endpoints) {
+        System.out.println("broadcasting message to " + endpoints.size() + " endpoints");
+        System.out.println(msg);
+        for (Endpoint endpoint : endpoints) {
+            try
+            {
+                endpoint.sendMessageOnDataChannel(msg);
+            }
+            catch (IOException e)
+            {
+                logger.error("Failed to send message on data channel.", e);
+            }
+        }
+    }
+
+
+    /**
      * Broadcasts string message to al participants over default data channel.
      *
      * @param msg the message to be advertised across conference peers.
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -247,17 +247,7 @@ public class Conference
      */
     private void broadcastMessageOnDataChannels(String msg)
     {
-        for (Endpoint endpoint : getEndpoints())
-        {
-            try
-            {
-                endpoint.sendMessageOnDataChannel(msg);
-            }
-            catch (IOException e)
-            {
-                logger.error("Failed to send message on data channel.", e);
-            }
-        }
+        sendMessageOnDataChannels(msg, getEndpoints());
     }
 
     /**
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -437,27 +437,34 @@ public class Endpoint
     }
 
     /**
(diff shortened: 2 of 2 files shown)


bgrozev:
Sending messages between endpoints over the data channel can have many different use-cases, so I think defining a generic mechanism as you have done is a very good idea. I think we should include it.

In the format for a broadcast, I would rather use 'to: ""', or use a separate colibriClass (e.g. BroadcastMessage), so that "broadcast" is free for use as an endpoint ID.

fippo:
the right thing to do (from a protocol perspective) would not be to put everything on the same datachannel and reinvent the whole (de)multiplexing done by sctp and webrtc datachannels.

brianh5:
thanks guys.

@fippo sounds interesting...can you elaborate on what you mean?

fippo:
@brianh5 the webrtc peerconnection supports a (nearly) arbitrary number of different datachannels running over it. Using different datachannels would prevent issues like filetransfers blocking active speaker notifications.

brianh5:
Ah, I understand now.  The intent for this message wrapper was for messages that are very similar to the events Jitsi already passes over this datachannel (normal signaling-type messages, not things like file transfer), but that are application specific; so to me it felt like re-using the existing connection made sense.  I do totally agree for things of a different nature (like file transfer) would be best handled with a separate one.

fippo:
the thing about being application specific is that you need to demultiplex them somehow. Which would be  best done by the datachannel's protocol property. That way, you've separated your application concerns from the jvb's concerns.

brianh5:
The way I saw it, demuxing these messages would be no different than the ones already sent by the bridge.  Each type (dominantlastspeaker, lastn, etc.) needs to be tested for and handled with a specific method, I saw the things going into the application message being handled the same way.  Or, detected as an application method, and a handler is invoked that the application can register for to handle.  To me it didn't feel like that would be muddying the waters too much.  I had viewed this more as a generic wrapper to support messages over an existing channel.

Admittedly, the idea of spinning up a new data channel didn't occur to me.  I don't know if there are any other issues there (e.g. any overhead, any testing/verification the bridge may need with regards to setting up multiple data channels, etc.).  It seems like the jitsi code (referring to the jitsi javascript library, which the application uses) would have to at least give the hooks for creating this second data channel and allow the application to register a handler for it.

Not sure how things like this are typically settled, I'm happy to defer to the Jitsi team if they think this isn't what's desired upstream.

paweldomas:
Also I agree with Boris that we should either use an empty endpoint "" instead of "broadcast".

brianh5:
@bgrozev @paweldomas 
So I've just made a few tweaks:
1) Cleaned up the system.out.println logging
2) Used empty string to denote 'broadcast' rather than a hard-coded string

I also made an attempt at documenting the behavior and some of the caveats we talked about (in onEndpointMessage)--let me know if you guys think that strikes the right tone.

If it looks good, let me know and I can squash the commits before you merge it in.

bgrozev:
@brianh5 LGTM
Two minor things: by our code convention, all lines should be wrapped at 80 chars. To avoid code duplication, the existing broadcastMessageOnDataChannels should be changed to use sendMessageOnDataChannels.

brianh5:
@bgrozev thanks boris, made those tweaks.

xdumaine:
I'd like to comment on this too, since it's (fairly recent):

> The way I saw it, demuxing these messages would be no different than the ones already sent by the bridge. Each type (dominantlastspeaker, lastn, etc.) needs to be tested for and handled with a specific method, I saw the things going into the application message being handled the same way. Or, detected as an application method, and a handler is invoked that the application can register for to handle. To me it didn't feel like that would be muddying the waters too much. I had viewed this more as a generic wrapper to support messages over an existing channel.

I agree with Fippo. The argument you're making here is that because the bridge is already using a single channel for multiple things, it justifies continuing to do so. In line with Fippo's suggestions, the `dominantEndpoint` and `lastN` messages should be on separate channels too. 

I find it quite confusing that I'm getting arbitrary messages, `dominantEndpoint` messages, and `lastN` messages, all over a single channel called, `lastN` - they should each have a discrete channel.

That said, I do appreciate you doing this, as I was able to make use of it today.

Additionally, I'd like to see more `docs` writeups for features like this. I can help contribute to those missing as well, but new features should always come with documentation.

brianh5:
I'm not sure a unique data channel per message type feels like the way to go.  To me it seems like having them at a higher level based on the context and category of traffic makes more sense.  I don't think we need unique channels for `dominantEndpoint` and `lastN`--not only do these messages have a heavy contextual overlap, but their transmission characteristics (frequency, size, etc.) are similar.  Now, should we introduce something like file transfer, which has not only a different context but very different transmission characteristics that could cause issues for other messages, that seems like something deserving of its own channel.

I didn't actually know the current channel was named `lastN`, but it just seems like a naming problem to me.

Just my thoughts.  I definitely agree it would be nice to have support for easily spinning up new data channels as needed.

PR #98, opened 2015-11-23
flesh out long-polling support in the rest api and add a endpoint exp…

brianh5:
…iration message.  NEEDS CLEANUP.

There's some ugly, temporary hacks here; I'd consider this basically a straw man implementation for the purpose of review.  Hopefully the basic ideas are sound, but I figured I'd just get this review going in the background so the team can take a look and we can get it cleaned up and committed upstream.  Mainly, need input on the mechanism by which the core code will push events to the long-polling channel--ideally this would be the same as how it's done with xmpp.

@lyubomir

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,11 @@
       <artifactId>jetty-rewrite</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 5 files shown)


brianh5:
did this on the wrong branch by accident, opened a new one here: https://github.com/jitsi/jitsi-videobridge/pull/106

PR #100, opened 2015-11-23
Use exec in wrapper scripts for launching java

leedm777:
When launching Java from a wrapper script, it's usually better to use
exec, so that the extra bash process isn't hanging around. It also helps
shutdown and process monitoring (like with upstart, systemd, docker,
etc.) since they would be monitoring (and killing) the Java process
itself, instead of the wrapper script.

### DIFF
--- a/resources/collect-dump-logs.sh
+++ b/resources/collect-dump-logs.sh
@@ -5,14 +5,13 @@
 
 PID=$(cat /var/run/jitsi-videobridge.pid)
 if [ $PID ]; then
-    PROC_PID=$(pgrep -P $PID)
-    echo "Jvb at pid $PROC_PID"
+    echo "Jvb at pid $PID"
     STAMP=`date +%Y-%m-%d-%H%M`
-    THREADS_FILE="/tmp/stack-${STAMP}-${PROC_PID}.threads"
-    HEAP_FILE="/tmp/heap-${STAMP}-${PROC_PID}.bin"
-    sudo -u jvb jstack ${PROC_PID} > ${THREADS_FILE}
-    sudo -u jvb jmap -dump:live,format=b,file=${HEAP_FILE} ${PROC_PID}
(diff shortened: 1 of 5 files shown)


bgrozev:
I think that's a nice fix, but we need to make sure that it doesn't brake anything. [This script](https://github.com/jitsi/jitsi-videobridge/blob/master/resources/collect-dump-logs.sh) will need an update. @damencho WDYT?

@leedm777 Before we merge any of your contributions we need you to sign our contributor agreement ([individual](https://jitsi.org/icla) or [corporate](https://jitsi.org/ccla)). Apologies if you've already done this.

leedm777:
@bgrozev I'm not sure what needs to change in `collect-dump-logs.sh`. Can you be more specific?

I think my company (Digium) already has a corporate CLA on file. Can you check to see if that would cover my contribution?

bgrozev:
You do need to sign the CLA again (as the currently signed one doesn't include your name). 

For the script, we need to remove PROC_PID and just use PID instead. @damencho just pointed out that we will need to do the same in the [init script](https://github.com/leedm777/jitsi-videobridge/blob/master/resources/install/debian/init.d)

leedm777:
@bgrozev I've fixed the init and collect dump logs scripts.

I'm a little unsure about the changes to the init script; especially the status option. Do you have a setup you can test that on?

bgrozev:
@leedm777 Thank you! We'll get this tested when we get a chance.

leedm777:
BTW, I've submitted the signed CLA.

leedm777:
bump.

bgrozev:
Will test and merge this or next week. Thanks!

bgrozev:
Merged. Also applied the same to jicofo: https://github.com/jitsi/jicofo/commit/57563e815b60cc5a5e7ca6091d7ffac332c82e77

leedm777:
@bgrozev Very cool. Thanks!

PR #106, opened 2015-12-03
flesh out long-polling support in the rest api and add a endpoint exp…

brianh5:
…iration message.  NEEDS CLEANUP.

There's some ugly, temporary hacks here; I'd consider this basically a straw man implementation for the purpose of review. Hopefully the basic ideas are sound, but I figured I'd just get this review going in the background so the team can take a look and we can get it cleaned up and committed upstream. Mainly, need input on the mechanism by which the core code will push events to the long-polling channel--ideally this would be the same as how it's done with xmpp.

@lyubomir

NOTE: re-opened this as I accidentally did the previous one from the wrong branch.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -21,6 +21,7 @@ import java.lang.ref.*;
 import java.lang.reflect.*;
 import java.text.*;
 import java.util.*;
+import java.util.concurrent.*;
 
 import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.*;
 import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.ColibriConferenceIQ.Recording.*;
@@ -158,6 +159,14 @@ public class Conference
      */
     private final Map<String, IceUdpTransportManager> transportManagers
         = new HashMap<>();
+
+    /**
+     * Queue to which any push events are added.  These events will be pulled by another
+     *  object who will send them out.
+     * NOTE: this sucks.  this should use whatever mechanism is used to push events out
+     *  the xmpp channel.
+     */
+    private final BlockingQueue<JSONObject> pushEventQueue = new LinkedBlockingQueue();
 
     /**
      * The <tt>Videobridge</tt> which has initialized this <tt>Conference</tt>.
@@ -1677,5 +1686,23 @@ public class Conference
     public String getName()
     {
         return name;
+    }
+
+    public void addPushEvent(JSONObject event)
+    {
+      System.out.println("CONF ADDING PUSH EVENT");
+      try {
+        pushEventQueue.put(event);
+      } 
+      catch (InterruptedException e)
+      {
+        logger.error("Conference " + getID() + " failed to add push event to queue: " + event.toString());
+      }
+    }
+
+    public JSONObject getPushEvent() throws InterruptedException
+    {
+      System.out.println("CONF GETTING PUSH EVENT");
+      return pushEventQueue.take();
     }
 }
--- a/src/main/java/org/jitsi/videobridge/rest/LongPollingServlet.java
+++ b/src/main/java/org/jitsi/videobridge/rest/LongPollingServlet.java
@@ -19,6 +19,19 @@ import java.io.*;
 import javax.servlet.*;
 import javax.servlet.http.*;
 
+import org.json.simple.*;
+import org.osgi.framework.*;
+
+import org.jitsi.videobridge.*;
+import org.jitsi.osgi.*;
+import org.eclipse.jetty.continuation.*;
+
+
+//TEMP
+import java.util.concurrent.*;
+import java.util.*;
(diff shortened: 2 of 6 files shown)


src/main/java/org/jitsi/videobridge/rest/LongPollingServlet.java
```diff
@@ -38,7 +59,55 @@ protected void doGet(
         throws IOException,
                ServletException
     {
-        // TODO Auto-generated method stub
-        response.setStatus(HttpServletResponse.SC_NOT_FOUND);
+        System.out.println("BB: IN LONG POLLING");
+        Continuation continuation = ContinuationSupport.getContinuation(request);
```
lyubomir: The last bit of information on the subject of asynchronous processing of HTTP requests that we exchanged via e-mail was:

We’ll use asynchronous processing of HTTP requests defined by the Servlet 3.0 specification. Please refer to https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#startAsync(javax.servlet.ServletRequest, javax.servlet.ServletResponse).

I know I mentioned Jetty Continuations in person prior to that e-mail exchange when we were still in the support investigation phase. I'd like us to go with the Servlet 3.0 API because we have it already in our dependency list (while Jetty Continuations is yet another dependency) and is equally easy to work with.

src/main/java/org/jitsi/videobridge/rest/LongPollingServlet.java
```diff
@@ -38,7 +59,55 @@ protected void doGet(
...
+        
+        // We don't get the target here (TODO: is that something that could
+        //  be fixed?), so we need to do a bit more work to extract the bits of the path
+        //  we're interested in.
+        // The stuff we're interested in starts after the /colibri/
+        String colibriPath = request.getRequestURL().toString().split("colibri/")[1];
+        String target = colibriPath.split("/")[0];
+
+        continuation.suspend();
```
lyubomir: Starting an asynchronous processing of the HTTP request here is too early because you don't even know whether the target is supported and is to be asynchronously processed.

lyubomir: Once you've identified a target/HTTP request for asynchronous processing and started it i.e. suspended the request, you are to register the continuation (either Jetty Continuations or Servlet 3.0 API) with an asynchronous service which is to resume the continuation when the waited-for event occurs. In other words, the code bellow should return the control to the calling method as soon as possible and without blocking.

src/main/java/org/jitsi/videobridge/rest/LongPollingServlet.java
```diff
@@ -38,7 +59,55 @@ protected void doGet(
...
+          {
+            System.out.println("Couldn't find conference " + confId);
+          }
+          else
+          {
+            System.out.println("found conf object");
+            System.out.println("Waiting for event");
+            try
+            {
+              JSONObject pushEvent = conference.getPushEvent();
```
lyubomir: That's a blocking method call and is not the way to go with asynchronous processing (as commented above). What you want to do is suspend the request, register the suspension/continuation with another entity which monitors the blocking queue and resumes the continuation when at least one element is queued. When the request is resumed, read from the queue without blocking.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1673,4 +1682,22 @@ public String getName()
     {
         return name;
     }
+
+    public void addPushEvent(JSONObject event)
+    {
+      System.out.println("CONF ADDING PUSH EVENT");
+      try {
+        pushEventQueue.put(event);
```
lyubomir: Limiting the number of events stored in the queue is probably one of the issues that we'll want to look into sooner rather than later; otherwise, we're opening ourselves to running out of memory. We could do a maximum number of events, timing events out, etc.

brianh5: yeah i agree...a max size that returns some value on failure (so the caller can potentially try again) seems good.  but at a higher level, this queue (especially in this location) was really just a hack.  doesn't the conference already have a mechanism for pushing events out on the xmpp channel?  i'd think there should be some code we can re-use from there

src/main/java/org/jitsi/videobridge/rest/LongPollingServlet.java
brianh5: i think that i did look into the async context stuff, but ended up going down the continuation route--i don't remember what the reason was.  but i'll work on moving it over to the async servlet stuff.

src/main/java/org/jitsi/videobridge/rest/LongPollingServlet.java
brianh5: yep, sounds good.  again this was more just shortest-path to get things working.

brianh5:
@lyubomir finally got some time to work on this.  still need some cleanup, but let me know if this is what you had in mind.

PR #119, opened 2016-01-04
Removes an unnecessary lock acquisition in order to prevent a deadlock

bgrozev:
reported by Swapnil Bagadia (and reduce the risk of deadlocks in
general).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/TransportManager.java
+++ b/src/main/java/org/jitsi/videobridge/TransportManager.java
@@ -244,10 +244,7 @@ public abstract class TransportManager
      */
     protected List<Channel> getChannels()
     {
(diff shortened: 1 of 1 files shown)


lyubomir:
Well... I agree that the original code before the modification wasn't good at protecting the TransportManager.channels List and, consequently, removing the synchronized block shouldn't make a (theoretically) bad situation worse. I'd rather we at least fix the (theoretically) possible ConcurrentModificationException (which I've fixed locally and I'm currently testing) in addition to fixing the deadlock. Anyway, @bgrozev, please feel free to merge your pull request.

gpolitis:
This looks fairly innocent to me as well, especially given that protecting the list is broken anyway.

PR #122, opened 2016-01-07
Split last n

bgrozev:


src/main/java/org/jitsi/videobridge/ratecontrol/BitrateController.java
```diff
@@ -210,7 +214,7 @@ public int calcNumEndpointsThatFitIn()
         long remainingBandwidth = availableBandwidth;
         int numEndpointsThatFitIn = 0;
 
-        final Iterator<Endpoint> it = channel.getReceivingEndpoints();
+        final Iterator<Endpoint> it = null; //channel.getReceivingEndpoints();
```
gpolitis: Is the BitrateController actually used? If it is this would cause an NPE in a few in the while statement.

src/main/java/org/jitsi/videobridge/ratecontrol/BitrateController.java
```diff
@@ -370,19 +382,22 @@ private void initializeConfiguration()
...
             // A bitrate adaptor is not set. It makes no sense to continue.
             return;
         }
+        logger.warn(hashCode()+" YYY new bw estimate: " + remb);
 
         long now = System.currentTimeMillis();
 
         // The number of endpoints this channel is currently receiving
-        int receivingEndpointCount = channel.getReceivingEndpointCount();
+        int receivingEndpointCount = 0;//channel.getReceivingEndpointCount();
```
gpolitis: Is this going to change in subsequent commits?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -525,59 +537,92 @@ private void onPinnedEndpointChangedEvent(
...
-                if (newPinnedEndpoint == null)
-                {
-                    this.weakPinnedEndpoint = null;
-                }
-                else
+                for (int i = 0; i < pinnedEndpoints.size(); i++)
                 {
-                    this.weakPinnedEndpoint
-                        = new WeakReference<>(newPinnedEndpoint);
+                    if (!pinnedEndpoints.get(i).
```
lsevans: I could be misunderstanding, but can pinnedEndpoints.get(i) be null? 

In our case, if you pin a stream then unpin the same stream (back to 0 pinnedEndpoints) this method will get called with a singleton list of [null].

It then seems like the resulting NPE here gets swallowed by someone and a Pinned Endpoint property event never fires.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -525,59 +537,92 @@ private void onPinnedEndpointChangedEvent(
             sc.bind("this", this);
             logger.debug(sc.c(
                     "Endpoint {this.id} notified us that it has pinned"
-                        + " {pinnedId}."));
+                            + " {pinnedId}."));
         }
 
-        Conference conference = weakConference.get();
+        pinnedEndpointsChanged(Collections.singletonList(newPinnedEndpointID));
```
lsevans: Here's how I fixed the bug mentioned in my previous comment: 

List<String> newPinnedIDList = (newPinnedEndpointID == null) ? Collections.EMPTY_LIST : Collections.singletonList(newPinnedEndpointID);
        pinnedEndpointsChanged(newPinnedIDList);

bgrozev:
Thanks, I'll incorporate the fix! 

Personal opinion: I think using the list syntax ("pinnedEndpoints: []") even for a single endpoint is a better approach. It covers all cases, and I don't think we have any reason apart from backward compatibility to maintain the old one, so I would argue for deprecating it at some point.

Issue #102, opened 2015-11-26
Error in reply to raw UDP channel request


joqn:
When posting the following (note the _raw_ in the namespace)

```
{
    "contents": [
        {
            "channels": [
                {
                    "expire": 60,
                    "rtp-level-relay-type": "mixer",
                    "transport": {
                        "xmlns": "urn:xmpp:jingle:transports:raw-udp:1"
                    }
                }
            ],
            "name": "audio"
        }
    ]
}
```

to the REST interface,I get the following reply:

```
{
"contents":[
    {
        "channels":[
            {
            "sources":[
                783642886
            ],
            "rtp-level-relay-type":"translator",
            "expire":60,
            "initiator":true,
            "id":"65f2743ac85bd3f9",
            "transport":{
                "candidates":[
                    {
                        "generation":0,
                        "component":1,
                        "port":10002,
                        "ip":"127.0.0.1",
                        "id":"3",
                        "type":"host"
                    },
                    {
                        "generation":0,
                        "component":2,
                        "port":10003,
                        "ip":"127.0.0.1",
                        "id":"4",
                        "type":"host"
                    }
                ],
                "xmlns":"urn:xmpp:jingle:transports:raw-udp:1"
            },
            "direction":"sendrecv"
            }
        ],
        "name":"audio"
    }
],
"id":"8c68373d95edc463"
}
```

which has 2 drawbacks:
1. I get "translator" back although a "mixer" was requested.  Is translator mode generally not supported in this case?
2. The candidates reported back have 127.0.0.1 as the IP address, which is not very useful in real cases.  Note that I don't have any XMPP set up, so the helper lookup by using XMPP domains fails.
   It could be fixed by using getLocalHostLANAddress in RawUdpTransportManager.createStreamConnector, as proposed here
   http://stackoverflow.com/questions/9481865/getting-the-ip-address-of-the-current-machine-using-java

joqn:
I'm happy to fix these problems myself, but please respond whether there are fundamental problems with it (translator vs. mixer case for raw UDP), and whether you consider the 127.0.0.1 a bug that should be fixed.

bgrozev:
I think both of these are bugs. Note that while not unsupported, both the raw-udp transport and the mixer mode are not actively used, so regressions might have crept in.

joqn:
Isn't "mixer" at least the standard way for audio?

The control flow causing the first problem is actually the very same as in 
[...]

PR #128, opened 2016-02-03
Use getLocalHostLanAddress to avoid using 127.0.0.1 as the local IP

joqn:
The code is mostly taken from here, which should also be Apache-licensed
https://issues.apache.org/jira/browse/JCS-40

It fixes the second problem in https://github.com/jitsi/jitsi-videobridge/issues/102

It might be desirable anyway to allow to configure "the" IP address. Suggestions for a property name?


bgrozev:
This looks good to me, but can you please fix the formatting? We have a strict limit of 80 chars per line and use package imports (e.g. import org.junit.*) everywhere.

joqn:
Okay. Better now?

bgrozev:
Yes, perfect. Thank you!

Issue #9, opened 2014-07-03
Problem with setting "mixer" rtpLevelRelayType for RTPChannels


romix:
I think recent changes to Jitsi-videobridge introduced a small bug.

Constructor of RTPChannel performs this call even before the XML attributes of the channel are parsed:

https://github.com/jitsi/jitsi-videobridge/blob/master/src/org/jitsi/videobridge/RtpChannel.java#L205

``` java
 if (RTPLevelRelayType.MIXER.equals(getRTPLevelRelayType()))
```

As a result, getRTPLevelRelayType sets the type to a default, which is a "translator". When attributes of the channel's XML description are parsed later and the attribute `rtp-level-relay-type="mixer"` is discovered later in Videobridge.handleColibriConferenceIQ, the code tries to set the type to "mixer". But it does not work, because the logic in RtpChannel.setRTPLevelRelayType allows to set a type only if it is not set set or it is the same. Since each RTPChannel has now "translator" as an initially set type, it is impossible to set "mixer" at all, as far as I understand. 

This seems to be a regression from previous versions of the video bridge. I guess it should be still possible to set a "mixer" mode during conference establishment.

PR #129, opened 2016-02-03
Fix setting RTP relay type to mixer

joqn:
(This PR includes https://github.com/jitsi/jitsi-videobridge/pull/128 in terms of test infrastructure, please accept that one before).

This fixes the first part of https://github.com/jitsi/jitsi-videobridge/issues/102 and presumably also https://github.com/jitsi/jitsi-videobridge/issues/9

The problem was:
RtpChannel.initialize calls transportConnected because for RawUdpTransportManager, isConnected is immediately true. Then transportConnected -> maybeStartStream -> getRTPLevelRelayType
As the RTP level type is not yet set at this point in time, it will be set to the default TRANSLATOR in getRTPLevelRelayType.

There is explicitly a comment stating

            /*
             * We've postponed the invocation of the method
             * getRTPLevelRelayType() in order to make sure that the conference
             * focus has had a chance to set the RTP-level relay type. We have
             * to invoke the method MediaStream#setSSRCFactory(SSRCFactory)
             * before starting the stream.
             */

but it's just not true before the patch proposed here.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Content.java
+++ b/src/main/java/org/jitsi/videobridge/Content.java
@@ -301,8 +301,6 @@ public class Content
         // Initialize channel
         channel.initialize(rtpLevelRelayType);
 
-        Videobridge videobridge = getConference().getVideobridge();
-
         if (logger.isInfoEnabled())
         {
             /*
@@ -311,6 +309,7 @@ public class Content
              * of causing deadlocks.
              */
 
+            Videobridge videobridge = getConference().getVideobridge();
             logger.info(
                     "Created channel " + channel.getID() + " of content "
                         + getName() + " of conference " + conference.getID()
--- a/src/main/java/org/jitsi/videobridge/Content.java
+++ b/src/main/java/org/jitsi/videobridge/Content.java
@@ -19,6 +19,7 @@ import java.io.*;
 import java.lang.ref.*;
 import java.util.*;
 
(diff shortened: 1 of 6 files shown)


bgrozev:
This also looks good (and passes our torture tests), but same note as for #128: please fix the formatting.

joqn:
Formatting fixed.

bgrozev:
Thank you!

brianh5:
@joqn is mixer mode working for you on the bridge?  i'm on master as of yesterday and when i set the relay mode to mixer i get no audio at all on any of the participants.

joqn:
Mixer is working for us, but we are not at very latest master, and also, we use our own client, and set up Jitsi via the REST interface.

brianh5:
@joqn thanks, could you tell me what revision you're on?  we're using our own client and the rest interface as well.  i see the channel relay type gets set to mixer, but we don't get any audio.

joqn:
So far we are based on c55c809b6e8b0dc3c26372b25eff861c69c11fe3.

brianh5:
@joqn thanks...i tried that revision but still no luck.  tried with and without last-n configured.  do you have any other settings applied? anything in sip-communicator.properties maybe?

brianh5:
also, what audio codec are your clients sending?  we're doing opus, wondering if it's related to that.

PR #130, opened 2016-02-04
Synchronize access to RtpChannel.stream.

joqn:
This fixes a nasty race condition during media channel establishment. We noticed it in mixer mode using an automated test with all components on a single machine, so it might be due to very tight timing.

In RtpChannel, during the execution of 

    stream.start()

a media packet from the peer comes in already, indirectly triggering 

    stream.setDirection(MediaDirection.SENDRECV);

in another thread.
However, as MediaStreamImpl.started is not yet set to true, this call effectively does not start the outgoing stream, resulting in a one-way media stream.

This could alternatively be fixed by declaring MediaStreamImpl.start/stop/setDirection as "synchronized" in libjitsi.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -544,7 +544,10 @@ public class RtpChannel
                                     && !format.equals(stream.getFormat()))
                             {
                                 stream.setFormat(format);
-                                stream.setDirection(MediaDirection.SENDRECV);
+                                synchronized (streamSyncRoot)
+                                {   // otherwise races with stream.start()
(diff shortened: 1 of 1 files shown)


cmeng-git:
Your described problem with all components installed on the same machine seems similar to the one I encountered as described in: [jitsi-dev] [jitsi-videobridge] JVB downstream RTP/RTCP link Setup implementation

that leads to only one-way media stream with missing outgoing stream.

I tried your patches to the RtpChannel.java. However it does not fix my problem even after the fixes.

Note: I am using the latest jitsi-videobridge source pull on Feb 6, 2016.

Any further advice that I can try to pin-point the problem?

bgrozev:
Merged. Thank you, @joqn!

PR #132, opened 2016-02-08
Fixes issues with lastN when the N+1th person joins

lsevans:
Removed call to updateInLastN() when SCTP connection is ready.
This appeared to make all clients disregard the N+1th person's video feed.

Also, Fixes logic for when pinned endpoints change. It was previously passing only null or empty lists of endpoints.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -540,12 +540,7 @@ public class Endpoint
                             + " {pinnedId}."));
         }
 
-        List<String> newPinnedIDList = Collections.EMPTY_LIST;
-        if (newPinnedEndpointID == null || "".equals(newPinnedEndpointID))
-        {
-            newPinnedIDList = Collections.singletonList(newPinnedEndpointID);
-        }
-
+        List<String> newPinnedIDList = (newPinnedEndpointID == null) ? Collections.EMPTY_LIST : Collections.singletonList(newPinnedEndpointID);
         pinnedEndpointsChanged(newPinnedIDList);
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -540,12 +540,7 @@ private void onPinnedEndpointChangedEvent(
                             + " {pinnedId}."));
         }
 
-        List<String> newPinnedIDList = Collections.EMPTY_LIST;
-        if (newPinnedEndpointID == null || "".equals(newPinnedEndpointID))
-        {
-            newPinnedIDList = Collections.singletonList(newPinnedEndpointID);
-        }
-
+        List<String> newPinnedIDList = (newPinnedEndpointID == null) ? Collections.EMPTY_LIST : Collections.singletonList(newPinnedEndpointID);
```
lsevans: This was previously only sending a Collections.EMPTY_LIST or a singletonList with a null or empty value

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -494,8 +494,6 @@ void sctpConnectionReady(Endpoint endpoint)
                 sendLastNEndpointsChangeEventOnDataChannel(
                         lastNController.getForwardedEndpoints(), null);
             }
-
-            updateInLastN(this);
```
lsevans: This one I don't have the best understanding of.. but I was seeing an issue when lastN=1 and clients A and B are in a conference. Then a 3rd client, C, tries to join.

Client C downloads 1 stream like we expect and can switch between A and B's feeds, but A or B cannot view C's stream (even when pinned)

It looks like this call sets C's inLastN value to false for all the other channels, which causes the LastNTransformEngine to drop all the RTP packets. 

This results in A and B receiving a muted video stream from client C. 

Is there somewhere that the inLastN value should be set back to true?

bgrozev:
@lsevans Thank you for the report and suggested solutions! I'll fix the problems separately. For the first one, I want to keep the check for "". For the second one, the inLastN field should be updated properly.

PR #134, opened 2016-02-12
Prevent NullPointerException when there are no simulcast source groups.

legastero:
The main for loop at line 894 short circuits if a source group does not
have a simulcast semantic. If there are no simulcast source groups at
all, the ssrc2stream hash map is never populated, triggering a NPE
after the main for loop.

This change detects that there were no simulcast groups and skips
running those final steps.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/VideoChannel.java
+++ b/src/main/java/org/jitsi/videobridge/VideoChannel.java
@@ -915,16 +915,18 @@ public class VideoChannel
 
         }
 
-        // FID groups have been saved in RtpChannel. Make sure any changes are
-        // propagated to the appropriate SimulcastStream-s.
-        for (Map.Entry<Long, Long> entry : this.fidSourceGroups.entrySet())
-        {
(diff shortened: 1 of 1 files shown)


bgrozev:
Thank you for the fix!

Do I correctly understand that the NPE is due to simulcastStream being null on line 923? If that's the case, I think an NPE could still occur in the case where there are some SIM groups, but there is also a FID group for an SSRC not in a SIM group, for example something like:
SIM 1 2
FID 1 3
FID 2 4
FID 100 200

legastero:
Correct. 

The case we've been dealing with is when we have FID groups, but no SIM groups.

bgrozev:
Cool, thanks! 

I need your contributor agreement (CLA) before I can merge, and I can't find it. You can sign the [individual](https://jitsi.org/icla) or [corporate](https://jitsi.org/ccla) version. If you have already signed it, please let me know of the name that was used.

legastero:
Just signed as individual, under Lance Stout

bgrozev:
Thank you!

PR #137, opened 2016-02-17
Updates the "inLastN" property of a VideoChannel when changes to the …

bgrozev:
…"lastN" set occur.

Torture passed, manually tested with the scenario described in #132.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/LastNController.java
+++ b/src/main/java/org/jitsi/videobridge/LastNController.java
@@ -15,6 +15,7 @@
  */
 package org.jitsi.videobridge;
 
+import org.jitsi.service.neomedia.*;
 import org.jitsi.util.*;
 import org.jitsi.videobridge.ratecontrol.*;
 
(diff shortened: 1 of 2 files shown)


davidertel:
I will do some testing on this tomorrow, thanks for the fix!

src/main/java/org/jitsi/videobridge/LastNController.java
```diff
@@ -544,6 +551,91 @@ else if (newForwardedEndpoints.size() < currentLastN)
...
+        return getVideoChannels(endpoint);
+    }
+
+    /**
+     * Gets the list of {@link VideoChannel}s belonging to an endpoint.
+     *
+     * @param endpoint the endpoint.
+     * @return the {@link VideoChannel}s of the endpoint.
+     */
+    private List<VideoChannel> getVideoChannels(Endpoint endpoint)
```
gpolitis: Do we really need this since we have `endpoint.getChannels(MediaType.VIDEO)`?

gpolitis:
LGTM. @davidertel do you still want to do some testing before I merge this?

lsevans:
I just ran some tests and confirmed this fixes the issue I was seeing. Thanks!

davidertel:
George, I did some testing and it looks good to me. Thank you again

gpolitis:
@lsevans, @davidertel thank you!

PR #160, opened 2016-03-03
Fix a NullPointerException in IceUdpTransportManager

lyubomir:
Fixes a NullPointerException in IceUdpTransportManager which occurs repeatedly and regularly. That's likely happening during health checks (from Jicofo, for example). No functionality appears to be broken.


damencho:
It is ok to test now by jenkins.

damencho:
It is ok to test now by jenkins.

lyubomir:
It is ok to test now by jenkins.

PR #161, opened 2016-03-04
Adaptive simulcast

bgrozev:
This adds a new implementation of adaptive simulcast in AdaptiveSimulcastBitrateController. The class previously named BitrateController is renamed to LastNBitrateController. The intention is to eventually refactor the classes in the ratecontrol package.

Note that this depends on libjitsi PR 95 and shouldn't be merged without it (it won't compile): skip ci


src/main/java/org/jitsi/videobridge/ratecontrol/AdaptiveSimulcastBitrateController.java
```diff
@@ -0,0 +1,546 @@
...
+    /**
+     * Notifies this instance that the estimation of the available bandwidth
+     * has changed.
+     *
+     * @param bwe the new estimation of the available bandwidth in bits per second.
+     */
+    @Override
+    public void bandwidthEstimationChanged(long bwe)
+    {
+        latestBwe = bwe;
```
gpolitis: Why do we care about the remote send bandwidth estimation (that's what the bandwidth estimation algorithm running at the bridge is calculating)? We can't use those estimations to figure out if we're oversending, instead we need to take into account the values in the REMBs. Or am I missing something obvious?

gpolitis: Thanks for explaining @bgrozev, I thought it was the remote bitrate estimator that was giving those values.

bgrozev:
Thanks, @gp! I think the latest commits should address these issues.

PR #168, opened 2016-03-15
fix the list comparison when checking for new last-n to be order agno…

brianh5:
…stic

@gpolitis @bgrozev


bgrozev:
Thanks, @brianh5 ! We do have your CLA, I guess the comment above is from some of the newly added automations (@damencho?).

Jenkins: add to whitelist

PR #171, opened 2016-03-16
Build Improvements: put the git SHA-1 in the jar manifest and in the package filename 

matteocampana-zz:


bgrozev:

PR #176, opened 2016-03-19
Reduce syncronization on Channel instances.

bgrozev:


bgrozev:
Note that this depends on libjitsi [PR 112](https://github.com/jitsi/libjitsi/pull/112), and pom.xml should be updated before merging.

PR #177, opened 2016-03-19
Uses the thread pool to handle messages coming from the usrsctp stack,

bgrozev:
in order to ensure that the calling thread does not block. We have
observed that blocking this thread (e.g. by having it eventually try
to write to a Socket) results in calls to usrsctp_socket to block (i.e.
to our inability to initialize any new SCTP sockets).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/SctpConnection.java
+++ b/src/main/java/org/jitsi/videobridge/SctpConnection.java
@@ -927,7 +927,7 @@ public class SctpConnection
         sctpSocket.setLink(new NetworkLink()
         {
             @Override
-            public void onConnOut(SctpSocket s, byte[] packet)
+            public void onConnOut(SctpSocket s, final byte[] packet)
                 throws IOException
             {
                 if (LOG_SCTP_PACKETS)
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -943,8 +943,16 @@ public void onConnOut(SctpSocket s, byte[] packet)
                             packet);
                 }
 
-                // Send through DTLS transport
-                transformer.sendApplicationData(packet, 0, packet.length);
+                // Send through DTLS transport.
+                threadPool.execute(new Runnable()
```
lyubomir: I'm not absolutely comfortable with sending each "packet" in a separate Thread of "threadPool". You said the TCP socket would block for very roughly 30 seconds. DominantSpeakerIdentification runs at an interval of 300 milliseconds which in the worst case would result in 100 "threadPool" Threads. I'd rather we use a packet queue and send it in a single Thread (of "threadPool"). Anyway, I'll think about it a tiny bit more.

bgrozev: This would be a queue and a thread for each SctpConnection instance, correct?

lyubomir: Yup. If you think we're fine though, I'll merge this the way it is.

bgrozev:
After a discussion with @lyubomir we agreed to go with a queue.

PR #181, opened 2016-03-21
Calls MediaStream#setTarget() before setConnector().

bgrozev:
Tries to solve a problem reported by Somil Bansal on the dev mailing list. Calls setTarget() before setConnector(), because the latter starts the DTLS server or client, which shouldn't be started without a target. See a discussion here: http://lists.jitsi.org/pipermail/dev/2016-March/027119.html


bgrozev:
The failure is probably not related to the changes in the PR. Will try the tests again later.

damencho:
Jenkins: it's ok to test.

PR #183, opened 2016-03-23
Uses a PacketQueue for packets coming out of the SCTP stack in order to

bgrozev:
ensure that the calling thread does not block. We have
observed that blocking this thread (e.g. by having it eventually try
to write to a Socket) results in calls to usrsctp_socket to block (i.e.
to our inability to initialize any new SCTP sockets).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/SctpConnection.java
+++ b/src/main/java/org/jitsi/videobridge/SctpConnection.java
@@ -211,25 +211,7 @@ public class SctpConnection
      * The instance which we use to handle packets read from
      * {@link #packetQueue}.
      */
-    private final PacketQueue.PacketHandler<RawPacket> handler
-        = new PacketQueue.PacketHandler<RawPacket>()
-    {
-        @Override
-        public boolean handlePacket(RawPacket pkt)
-        {
-            if (pkt == null)
-                return true;
-
-            if (transformer == null)
-                return false;
-
-            System.err.println("XXX sending sctp data over DTLS");
-            transformer.sendApplicationData(
-                pkt.getBuffer(), pkt.getOffset(), pkt.getLength());
-
-            return true;
-        }
-    };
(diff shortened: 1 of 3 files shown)


brianh5:
could this potentially result in long datachannel connection times (~30 seconds)?  we've seen that happen on occasion...not sure if that's just the nature of sctp or if it may be related to something like this.

bgrozev:
Possibly, but I think you would have noticed other problems and not just slow SCTP establishment. In what we observed, while the thread was blocked health checks never returned, at least some COLIBRI requests never returned and the statistics were broken.

src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -195,6 +196,42 @@ private static synchronized int generateDebugId()
...
+        @Override
+        public boolean handlePacket(RawPacket pkt)
+        {
+            if (pkt == null)
+                return true;
+
+            if (transformer == null)
+                return false;
+
+            System.err.println("XXX sending sctp data over DTLS");
```
paweldomas: System.err.println ? Forgot to remove before commit ?

src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -195,6 +196,42 @@ private static synchronized int generateDebugId()
...
+     * SCTP packets.
+     */
+    private DtlsPacketTransformer transformer = null;
+
+    /**
+     * The instance which we use to handle packets read from
+     * {@link #packetQueue}.
+     */
+    private final PacketQueue.PacketHandler<RawPacket> handler
+        = new PacketQueue.PacketHandler<RawPacket>()
```
paweldomas: Don't really like creating anonymous class here, but that's just my opinion. Would prefer to have named class at the bottom, does the same thing.

PR #185, opened 2016-03-24
Simulcast fixes and optimizations.

gpolitis:


src/main/java/org/jitsi/videobridge/simulcast/SimulcastReceiver.java
```diff
@@ -188,32 +189,76 @@ public void setSimulcastStreams(long[] ssrcs)
...
+    public void setSimulcastStreams(SimulcastStream[] newSimulcastStreams)
     {
-        this.simulcastStreams = simulcastStreams;
+        SimulcastStream[] oldSimulcastStreams = this.simulcastStreams;
+        int oldLen
+            = oldSimulcastStreams == null ? 0 : oldSimulcastStreams.length;
+        int newLen
+            = newSimulcastStreams == null ? 0 : newSimulcastStreams.length;
+
+        if (oldLen == newLen)
```
bgrozev: Can we simplify this? Maybe use something like:
boolean equal = (oldLen == 0 && newLen == 0) || Arrays.equals(oldSimulcastStreams, newSimulcastStreams);

src/main/java/org/jitsi/videobridge/simulcast/SimulcastReceiver.java
```diff
@@ -571,9 +630,11 @@ private void maybeTimeout(
             SimulcastStream cause,
             RawPacket pkt,
             SimulcastStream effect,
+            List<SimulcastStream> localSimulcastStreamFrameHistory,
```
bgrozev: I think this (and endIndexInSimulcastStreamFrameHistory, too) needs a javadoc. It is not obvious what it is.

src/main/java/org/jitsi/videobridge/simulcast/SimulcastStream.java
```diff
@@ -243,4 +243,29 @@ public void askForKeyframe()
     {
         simulcastReceiver.askForKeyframe(this);
     }
+
+    @Override
```
bgrozev: These need javadocs, and formatting fixes. I don't think we have one-line ifs anywhere else.

src/main/java/org/jitsi/videobridge/simulcast/SimulcastStream.java
```diff
@@ -243,4 +243,29 @@ public void askForKeyframe()
...
+
+    }
+
+    @Override
+    public int hashCode()
+    {
+        int result = (int) (primarySSRC ^ (primarySSRC >>> 32));
+        result = 31 * result + (int) (rtxSSRC ^ (rtxSSRC >>> 32));
+        result = 31 * result + (int) (fecSSRC ^ (fecSSRC >>> 32));
+        result = 31 * result + order;
```
bgrozev: This makes me nervous. These fields should all be final if they are used in hashCode(). For primarySSRC, fecSSRC and order this can be made easily, the rtxSSRC will require a bit of refactoring.

src/main/java/org/jitsi/videobridge/simulcast/SimulcastReceiver.java
```diff
@@ -562,6 +571,10 @@ else if (simStream == acceptedStream)
         }
```
bgrozev: Ideally we should split this method because it is way too long, but that's not related to the PR. It isn't obvious that changedStreams doesn't contain duplicates, and we probably want it to not have duplicates. I think we should use a Set here (after fixing SimulcastStream#hashCode)

src/main/java/org/jitsi/videobridge/simulcast/SimulcastReceiver.java
gpolitis: Good point, thanks!

src/main/java/org/jitsi/videobridge/simulcast/SimulcastReceiver.java
gpolitis: Good point, thanks!

src/main/java/org/jitsi/videobridge/simulcast/sendmodes/RewritingSendMode.java
```diff
@@ -115,42 +118,56 @@ public void receive(SimulcastStream simStream, boolean urgent)
...
+            this.state
+                = new State(oldState.weakCurrent, new WeakReference<>(simStream));
         }
     }
 
     /**
-     * Gets the <tt>SimulcastStream</tt> that is currently being received.
-     *
-     * @return
+     * A simple class that holds the state of a {@RewritingSendMode}.
```
bgrozev: Typo?

src/main/java/org/jitsi/videobridge/simulcast/sendmodes/RewritingSendMode.java
```diff
@@ -115,42 +118,56 @@ public void receive(SimulcastStream simStream, boolean urgent)
...
-     */
-    public SimulcastStream getNext()
-    {
-        WeakReference<SimulcastStream> wr = this.weakNext;
-        return (wr != null) ? wr.get() : null;
+        /**
+         * A <tt>WeakReference</tt> to the <tt>SimulcastStream</tt> that is
+         * currently being received.
+         */
+        private final WeakReference<SimulcastStream> weakCurrent;
```
bgrozev: Since this is moved to a class of it's own, can we hide the WeakReference complexity with a getter and setter?

src/main/java/org/jitsi/videobridge/simulcast/SimulcastStream.java
gpolitis: Good point, thanks!

src/main/java/org/jitsi/videobridge/simulcast/SimulcastStream.java
gpolitis: Good point, thanks!

src/main/java/org/jitsi/videobridge/simulcast/sendmodes/RewritingSendMode.java
gpolitis: Yep :)

PR #190, opened 2016-03-25
Set target before connector

bgrozev:


bgrozev:
Jenkins: add to whitelist

PR #205, opened 2016-04-04
Enable adaptive simulcast by default

bgrozev:
Adds null checks which could occur if adaptive simulcast is enabled, but
simulcast is disabled.


bgrozev:
Jenkins: ok to test

PR #216, opened 2016-04-14
Don't use Map<>.containsKey before Map<>.get

champtar:
if the key doesn't exist Map<>.get return null,
and we are not putting null as value, so use that.


lyubomir:
Jenkins, it's ok to test this.

lyubomir:
Jenkins, test this please.

lyubomir:
The first automatic PR test failed because of a known Jenkins failure condition related to an SSH library. The second automatic PR test failed because of known test failures.

PR #225, opened 2016-04-20
Fixes a problem where you may get stuck with low quality video

gpolitis:
When data channel connection establishment is fast and if the selected
endpoint data channel message is sent early enough then when one user
(re)joins a call, it's getting a good quality stream from the participant
that was already in the room. But the user that was already in the room,
gets a low quality stream from the newly joined client. So effectively,
in a 1:1 session, one user always ends up with a low quality stream of
the remote... The user that joined first.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/simulcast/SimulcastSenderManager.java
+++ b/src/main/java/org/jitsi/videobridge/simulcast/SimulcastSenderManager.java
@@ -155,18 +155,15 @@ public class SimulcastSenderManager
 
         if (simulcastSender == null) // Create a new sender.
         {
-            int targetOrder = simulcastEngine
-                .getVideoChannel() // never null.
-                .getReceiveSimulcastLayer();
+            VideoChannel videoChannel = simulcastEngine.getVideoChannel();
+            int targetOrder = videoChannel.getReceiveSimulcastLayer();
 
-            Endpoint sendingEndpoint = simulcastReceiver // never null.
-                .getSimulcastEngine() // never null.
-                .getVideoChannel() // never null.
-                .getEndpoint(); // might be null.
+            Endpoint sendingEndpoint = simulcastReceiver
+                .getSimulcastEngine()
+                .getVideoChannel()
+                .getEndpoint();
 
-            Endpoint receivingEndpoint = simulcastEngine // never null.
-                .getVideoChannel() // never null.
-                .getEndpoint(); // might be null.
+            Endpoint receivingEndpoint = videoChannel.getEndpoint();
 
             if (receivingEndpoint != null && sendingEndpoint != null)
             {
--- a/src/main/java/org/jitsi/videobridge/simulcast/SimulcastSenderManager.java
+++ b/src/main/java/org/jitsi/videobridge/simulcast/SimulcastSenderManager.java
@@ -134,8 +134,15 @@ public class SimulcastSenderManager
     private synchronized SimulcastSender getOrCreateSimulcastSender(
         SimulcastReceiver simulcastReceiver)
     {
-        if (simulcastReceiver == null
-                || !simulcastReceiver.isSimulcastSignaled())
+        if (simulcastReceiver == null)
+        {
+            return null;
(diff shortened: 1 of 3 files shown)


bgrozev:
LGTM

src/main/java/org/jitsi/videobridge/simulcast/SimulcastSenderManager.java
```diff
@@ -134,8 +134,15 @@ public boolean accept(RawPacket pkt)
...
-                || !simulcastReceiver.isSimulcastSignaled())
+        if (simulcastReceiver == null)
+        {
+            return null;
+        }
+
+        SimulcastStream[] simulcastStreams
+            = simulcastReceiver.getSimulcastStreams();
+
+        if (simulcastStreams == null || simulcastStreams.length == 0)
```
lyubomir: What?! Could you please write a comment why you're checking for simulcastStreams? I get it that it's necessary later on but that's under 2 other ifs there so I don't think it's obvious why you need to perform the check and return early and in all cases regardless of the later ifs.

gpolitis: Thank you @lyubomir !

src/main/java/org/jitsi/videobridge/simulcast/SimulcastSenderManager.java
```diff
@@ -144,11 +151,31 @@ private synchronized SimulcastSender getOrCreateSimulcastSender(
...
+                .getReceiveSimulcastLayer();
 
+            Endpoint sendingEndpoint = simulcastReceiver // never null.
+                .getSimulcastEngine() // never null.
+                .getVideoChannel() // never null.
+                .getEndpoint(); // might be null.
+
+            Endpoint receivingEndpoint = simulcastEngine // never null.
+                .getVideoChannel() // never null.
+                .getEndpoint(); // might be null.
```
lyubomir: (1) Could I please ask you to rewrite that without repeating field accesses, methods calls and comments? I believe the source code will become more readable. For example,

VideoChannel videoChannel = simulcastEngine.getVideoChannel();
int targetOrder = videoChannel.getReceiveSimulcastLayer();
...
Endpoint receivingEndpoint = videoChannel.getEndpoint();

(2) I find these comments about "never null" and "might be null" distracting and unnecessary at times. For example, you check for receivingEndpoing right afterwards so it's pointless to me to say here that it might be null because your source code says it.

Additionally, with the repeating field accesses and method calls, you get into inconsistencies like the first simulcastEngine access doesn't have a comment but later on it has a comment.

gpolitis: Will do, thank you @lyubomir !

src/main/java/org/jitsi/videobridge/simulcast/SimulcastSenderManager.java
```diff
@@ -134,21 +134,49 @@ public boolean accept(RawPacket pkt)
...
+        {
+            return null;
+        }
+
+        SimulcastStream[] simulcastStreams
+            = simulcastReceiver.getSimulcastStreams();
+
+        if (simulcastStreams == null || simulcastStreams.length == 0)
         {
+            // This is equivalent to !simulcastReceiver.isSimulcastSignaled() in
```
lyubomir: I don't want you to modify this any further, I merely want to throw in my brief observations:

You're in a private method which is called by a single (at the time of this writing) other method. The caller has the check:

(simulcastReceiver == null || !simulcastReceiver.isSimulcastSignaled())

And the caller appears to perform the check just because it wants to call this method, the caller doesn't need simulcastReceiver itself.

But then you have the same checks in this method:
(1) (simulcastReceiver == null)
(2) Another, more verbose alternative to !simulcastReceiver.isSimulcastSignaled().

Again, I don't want to modify this PR any further and I merely wanted to state that I think there's repetition which adds complexity and, consequently, increases maintenance costs.

PR #226, opened 2016-04-20
Breaks out of the loop if the SCTP socket early

bgrozev:
Breaks out of the loop if the SCTP socket is closed before trying to receive from the ICE socket.

This potentially avoids an infinite loop, if the ICE socket
implementation is faulty and returns something (e.g. a packet with
length 0) when the socket is closed. More importantly, it makes it
obvious to anyone reading the code that such a scenario is not possible.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/SctpConnection.java
+++ b/src/main/java/org/jitsi/videobridge/SctpConnection.java
@@ -1080,6 +1080,9 @@ public class SctpConnection
         {
             do
             {
+                if (sctpSocket == null)
+                    break;
+
                 iceSocket.receive(recv);
 
                 RawPacket[] send
@@ -1117,9 +1120,6 @@ public class SctpConnection
                                 s.getBuffer(), s.getOffset(), s.getLength());
                     }
                 }
-
-                if (sctpSocket == null)
-                    break;
 
                 // Pass network packet to SCTP stack
                 for (RawPacket s : send)


src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -1080,6 +1080,9 @@ public void run()
         {
             do
             {
+                if (sctpSocket == null)
+                    break;
+
```
lyubomir: Could I please ask you to write a comment explaining why the check's there? I got the reason from your PR comment but I don't think that's obvious (to me).

src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -1118,9 +1121,6 @@ public void run()
                     }
                 }
 
-                if (sctpSocket == null)
-                    break;
-
```
lyubomir: I get the reason why the same check is necessary before the receive (after reading your PR comment). But I don't get why it's no longer necessary after the receive. Generally, a receive usually takes significantly long time to return so checks from before may not hold afterwards. In the case that may or may not be true, I'm just saying that it's not obvious to me.

On the other hand, the for loop bellow will repeatedly access sctpSocket so it's not like it will not become null inside the loop (I guess).

Anyway, I suppose I mostly wanted to express my feeling that something's amiss around these null checks outside a loop and accesses to a field in a loop.

It's outside the scope of this PR, of course, but I really feel that this method should be broken down into multiple smaller ones in order to aid reading and understanding.

bgrozev: I will close this PR. I agree that we should refactor this anyway, and doing so is on the roadmap, so I'll leave it for now.

lyubomir: I'm sorry.

PR #231, opened 2016-04-28
Adds callstats user statistics per endpoint

damencho:


damencho:
Depends on https://github.com/jitsi/libjitsi/pull/154. And needs it first to be merged and then this one needs an update in pom.xml with the correct libjitsi version.

src/main/java/org/jitsi/videobridge/eventadmin/callstats/Activator.java
```diff
@@ -0,0 +1,171 @@
...
+                    serviceRegistration
+                        = EventUtil.registerEventHandler(
+                                bundleContext,
+                                topics,
+                                conferenceStatsHandler);
+                }
+            }
+            break;
+
+        case ServiceEvent.UNREGISTERING:
```
lyubomir: @damencho, shouldn't you be listening to UNREGISTERING of CallStats only here and not just any OSGi service? Like you do in REGISTERED?

PR #235, opened 2016-05-04
Fix a regression after #190 when RawUdpTransport is used.

matteocampana:
After #190 when RawUdpTransport is used the media stream never starts, because with RawUdpTransport the streamTarget is always null by definition, and so maybeStartStream() returns too early without reaching stream.start().
I mantained the setTarget / setConnector order as in #190.


bgrozev:
Jenkins: add to whitelist

PR #240, opened 2016-05-11
Fix #227: Fix SSRC Mapping when people join a recorded conversation

### LINKED ISSUE #227: No entry in json files for people who join after recording has started
When someone joins a session which is recording, jetsi meet will not record his track in metadata.json nor endpoints.json.

roidelapluie:
Add the mapping from SSRC to EndpointID to the Synchronizer instance used by
RecorderRtpImpl when new SSRCs are signalled over colibri.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -1726,6 +1726,17 @@ public class RtpChannel
         addedSSRCs.removeAll(oldSignaledSSRCs);
         if (!addedSSRCs.isEmpty())
         {
+
+            Recorder recorder = null;
+            Synchronizer synchronizer = null;
+            Endpoint endpoint = null;
+            if (getContent().isRecording())
+            {
+                recorder = getContent().getRecorder();
+                synchronizer = recorder.getSynchronizer();
+                endpoint = getEndpoint();
+            }
+
             for (Integer addedSSRC : addedSSRCs)
             {
                 try
@@ -1733,6 +1744,15 @@ public class RtpChannel
                     // Do allow the number of explicitly signalled SSRCs to
                     // exceed the limit.
                     addReceiveSSRC(addedSSRC, false);
+
+                    // If recording is enabled, we need to add the mapping from
+                    // SSRC to EndpointID into the Synchronizer
+                    // instance used by RecorderRtpImpl.
+                    if (recorder != null && endpoint != null && synchronizer != null)
+                    {
+                        synchronizer.setEndpoint(addedSSRC & 0xffffffffl,
+                            endpoint.getID());
+                    }
                 }
                 catch (SizeExceededException see)
                 {


roidelapluie:
It is signes now

roidelapluie:
ping @gavllew

gavllew:
I can't help much, as I'm just another third party - I can't approve pull requests.  However I do at least have a comment on the diff which I'll add in a second.

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1857,13 +1857,38 @@ public void setSources(List<SourcePacketExtension> sources)
...
                     // Do allow the number of explicitly signalled SSRCs to
                     // exceed the limit.
                     addReceiveSSRC(addedSSRC, false);
+
+                    // If recording is enabled, we need to add the mapping from
+                    // SSRC to EndpointID into the Synchronizer
+                    // instance used by RecorderRtpImpl.
+                    if (addToSyncronizer)
+                    {
+                        synchronizer.setEndpoint(addedSSRC & 0xffffffffl,
```
gavllew: The `synchronizer` variable will be out-of-scope before you reach this code block.

bgrozev:
Looks OK to me, except that it doesn't compile :). Should be a quick fix.

Are you actually using some variant of this? I don't understand what problem it solves. If I understand correctly, it only makes a difference when recording SSRCs that are signaled, but don't send and RTP/RTCP packets.

gavllew:
The core problem was that new streams added after recording had started were not appearing in metadata.json if the SSRCs were signalled before arriving in media packets.  This is because the metadata attempts to calculate the wall-clock instant that the stream starts, which depends on the per-Endpoint RTP timestamp mapping maintained by the Synchronizer.

Xura is using a variant of this fix, but we're currently unable to contribute anything because we've never managed to get the CLA sorted out. :-(

bgrozev:
I see now that the code in acceptDataInputStreamDatagramPacket() only updates the synchronizer if the SSRC is new. Thanks for the explanation.

We have the CLA for @roidelapluie so we can get this merged when the PR is fixed.

roidelapluie:
updated

src/main/java/org/jitsi/videobridge/RtpChannel.java
roidelapluie: fixed

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1726,13 +1726,38 @@ public void setSources(List<SourcePacketExtension> sources)
         addedSSRCs.removeAll(oldSignaledSSRCs);
         if (!addedSSRCs.isEmpty())
         {
+
+            boolean addToSyncronizer = false;
+            Recorder recorder = getContent().getRecorder();
```
bgrozev: Calling getRecorder will create a Recorder instance if one doesn't exist (regardless of whether recording is enabled or not). It should only be called if isRecording() returns true (yeah, the design of the API is not good...).

roidelapluie: Fixed

bgrozev:
@zbettenbuk this looks like it may solve one of the issues you described

zbettenbuk:
@bgrozev which one you mean?

bgrozev:
The missing entries from metadata.json

PR #254, opened 2016-06-01
Use harvester local address to configure RawUdpTransport bind address.

matteocampana:
Basically this is a fix to my previous fix: if JVB is NATed, the RawUdpTransport cannot bind to the public address so use the harvester local address instead.
With this change we can use NATed JVB and RawUdpTransport.


bgrozev:
Thank you for the contribution!

PR #257, opened 2016-06-06
Adds maven based exec script.

gpolitis:
Adds a startup script file for the Jitsi Videobridge that uses maven.
Maven is configured to run with a JVB-specific repository path. Multiple
required and optional arguments are supported.


matteocampana:
Just a note:  in the current version the script on my Ubuntu 14.04 machine doesn't work, it complains about "function: not found".
It works if I use for the two functions the syntax "error_exit ()".
I'll create a PR.

PR #258, opened 2016-06-07
Use a more portable syntax for bash functions in mvn_exec.sh

matteocampana:


gpolitis:
Hi @matteocampana, I missed this PR and pushed a fix for this a while ago. Thanks for your contribution anyway :)

PR #260, opened 2016-06-22
[video channe] sequence number rewriting for video rtp packets

mdaneshi:
The change includes sequence number rewriting on the video channel. If an rtp packet is getting forwarded, we update the sequence number. otherwise, we keep track of the delta since the last transmitted sequence number.
We also take care of seq. no. and delta wrap arounds


mdaneshi:
compile error due to dependency on libjitsi change.

bgrozev:
Can you please re-open this from a source branch with the same name like the one in libjitsi (seq_rewrite instead of seqno_rewrite)? That's to make Jenkins use match the two PRs and test them together.

PR #261, opened 2016-06-22
[video channel] sequence number rewriting for video rtp packets

mdaneshi:
The change includes sequence number rewriting on the video channel. If an rtp packet is getting forwarded, we update the sequence number. otherwise, we keep track of the delta since the last transmitted sequence number.
We also take care of seq. no. and delta wrap arounds


mdaneshi:
Ping on this PR, thank you!

PR #265, opened 2016-06-30
Allow the stream direction to be changed before connected if using ICE.

brianh5:
if we're using ice, don't block setting the direction if the connection isn't made yet.

@emcho here's a sample of what some checks for this around the code may look like (and this is a place where we've commented out the check altogether since we're not using latching).  if it looks good, we can start taking a look at other places where checks like this would be useful.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -1458,9 +1458,12 @@ public class RtpChannel
      */
     public void setDirection(MediaDirection direction)
     {
-        // XXX We modify the stream direction only after latching has finished.
-        if (streamTarget.getDataAddress() != null)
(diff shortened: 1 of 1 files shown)


brianh5:
wanted to add that the 'if' condition here could probably be a bit more concise, but i thought this best communicated the situation.  open to something more terse though if you guys think it'd be better.

brianh5:
cc @lyubomir

brianh5:
ping @lyubomir

bgrozev:
FWIW I think this is fine. My understanding is that the original reason for the check was to prevent a direction which doesn't allow receiving to prevent latching from working, and this isn't a concern with ICE.

Jenkins: test again please

brianh5:
which check is failing?  i can't see the full url of the report

damencho:
Jenkins: test this please

PR #269, opened 2016-07-15
set the media format of a mixed audio channel when we get the answer,

brianh5:
and don't force the channel to go to recvonly until it receives a packet

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -548,35 +548,6 @@ public class RtpChannel
                         }
                     }
 
-                    /*
-                     * When performing content mixing (rather than RTP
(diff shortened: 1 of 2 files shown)


brianh5:
first crack at this to show the general idea.  maybe a decision to be made as to how the bridge chooses the format based on the available payload types.

@emcho we talked about this a little while back, unfortunately i don't remember the details of what we agreed on, but figured this is a good enough start to get some feedback.

cc @lyubomir

matteocampana:
Could this change make the bridge starting to send packets without waiting for the first incoming packet?

brianh5:
@matteocampana for an audio channel that was signaled as using the mixer, yes it would cause the bridge to send before waiting to receive a packet.

bgrozev:
Is this still relevant?

bbaldino:
doesn't look like it

PR #272, opened 2016-07-22
Endpoint connection monitoring

paweldomas:
This PR adds endpoint connection status monitoring which sends notifications over data channels when connectivity status changes. Endpoint connection status is considered inactive when there is no media or ICE related traffic on any of it's channels. This is tracked in new 'lastTransportActivityTime' timestamp added to every Channel. It is updated whenever RTP/RTCP packet is received or when ICE "consent freshness" check succeeds(which is now sent more often - every 3 seconds).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Channel.java
+++ b/src/main/java/org/jitsi/videobridge/Channel.java
@@ -65,6 +65,11 @@ public abstract class Channel
     private final String channelBundleId;
 
     /**
+     * Remembers when this <tt>Channel</tt> instance was created.
+     */
+    private final long creationTimestamp = System.currentTimeMillis();
+
+    /**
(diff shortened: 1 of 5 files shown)


brianh5:
just curious, what's the goal of this work?  is it to be able to detect clients who have timed out more quickly?  that's definitely something we've been wanting to do...glad to see the consent check interval shortened, that was one of the main blockers.

paweldomas:
So right now we have this blurred screen effect added to the UI of Jitsi Meet when ICE disconnected state is triggered. But that happens only locally for the user and others do not see that. After this PR the JVB will send notification to everyone in the conference when connectivity issues are detected. There is ongoing work on handling those events in the lib-jitsi-meet and the meet app.

PR #273, opened 2016-07-22
slight rework to LastNController to make it easier to override its

brianh5:
behavior.

idea here is to tweak a couple things to make it easier to override the specific logic of determining which endpoints should be in last-n (we do some different things here, so have a custom HighfiveLastNController to implement the logic we need).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/LastNController.java
+++ b/src/main/java/org/jitsi/videobridge/LastNController.java
@@ -42,7 +42,7 @@ public class LastNController
     /**
      * An empty list instance.
      */
-    private static final List<String> INITIAL_EMPTY_LIST
+    protected static final List<String> INITIAL_EMPTY_LIST
             = Collections.unmodifiableList(new LinkedList<String>());
 
     /**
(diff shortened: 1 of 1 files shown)


brianh5:
cc @bgrozev

bgrozev:
Jenkins test again please

bgrozev:
LGTM, will wait for the tests to run again.

damencho:
Jenkins, test this please

bgrozev:
jenkins test this please

bgrozev:
@brianh5 looking at the base determineLastNList method, it's parameter is unused. Do you use the parameter in your extended implementation, or can we remove it?

brianh5:
@bgrozev yeah i do use it.  memory is a bit fuzzy, but i think since we drive new last-n changes on new endpoints joining (to fill without activity if we have empty slots), there are times where an endpoint is on the new endpoints list, but isn't in conferenceSpeechActivityEndpoints

bgrozev:
Thanks, that makes sense.

PR #276, opened 2016-07-27
Add logs

bgrozev:


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -493,16 +493,12 @@ private void dominantSpeakerChanged()
     {
         Endpoint dominantSpeaker = speechActivity.getDominantEndpoint();
 
-        if (logger.isTraceEnabled())
-        {
-            logger.trace(
-                    "The dominant speaker in conference " + getID()
+        logger.info("The dominant speaker in conference " + getID()
```
gpolitis: We could potentially keep the if statement to guard against unnecessary string concatenation.

src/main/java/org/jitsi/videobridge/simulcast/SimulcastReceiver.java
```diff
@@ -399,6 +400,12 @@ public void accepted(RawPacket pkt)
         // timestamps of the RTP packets.
         long pktTimestamp = pkt.getTimestamp();
         boolean frameStarted = false;
+        boolean isKeyFrame = isKeyFrame(pkt);
```
gpolitis: We could potentially have this inside an if (INFO) statement.

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: We don't check for INFO level anywhere else. Would you prefer if I add the check here?

bgrozev: I stand corrected. Adding the check.

PR #277, opened 2016-07-27
Implements LS hack.

gpolitis:
Implements a hack to mitigate https://bugs.chromium.org/p/chromium/issues/detail?id=403710. The hack sends black VP8 key frames so that audio playback is unstuck at the client.

UPDATE: This PR depends on https://github.com/jitsi/libjitsi/pull/181. Please do not merge this before the LJ pr is merged. Also, the pom.xml file needs to be updated before merging this.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -297,7 +297,7 @@ public class LipSyncHack
                 = targetVC.getStream().getStreamRTPManager();
 
             ResumableStreamRewriter rewriter = streamRTPManager
-                .getResumableStreamRewriter(acceptedVideoSSRC);
+                .getResumableStreamRewriter(acceptedVideoSSRC, false);
 
             if (rewriter == null)
             {
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -89,6 +89,12 @@ public class LipSyncHack
     private static final boolean DEBUG = logger.isDebugEnabled();
 
     /**
+     * The {@link Random} that will be used to generate the random sequence
+     * number and RTP timestamp offsets.
+     */
+    private static final Random RANDOM = new Random();
+
+    /**
      * The owner of this hack.
      */
     private final Endpoint endpoint;
@@ -282,15 +288,15 @@ public class LipSyncHack
                 // Pretend we have dropped all the packets prior to the one
                 // that's about to be written by the translator.
                 int lastSeqnumDropped = RTPUtils.subtractNumber(seqnum, 1);
+                int highestSeqnumSent = state.getNextSequenceNumber();
                 int seqnumDelta = RTPUtils.subtractNumber(
-                    lastSeqnumDropped, state.numOfKeyframesSent);
+                    lastSeqnumDropped, highestSeqnumSent);
 
                 long timestamp
                     = RawPacket.getTimestamp(buffer, offset, length);
 
                 // Timestamps are calculated.
-                long highestTimestampSent
-                    = state.numOfKeyframesSent * TS_INCREMENT_PER_FRAME;
+                long highestTimestampSent = state.getNextTimestamp();
 
                 // Pretend we have dropped all the packets prior to the one
                 // that's about to be written by the translator.
@@ -300,7 +306,7 @@ public class LipSyncHack
                     = (lastTimestampDropped - highestTimestampSent) & 0xffffffff;
 
                 rewriter = new ResumableStreamRewriter(
-                    state.numOfKeyframesSent, seqnumDelta,
+                    highestSeqnumSent, seqnumDelta,
                     highestTimestampSent, timestampDelta);
 
                 streamRTPManager.ssrcToRewriter.put(acceptedVideoSSRC, rewriter);
@@ -373,11 +379,15 @@ public class LipSyncHack
                     byte[] buf = KEY_FRAME_BUFFER.clone();
                     RawPacket keyframe = new RawPacket(buf, 0, buf.length);
 
-                    long timestamp
-                        = injectState.numOfKeyframesSent * TS_INCREMENT_PER_FRAME;
+                    // Set SSRC.
                     keyframe.setSSRC(injectState.ssrc.intValue());
-                    keyframe.setSequenceNumber(
-                        injectState.numOfKeyframesSent);
+
+                    // Set sequence number.
+                    int seqnum = injectState.getNextSequenceNumber();
+                    keyframe.setSequenceNumber(seqnum);
+
+                    // Set RTP timestamp.
+                    long timestamp = injectState.getNextTimestamp();
                     keyframe.setTimestamp(timestamp);
 
                     if (DEBUG)
@@ -419,9 +429,19 @@ public class LipSyncHack
         private final Long ssrc;
 
         /**
-         * The
+         * The target to inject RTP packets to.
          */
         private final WeakReference<MediaStream> target;
+
+        /**
+         * The random offset for the sequence numbers.
+         */
+        private final int seqnumOffset;
+
+        /**
+         * The random offset for the RTP timestamps.
+         */
+        private final long timestampOffset;
 
         /**
          *
@@ -434,6 +454,31 @@ public class LipSyncHack
         private int numOfKeyframesSent = 0;
 
         /**
+         * Gets the next sequence number to use based on the number of key
+         * frames that have already been sent.
+         *
+         * @return the next sequence number to use based on the number of key
+         * frames that have already been sent.
+         */
+        public int getNextSequenceNumber()
+        {
+            return (seqnumOffset + numOfKeyframesSent) & 0xffff;
+        }
+
+        /**
+         * Gets the next timestamp to use based on the number of key frames
+         * that have already been sent.
+         *
+         * @return Gets the next timestamp to use based on the number of key frames
+         * that have already been sent.
+         */
+        public long getNextTimestamp()
+        {
+            return (timestampOffset
+                + numOfKeyframesSent * TS_INCREMENT_PER_FRAME) & 0xffffffffl;
+        }
+
+        /**
          * Ctor.
          *
          * @param ssrc
@@ -444,6 +489,8 @@ public class LipSyncHack
             this.ssrc = ssrc;
             this.active = active;
             this.target = new WeakReference<>(target);
+            this.seqnumOffset = RANDOM.nextInt(Short.MAX_VALUE + 1);
+            this.timestampOffset = RANDOM.nextInt() & 0xffffffffl;
         }
     }
 }
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -296,8 +296,8 @@ public class LipSyncHack
             StreamRTPManager streamRTPManager
                 = targetVC.getStream().getStreamRTPManager();
 
-            ResumableStreamRewriter rewriter
-                = streamRTPManager.ssrcToRewriter.get(acceptedVideoSSRC);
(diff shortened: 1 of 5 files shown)


brianh5:
is there a way to disable this?

gpolitis:
I will make this configurable, thanks @brianh5 !

brianh5:
thank YOU :)

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -536,8 +536,13 @@ boolean rtpTranslatorWillWrite(
         // XXX(gp) we could potentially move this into a TransformEngine.
         boolean accept = lastNController.isForwarded(source);
 
-        getEndpoint().getLipSyncHack().onRTPTranslatorWillWriteVideo(
-            accept, data, buffer, offset, length, this);
+        LipSyncHack h = getEndpoint().getLipSyncHack();
```
bgrozev: Can we rename this to lipSyncHack?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+
+    /**
+     * The rate (in ms) at which we are to send black key frames.
+     */
+    private static final long KEY_FRAME_RATE_MS = 33;
+
+    /**
+     * Wait for media for WAIT_MS before sending frames.
+     */
+    private final long WAIT_MS = 1000;
```
bgrozev: This could be static.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+        // because in the current architecture this method is called by a single
+        // thread at the time.
+        if (acceptedAudioSSRCs.contains(acceptedAudioSSRC))
+        {
+            // We've already triggered the hack for this audio stream and its
+            // associated video streams.
+            return;
+        }
+
+        acceptedAudioSSRCs.add(acceptedAudioSSRC);
```
bgrozev: Would it make sense to move this below the sanity checks (e.g. right before the for loop), in order to avoid any race condition with a video channel being created and added to its Endpoint? Or is that for the case where an endpoint only has an audio channel?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+            return;
+        }
+
+        VideoChannel sourceVC = (VideoChannel) sourceVideoChannels.get(0);
+        if (sourceVC == null)
+        {
+            return;
+        }
+
+        // FIXME this will include rtx ssrcs and whatnot.
```
bgrozev: If I am reading this correctly, this injects keyframes for each known video SSRC, with each audio packet from an SSRC it hasn't seen before. Is there a reason that the audio SSRC is taken into account and not just the source audio Channel?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+                    scheduledFuture.cancel(true);
+                    return;
+                }
+
+                MediaStream mediaStream = injectState.target.get();
+                if (mediaStream == null || !mediaStream.isStarted())
+                {
+                    logger.debug("Waiting for the media stream to become" +
+                        "available.");
+                    return;
```
bgrozev: Here numOfKeyframesSent has already been incremented, although we didn't send a keyframe. Is that on purpose?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+                }
+
+                try
+                {
+                    // FIXME maybe grab from the write pool and copy the array?
+                    byte[] buf = KEY_FRAME_BUFFER.clone();
+                    RawPacket keyframe = new RawPacket(buf, 0, buf.length);
+
+                    long timestamp = injectState.numOfKeyframesSent * 3000;
```
bgrozev: Is "3000" meant to approximate 90000 / KEY_FRAME_RATE_MS?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+ * https://bugs.chromium.org/p/chromium/issues/detail?id=403710. The hack
+ * injects black video key frames to unstuck the playback of audio for composite
+ * media streams.
+ *
+ * @author George Politis
+ */
+public class LipSyncHack
+{
+    /**
+     * A byte array holding a black key frame.
```
bgrozev: Is that just the VP8 frame? Or a frame encapsulated in RTP?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+                    // FIXME maybe grab from the write pool and copy the array?
+                    byte[] buf = KEY_FRAME_BUFFER.clone();
+                    RawPacket keyframe = new RawPacket(buf, 0, buf.length);
+
+                    long timestamp = injectState.numOfKeyframesSent * 3000;
+                    keyframe.setSSRC(injectState.ssrc.intValue());
+                    keyframe.setSequenceNumber(
+                        injectState.numOfKeyframesSent);
+                    keyframe.setTimestamp(timestamp);
+
```
bgrozev: Assuming KEY_FRAME_BUFFER contains space for the RTP header (and VP8 descriptor), how does the Payload Type value get set? If it is hard-coded in KEY_FRAME_BUFFER, then it's "52", or am I missing something?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+                try
+                {
+                    // FIXME maybe grab from the write pool and copy the array?
+                    byte[] buf = KEY_FRAME_BUFFER.clone();
+                    RawPacket keyframe = new RawPacket(buf, 0, buf.length);
+
+                    long timestamp = injectState.numOfKeyframesSent * 3000;
+                    keyframe.setSSRC(injectState.ssrc.intValue());
+                    keyframe.setSequenceNumber(
+                        injectState.numOfKeyframesSent);
```
bgrozev: This will probably affect the sequence number space used for video packets, if video is enabled later. It should start with a randomly chosen value per the RTP specs.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+                    keyframe.setTimestamp(timestamp);
```
bgrozev: Same here

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+
+                    logger.debug("Injecting black key frame ssrc=" + injectState.ssrc
```
bgrozev: Check for isDebugEnabled()?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+
+        InjectState state;
+        synchronized (states)
+        {
+            state = states.get(acceptedVideoSSRC);
+            if (state == null)
+            {
+                // The hack has never been triggered for this stream.
+                states.put(acceptedVideoSSRC, new InjectState(acceptedVideoSSRC,
+                    ((RtpChannel) target).getStream(), false));
```
bgrozev: There's two casts of 'target'. Maybe cast once in a local variable?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: We create a state object for a given video SSRC only once. So, however many audio SSRCs we receive from the a given audio channel, the SSRCs from the associated video channel are added only once.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: The KEY_FRAME_BUFFER contains the full UDP payload -> the full RTP packet. The payload type is the `KEY_FRAME_BUFFER[1] & 7f` which is equal to 100.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: If video is enabled later, the sequence numbers of the (real) video packets will be re-written (as well as their timestamps). Shouldn't that solve the problem?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
bgrozev: OK, please add documentation for KEY_FRAME_BUFFER

gpolitis: Done :)

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Fixed, thanks :)

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Made this a constant and added some JavaDoc.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: This is the full UDP payload of an RTP packet that contains a VP8 key frame. Added some clarifying JavaDoc.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Added random offsets for sequence numbers and RTP timestamps. Thank you Boris :)

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Done.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Fixed.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Fixed.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -444,6 +489,8 @@ public InjectState(Long ssrc, MediaStream target, boolean active)
             this.ssrc = ssrc;
             this.active = active;
             this.target = new WeakReference<>(target);
+            this.seqnumOffset = RANDOM.nextInt(Short.MAX_VALUE + 1);
```
bgrozev: Short.MAX_VALUE is 2^15, can we use 0xffff (or something with the same value) instead?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -0,0 +1,423 @@
...
+                    = RawPacket.getTimestamp(buffer, offset, length);
+
+                // Timestamps are calculated.
+                long highestTimestampSent = state.numOfKeyframesSent * 3000;
+
+                // Pretend we have dropped all the packets prior to the one
+                // that's about to be written by the translator.
+                long lastTimestampDropped = (timestamp - 3000) & 0xffffffffl;
+                long timestampDelta
+                    = (lastTimestampDropped - highestTimestampSent) & 0xffffffff;
```
bgrozev: The literal needs to be long.

PR #284, opened 2016-08-11
Statistics improvements

bgrozev:

PR #285, opened 2016-08-15
Failed channels stats.

gpolitis:
Exposes the failed channels statistics in colibri stats.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -70,6 +70,12 @@ public class Conference
     private final List<Content> contents = new LinkedList<>();
 
     /**
+     *  A boolean flag that determines whether logging should be enabled for
+     *  this conference.
+     */
+    private final boolean loggingEnabled;
+
+    /**
      * An instance used to save information about the endpoints of this
      * <tt>Conference</tt>, when media recording is enabled.
      */
@@ -231,6 +237,7 @@ public class Conference
         this.focus = focus;
         this.eventAdmin = loggingEnabled ? videobridge.getEventAdmin() : null;
         this.name = name;
+        this.loggingEnabled = loggingEnabled;
 
         if (!loggingEnabled)
         {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -201,6 +201,13 @@
     private Pattern shutdownSourcePattern;
 
     /**
+     * Keeps track of the total number of failed channels. (look at
+     * @{link RtpChannel} for the definition of "failed").
+     */
+    private Map<MediaType, Integer> totalNumberOfFailedChannelsMap
```
bgrozev: Why not just use two atomic integers and remove the synchronization logic?

src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -138,6 +138,18 @@
      */
     public static final String VIDEOSTREAMS = "videostreams";
 
+    /**
+     * The name of the total number of failed audio channels statistic.
+     */
+    private static final String TOTAL_FAILED_AUDIO_CHANNELS
```
bgrozev: Can we use names that fit the convention?

gpolitis:
@paweldomas has a better way of doing some of these things here https://github.com/jitsi/jitsi-videobridge/pull/272/commits/b3bdcd81193af19555e78c925475923393fdde16. Will cherry-pick his commit and adapt my changes to his.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -70,6 +70,12 @@
     private final List<Content> contents = new LinkedList<>();
 
     /**
+     *  A boolean flag that determines whether logging should be enabled for
+     *  this conference.
+     */
+    private final boolean loggingEnabled;
```
bgrozev: This clashes with the implementation in #284 (which implements pretty much the same). Do you prefer to use this version?

PR #286, opened 2016-08-15
Adds statistics about total number of failed/succeeded conferences/channels.

gpolitis:


src/main/java/org/jitsi/videobridge/Channel.java
```diff
@@ -117,6 +117,16 @@
         = new MonotonicAtomicLong();
 
     /**
+     * The time in milliseconds of the last transport related activity to this
```
bgrozev: The doc needs an update

src/main/java/org/jitsi/videobridge/Channel.java
```diff
@@ -127,6 +127,14 @@
         = new MonotonicAtomicLong();
 
     /**
+     * The time in milliseconds of the last payload related activity to this
+     * <tt>Channel</tt>. Currently this means when for the last time there were
+     * any RTP packets seen for this channel.
```
bgrozev: Can we replace "seen" with "received"?

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1193,7 +1193,7 @@ protected void maybeStartStream()
         }
 
         // It seems this Channel is still active.
-        touch(true /* triggered by transport related event */);
+        touch(ActivityType.TRANSPORT /* transport connected */);
```
bgrozev: Why don't we move that to transportConnected()?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -676,11 +690,61 @@ else if (t instanceof ThreadDeath)
...
+                    int[] metrics
+                        = videobridge.getConferenceChannelAndStreamCount();
+
+                    logger.info(
+                        "Expired conference id=" + getID()
+                            + ", conferenceCount="
+                            + metrics[0]
+                            + ", channelCount="
+                            + metrics[1]
+                            + ", streamCount="
```
bgrozev: Can we revert to the original "video streams" here? It has special meaning (I agree that it isn't accurate) and I'd rather just use the term that we already use.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1560,4 +1576,42 @@ private int getContentStreamCount(Content content, int contentChannelCount)
         }
         return contentStreamCount;
     }
+
+    /**
+     * Basic statistics about the videobridge.
+     */
+    public static class Statistics
+    {
+        /**
```
bgrozev: Maybe add docs?

src/main/java/org/jitsi/videobridge/Channel.java
```diff
@@ -800,15 +800,21 @@ public void setTransport(IceUdpTransportPacketExtension transport)
...
+                break;
             case TRANSPORT:
-                touchTransport();
+                lastActivityTime.increase(now);
+                lastTransportActivityTime.increase(now);
                 break;
-            case PAYLOAD:
-                touchPayload();
+            default:
+                lastActivityTime.increase(now);
```
bgrozev: Maybe merge the three lastActivityTime.increase(now) calls outside the "switch"?

bgrozev:
Can we add the new fields to docs/using_statistics.md? Or maybe wait until #284 is merged to avoid dealing with conflicts?

src/main/java/org/jitsi/videobridge/Channel.java
gpolitis: @bgrozev could you please be a little more precise?

gpolitis:
I think we would like to have this merged as soon as possible to start gathering data. Would you mind merging this if I push a commit to https://github.com/jitsi/jitsi-videobridge/pull/284 to update the statistics documentation?

src/main/java/org/jitsi/videobridge/Channel.java
jitsi-developers: After the latest commits the "transport" activity does not include rtp, but
the doc states that it does. I may be missing something, will check again
when I have a computer.

Boris

gpolitis: Are you talking about the change in the switch statement?

bgrozev: No, I mis-read the code.

bgrozev:
> I think we would like to have this merged as soon as possible to start gathering data. Would you mind merging this if I push a commit to #284 to update the statistics documentation?

Yeah, sure

PR #287, opened 2016-08-16
Do not use WeakReference on Endpoint and Conference

paweldomas:
Both Conference and Endpoint have "expired" properties which can be used to determine if instances are valid, so using WeakReference only adds extra complexity.

I need these changes for another work which will require to store expired Endpoint instances until Conference is not expired and it may break logic which depends on WeakReferences<Endpoint>.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
+++ b/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
@@ -185,7 +185,7 @@ public class ConferenceSpeechActivity
      * The <tt>Endpoint</tt> which is the dominant speaker in
      * {@link #conference}.
      */
-    private WeakReference<Endpoint> dominantEndpoint;
+    private Endpoint dominantEndpoint;
 
     /**
      * The indicator which signals to {@link #eventDispatcher} that
@@ -205,7 +205,7 @@ public class ConferenceSpeechActivity
      * {@link #conference} with the dominant (speaker) <tt>Endpoint</tt> at the
      * beginning of the list i.e. the dominant speaker history.
      */
-    private List<WeakReference<Endpoint>> endpoints;
+    private List<Endpoint> endpoints;
 
     /**
      * The indicator which signals to {@link #eventDispatcher} that the
@@ -306,7 +306,7 @@ public class ConferenceSpeechActivity
 
                     if (!endpoint.equals(dominantEndpoint))
                     {
-                        this.dominantEndpoint = new WeakReference<>(endpoint);
+                        this.dominantEndpoint = endpoint;
                         maybeStartEventDispatcher = true;
                     }
                 }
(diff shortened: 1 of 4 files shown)


gpolitis:
The comment at line Conference.java:557 is no longer valid.

gpolitis:
I'm wondering why did we even have weakreferences to endpoints in the first place. There must have been a reason, that's lost by now, and I'm worried that we're not taking that into account.

gpolitis:
I guess we just weren't using the Endpoint.expired field as we should. The functional changes look good to me, I'll merge this after the very few editorial fixes.

paweldomas:
So @lyubomir said that at the time when this code was written there was no "expired" property on Endpoint

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -874,7 +876,7 @@ else if (conferenceEndpoints.contains(endpoint))
                 if ((dominantEndpointIndex != -1)
                         && (dominantEndpointIndex != 0))
                 {
-                    WeakReference<Endpoint> weakReference
+                    Endpoint weakReference
```
gpolitis: Maybe rename this to something else.

paweldomas: Oh yes, sure I have missed that one

gpolitis:
> So @lyubomir said that at the time when this code was written there was no "expired" property on Endpoint

got it, thanks @paweldomas 👍

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -276,7 +278,7 @@ private void activeSpeakerChanged(long ssrc)
     {
         Conference conference = getConference();
 
-        if ((conference != null) && !conference.isExpired())
+        if (conference != null)
```
gpolitis: Do we really not want the isExpired check here?

gpolitis: I'm sorry, please disregard this, I just noticed that you're taking care of that in the `getConference` method.

paweldomas: getConference() returning null equals conference has expired, but if you think it's not readable I will revert that.

paweldomas: Ok

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -551,8 +555,8 @@ public Endpoint getDominantEndpoint()
             }
```
gpolitis: Here we can now remove the  `if (this.dominantEndpoint == null)` clause.

paweldomas: Are you sure ? It can be null. Is it ok to remove and add null ?

gpolitis: Sorry about that, I misread the code. We need that there!

PR #289, opened 2016-08-17
added possibility to generate unique source (SSRC) for channel

sidorovis:
added generate_sources boolean parameter as channel setting - so jitsi videobridge will generate unique SSRC for it

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -167,6 +167,9 @@ public class Conference
      */
     private final Map<String, IceUdpTransportManager> transportManagers
         = new HashMap<>();
+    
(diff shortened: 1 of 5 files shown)


sidorovis:
I don't have Attlasian code and don't plan to use it.

gpolitis:
Hi @sidorovis, could you please help us understand what are you trying to achieve with this? i.e. why do you need to generate unique ssrcs for the video channel.

sidorovis:
@gpolitis we had case when we can't define SSRC on client side, because amount of users in  conference is not predefined. So users could connect to conference without any SSRC and receive that from conference server side.

gpolitis:
I think I would prefer to see this functionality in the signaling server (for example, Jicofo).

PR #296, opened 2016-08-24
fix: Fixes a bug reported by @aaronkvanmeerten

bgrozev:
Only expires channels once. Fixes the values reported in statistics.


src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1833,8 +1833,14 @@ else if (second == null)
      * assumes the responsibility of releasing its resources.
      */
     @Override
-    public void expire()
+    public boolean expire()
```
paweldomas: Please add missing docs for return statement. Eventually {@inheritDocs} may work as well

PR #299, opened 2016-08-29
Stream switching

bgrozev:
Sends keyframe requests preemptively when an endpoint becomes the dominant speaker, in anticipation of receivers selecting that endpoint.


brianh5:
last-n already has code for this i think, right? does that mean it needs to be removed from the last-n logic or else 2 idrs will get requested?

bgrozev:
RTCPFeedbackMessageSender manages the actual requests that are sent, and should take care of not sending them unless necessary.

PR #300, opened 2016-08-29
Respect last-n when adding pinned endpoints...

bgrozev:
...to the list of forwarded endpoints.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/LastNController.java
+++ b/src/main/java/org/jitsi/videobridge/LastNController.java
@@ -455,11 +455,12 @@ public class LastNController
      * @param newConferenceEndpoints A list of endpoints which entered the
      * conference since the last call to this method. They need not be asked
      * for keyframes, because they were never filtered by this
-     * {@link #LastNController(VideoChannel)}.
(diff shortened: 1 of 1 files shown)


brianh5:
@bgrozev i'm curious, is there something that breaks by letting pin go beyond the last-n list?  i know the call on what to do there has been a question for a while (let it go over or don't), and we don't currently use it, but it didn't seem like the worst thing to allow for an explicit override.  was there something that finally pushed the decision to limit it no matter what?

bgrozev:
Not sure if there was something in particular that triggered a decision. We were thinking of how this will play with adaptive last N (or changes to last-n from the bridge), and this seemed like the correct decisions, since the bridge explicitly signals to the client which endpoints it is currently forwarding anyway.
Do you see a use-case where honoring the client's request is an advantage?

brianh5:
@bgrozev basically just to allow the client to say "i know i told you a last-n of 3 when i joined, but now i'm telling you i want to pin 4 clients, trust me".  no passionate feelings about it being one way or another right now, was just curious if there was a specific reason to limit it in case we ever tinker with letting a client go over--wanted to know if there's anything major relying on it.

PR #301, opened 2016-08-31
Sync fixes

bgrozev:


src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -153,7 +153,8 @@ private static synchronized int generateDebugId()
     /**
      * Data channels mapped by SCTP stream identified(sid).
      */
-    private final Map<Integer,WebRtcDataStream> channels = new HashMap<>();
+    private final Map<Integer,WebRtcDataStream> channels
+        = new ConcurrentHashMap<>();
```
lyubomir: I see in the source code of SctpConnection that the access to the field channels is synchronized on the field syncRoot. Did you see an inconsistency that prompted the necessity of ConcurrentHashMap?

src/main/java/org/jitsi/videobridge/ratecontrol/LastNBitrateController.java
```diff
@@ -573,7 +573,7 @@ private void add(long time, long rate)
         /**
          * Removes values added before <tt>time - period</tt>.
          */
-        private void clean(long time)
+        private synchronized void clean(long time)
```
lyubomir: The synchronization seems inconsistent to me because the method clean modifies both receivedRembs.size() and sum which are both read from the unsynchronized method getAverage.

On a side note, I find the implementation of the clean method to be inefficient and more complex than really necessary:
- toRemove is an unnecessary collection that takes up instance space, grows and shrinks;
- toRemove retains elements even after it has lived up its purpose;
- Because of toRemove, more auto(un)boxing happens;
- receivedRembs is searched upon removal.
I believe a solution based on an Iterator over receivedRembs.entrySet() will be more efficient and the source code will be easier to read.

src/main/java/org/jitsi/videobridge/SctpConnection.java
bgrozev: No, removing the commit (or rather, not including it in the new version of the PR).

src/main/java/org/jitsi/videobridge/ratecontrol/LastNBitrateController.java
bgrozev: Thanks Lyubo! I don't consider the implementation of this class complete. I'll add your specific notes as comments in the code, to be addressed when we next work on it.

bgrozev:
Moved to #305

PR #304, opened 2016-09-06
Enable lipsync by default

paweldomas:


brianh5:
will this being enabled break anything for users who don't need it?  we can add something to the config to disable it, but we only update every so often and i may forget when the time comes around.  just curious what the behavior will be if this is enabled but mixing is being used.

paweldomas:
@brianh5 sorry about that, but we need that change in the master. If you add this config option now, then older JVB will just ignore it until updated. Does that make sense ?

I can not tell about mixing. Maybe @gpolitis will know more

PR #307, opened 2016-09-09
feat: Uses more buckets for conference size statistics.

bgrozev:


gjaekel:
You forgot to update the docs, see #1550.

PR #310, opened 2016-09-13
fix: Increases the delay before a keyframe request.

bgrozev:


brianh5:
@bgrozev i was talking with @gpolitis a bit about this...i was wondering why add the extra delay and he reminded me in non-last-n there's an extra round trip to wait for the 'selectedEndpoint' even to come around, hence why IDRs may be coming too early.  in last-n though we don't have that extra roundtrip, and really want the IDRs as quickly as possible.  maybe we could add a check for a last-n value and, if present, minimize the delay?

bgrozev:
@brianh5 this doesn't affect the last-n logic. For last-n, FIRs will still be sent as soon as the dominant speaker changes through the code here: https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/Conference.java#L1672

brianh5:
ah great, thanks @bgrozev.  so does that mean in last-n we'll double request the IDR (once due to the code you just cited, and again due to the code in the diff)?  or is there some throttling for requests on the bridge side that will prevent that?  what's that value set to?

bgrozev:
Yes, there is a possibility for a double request, but it should happen very rarely. Both use cases (the last-n one, and the one with the delay for simulcast) go through RTCPFeedbackMessageSender#sendFIR(), which will throttle the FIRs to at most one every 300ms if I am reading the code correctly.

A double request will happen if the delay is big:
max(receiver_rtts) - sender_rtt + 10ms  > 300ms

and also the keyframe arrives in less than 300ms.

brianh5:
sounds good.  thanks boris.

PR #316, opened 2016-09-25
Rewriting refactoring

gpolitis:


src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1991,6 +1968,121 @@ public RtpChannelTransformEngine getTransformEngine()
...
+        synchronized (remoteTracks)
+        {
+            remoteTracks.clear();
+            remoteTracks.putAll(tracksBySSRC);
+
+            for (Long primarySSRC : freeSSRCs)
+            {
+                MediaStreamTrack mst = new MediaStreamTrack();
+                mst.addEncoding(primarySSRC, -1, -1, RTPEncoding.BASE_ORDER);
+                remoteTracks.put(primarySSRC, mst);
```
lyubomir: That's murky business. I saw that you exposed the private field/state of `MediaStreamImpl` through `#getRemoteTracks()` and I let it be because I didn't want to stop your libjisti PR. However, this is a bit too much for me to swallow without mentioning it at least. You get the Map from `stream` without being 100% sure as a client that the implementation is not a copy of the private state, then you synchronize on it, and then you go an modify it hoping that you're modifying the private state of `MediaStreamImpl`!

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1991,6 +1968,121 @@ public RtpChannelTransformEngine getTransformEngine()
...
+                        long[] encoding = new long[] { -1, -1, -1 };
+                        encoding[0] = primarySSRC;
+                        encoding[1] = order++;
+
+                        encodings.put(primarySSRC, encoding);
+                        tracksBySSRC.put(primarySSRC, track);
+                    }
+                }
+                else if ("fid".equalsIgnoreCase(sgpe.getSemantics())
+                    && sgpe.getSources() != null
```
lyubomir: `sgpe.getSources() != null` is checked in both `if`s. That's an unnecessary both as a runtime operation and as source code to be read.

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1991,6 +1968,121 @@ public RtpChannelTransformEngine getTransformEngine()
...
+                    && sgpe.getSources().size() == 2)
+                {
+                    Long primarySSRC = sgpe.getSources().get(0).getSSRC();
+                    freeSSRCs.remove(primarySSRC);
+                    Long rtxSSRC = sgpe.getSources().get(1).getSSRC();
```
lyubomir: I counted 6 occurrences of `sgpe.getSources()` in this loop. I believe a local variable will make the source code easier to read.

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1991,6 +1968,121 @@ public RtpChannelTransformEngine getTransformEngine()
...
+
+                    int order = RTPEncoding.BASE_ORDER;
+                    for (SourcePacketExtension spe : sgpe.getSources())
+                    {
+                        long primarySSRC = spe.getSSRC();
+                        freeSSRCs.remove(primarySSRC);
+
+                        long[] encoding = new long[] { -1, -1, -1 };
+                        encoding[0] = primarySSRC;
+                        encoding[1] = order++;
```
lyubomir: Ugh, doesn't Java allow you to do the initialization without constants? For example,

```` java
long[] encoding = new long[] { primarySSRC, order++, -1 };
````

I mean, there are magic indices, `-1` that get immediately overwritten.... Even the following would be better

```` java
long[] encoding = new long[3];
encoding[0] = primarySSRC;
encoding[1] = order++;
encoding[2] = -1;
````

File: src/main/java/org/jitsi/videobridge/RtpChannel.java [hunk 5]
```diff
@@ -1991,6 +1968,121 @@ public RtpChannelTransformEngine getTransformEngine()
...
+
+        this.setSources(sources); // TODO remove and rely on MSTs.
+        this.setSourceGroups(sourceGroups); // TODO remove and rely on MSTs.
+
+        Map<Long, MediaStreamTrack> tracksBySSRC = new TreeMap<>();
+        Map<Long, long[]> encodings = new TreeMap<>();
+
+        List<Long> freeSSRCs = new ArrayList<>();
+
+        if (sources != null && sources.size() != 0)
```
lyubomir: I find `!sources.isEmpty()` clearer than `sources.size() != 0`.

lyubomir: And why do we have to write these checks twice. Cannot we write them once and use local variables?

src/main/java/org/jitsi/videobridge/RtpChannel.java
gpolitis: Thank you Lyubo. The idea here is that the signaling manages the map, as you suggested in the LJ pr. But how would you like to change this?

src/main/java/org/jitsi/videobridge/RtpChannel.java
gpolitis: Fixed, thank you Lyubo!

src/main/java/org/jitsi/videobridge/RtpChannel.java
gpolitis: Fixed, thank you Lyubo!

src/main/java/org/jitsi/videobridge/RtpChannel.java
gpolitis: Fixed, thank you Lyubo!

src/main/java/org/jitsi/videobridge/RtpChannel.java
gpolitis: Fixed, thank you Lyubo!

gpolitis: Fixed, thank you Lyubo!

Issue #178, opened 2016-03-20
Jitsi videobridge plugin for openfire is broken


deleolajide:
Followed the notes in build.xml

    

STEP 3 is BROKEN

````
 [of.javac] C:\Projects\ignite\openfire\src\plugins\jitsivideobridge\src\java\org\jitsi\videobridge\openfire\PluginImpl.java:149: error: constructor ComponentImpl in class ComponentImpl cannot be applied to given types;
 [of.javac]         Component component = new ComponentImpl();
 [of.javac]                               ^
 [of.javac]   required: String,int,String,String,String
 [of.javac]   found: no arguments
 [of.javac]   reason: actual and formal argument lists differ in length
 [of.javac] 1 error
     [echo] Error building plugin: jitsivideobridge. Exception:
````

bgrozev:
The plugin for openfire is not maintained here. It looks like the plugin hasn't been updated after [a change](https://github.com/jitsi/jitsi-videobridge/commit/8e37fd) in jitsi-videobridge over 5 months ago.

bgrozev:
My bad! I wasn't aware that it's here. Apologies.

thedjnK:
To add to this, the jar files on https://download.jitsi.org/jitsi-videobridge/openfire/ have apparently been created in june but are also broken. They can be installed but attempting to uninstall them from the openfire web interface fails. They continually error as can be seen in http://pastebin.com/uHVEaF98 and the title in in the server tab is 'plugin.title' - this page is just a broken table.

guusdk:
Most of the issues described here should be fixed by #328 and #330. Could you guys retest?

deleolajide:
@guusdk thanks for the PR. I have done a retest and can confirm build now works and plugin gets built ok. I have tested the jitsivideobridge in Openfire 4.0.3, it loads ok and I can see heartbeats to the jitsi-videobridge component, but there are some errors at startup

````
2016.10.21 21:27:22 org.jitsi.impl.neomedia.device.DeviceSystem - Failed to initialize org.jitsi.impl.neomedia.device.PortAudioSystem
java.lang.reflect.UndeclaredThrowableException
	at org.jitsi.impl.neomedia.device.DeviceSystem.invokeDeviceSystemInitialize(DeviceSystem.java:441)
	at org.jitsi.impl.neomedia.device.DeviceSystem.invokeDeviceSystemInitialize(DeviceSystem.java:348)
[...]

PR #328, opened 2016-10-17
Addressing various Openfire plugin issues

guusdk:
The changes in this pull request address various issues (including, but not limited to #178) with the Openfire plugin, that is part of the Jitsi-videobridge project.

I've tried to address each issue (which range from bugs to minor improvement) in separate and concise commits.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -347,8 +347,21 @@
         <copy todir="${dist}/openfire/jitsiVideobridge/lib">
             <fileset dir="lib">
                 <exclude name="native/**"/>
-                <exclude name="bouncycastle.jar"/>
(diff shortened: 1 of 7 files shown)


build.xml
```diff
@@ -347,8 +347,19 @@
         <copy todir="${dist}/openfire/jitsiVideobridge/lib">
             <fileset dir="lib">
                 <exclude name="native/**"/>
-                <exclude name="bouncycastle.jar"/>
+                <exclude name="bc*.jar"/>
```
damencho: I think this will also exclude bccontrib-....jar which is needed when running jitsi videobridge.

damencho:
Jenkins, add to whitelist.

PR #333, opened 2016-10-18
Updates callstats to 4.0.0 and add reports for packet loss.

damencho:


src/main/java/org/jitsi/videobridge/eventadmin/callstats/CallStatsConferenceStatsHandler.java
```diff
@@ -340,6 +340,9 @@ private void processChannelStats(RtpChannel channel)
                     = new ConferenceStatsBuilder()
                         .bytesSent(receiveStat.getBytes())
                         .packetsSent(receiveStat.getPackets())
+                        .packetsLost(receiveStat.getPacketsLost())
+                        .fractionalPacketLost(
+                            receiveStat.getCurrentPacketsLost())
```
bgrozev: This seems weird -- I would expect fractionalPacketLost to expect a fraction (although I haven't checked) and we're reporting a number of packets.

Issue #342, opened 2016-10-21
Exceptions should not be completely ignored


guusdk:
When an exception occurs, it should not be ignored completely. If handling is not needed, the exception should be logged (if only at the lowest of log levels). This greatly reduces confusion while debugging a situation where a bug occurs caused by an exception that the original developer believed was "impossible".

PR #343, opened 2016-10-21
Exceptions should not be completely ignored

guusdk:
This is a fix for https://github.com/jitsi/jitsi-videobridge/issues/342

When an exception occurs, it should not be ignored completely. If handling is not needed, the exception should be logged (if only at the lowest of log levels). This greatly reduces confusion while debugging a situation where a bug occurs caused by an exception that the original developer believed was "impossible".

### DIFF
--- a/src/main/java/org/jitsi/videobridge/stats/PubSubStatsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/stats/PubSubStatsTransport.java
@@ -322,15 +322,8 @@ public class PubSubStatsTransport
                     service
                         = bundleContext.getService(ev.getServiceReference());
                 }
-                catch (IllegalArgumentException ex)
-                {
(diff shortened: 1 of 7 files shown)


src/main/java/org/jitsi/videobridge/stats/PubSubStatsTransport.java
```diff
@@ -324,12 +324,15 @@ private void serviceChanged(ServiceEvent ev)
                 }
                 catch (IllegalArgumentException ex)
                 {
+                    logger.debug( "An unexpected exception occurred.", ex );
```
bgrozev: Can you please collapse these three catch block into one with the java 7 "|" syntax?

guusdk:
I fixed the one you referred, as well as a couple of others that were in the code.

bgrozev:
Thanks! Will merge after the tests pass.

Issue #347, opened 2016-10-24
Property name constants should be publicly accessible


guusdk:
When using a constant to refer to a property, this is typically done to avoid typos and add configuration. Keeping such constants non-public is counterproductive, as it forces people to revert to manually typing the property name (and typically hides javadoc).

Constants that are used to store property names should be publicly accessible.

PR #348, opened 2016-10-24
Property name constants should be publicly accessible.

guusdk:
This is a fix for https://github.com/jitsi/jitsi-videobridge/issues/347

When using a constant to refer to a property, this is typically done to avoid typos and add configuration. Keeping such constants non-public is counterproductive, as it forces people to revert to manually typing the property name (and typically hides javadoc).

Constants that are used to store property names should be publicly accessible.


bgrozev:
Thank you for all the contributions, @guusdk !

guusdk:
Thanks for all the merges @bgrozev !

PR #349, opened 2016-10-24
fix(ls-hack): Sends 3 waves of black key frames.

gpolitis:
Sends 3 waves of black key frames, each separated by 1 second, to make
sure that at least one black key frame reaches the receiver after
signaling has propagated.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -17,6 +17,7 @@ package org.jitsi.videobridge;
 
 import net.sf.fmj.media.rtp.*;
 import org.jitsi.impl.neomedia.*;
+import org.jitsi.impl.neomedia.rtcp.*;
 import org.jitsi.impl.neomedia.rtp.*;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.util.*;
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -190,8 +190,11 @@ public class LipSyncHack
         }
 
         VideoChannel targetVC = (VideoChannel) targetVideoChannels.get(0);
-        if (targetVC == null)
+        MediaStream stream;
+        if (targetVC == null || (stream = targetVC.getStream()) == null
+            || !stream.isStarted())
         {
+            // It seems like we're not ready yet to trigger the hack.
             return;
         }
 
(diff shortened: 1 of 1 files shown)


bgrozev:
This was not introduced with this PR but needs fixing: the catch block on line 494 decrements injectState.numOfKeyframesSent, but the increment itself is inside the try block.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -497,12 +498,33 @@ public void run()
...
+        public long getTimeUntilNextRun()
         {
-            this.scheduledFuture = scheduler.scheduleAtFixedRate(
-                this, WAIT_MS , KEY_FRAME_RATE_MS, TimeUnit.MILLISECONDS);
+            synchronized (injectState)
+            {
+                if (!injectState.active
+                    || injectState.numOfKeyframesSent >= MAX_KEY_FRAMES)
+                {
+                    return -1L;
```
bgrozev: It is not clear from the documentation of RecurringRunnable what the semantics of returning -1 are, but from a quick look at the code it looks like this would make RecurringRunnableExecutor run the task perpetually.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -497,12 +498,33 @@ public void run()
...
+                }
+
+                // send 3 "waves" of 10 packet bursts every 1 second. This makes
+                // sure that the signaling has propagated and that eventually a
+                // packet gets through SRTP.
+                long delay = 0;
+                if (injectState.numOfKeyframesSent % 10 == 0)
```
bgrozev: RecurringRunnableExecutor calls getTimeUntilNextRun() before it calls run() and injectState.numOfKeyframesSent is initialized to 0. Wouldn't this cause an initial delay of WAIT_MS?

bgrozev: Nevermind, it wouldn't because of the clause below. Can we check "lastRunTime < 0L" in the beginning of the method and return 0 immediately, though?

gpolitis: I want to wait for an initial WAIT_MS, so the fact that we don't is a bug. Thanks for catching it.

gpolitis: The idea for waiting an initial WAIT_MS was to wait for actual video packets from the sender, but we risk of clipping the first few seconds of audio, in case the sender is video muted. I think I’m going to remove this initial wait.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -21,10 +21,10 @@
 import org.jitsi.impl.neomedia.rtp.*;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.util.*;
+import org.jitsi.util.concurrent.*;
 
 import java.lang.ref.*;
```
bgrozev: This is now an unused import, can you please remove it?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -497,12 +524,58 @@ public void run()
...
+         * De-registers this {@code RecurringRunnable} and optionally prints a
+         * debug message.
+         *
+         * @param reason the de-registration reason
+         */
+        private void deregister(String reason)
+        {
+            if (DEBUG)
+            {
+                logger.debug("ls_hack_dererister"
```
bgrozev: s/dererister/deregister /

bgrozev:
Jenkins: ok to test

PR #352, opened 2016-11-01
Process XMPP packets in order

paweldomas:
Updates Jicoco in order to use single thread packet processor. DO NOT MERGE(Jicoco is referenced by the SNAPSHOT version)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jicoco</artifactId>
-      <version>1.0-20160822.191856-34</version>
+      <version>1.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -79,7 +79,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jicoco</artifactId>
-      <version>1.0-20160822.191856-34</version>
+      <version>1.0-SNAPSHOT</version>
```
bgrozev: Is this intentionally SNAPSHOT and not just the new version?

paweldomas: yes, for now, Jicoco needs to be merged first

bgrozev:
Jenkins: ok to test

PR #353, opened 2016-11-02
fix(ls-hack): Prevents a swallowed NPE when simulcast is not enabled.

gpolitis:


paweldomas:
Where is the NPE swallowed ?

gpolitis:
@paweldomas In AbstractRTPTranslator:227. We should print an error there.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -210,17 +211,37 @@ public void onRTPTranslatorWillWriteAudio(
...
         }
 
+        SimulcastReceiver recv = sourceVC.getTransformEngine()
+            .getSimulcastEngine().getSimulcastReceiver();
+
+        Long receiveVideoSSRC;
+        if (recv != null && recv.isSimulcastSignaled())
+        {
+            // FIXME this is a little ugly
+            receiveVideoSSRC = sourceVC.getTransformEngine()
```
paweldomas: you have "SimulcastReceiver recv" var here already or am I missing something ?

paweldomas: the whole logic in the spaghetti here fells like a candidate for a method in the VideoChannel, no ?

gpolitis: Thanks for the suggestion Pawel, I really appreciate it and I agree with you that this looks ugly. I plan to simplify it like this https://github.com/jitsi/jitsi-videobridge/blob/cc/src/main/java/org/jitsi/videobridge/LipSyncHack.java#L213 (please do not pay attention to the unchecked array accesses as the code is not finished yet).

PR #356, opened 2016-11-04
logd() should log to debug, not info.

guusdk:
The current implementation causes all XMPP traffic to be printed to INFO, instead of the intended DEBUG level. Verbosity should reduced.

### DIFF
--- a/lib/logging.properties
+++ b/lib/logging.properties
@@ -8,7 +8,7 @@ net.java.sip.communicator.util.ScLogFormatter.programname=JVB
 
 .level=INFO
 
-org.jitsi.videobridge.xmpp.ComponentImpl.level=DEBUG
+org.jitsi.videobridge.xmpp.ComponentImpl.level=FINE
 
 # All of the INFO level logs from MediaStreamImpl are unnecessary in the context of jitsi-videobridge.
 org.jitsi.impl.neomedia.MediaStreamImpl.level=WARNING
--- a/lib/logging.properties
+++ b/lib/logging.properties
@@ -8,6 +8,8 @@ net.java.sip.communicator.util.ScLogFormatter.programname=JVB
 
 .level=INFO
 
+org.jitsi.videobridge.xmpp.ComponentImpl.level=DEBUG
+
 # All of the INFO level logs from MediaStreamImpl are unnecessary in the context of jitsi-videobridge.
 org.jitsi.impl.neomedia.MediaStreamImpl.level=WARNING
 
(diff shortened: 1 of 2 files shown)


guusdk:
The `logd()` method accepts a String only, which suggests that the message has already been compiled. It'd make somewhat sense if that method was overloaded to allow for objects to be passed (but even then, I say it should be the responsibility of the underlying framework to optimize stuff like that).

damencho:
I mean just:
        if (logger.isDebugEnabled())
            logger.debug(s);
I meant : http://www.slf4j.org/api/org/slf4j/bridge/SLF4JBridgeHandler.html
The statement there:
Please note that translating a java.util.logging event into SLF4J incurs the cost of constructing LogRecord instance regardless of whether the SLF4J logger is disabled for the given level. Consequently, j.u.l. to SLF4J translation can seriously increase the cost of disabled logging statements (60 fold or 6000% increase) and measurably impact the performance of enabled log statements (20% overall increase).

damencho:
We normally do not put such statements for info level, cause its default and normally is always enabled. But put those if for trace and debug.

bgrozev:
We've discussed the usage of "if (debugEnabled)" before, and have agreed to always use it for debug logs to avoid unnecessary evaluation of the arguments. So if we want to move these logs to DEBUG level, I would prefer to get rid of the logd helper (especially given all the toXML() calls).

That being said, it might have been the original intention to log them at DEBUG, but I think that nowadays we want the logs at INFO, because they are often very helpful.

guusdk:
I modified the `logd` statement, it now checks if debug is enabled.

The XMPP data that's being transmitted to and from the component is very verbose. A lot of the data are XMPP pings and their responses. Other XMPP data is only of interest to people that are actively debugging the system. For day to day operation, logging XMPP data isn't that useful. I suggest keeping that on debug, as you'll drown other, more useful, log statements otherwise.

bgrozev:
> I modified the logd statement, it now checks if debug is enabled.

This way the arguments passed to logd are evaluated unconditionally.

> The XMPP data that's being transmitted to and from the component is very verbose. A lot of the data are XMPP pings and their responses. Other XMPP data is only of interest to people that are actively debugging the system. For day to day operation, logging XMPP data isn't that useful. I suggest keeping that on debug, as you'll drown other, more useful, log statements otherwise.

I disagree. These logs provide very useful context for the rest of the logs, and we've used them multiple times in the past to analyze issues retrospectively (not while debugging).

I don't have a problem with moving these to DEBUG and changing the default level (in the logging.properties file) for ComponentImpl to DEBUG. Would that work for your purposes?

guusdk:
Like this?

bgrozev:
Actually it should be "FINE", not "DEBUG" (just tested to see if both work -- they don't), sorry for misleading.

guusdk:
Fine! ;)

bgrozev:
Thank you!

PR #363, opened 2016-11-17
Fix: Update last-n on dominant speaker change

dazaar:
Clients were not receiving updates to the last-n endpoints when the
active speaker changed. Now, the LastNController for each
VideoChannel should be notified as active speaker updates and will
push out last-n change events, if necessary.


dwilson6:
dazaar should be added to our CLA.

bgrozev:
Thanks you for the contribution, @dazaar !

Jenkins, ok to test

PR #370, opened 2016-12-02
Provide more info when the data channel transfer fails

paweldomas:
More details about the message and Endpoint will be logged when SCTP transfer fails. Also SCTP stack events will be logged.


bgrozev:
Jenkins ok to test

PR #372, opened 2016-12-09
stats: Adds the number of messages sent and received over data channels

bgrozev:
to the statistics.


gpolitis:
jenkins it's ok to test

PR #373, opened 2016-12-09
ice4j VSL

bgrozev:


src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -1158,8 +1158,11 @@ public void run()
         }
         catch (SocketException ex)
         {
-            if (!"Socket closed".equals(ex.getMessage()))
+            if (!"Socket closed".equals(ex.getMessage())
```
bbaldino: the string comparison seems like a bit of a bummer.  is it possible other exceptions (which we don't know the types of) might throw that same error message?

bgrozev: But we are only catching SocketException-s in this block. Am I missing something?

bbaldino: i'm just wondering why we have to do the string comparison and why the 'instanceof' check isn't enough?

bgrozev: The jdk sockets throw a vanilla SocketException with that string, and this is how we've been recognizing it until now, which is awful. The SocketClosedException is just something I added in ice4j, but it is only thrown from MergingDatagramSocket right now. I guess we could make all of the ice4j sockets throw this, and even recognize the string there and re-throw, but I think if we go that way it should be a separate effort.

bbaldino: ok gotcha, sounds like we're stuck then. thanks for the background.

bgrozev:
jenkins ok to test

PR #379, opened 2016-12-20
logs: Uses a strudcured format for some of the log messages.

bgrozev:


davidertel:
My only suggestion would be to expand the variable names in the logs instead of abbreviating them. So conf_id would be conference_id and ch_id would be channel_id etc. The advantage of expanding them to the full words is that they become pretty much self documenting and it removes any ambiguity as to what they refer to. They logging improvement will be a big help in debugging multiple conferences.

gpolitis:
I tend to prefer the shorter names, but I would let it up to @bgrozev to make the final decision.

bgrozev:
Yes, let's stick to shorter names. I'll add documentation later on.

PR #380, opened 2016-12-20
fix: Request keyframes from last-n endpoints for newly joined video channels

dazaar:
As users join a meeting with last-n enabled they need to request
keyframes from the currently forwarded endpoints in order to
begin receiving video. Prior to this change a user joining the
meeting, that did not result in an update to the last-n endpoints,
would not receive keyframes for the forwarded endpoints. The
result was the newly joined user being unable to display video
until a dominant speaker event, pinned endpoint event, or
the next keyframe interval occurred.


gpolitis:
@dazaar we have rewritten the lastN implementation and this should be fixed now as part of that.

gpolitis:
(Please open an issue or a new PR if this is still broken for you.)

PR #382, opened 2016-12-21
Sctp close fix

bgrozev:


src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -1127,7 +1129,11 @@ public void run()
                 if (send == null || send.length == 0)
                     continue;
 
-                if(LOG_SCTP_PACKETS)
+                // We received data for the SCTP socket, this SctpConnection
+                // is still alive
+                touch(ActivityType.PAYLOAD);
```
bbaldino: to make sure i'm understanding this right...previously data channel messages weren't counting as channel activity?  and the stun checks counting as channel activity is handled elsewhere i guess?

bgrozev: Yeah, exactly: https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java#L1591

PR #383, opened 2016-12-22
fix: Prepend the video stream with black VP8 kfs.

gpolitis:
If the lip sync hack is enabled, always send black VP8 kfs before the
start of a stream.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -502,12 +502,17 @@ public class LipSyncHack
             {
                 ssrcsWithoutBlackKeyframes.remove(ssrc);
 
+                boolean isSOF = channel.getStream().isStartOfFrame(
+                    pkts[i].getBuffer(), pkts[i].getOffset(), pkts[i].getLength());
+
+                int sofDistance = isSOF ? 0 : 10;
+
                 int seqNum = pkts[i].getSequenceNumber();
                 long ts = pkts[i].getTimestamp();
                 RawPacket[] kfs = new RawPacket[MAX_KEY_FRAMES];
                 for (int j = 0; j < kfs.length; j++)
                 {
-                    int relativeIdx = j - kfs.length;
+                    int relativeIdx = j - kfs.length - sofDistance;
                     byte[] buf = KEY_FRAME_BUFFER.clone();
                     RawPacket kf = new RawPacket(buf, 0, buf.length);
 
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -505,7 +505,7 @@ public class LipSyncHack
                 RawPacket[] kfs = new RawPacket[MAX_KEY_FRAMES];
                 for (int j = 0; j < kfs.length; j++)
                 {
-                    int relativeIdx = (j - 1 - kfs.length);
+                    int relativeIdx = j - kfs.length;
                     byte[] buf = KEY_FRAME_BUFFER.clone();
                     RawPacket kf = new RawPacket(buf, 0, buf.length);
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -434,6 +445,107 @@ public void onRTPTranslatorWillWriteVideo(
...
+            {
+                ssrcsWithoutBlackKeyframes.remove(ssrc);
+
+                int seqNum = pkts[i].getSequenceNumber();
+                long ts = pkts[i].getTimestamp();
+                RawPacket[] kfs = new RawPacket[MAX_KEY_FRAMES];
+                for (int j = 0; j < kfs.length; j++)
+                {
+                    int relativeIdx = (j - 1 - kfs.length);
+                    byte[] buf = KEY_FRAME_BUFFER.clone();
```
bbaldino: is KEY_FRAME_BUFFER a static instance of a black key frame?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -434,6 +445,107 @@ public void onRTPTranslatorWillWriteVideo(
...
+     * {@inheritDoc}
+     */
+    @Override
+    public RawPacket[] reverseTransform(RawPacket[] pkts)
+    {
+        return pkts;
+    }
+
+    /**
+     * {@inheritDoc}
```
bbaldino: i think it'd be nice to have some general summary of what this transform method does in the comments

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -434,6 +445,107 @@ public void onRTPTranslatorWillWriteVideo(
...
+
+        for (int i = 0; i < pkts.length; i++)
+        {
+            if (pkts[i] == null)
+            {
+                continue;
+            }
+
+            long ssrc = pkts[i].getSSRCAsLong();
+            if (!ssrcsWithoutBlackKeyframes.isEmpty()
```
bbaldino: so is the idea that this logic only handles sending the very first black keyframe and doesn't ever kick in again?  just trying to get a sense of how frequently we'll have to do that array copy below

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Yes, that's right.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Will do, thanks!

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: That's right. If the participant joins video muted/hidden, then we inject black key frames with the inject task. If the participant joins with his video on, then the inject task never kicks in because it sees video packets, and that's when the transformer comes into play.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
bbaldino: does it include sps/pps?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
bbaldino: which part is the 'inject task' you're referring to?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: The byte array holds a full RTP packet and not just the VP8 payload, if that's what you're referring to.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: It's a class at the bottom of the file. It existed before this PR so you won't see it in the diff.

bbaldino: whoops, see it now.

PR #386, opened 2016-12-28
Cache improvements

bgrozev:


src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -2144,4 +2166,56 @@ public static String getLoggingId(RtpChannel rtpChannel)
...
+        /**
+         * Number of packets received. Only updated when the {@link MediaStream}
+         * is closed.
+         */
+        protected long packetsReceived = -1;
+
+        /**
+         * Number of bytes retransmitted.
+         */
+        protected AtomicLong bytesRetransmitted = new AtomicLong();
```
bickelj: Could these be final?

bgrozev: Good point, made them final.

PR #387, opened 2016-12-29
implement configuration for custom channel and conference ids

davidertel:
Allows custom ids to be used for conference and channel if the conference "name" attribute is set and the configuration property org.jitsi.videobridge.useNameForConferenceId is set to true. Allows the endpoint to be used as the channel id if the configuration property org.jitsi.videobridge.useEndpointForChannelId is set to true and endpoint is provided.

Also properly deserialize and serialize the name attribute of the conference.

Add documentation for new properties

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Content.java
+++ b/src/main/java/org/jitsi/videobridge/Content.java
@@ -276,9 +276,18 @@ public class Content
         do
         {
             String id;
-            if(!StringUtils.isNullOrEmpty(endpointId)) {
+            if(!StringUtils.isNullOrEmpty(endpointId))
+            {
+                if (channels.containsKey(endpointId))
+                {
+                    throw new Exception(String.format("Cannot create channel " +
+                            "for endpoint %s, channel already exists",
+                            endpointId));
+                }
                 id = endpointId;
-            } else {
+            }
+            else
+            {
                 id = generateChannelID();
             }
 
--- a/src/main/java/org/jitsi/videobridge/Content.java
+++ b/src/main/java/org/jitsi/videobridge/Content.java
@@ -379,14 +379,29 @@ public class Content
             Endpoint endpoint,
             int sctpPort,
             String channelBundleId,
-            Boolean initiator)
+            Boolean initiator,
+            String endpointId)
         throws Exception
     {
         SctpConnection sctpConnection;
 
         synchronized (channels)
         {
-            String id = generateChannelID();
+            String id;
+            if(!StringUtils.isNullOrEmpty(endpointId))
+            {
+                if (channels.containsKey(endpointId))
+                {
+                    throw new Exception(String.format("Cannot create channel " +
+                                    "for endpoint %s, channel already exists",
+                            endpointId));
+                }
+                id = endpointId;
+            }
+            else
+            {
+                id = generateChannelID();
+            }
 
             sctpConnection
                 = new SctpConnection(id, this, endpoint, sctpPort,
--- a/src/main/java/org/jitsi/videobridge/Content.java
+++ b/src/main/java/org/jitsi/videobridge/Content.java
@@ -259,20 +259,28 @@ public class Content
      * @param initiator the value to use for the initiator field, or
      * <tt>null</tt> to use the default value.
      * @param rtpLevelRelayType
+     * @param endpointId the id of the endpoint this channel is for, if provided
+     * the endpoint id will be used as the channel id
      * @return the created <tt>RtpChannel</tt> instance.
      * @throws Exception
      */
     public RtpChannel createRtpChannel(String channelBundleId,
                                        String transportNamespace,
                                        Boolean initiator,
-                                       RTPLevelRelayType rtpLevelRelayType)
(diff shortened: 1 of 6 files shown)


bgrozev:
Hey David, this is quite neat, avoiding adding new fields in this way! I'll add some comments inline.

PS I don't know why Jenkins keeps asking you for the CLA. Jenkins, add to whitelist, please.

src/main/java/org/jitsi/videobridge/Content.java
```diff
@@ -259,20 +259,28 @@ void askForKeyframes(Collection<Endpoint> endpoints)
...
 
         do
         {
-            String id = generateChannelID();
+            String id;
+            if(!StringUtils.isNullOrEmpty(endpointId)) {
+                id = endpointId;
+            } else {
+                id = generateChannelID();
+            }
```
bgrozev: If it so happens that there is already a channel with the ID of the endpoint, this will enter an infinite loop. Can we throw an exception instead?
Also, can you please place the braces on separate lines, as per our convention?

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -99,6 +99,20 @@
         = "org.jitsi.videobridge.MEDIA_RECORDING_TOKEN";
 
     /**
+     * If an endpoint is provided at time of channel creation, use the
+     * endpoint for the channel id
+     */
+    public static final String USE_ENDPOINT_FOR_CHANNEL_ID
+            = "org.jitsi.videobridge.useEndpointForChannelId";
```
bgrozev: Can we rename that to org.jitsi.videobridge.USE_ENDPOINT_FOR_CHANNEL_ID for consistency? I know some of the REST properties use camel case, but everything else uses all caps with underscores.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -99,6 +99,20 @@
...
+
+    /**
+     * If a name is provided at time of conference creation
+     * use the name for the conference id
+     */
+    public static final String USE_NAME_FOR_CONFERENCE_ID
+            = "org.jitsi.videobridge.useNameForConferenceId";
```
bgrozev: Same as above.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -263,7 +289,12 @@ public Conference createConference(
 
         do
         {
-            String id = generateConferenceID();
+            String id;
+            if(useNameForConferenceId && !StringUtils.isNullOrEmpty(name)) {
+                id = name;
+            } else {
+                id = generateConferenceID();
+            }
```
bgrozev: Same comment as above re: the possibility of an infinite loop (and braces).

davidertel: sure I can fix the braces add a check for the id, do you have a preference on the exception that should be thrown?

bgrozev: No preference. Looks like anything will be OK for the XMPP part -- throwing any exception or returning null from createConference() or createRtpChannel() should result in an error IQ being returned.

src/main/java/org/jitsi/videobridge/Content.java
```diff
@@ -276,9 +276,18 @@ public RtpChannel createRtpChannel(String channelBundleId,
         do
         {
             String id;
-            if(!StringUtils.isNullOrEmpty(endpointId)) {
+            if(!StringUtils.isNullOrEmpty(endpointId))
+            {
+                if (channels.containsKey(endpointId))
```
bgrozev: I think this check requires the lock on "channels" as below.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -281,18 +284,30 @@ public Conference createConference(String focus, String name)
...
     {
         Conference conference = null;
 
         do
         {
             String id;
-            if(useNameForConferenceId && !StringUtils.isNullOrEmpty(name)) {
+            if(useNameForConferenceId && !StringUtils.isNullOrEmpty(name))
+            {
+                if (conferences.containsKey(name))
```
bgrozev: Same as above, needs the lock on "conferences".

bgrozev:
LGTM. 

Would anyone else like to comment on this approach -- @lyubomir , @emcho , @paweldomas ?

davidertel:
bump

bgrozev:
@emcho had some comments after this week's community call

davidertel:
@bgrozev is he going to add the comments to this PR? if not were they recorded on the call and where are the recordings being kept now?

bgrozev:
> @bgrozev is he going to add the comments to this PR? if not were they recorded on the call and where are the recordings being kept now?

I'm sorry, we don't have a recording from this week, because I was recording locally and I messed it up.

bgrozev:

PR #390, opened 2017-01-03
fix: Reduces synchronization.

gpolitis:
Removes some unnecessary synchronized blocks.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -19,8 +19,6 @@ import java.io.*;
 import java.lang.ref.*;
 import java.util.*;
 
-import org.jitsi.service.configuration.*;
-import org.jitsi.service.libjitsi.*;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.util.*;
 import org.jitsi.util.event.*;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -580,10 +577,10 @@ private void onPinnedEndpointChangedEvent(
         // Find the new pinned endpoint.
         String newPinnedEndpointID = (String) jsonObject.get("pinnedEndpoint");
 
-        List<String> newPinnedIDList = Collections.EMPTY_LIST;
+        Set<String> newPinnedIDList = Collections.EMPTY_SET;
```
bbaldino: maybe we want to change the variable to something like `newPinnedIDs` (or just `newPinnedEndpoints` like used in `pinnedEndpointsChanged` below)?  also...doesn't look like we get to take advantage of multiple endpoints being pinned here, we only parse a single id?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -674,59 +652,67 @@ private void pinnedEndpointsChanged(List<String> pinnedEndpoints)
...
+        if (newSelectedEndpointID != null && !"".equals(newSelectedEndpointID))
         {
-            logger.debug(Logger.Category.STATISTICS,
-                         "selected_message," + getLoggingId()
-                         + " selected=" + newSelectedEndpointIDs);
+            newSelectedIDList = Collections.singleton(newSelectedEndpointID);
         }
 
-        Set<Endpoint> newSelectedEndpoints = new HashSet<>();
+        selectedEndpointsChanged(newSelectedIDList);
```
bbaldino: same question here...we only parse a single one and override any existing, so is the list useful?

src/main/java/org/jitsi/videobridge/Endpoint.java
gpolitis: I should've changed the variable names, thanks @bbaldino. It is true that we only parse a single ID; meet is sending a single ID at a time, but the JVB code is ready to handle the general case where you have a single data channel message with multiple pinned/selected endpoints.

gpolitis:
thanks @bgrozev 👍

PR #391, opened 2017-01-03
collect crash dumps during collect-dump-logs.sh

aaronkvanmeerten:
check for and archive up the crash dumps from /var/crash if they exist for our user in the past 20 minutes


gpolitis:
Jenkins, add to whitelist, please.

PR #392, opened 2017-01-04
improve logging for creating and patching conferences through the RES…

davidertel:
…T inteface. Return an Error header with more information about why the request failed. This helps debug failures that occur when using the REST API.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/rest/HandlerImpl.java
+++ b/src/main/java/org/jitsi/videobridge/rest/HandlerImpl.java
@@ -213,7 +213,7 @@ class HandlerImpl
      * The HTTP resource which is used to trigger graceful shutdown.
      */
     private static final String SHUTDOWN = "shutdown";
-    
+
     /**
      * The HTTP resource which lists the JSON representation of the
      * <tt>VideobridgeStatistics</tt>s of <tt>Videobridge</tt>.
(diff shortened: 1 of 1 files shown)


davidertel:
Adding the error header is just one way of reporting the errors back. Currently all of the errors are quietly swallowed through the REST API.

davidertel:
any comments or suggestions on this PR?

bbaldino:
looks good to me and this would give some good feedback when using the rest api... @lyubomir any thoughts?

bgrozev:
Just an opinion: for health checks (/about/health) we include similar information in the content (if the response code is 4xx/5xx). For consistency, I'd prefer to use the same approach in both places, whether that's using the content or a header field.

davidertel:
I'll update to put the content in the body and remove the header to be consistent with the health check controller

davidertel:
Moved the error messages to the response body as requested, keeps it consistent with the health check

src/main/java/org/jitsi/videobridge/rest/HandlerImpl.java
```diff
@@ -667,16 +670,18 @@ else if (RESTUtil.isJSONContentType(request.getContentType()))
                         }
                     }
                 }
-                if (status != 0)
+                else {
                     response.setStatus(status);
             }
```
bbaldino: mis-aligned brace here?  also the opening brace for the else should be on a newline

bbaldino:
lgtm.  @bgrozev ?

bgrozev:
No objections here

PR #394, opened 2017-01-05
Set a default lastActivityTime on Conference creation

virtuacoplenny:
Summary:
Conferences have a default lastActivityTime of zero.
It could be possible for VideobridgeExpireThread
to run immediately after a conference is created,
thereby expiring the conference.


bbaldino:
@bgrozev lenny had told me about this before christmas and i took a look then...i don't remembering finding anything that would prevent this scenario from happening so it seemed like a legitimate race condition.  have you guys ever seen this?

bgrozev:
I haven't seen the condition occur, but the fix makes sense to me.

PR #396, opened 2017-01-12
Refactors LastN, simulcast and RTCP termination.

gpolitis:
This needs a very long description. Depends on https://github.com/jitsi/libjitsi/pull/238


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -929,4 +930,39 @@ public String getLoggingId()
...
+
+        if (videoChannels == null || videoChannels.isEmpty())
+        {
+            return null;
+        }
+
+        MediaStreamTrackImpl[] ret = videoChannels.get(0)
+            .getStream().getMediaStreamTrackReceiver().getMediaStreamTracks();
+
+        if (videoChannels.size() > 1)
```
bbaldino: it might be slightly more efficient to call `if (!videoChannels.empty())`

bbaldino: EDIT: nevermind, i better understand your usage below now.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -284,15 +257,10 @@ public void onRTPTranslatorWillWriteAudio(
      * <tt>offset</tt> which represent the received RTP or RTCP packet.
      * @param target the {@link Channel} where this packet is going.
      */
-    public void onRTPTranslatorWillWriteVideo(
-        boolean accept, boolean data, byte[] buffer,
+    void onRTPTranslatorWillWriteVideo(
+        boolean data, byte[] buffer,
```
bbaldino: what is `boolean data`? should add to the javadoc probably

bbaldino: and maybe `data` should be renamed to something like `isData` to better match it being a boolean

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -93,8 +94,11 @@ else if (SourceGroupPacketExtension.SEMANTICS_FID
                     List<SourcePacketExtension> simSources
                         = simGroup.getSources();
 
+                    int encodingslen = includeTemporal
+                        ? simSources.size() * 3 : simSources.size();
```
bbaldino: worth defining a constant for the `3` here?  i assume it's something like "default num temporal layers"?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -111,24 +115,34 @@ else if (SourceGroupPacketExtension.SEMANTICS_FID
...
-                        encoding = new RTPEncodingImpl(
-                            track, subjectiveQualityIdx, primarySSRC, rtxSSRC,
-                            1, new RTPEncodingImpl[] { encoding });
-                        rtpEncodings[subjectiveQualityIdx] = encoding;
+                        if (includeTemporal)
+                        {
+                            int subjectiveQualityIdx = i * 3;
+                            RTPEncodingImpl encoding = new RTPEncodingImpl(
+                                track, subjectiveQualityIdx, primarySSRC, rtxSSRC,
+                                0, null);
```
bbaldino: could we describe the 0, 1, 2 used in the calls to `new RTPEncodingImpl` here?  even if it was with an inline comment like `..., 0 /* layer index */, ...`

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,568 @@
...
+            long ret = RawPacket.getSSRCAsLong(buf, off, len);
+            if (ret < 0)
+            {
+                return false;
+            }
+            ssrc = (int) ret;
+        }
+        else
+        {
+            // NOTE(gp) This will need to be adjusted when we enable
```
bbaldino: should this comment be on the RTCPHeaderUtils.getSenderSSRC method itself? rather than where it's being called?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,568 @@
...
+
+        return subStrl != null
+            && subStrl.rtpTranslatorWillWrite(data, buf, off, len);
+    }
+
+    /**
+     *
+     * @param conferenceEndpoints
+     */
+    public void update(List<Endpoint> conferenceEndpoints)
```
bbaldino: always thought this method name could be more descriptive

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,568 @@
...
+            long ret = RTCPHeaderUtils.getSenderSSRC(buf, off, len);
+            if (ret < 0)
+            {
+                return false;
+            }
+
+            ssrc = (int) ret;
+        }
+
+        SimulcastController subStrl = subCtrls.get(ssrc);
```
bbaldino: i'm confused by the naming of this variable...subStrl?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,568 @@
...
+            {
+                ctrl = new SimulcastController(allocations[i].track);
+
+                RTPEncodingImpl[] rtpEncodings
+                    = allocations[i].track.getRTPEncodings();
+
+                // Route all encodings to the specified bitrate controller.
+                for (int j = 0; j < rtpEncodings.length; j++)
+                {
+                    subCtrls.put((int) rtpEncodings[j].getPrimarySSRC(), ctrl);
```
bbaldino: every time you cast an ssrc to an int a kitten dies

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,568 @@
...
+
+        Set<String> selectedEndpoints = destEndpoint.getSelectedEndpoints();
+        int offPinned = 0, szSelected
+            = selectedEndpoints == null ? 0 : selectedEndpoints.size();
+
+        int maxQuality = 0;
+        long oldBweBps = bweBps;
+
+        int idx = -1;
+        for (int i = 0; i < szConference; i++)
```
bbaldino: any reason not to just iterate directly over the endpoints in the conference?

   for (Endpoint sourceEndpoint : conferenceEndpoints)
   {

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,568 @@
...
+            {
+                continue;
+            }
+
+            idx++;
+
+            boolean pinned = szPinned != 0
+                && pinnedEndpoints.contains(sourceEndpoint.getID());
+
+            int priority; // Higher priority endpoints are optimized first.
```
bbaldino: i think we could get rid of this priority thing altogether if we just did a pre-step of moving the pinned endpoints to the front, which might be easier?

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1801,17 +1761,18 @@ public boolean expire()
                 conferenceStatistics.totalNoPayloadChannels.incrementAndGet();
             }
 
+            MediaStreamStats streamStats = stream.getMediaStreamStats();
```
bgrozev: Can we add a null check here?

src/main/java/org/jitsi/videobridge/BitrateController.java
bgrozev: I agree with Brian, I can't even figure out what it is an abbreviation of.

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,601 @@
...
+    /**
+     * The {@link VideoChannel} which owns this {@link BitrateController}.
+     */
+    private final VideoChannel dest;
+
+    /**
+     * The bitrate controller of the {@link Conference}.
+     */
+    private final Map<Integer, SimulcastController>
+        subCtrls = new ConcurrentHashMap<>();
```
bgrozev: Can we use a more descriptive name?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,601 @@
...
+    /**
+     *
+     * @param data
+     * @param buf
+     * @param off
+     * @param len
+     * @param source
+     * @return
+     */
+    public boolean rtpTranslatorWillWrite(
```
bgrozev: Docs

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,601 @@
...
+        boolean data, byte[] buf, int off, int len, RtpChannel source)
+    {
+        Integer ssrc;
```
bgrozev: Why not use an "int" and make the autoboxing implicit? Seems that we wouldn't perform more autoboxing this way

src/main/java/org/jitsi/videobridge/BitrateController.java
bgrozev: Docs

src/main/java/org/jitsi/videobridge/BitrateController.java
bgrozev: And a koala, too

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,601 @@
...
+
+        long bweBps = Long.MAX_VALUE;
+
+        // Compute the bitrate allocation.
+        EndpointBitrateAllocation[]
+            allocations = allocate(bweBps, conferenceEndpoints);
+
+        List<String> newForwardedEndpoints = new ArrayList<>();
+
+        for (int i = 0; i < allocations.length; i++)
```
bgrozev: Foreach would be much readable here. Are you avoiding it on purpose?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,601 @@
...
+                SimulcastController subCtrl = subCtrls.get(ssrc);
+
+                if (subCtrl == null)
+                {
+                    pkts[i] = null;
+                    continue;
+                }
+
+                RawPacket[] transformedPkts = subCtrl
+                    .getRTPTransformer().transform(new RawPacket[]{pkts[i]});
```
bgrozev: If we use the SinglePacketTransformer here (and in SimulcastController) it will be simpler and we will not unnecessarily create a new array instance on every packet.

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,601 @@
...
+        {
+            if (ArrayUtils.isNullOrEmpty(pkts))
+            {
+                return pkts;
+            }
+
+            for (int i = 0; i < pkts.length; i++)
+            {
+                if (pkts[i] == null
+                    || !RTPPacketPredicate.INSTANCE.test(pkts[i]))
```
bgrozev: Shouldn't this be RTCP instead of RTP? And same comment as above for reusing SinglePacketTransforme.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Removed.

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
gpolitis: Good point; I introduced a constant and re-organized the code.

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: It's for performance because declaring an integer local variable is cheaper than creating a new iterator instance. This isn't a busy method so I can probably replace it with an iterator, if you prefer it.

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: Casting a primitive long into a primitive int is dirty cheap (if not optimized out entirely by the compiler) compared to the unavoidable implicit cast of an int to Integer that takes place here..

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -629,23 +442,19 @@ void sctpConnectionReady(Endpoint endpoint)
 
         if (endpoint.equals(getEndpoint()))
         {
-            if (lastNController.getLastN() >= 0 ||
-                    lastNController.getCurrentLastN() >= 0)
+            if (lastN > 0)
```
gpolitis: Thanks @bgrozev. No, it doesn't and I think I've completely broken the lastN notifications..

src/main/java/org/jitsi/videobridge/RtpChannel.java
gpolitis: Good point, but we do assume the existence of MediaStreamStats in various places and the MediaStream JavaDoc doesn't include null as a valid return value.

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: Thanks @bgrozev, you're right.

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: Thanks @bgrozev.

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -50,10 +50,10 @@
     private final VideoChannel dest;
 
     /**
-     * The bitrate controller of the {@link Conference}.
+     * The bitrate controllers for all SSRCs that this instance has seen.
      */
     private final Map<Integer, SimulcastController>
-        subCtrls = new ConcurrentHashMap<>();
+        ssrcToBitrateController = new ConcurrentHashMap<>();
```
bgrozev: Should this be ssrcToSimulcastController instead?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+            {
+                return false;
+            }
+
+        }
+
+        SimulcastController simulcastController
+            = ssrcToBitrateController.get((int) ssrc);
+
+        return simulcastController != null
```
bgrozev: Does this mean that we are effectively dropping SSRCs which are not signaled to us?

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: Thanks for pointing this out @bgrozev, this needs some work indeed.

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+                ctrl.update(targetIdx);
+            }
+
+            if (targetIdx > -1)
+            {
+                newForwardedEndpoints.add(allocations[i].endpointID);
+            }
+        }
+
+        this.forwardedEndpoints = newForwardedEndpoints;
```
bgrozev: Shouldn't wewe notify the receiver that the forwardedEndpoints changed (if they did)?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+
+        /**
+         *
+         */
+        private final long[] rates;
+
+        /**
+         *
+         */
+        private int targetIdx = -1;
```
bgrozev: docs

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+        }
+
+        Endpoint destEndpoint = dest.getEndpoint();
+        Set<String> pinnedEndpoints = destEndpoint.getPinnedEndpoints();
+        int szPinned = pinnedEndpoints == null ? 0 : pinnedEndpoints.size();
+        if (szPinned > 0)
+        {
+            // The lastN value should be large enough to accommodate all the
+            // pinned endpoints.
+            lastN = Math.max(lastN, szPinned);
```
bgrozev: This changes the pinned semantics. Is it intentional? https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/LastNController.java#L494

src/main/java/org/jitsi/videobridge/BitrateController.java
bgrozev: This is hard to verify, but it seems correct to me. I would love to see unit tests for it (but I don't expect to). 
If I understand it correctly, then I think Brian is right. We would get the same result if we: Extract all pinned, ordered by speech activity. Extract all non-pinned ordered by speech activity. Concatenate them.

src/main/java/org/jitsi/videobridge/BitrateController.java
bbaldino: well, maybe i'm misunderstanding here, but my point was that using 'int' for ssrcs is annoying because, even though int is 32bits, it's signed: so it doesn't always display properly when printing (some show up as negative).  personally i'd rather either standardize on something like: hold all ssrcs as longs (all the time) then they're easy to print directly for debugging or whatever or: hold all ssrcs as ints and provide one function (somewhere) to be used for displaying them...so just consistency one way or the other.  (my vote would be for just using longs that way we don't have to have the separate function for printing them).  is there something about storing them as longs that doesn't work well?  is there a reason `subCtrls` doesn't just hold a `long`?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+            boolean forwarded = priority < lastN;
+
+            EndpointBitrateAllocation endpointBitrateAllocation
+                = new EndpointBitrateAllocation(
+                sourceEndpoint, forwarded, selected);
+
+            endpointBitrateAllocations[priority] = endpointBitrateAllocation;
+
+            // First pass.
+            endpointBitrateAllocation.allocate(bweBps, maxQuality);
```
bgrozev: What is the idea here? maxEndpoint gets incremented on each iteration, and we are not necessarily iterating according to priority.

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+        }
+
+        /**
+         * Computes the optimal and the target bitrate, limiting the target to
+         * be less than bandwidth estimation specified as an argument.
+         *
+         * @param maxBps the maximum bitrate (in bps) that the target subjective
+         * quality can have.
+         * @param maxQuality the maximum subjective quality that the target
+         * subjective quality can have.
```
bgrozev: What is this, an index in the encodings array?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+                    continue;
+                }
+
+                int ssrc = (int) RTCPHeaderUtils.getSenderSSRC(pkts[i]);
+                SimulcastController subCtrl = ssrcToBitrateController.get(ssrc);
+
+                if (subCtrl == null)
+                {
+                    pkts[i] = null;
+                    continue;
```
bgrozev: Again, are we dropping non-signaled SSRCs?

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: I'm favoring for loops everywhere for performance. This isn't a busy method so I'll change this to use an iterator.

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: Yes.

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: Yes!

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -0,0 +1,613 @@
...
+     * eventually met.
+     *
+     * @param conferenceEndpoints the ordered list of {@link Endpoint}s
+     * participating in the multipoint conference with the dominant (speaker)
+     * {@link Endpoint} at the beginning of the list i.e. the dominant speaker
+     * history. This parameter is optional but it can be used for performaance;
+     * if it's omitted it will be fetched from the
+     * {@link ConferenceSpeechActivity}.
+     */
+    public void update(List<Endpoint> conferenceEndpoints)
```
bgrozev: I think we need to protect the iterate/write access to ssrcToBitrateController, since multiple threads could execute update()

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: I wanted to avoid all these sorting operations and do everything in a single loop, but I'm happy to sacrifice a little performance for improved readability.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -249,195 +193,108 @@ public void onRTPTranslatorWillWriteAudio(
...
+
+            // Set SSRC.
+            kfs[i].setSSRC(ssrc);
+
+            // Set sequence number.
+            int seqnum = (seqNumOff + i) & 0xFFFF;
+            kfs[i].setSequenceNumber(seqnum);
+
+            // Set RTP timestamp.
+            long timestamp = (tsOff + i * 3000) & 0xFFFFFFFFL;
```
bgrozev: Why did we remove the constant for 3000?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -457,333 +314,192 @@ public PacketTransformer getRTPTransformer()
...
+
+            return translate(pkts);
         }
 
-        /**
-         * {@inheritDoc}
-         */
-        @Override
-        public long getTimeUntilNextRun()
+        private RawPacket[] translate(RawPacket[] pkts)
```
bgrozev: Any reason not to use the SinglePacketTransformer for RTCP?

src/main/java/org/jitsi/videobridge/BitrateController.java
bgrozev: Agreed. The concern with doing long->int->Integer instead of long->Long is not performance.

src/main/java/org/jitsi/videobridge/BitrateController.java
bgrozev: We already have the list ordered by speech activity, so it is a matter of traversing it twice, not ordering.

src/main/java/org/jitsi/videobridge/RtpChannel.java
bgrozev: My concern is that 'stream could be null.

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -270,81 +331,81 @@ boolean isForwarded(RtpChannel source)
...
+                Endpoint sourceEndpoint = it.next();
+                if (sourceEndpoint == destEndpoint
+                    || !pinnedEndpoints.contains(sourceEndpoint.getID()))
+                {
+                    continue;
+                }
 
-            endpointBitrateAllocations[priority] = endpointBitrateAllocation;
+                endpointBitrateAllocations[priority++]
+                    = new EndpointBitrateAllocation(
```
bgrozev: I think this assumes that pinned and selected have no common elements.

bgrozev: Please disregard, I am wrong.

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -616,13 +602,19 @@ public RawPacket transform(RawPacket pkt)
         @Override
         public RawPacket transform(RawPacket pkt)
         {
+            long ssrc = pkt.getRTCPPacketType();
```
bgrozev: This looks like a typo.

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -59,10 +59,16 @@
...
+     */
+    final static String ENABLE_LIPSYNC_HACK_PNAME
+        = VideoChannel.class.getName() + ".ENABLE_LIPSYNC_HACK";
+
+    /**
      * The name of the property which controls whether {@link VideoChannel}s
      * periodically log statistics related to oversending data.
      */
-    public static final String LOG_OVERSENDING_STATS_PNAME
+    private static final String LOG_OVERSENDING_STATS_PNAME
```
bgrozev: Why is the property name private? We've specifically changed them from private to public in the past (recently).

gpolitis: Maybe I'm missing something, but it seems like it's not used elsewhere. So why have it public?

bgrozev: If someone (either from another class in the bridge, or using the bridge as a dependency) wants to set the value, they should be able to use the constant.
https://github.com/jitsi/jitsi-videobridge/issues/347

PR #405, opened 2017-01-31
fix: Piggybacks missed initial key frame packets on the egress side.

gpolitis:


src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -577,41 +577,80 @@ long getTargetBitrate()
...
+                    int extrasLen
+                        = ArrayUtils.isNullOrEmpty(extras) ? 0 : extras.length;
+
+                    RawPacket[] newExtras
+                        = new RawPacket[extrasLen + ret.length - 1];
+
+                    System.arraycopy(
+                        ret, 1, newExtras, extrasLen, ret.length - 1);
+
+                    if (extrasLen > 1)
```
bgrozev: Why do we not copy if extrasLen == 1?

gpolitis: Thanks for catching that @bgrozev.

PR #414, opened 2017-02-09
Revert "cleanup: Removes the recording functionality."

jerry2013:
We actually use the recording functionality/capability in JVB. It does different things compared with the jibri/jirecon approach.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -18,16 +18,20 @@ package org.jitsi.videobridge;
 import java.beans.*;
 import java.io.*;
 import java.lang.reflect.*;
+import java.text.*;
 import java.util.*;
 import java.util.concurrent.atomic.*;
 import java.util.logging.*;
 
 import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.*;
+import net.java.sip.communicator.impl.protocol.jabber.extensions.colibri.ColibriConferenceIQ.Recording.*;
 import net.java.sip.communicator.util.*;
 
 import org.jitsi.eventadmin.*;
+import org.jitsi.service.configuration.*;
 import org.jitsi.service.libjitsi.*;
 import org.jitsi.service.neomedia.*;
+import org.jitsi.service.neomedia.recording.*;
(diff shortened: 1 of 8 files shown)


bgrozev:
Hi @jerry2013, 

Thank you for the feedback! This revert alone will not make recording work out of the box, because of the dependency on the jnffmpeg library. You will need to create your own build with the additional dependency. Would that work for you?

jerry2013:
Thanks @bgrozev - I/we have no problem doing our own build - so long as the core functions are still in the codebase.

BTW, 2 other members of my team are on the CCLA, while I'm not. What do I do now? I didn't realize CCLA is needed for reverting.

ibauersachs:
Could this be moved into a plugin instead of the main JVB codebase?

bgrozev:
Ingo, I don't see how we could extract this code at all (while still keeping it in the bridge as opposed to e.g. jirecon).

I'm closing this and will revert the commit in another PR. But note that recording is not actively maintained and may degrade or be removed in the future.

emcho:
I am curious to hear more about this problem and specifically: why is jirecon not an option here?

emcho:
This code is inconvenient to maintain and, given that it does not actually provide a reliable solution for a majority in the community, we will be removing it soon. 

we can try to help you migrate to another solution oand give you enough time to do so (say two or three months), but we'd need to understand your usecase in order to do that.

Hence my question

jerry2013:
Thanks Emil. We developed a solution to record the raw streams from JVB (with some minor patches in jvb/jicofo/libjitsi, plus additional control logic in our application) and then synchronize streams and do post-processing, etc.

Conceptually jirecon should be able to do the same but when we added extra MUC plus the network transmission, the result wasn't as good as doing it directly at JVB. We went with this route because both were/are valid options and in essence, most of the heavy lifting is in libjitsi anyway. We thought it had better result (our opinion?), simpler to control (existing REST API to start/stop), and more straightforward to setup.

I sure hope the JVB recording functionality won't be removed, but I also understand the need to streamline the codebase if no other community members use it - although I've seen other people discussing in issues and submitting patches related to the JVB recording.

Thanks for the consideration and enjoy the weekend.

PR #417, opened 2017-02-09
feat: First version of padding termination.

gpolitis:


src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -37,15 +38,34 @@
...
      * The name of the property used to disable LastN notifications.
      */
     public static final String DISABLE_LASTN_NOTIFICATIONS_PNAME
         = "org.jitsi.videobridge.DISABLE_LASTN_NOTIFICATIONS";
 
     /**
+     * The name of the property used to trust bandwidth estimations.
+     */
+    public static final String TRUST_BWE_PNAME
+        = "org.jitsi.videobridge.TRUST_BWE_NAME";
```
bgrozev: Why _NAME?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -178,7 +213,16 @@ public void update(List<Endpoint> conferenceEndpoints)
             conferenceEndpoints = new ArrayList<>(conferenceEndpoints);
         }
 
-        long bweBps = Long.MAX_VALUE;
+        if (bweBps == -1)
+        {
+            bweBps = ((VideoMediaStream) dest.getStream())
+                .getOrCreateBandwidthEstimator().getLatestEstimate();
```
bgrozev: Maybe we can use getRemoteBitrateEstimator here?
Also, I think some sanity check on the value from getLatestEstimate() might be useful, e.g. if (bweBps <0) bweBps = Long.MAX_VALUE

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -251,6 +302,20 @@ public void update(List<Endpoint> conferenceEndpoints)
             }
         }
 
+        long availablePaddingBps = bweBps - mediaBps;
+        paddingParams.bps = Math.min(optimalBps - mediaBps, availablePaddingBps);
```
bgrozev: Make sure this is >=0?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -37,15 +38,34 @@
  * @author George Politis
  */
 public class BitrateController
+    extends PeriodicRunnable
```
bbaldino: what's the difference between PeriodicRunnable and RecurringRunnable?  Does PeriodRunnable already implement RecurringRunnable?  Why not use that for the `Injection` class in https://github.com/jitsi/jitsi-videobridge/pull/418 ?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -37,15 +38,34 @@
...
+     * The {@link Logger} to be used by this instance to print debug
+     * information.
+     */
+    private final Logger logger = Logger.getLogger(BitrateController.class);
+    
+    /**
+     * the interval/period in milliseconds at which {@link #run()} is to be
+     * invoked.
+     */
+    private static final long PADDING_PERIODICITY_MS = 15;
```
bbaldino: this language also seems different from `PERIOD_MS` in https://github.com/jitsi/jitsi-videobridge/pull/418, but seems like they are similar and could be more consistent?

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -93,18 +113,32 @@
     private final boolean disableLastNNotifications;
 
     /**
+     * A boolean that indicates whether or not we should trust the bandwidth
+     * estimations.
+     */
+    private final boolean trustBwe;
```
bbaldino: maybe we could update this comment with the behavior if bwe is not trusted? (looks like we just hard-code it to max?)

bbaldino:
> i think i'm confused about this due to a lack of familiarity...
i see now that it looks like this code is mainly in libjitsi, so i'm guessing when i get through that pr this will make more sense

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: > Maybe we can use getRemoteBitrateEstimator here?

I don't see how that would work.

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: I guess we could use a PeriodicRunnable there..

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -93,18 +113,33 @@
...
     BitrateController(VideoChannel dest)
     {
+        super(PADDING_PERIOD_MS);
         this.dest = dest;
 
         ConfigurationService cfg = LibJitsi.getConfigurationService();
-        disableLastNNotifications = cfg == null
-            || cfg.getBoolean(DISABLE_LASTN_NOTIFICATIONS_PNAME, true);
+        disableLastNNotifications = cfg != null
+            && cfg.getBoolean(DISABLE_LASTN_NOTIFICATIONS_PNAME, false);
```
bgrozev: LastN notifications are now enabled by default?

gpolitis: Yes, per Emil's suggestion (so no PSA).

src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -251,6 +303,21 @@ public void update(List<Endpoint> conferenceEndpoints)
             }
         }
 
+        long availablePaddingBps = bweBps - mediaBps;
+        paddingParams.bps = Math.min(optimalBps - mediaBps, availablePaddingBps);
+        assert paddingParams.bps >= 0;
```
bgrozev: Isn't the "paddingParams.bps < 0" a valid case which could happen if "mediaBps > bweBps"?

gpolitis: `mediaBps > bweBps` is never true in this implementation.

bgrozev: Ah, ok

bgrozev:
LGTM

PR #418, opened 2017-02-10
fix(ls-hack): Implements continuous kf injection for video muted sources.

gpolitis:


kinnla:
Thx George! We will try it out on our server once it is released

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -540,28 +618,166 @@ public void close()
...
+
+            lastRunMs = System.currentTimeMillis();
+
+            MediaStream stream = dest.getStream();
+            if (stream == null)
+            {
+                return;
+            }
+
+            synchronized (this)
```
bbaldino: why do we need to synchronize this?  the only variable i can see that's also accessed elsewhere is `running`, but i don't think we actually need to synchronize in this case.  we only run the risk of an outdated value, not a potentially corrupt value, and there's always a chance running is set to false right after we check it anyway, so it seems like there isn't really a need?  but maybe i'm missing something here.

gpolitis: This is synchronized with the `stop` method (which updates the `running` field). The `stop` method is invoked when the video translator thread receives a video RTP packet and right before the `maxSeqNum` and `maxTs` fields are accessed by the video translator write thread.

The idea here is to serialize r/w of `maxSeqNum` and `maxTs`, to make sure the video translator thread is not using an outdated `maxSeqNum` value.

Does that make sense?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -204,52 +284,40 @@ void onRTPTranslatorWillWriteAudio(
             if (acceptedVideoSSRCs.contains(receiveVideoSSRC)
                 || injections.containsKey(receiveVideoSSRC))
             {
-                // No need to hack this SSRC.
+                // No need to hack this SSRC, as it's either already accepted
```
gpolitis: I have updated the explanatory comment. Please let me know if it still doesn't make sense.

bbaldino: ok one more stupid question, what is the qualification for 'acceptance' here?

src/main/java/org/jitsi/videobridge/LipSyncHack.java
bbaldino: gotcha.  i think it'd be good to put this guarantee in the java docs for stop so it's more clear (i.e.: "after stop is called, guaranteed that the injection logic is stopped and maxSeqNum and maxTs will not be further modified" or something)

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: So, an SSRC is assigned the accepted flag when at least 1 RTP video media packet has been accepted (or forwarded if you want).

src/main/java/org/jitsi/videobridge/LipSyncHack.java
gpolitis: Awesome, I added some documentation for that.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
bbaldino: ah ha, ok.  so this is basically saying "either we've already started forwarding real packets or we've already initialized the hack" and in both of those we don't need to initialize the hack.  think i got it, thanks george

PR #420, opened 2017-02-16
Add ability to dynamically change lastN over the datachannel

saghul:


bgrozev:
I'm really exited about this one, we've been talking about it for years :)

One concern, though: we're already overloading the term "last-n", and I was wondering if we can use a distinct name. In my mind, the system (i.e. jicofo) configures some value for "last-n", and the individual clients can only lower the number of streams they receive, not exceed it. With this in mind, I suggest CLIENT_MAX_VIDEO_STRAMS.

saghul:
Thanks for taking a look @bgrozev!  My understanding after discussing it with others was that some of these settings started as "global" settings (which jicofo sets) but would then transition to dynamic one like this one.

So, with that in mind, this doesn't overload the existing "last-n", it just makes it dynamically adjustable from the client side. Jicofo will still set the globally configured option, and then to each their own.

I guess there is indeed a clash with the concept of adaptive last N. I'm happy to change the name once we agree on the approach.

gpolitis:
I would prefer keeping the same name for lastN because that's what it is, however I won't oppose to using a different name.

bgrozev:
So individual clients will be able to (at least effectively) disable last-n, overriding what jicofo configured?

saghul:
> So individual clients will be able to (at least effectively) disable last-n, overriding what jicofo configured?

Correct. Example user case: when the mobile app goes into the background, we set lastN to 0, that is, disable all remote video. And when we come back we set it back to -1.

bgrozev:
I have two concerns with this. Suppose we have configured last-n=3 through jicofo. Then:
1. The client can disable last-n by setting it to -1.
2. Once the client sets last-n=0 the original value of 3 is lost (yeah, I know it is in config.js right now, but this is mostly a coincidence).

Because of this I prefer a separate client-side upper limit on last-n, and because of this I prefer another name.

bgrozev:
To clarify: I only suggest to change the name at this point.

gpolitis:
> Once the client sets last-n=0 the original value of 3 is lost (yeah, I know it is in config.js right now, but this is mostly a coincidence).

Maybe I'm missing something, but why can't the client store that value somewhere?

Since the effect of this data channel message is to change the `lastN` property in the `VideoChannel` (and of course update the bitrate controller so that the value becomes effective), I find it confusing to use another name.

bgrozev:
> Maybe I'm missing something, but why can't the client store that value somewhere?

How does it learn about it -- from jitsi-videobridge? From jicofo? 

> Since the effect of this data channel message is to change the lastN property

My point is, in the long-term, it shouldn't.

gpolitis:
> How does it learn about it -- from jitsi-videobridge? From jicofo?

From jicofo, just like it does now. Do you think that should change?

> My point is, in the long-term, it shouldn't.

I'm not sure how you see things in the future and I'd be really interested to hear that, however, right now that's all it does.

bgrozev:
> From jicofo, just like it does now. Do you think that should change?

I am not aware of this, and couldn't find it with a quick grep. Am I missing something?

> I'm not sure how you see things in the future and I'd be really interested to hear that

See above, a separate per-client, client configured limit on the number of video streams to be received.

> right now that's all it does.

This is am implementation detail. It will change when we implement the above. No need to hold the PR until then, but I don't want to rename properties later (or worse, be stuck with yet another usage of "last-n") when we do.

gpolitis:
I had a quick discussion with Boris and I now I see where he's coming from, so +1 to changing the name.

davidertel:
This is a pretty cool feature but it should probably be configurable whether or not to allow it. Take for instance the case where your focus limits certain customers to a set amount of active streams. Maybe if they are on a low bandwidth connection or you may limit based on your pricing tier etc. Allowing the client to override the focus setting would circumvent that.

saghul:
@bgrozev @gpolitis I'm not sure I get it 100% but I'll try to implement it with another name tomorrow morning. 3 questions:

- name suggestions?
- how does this play with lastN? we take the minimum of the 2, if lastN is -1?
- what message do we report back to the client? last N changed, XXX changed, ?

@davidertel Hum, interesting point. @bgrozev is that what you had in mind? FWIW, my understanding of the feature was not as a way to limit the number of received streams administratively, but to allow clients to self-manage.

Taking this one step further, why don't we get rid of the property in Jicofo altogether? Why do we want to globally limit the number of received streams for all clients?

I feel I miss some background on what lastN really represents.

saghul:
For the record:

I had a very productive and clarifying talk with @bgrozev, @gpolitis and @yanas.  The outcome was that we are going to build this as a new concept, under a different name: "max forwarded endpoints".

LastN will continue to exist but it will represent the value chosen by the infrastructure administrator. The max forwarded endpoints value, which the user can set, will not overrule lastN and will never be allowed go beyond it.

saghul:
Update: forget anything I said. New approach: keep the current PR as is, strictly related to lastN, and limited to allowing it to be changed dynamically.

Name changes shall be discussed at a later stage.

As for putting a maximum server enforced cap, that will happen in a follow-up PR and it will be a JVB config setting, not a config.js setting Jicofo passes along.

@bgrozev @gpolitis can you folks take another look? I have no further changes planned for this PR now that we finally reached agreement.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -706,6 +727,39 @@ private void selectedEndpointsChanged(Set<String> newSelectedEndpoints)
...
+        int lastN = ((Number) o).intValue();
+
+        List<RtpChannel> videoChannels = getChannels(MediaType.VIDEO);
+        if (videoChannels == null || videoChannels.isEmpty())
+        {
+            return;
+        }
+
+        for (int i = 0; i < videoChannels.size(); i++) {
+            videoChannels.get(i).setLastN(lastN);
```
bgrozev: How about foreach here?

bgrozev:
LGTM, I left a minor comment inline.

gpolitis:
LGTM modulo @bgrozev's comment.

saghul:
@bgrozev @gpolitis Updated and rebased. PTAL.

bgrozev:
Tests passed here: http://ci.jitsi.org/view/pr-testing/job/jitsi-videobridge-PR-test/418/console

PR #423, opened 2017-02-20
cleanup: remove code handling simulcast mode

saghul:
The option got axed, there is nothing to handle now.


saghul:
Jenkins test this please.

PR #428, opened 2017-03-02
feat: Adds a TransportCCEngine to the transport manager.

bgrozev:
This depends on libjitsi#276.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -863,6 +863,10 @@ public class RtpChannel
 
         synchronized (streamSyncRoot)
         {
+            TransportManager transportManager = getTransportManager();
+            TransportCCEngine transportCCEngine
+                = transportManager.getTransportCCEngine();
+
             stream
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -900,10 +908,9 @@ void initialize(RTPLevelRelayType rtpLevelRelayType)
 
             // The transport manager could be already connected, in which case
             // (since we just created the stream), any previous calls to
-            // transportConnected() have failed started the stream. So trigger
+            // transportConnected() have failed to start the stream. So trigger
             // one now, to make sure that the stream is started.
-            TransportManager transportManager = getTransportManager();
-            if (transportManager != null && transportManager.isConnected())
+            if (transportManager.isConnected())
```
bbaldino: why is the null check no longer needed?

bgrozev: Because Channel#transportManager, while not final, is always initialized to non-null in Channel#initialize()

PR #429, opened 2017-03-07
VP8 TL filtering & rewriting

gpolitis:


src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -47,15 +47,9 @@
...
-     * The name of the property used to disable LastN notifications.
-     */
-    public static final String DISABLE_LASTN_NOTIFICATIONS_PNAME
-        = "org.jitsi.videobridge.DISABLE_LASTN_NOTIFICATIONS";
-
-    /**
      * The name of the property used to trust bandwidth estimations.
      */
-    public static final String TRUST_BWE_PNAME
+    private static final String TRUST_BWE_PNAME
```
bgrozev: What is the reason for this?

gpolitis: The IDE flagged it as a warning (it's not used outside this class).

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -29,43 +31,150 @@
...
+        // The spatial layer is offered either by simulcast (VP8) or spatial
+        // scalability (VP9). Exotic cases might do simulcast + spatial
+        // scalability.
+
+        if (spatialLen < 2 /* simulcast */)
         {
-            int streamIdx = i / tempoLen, tempoIdx = i % tempoLen;
+            // this O(n) loop is more efficient than the O(n^3) loop below, and
+            // it works well for the simulcast case, so we keep it.
+            for (int i = 0; i < rtpEncodings.length; i++)
```
bgrozev: If the only reason for this code is optimization I would rather remove it (less logic for humans to understand, less space for a bug to hide). It seems like we expect n to be on the order of 9, so n^3 shouldn't be a problem.

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -181,11 +289,32 @@ else if (SourceGroupPacketExtension.SEMANTICS_FID
                     Long primarySSRC = fidEntry.getKey();
                     Long rtxSSRC = fidEntry.getValue();
 
-                    RTPEncodingDesc[] encodings = new RTPEncodingDesc[1];
+                    int numOfTemporal, numOfSpatial;
+                    if (false && ENABLE_SVC)
```
bgrozev: Is this a typo?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -202,10 +331,32 @@ else if (SourceGroupPacketExtension.SEMANTICS_FID
 
                 if (!grouped.contains(mediaSSRC))
                 {
-                    RTPEncodingDesc[] encodings = new RTPEncodingDesc[1];
+                    int numOfTemporal, numOfSpatial;
+                    if (false && ENABLE_SVC)
```
bgrozev: Ditto

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
gpolitis: No! `true  && ENABLE_SVC` is specific to VP9 SVC as it will make the bridge think that there are 6 layers (2 spatial * 3 temporal), but that is not true for plain VP8.

bgrozev: I'm lost. If this code is meant to be disabled for the time being, how about setting something like ENABLE_VP9_SUPPORT=false and then using the constant?

gpolitis: sure!

PR #431, opened 2017-03-17
feat: Bitrate allocation for the thumbnails.

gpolitis:


src/main/java/org/jitsi/videobridge/BitrateController.java
```diff
@@ -685,12 +685,24 @@ private EndpointBitrateAllocation(
 
             // Initialize rates.
             rates = new long[encodings.length];
+            int optimal180Idx = 0;
```
bbaldino: maybe this should be `optimalThumbnailIndex` and instead of hardcoding `180` below we make a `THUMBNAIL_MAX_RESOLUTION` field or something?

saghul: Agreed, I've seen Chrome send 240x135.

gpolitis:
Replaced by https://github.com/jitsi/jitsi-videobridge/pull/435

src/main/java/org/jitsi/videobridge/BitrateController.java
gpolitis: @saghul @bbaldino I followed your suggestion in https://github.com/jitsi/jitsi-videobridge/pull/435. Thanks!

Issue #147, opened 2016-02-23
Fine-Grained Control of Forwarded Simulcast Layers


bilhamil:
Hi there,
    I've been starting to work with the videobridge and lib-jitsi-meet to support a live video research project I'm working on. So far it has been awesome, especially with the recent lib-jitsi-meet support. Anyway, It would help me out if the bridge could support forwarding multiple high quality video streams to conference participants, as well as have mechanisms for the client to have stream by stream control of what layers were being forwarded.

The current selected participant mechanism is nice if you want to support a hangout like conference, but I'm particularly trying to design an experience where there may be multiple high, mid, or low quality videos that can change qualities somewhat dynamically. I don't think there is currently a means to support this kind of situation.

I currently have a mechanism implemented that adds a colibri message supporting a client requesting a particular simulcast layer on a video channel by ssrc. Is this a feature that people would be open to supporting? I realize that there are potentially different issues in terms of adaptive bandwidth control. I am just looking for some feedback on the feature. I can also work on preparing a pull request.

Thanks!

gpolitis:
Hello @bilhamil , we have been thinking about this for a while and various members of the community have expressed their interest as well. We know that some PRs are imminent that would allow for multiple selected endpoints by modifying the existing "selected endpoint" data channel message to hold an array of endpoints. Is this something that would fit your needs? You might want to join us next Monday in our next community call where we'll discuss some of this for sure.

bilhamil:
Hi @gpolitis, that would definitely be step in the right direction I think. I would like to potentially be able to access the in between simulcast layers eventually. I think with the selected endpoints approach it is not clear how to support that in particular. I would definitely be interested in being there for the community call. The call is next monday at 16:30 UTC?

gpolitis:
I see now, what you want to do is not going to be possible through the data channel message extension I had in mind (for which we have a PR already here https://github.com/jitsi/jitsi-videobridge/pull/148). There have been talks about exposing a more fine-grained control of forwarded simulcast layers in the last couple of months and we can definitely discuss that during the call (right, next Monday at 16:30 UTC).

markreg:
How does the algorithm for layer switching/selection currently work?

bgrozev:
It's controlled by the clients (highest quality for the "selected" endpoint, base layer for the rest).

jtg-gg:
any update on this? 
is there any api to manually select the simulcast layer from client using javascript ?

saghul:
[...]

PR #435, opened 2017-03-21
Adaptivity fixes

gpolitis:


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -704,14 +723,16 @@ private EndpointBitrateAllocation(
...
         void allocate(long maxBps, int maxQuality)
         {
-            if (!forwarded || rates.length == 0)
+            if (rates.length == 0)
             {
                 return;
             }
 
-            maxQuality = selected ? Math.min(maxQuality, rates.length - 1) : 0;
+            maxQuality = Math.min(maxQuality, optimalIdx);
```
bbaldino: does `maxQuality` refer to an index?  i think it'd be good to rename it to `maxQualityIdx`

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -113,7 +113,9 @@
         // scalability (VP9). Exotic cases might do simulcast + spatial
         // scalability.
 
+        int resolution = 180;
```
bbaldino: use the thumbnail resolution constant here?

gpolitis: This `resolution` should, in theory, be included in the SDP answer of a sender, while the thumbnail resolution constraint in the bitrate controller should come from the receiver as a data channel message (as it is requested here https://github.com/jitsi/jitsi-videobridge/issues/147) or even in the answer of a receiver.

So these two refer to different things and I feel uncomfortable using the same constant. How about I use another constant here (and I also include the above explanation)?

gpolitis:
@bbaldino thanks! Let's wait for a new LJ build and update the pom.xml file (otherwise the project won't build).

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -685,12 +692,24 @@ private EndpointBitrateAllocation(
...
             {
                 rates[i] = encodings[i].getLastStableBitrateBps();
+                if (encodings[i].getHeight() <= THUMBNAIL_MAX_HEIGHT)
+                {
+                    optimalThumbnailIndex = i;
+                }
             }
 
-            optimalIdx = selected ? encodings.length - 1 : (forwarded ? 0 : -1);
+            // TODO Determining the optimal index needs some work. The optimal
```
saghul: I'm not sure we'd ever want to do that. Mobile devices these days have so many pixels that a 360p video would look terrible.

PR #440, opened 2017-03-24
 Splits the REST interface into a private and public part

bgrozev:


damencho:
If we enable the private one by default, shouldn't we also turn on graceful shutdown command by default, with ALLOWED_SOURCE_REGEXP=127.0.0.1, WDYT?

bgrozev:
@damencho Hmm, I don't know.

Note that the default hasn't changed much, we still require "--apis=rest" (this is in contrast with the public HTTP API, which requires its port to be configured in addition to "--apis=rest").

PR #445, opened 2017-03-31
fix(ls-hack): Supports dynamic PT for VP8.

gpolitis:
Chrome 57 changed the default payload types (e.g. previously it was 100 for VP8 and now it's 96). We're still using the old payload types and now Chrome seems to not register the associated payload type for VP8 RTX which breaks retransmissions. The workaround is to use the new payload type defaults that Chrome 57 is using, so the LS hack needs to dynamically discover the VP8 payload type and not always assume it's 100.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/LipSyncHack.java
+++ b/src/main/java/org/jitsi/videobridge/LipSyncHack.java
@@ -31,6 +31,8 @@ import org.jitsi.util.function.*;
 import java.util.*;
 import java.util.concurrent.*;
 
+import org.jitsi.service.neomedia.codec.*;
+
 /**
  * Implements a hack for
  * https://bugs.chromium.org/p/chromium/issues/detail?id=403710. The hack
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -318,7 +325,31 @@ void onRTPTranslatorWillWriteAudio(
      * @param sz the number of black key frame packets to create.
      * @return the array of black key frame packets that were created.
      */
-    private static RawPacket[] make(int ssrc, int seqNumOff, long tsOff, int sz)
+    private RawPacket[] make(int ssrc, int seqNumOff, long tsOff, int sz)
```
bbaldino: is it not possible/practical to pull these from signaling instead?  not that this wouldn't work, but technically we should be following what has been signaled, right?  might also make things more straightforward if we actively don't accept media for payload types that haven't been signaled.

gpolitis: This is exactly what this patch does ;)

bbaldino: lol, sorry!  when i saw it pulling from the stream i just assumed it was reading from an incoming packet.  just ignore me then :)

bbaldino:
lgtm, will merge when tests finish

PR #446, opened 2017-04-07
Bandwidth estimation and probing for bandwidth fixes

gpolitis:
This PR contains fixes and features related to adaptivity (bandwidth estimations, bandwidth probing, simulcast). This PR depends on https://github.com/jitsi/libjitsi/pull/297

1. Various fixes in the padding budget calculation:
  a. We don't take into account expired endpoints
  b. we're more careful as to what we consider inactive, and this affects the padding calculation.
2. Simulcast related fixes, specifically about requesting key frames:
  a. when we detect that a stream has been suspended, we only ask for a key frame if the suspended stream is forwarded
  b. when temporal filtering is enabled we carelessly where requesting keyframes on temporal layer changes
3. makes a bunch of stuff configurable through java sys properties.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
@@ -18,6 +18,8 @@ package org.jitsi.videobridge.cc;
 import org.jitsi.impl.neomedia.*;
 import org.jitsi.impl.neomedia.rtp.*;
 import org.jitsi.impl.neomedia.transform.*;
+import org.jitsi.service.configuration.*;
+import org.jitsi.service.libjitsi.*;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.util.*;
 import org.jitsi.util.concurrent.*;
@@ -32,16 +34,45 @@ public class BandwidthProbing
     extends PeriodicRunnable
 {
     /**
+     * The system property name that holds a boolean that determines whether or
+     * not to activate the RTX bandwidth probing mechanism that implements
+     * stream protection.
+     */
+    public static final String
+        DISABLE_RTX_PROBING_PNAME = "org.jitsi.videobridge.DISABLE_RTX_PROBING";
+
+    /**
+     * The system property name that holds the interval/period in milliseconds
+     * at which {@link #run()} is to be invoked.
+     */
+    public static final String
+        PADDING_PERIOD_MS_PNAME = "org.jitsi.videobridge.PADDING_PERIOD_MS";
+
+    /**
      * The {@link Logger} to be used by this instance to print debug
      * information.
      */
-    private final Logger logger = Logger.getLogger(BandwidthProbing.class);
+    private static final Logger logger = Logger.getLogger(BandwidthProbing.class);
+
+    /**
+     * The ConfigurationService to get config values from.
+     */
+    private static final ConfigurationService
+        cfg = LibJitsi.getConfigurationService();
 
     /**
      * the interval/period in milliseconds at which {@link #run()} is to be
      * invoked.
      */
-    private static final long PADDING_PERIOD_MS = 15;
+    private static final long PADDING_PERIOD_MS =
+        cfg != null ? cfg.getInt(PADDING_PERIOD_MS_PNAME, 15) : 15;
+
+    /**
+     * A boolean that determines whether or not to activate the RTX bandwidth
+     * probing mechanism that implements stream protection.
+     */
+    private static final boolean DISABLE_RTX_PROBING =
+        cfg != null && cfg.getBoolean(DISABLE_RTX_PROBING_PNAME, false);
 
     /**
      * The {@link VideoChannel} to probe for available send bandwidth.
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
```diff
@@ -32,16 +34,45 @@
     extends PeriodicRunnable
 {
     /**
+     * The system property name that holds a boolean that determines whether or
+     * not to activate the RTX bandwidth probing mechanism that implements
+     * stream protection.
+     */
+    public static final String
+        DISABLE_RTX_PROBING_PNAME = "org.jitsi.videobridge.DISABLE_RTX_PROBING";
```
bbaldino: could we name this in the affirmative? `RTX_PROBING` or something?  the negative naming leads to weird double negatives like "if disable rtx probing is false..."

src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
```diff
@@ -32,16 +34,45 @@
...
      */
-    private static final long PADDING_PERIOD_MS = 15;
+    private static final long PADDING_PERIOD_MS =
+        cfg != null ? cfg.getInt(PADDING_PERIOD_MS_PNAME, 15) : 15;
+
+    /**
+     * A boolean that determines whether or not to activate the RTX bandwidth
+     * probing mechanism that implements stream protection.
+     */
+    private static final boolean DISABLE_RTX_PROBING =
```
bbaldino: same question as above

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -87,6 +87,12 @@
     implements TransformEngine
 {
     /**
+     * The property name that holds the bandwidth estimation threshold.
+     */
+    public static final String BWE_THRES_PNAME
```
bbaldino: i'd vote for just spelling out `THRESHOLD`

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -100,6 +106,20 @@
...
+    private static final ConfigurationService
+        cfg = LibJitsi.getConfigurationService();
+
+    /**
+     * In order to limit the resolution changes due to bandwidth changes we
+     * react to bandwidth changes greater BWE_THRES / 100 of the last bandwidth
+     * estimation.
+     */
+    private static final int BWE_THRES
+        = cfg != null ? cfg.getInt(BWE_THRES_PNAME, 15) : 15;
```
bbaldino: constant for default value?  also, we should have a unit of time here in the name i think (`BWE_CHANGETHRESHOLD_PCT`?) (i actually guessed _MS and _BPS first, but i think it's PCT? EDIT: and now actually i guess it isn't, since we still divide by 100....so i think units would help here :)

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -260,6 +287,17 @@ public boolean accept(byte[] buf, int off, int len)
      */
     public void update(List<Endpoint> conferenceEndpoints, long bweBps)
     {
+        if (bweBps > -1)
+        {
+            if (lastBwe != -1 &&
+                Math.abs(bweBps - lastBwe) < lastBwe * BWE_THRES / 100)
```
bbaldino: even though it's short, and probably only used here, i like the idea of helper functions for stuff like this...it makes reading it a bit clearer and makes it obvious where the adjustment needs to be made if we want to play with it, like:

```
public boolean isLargerThanBweThreshold(long currentBweBpsDelta, lastBweBps)
{
  return bweBpsDelta > lastBweBps * BWE_THRESHOLD_BPS / 100);
}
```
```
long bweBpsDelta = Math.abs(bweBps - lastBwe);
if (isLargerThanBweThreshold(bweBpsDelta)
{
  ...
```

src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
gpolitis: The default value of a boolean is false, and we want this feature on by default, hence the stupid naming.

bbaldino: is there a way to set the default to true?

gpolitis: We can set the default to whatever we want, but conceptually in general the default value of a boolean is false. I don't mind to changing this but we try to follow this rule across the projects.

bbaldino: yeah, i think it'd be nice to try and avoid the double-negative thing.  but ok, we can hold off on this for this pr.

gpolitis: It is a little bit counter intuitive, I agree :/

PR #450, opened 2017-04-17
Refactors the Endpoint class

bgrozev:
Extracts the code for handling sending and receiving messages for an Endpoint into EndpointMessageTransport in preparation for the introduction of a new underlying transport (web sockets).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
+++ b/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
@@ -1,0 +1,517 @@
+/*
+ * Copyright @ 2017 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.videobridge;
+
+import org.jitsi.service.neomedia.*;
+import org.jitsi.util.*;
+import org.json.simple.*;
+import org.json.simple.parser.*;
+
+import java.io.*;
+import java.util.*;
+
+/**
+ * @author Boris Grozev
+ */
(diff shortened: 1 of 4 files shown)


bgrozev:
Jenkins test this please

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -0,0 +1,517 @@
...
+
+import org.jitsi.service.neomedia.*;
+import org.jitsi.util.*;
+import org.json.simple.*;
+import org.json.simple.parser.*;
+
+import java.io.*;
+import java.util.*;
+
+/**
```
gpolitis: Maybe add a small description of this class?

bgrozev: Do you mind if I do this in another PR (coming in later today)?

gpolitis: That's fine 👍

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -0,0 +1,517 @@
...
+ * @author Boris Grozev
+ */
+class EndpointMessageTransport
```
bbaldino: is the idea that this class would handle the connection logic for both types (datachannel and websocket) as well as the high-level message parsing (determining selected vs pinned etc) ?

bgrozev: Yes. Endpoint will provide a "sendMessage" method, and EndpointMessageTransport will decide which of the underlying transport to send it over. For incoming messages, it will handle the JSON parsing (and e.g. call Endpoint#selectedEndpointsChanged(List<Endpoint>))

bgrozev: The two classes are quite coupled, but at least there's a separation between the messaging stuff and the other endpoint stuff (maintaining the list of channels, selected endpoints, etc)

bbaldino: ok, so currently it implements the datachannel methods, and soon it will implement the websocket messages?  so there will be some methods similar to `onStringData` and `onBinaryData` for websockets, is that right?

bgrozev: Exactly.

bbaldino:
ok, lgtm.  ready to merge?

bgrozev:
Yes

PR #453, opened 2017-04-18
Colibri web sockets

bgrozev:
Supports WebSockets as an alternative transport for COLIBRI messages (i.e. an alternative to WebRTC data channels).

### DIFF
--- a/doc/rest.md
+++ b/doc/rest.md
@@ -13,7 +13,7 @@ The _public_ interface inludes:
 
 * Support for serving static files (e.g. the HTML/js for jitsi-meet)
 * Support for proxying (for e.g. proxying BOSH connections to a prosody instance)
-* A WebSocket API for communicatoin with conference endpoints (```/colibri-ws/```)
+* A WebSocket API for communication with conference endpoints (```/colibri-ws/```)
 
 **For any of the HTTP interfaces to be enabled, jitsi-videobridge needs to be started with the ```--apis=rest```
 parameter (or ```--apis=rest,xmpp``` to also enable the XMPP interface to COLIBRI).** This is enough to enable the
--- a/doc/rest.md
+++ b/doc/rest.md
@@ -13,6 +13,7 @@ The _public_ interface inludes:
 
 * Support for serving static files (e.g. the HTML/js for jitsi-meet)
 * Support for proxying (for e.g. proxying BOSH connections to a prosody instance)
+* A WebSocket API for communicatoin with conference endpoints (```/colibri/ws/```)
 
 **For any of the HTTP interfaces to be enabled, jitsi-videobridge needs to be started with the ```--apis=rest```
 parameter (or ```--apis=rest,xmpp``` to also enable the XMPP interface to COLIBRI).** This is enough to enable the
--- a/doc/rest.md
+++ b/doc/rest.md
@@ -13,7 +13,7 @@ The _public_ interface inludes:
 
 * Support for serving static files (e.g. the HTML/js for jitsi-meet)
 * Support for proxying (for e.g. proxying BOSH connections to a prosody instance)
-* A WebSocket API for communicatoin with conference endpoints (```/colibri/ws/```)
+* A WebSocket API for communicatoin with conference endpoints (```/colibri-ws/```)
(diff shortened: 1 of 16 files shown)


doc/rest.md
```diff
@@ -13,6 +13,7 @@ The _public_ interface inludes:
 
 * Support for serving static files (e.g. the HTML/js for jitsi-meet)
 * Support for proxying (for e.g. proxying BOSH connections to a prosody instance)
+* A WebSocket API for communicatoin with conference endpoints (```/colibri/ws/```)
```
bbaldino: typo

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1312,7 +1312,8 @@ IceUdpTransportManager getTransportManager(
                 try
                 {
                     transportManager
-                        = new IceUdpTransportManager(this, initiator, 1);
+                        = new IceUdpTransportManager(
+                            this, initiator, 1, channelBundleId);
```
bbaldino: what was this change for?

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -103,46 +140,114 @@
...
+     * @param dst the {@link ColibriWebSocket} through which to send the message.
+     * @param message the message to send.
+     * @param errorMessage an error message to be logged in case of failure.
+     */
+    private void sendMessage(
+        ColibriWebSocket dst, String message, String errorMessage)
+    {
+        try
+        {
+            dst.getRemote().sendString(message);
```
bbaldino: should we add a stat for websocket sends?

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -464,54 +574,128 @@ public void onStringData(WebRtcDataStream src, String msg)
...
+    void sendMessage(String msg)
         throws IOException
     {
         SctpConnection sctpConnection = endpoint.getSctpConnection();
+        ColibriWebSocket webSocket = this.webSocket;
         String endpointId = endpoint.getID();
 
-        if (sctpConnection == null)
+        Object dst = null;
+        if (webSocketLastActive)
```
bbaldino: a lot of the logic around handling the destination makes me wonder if things might be cleaner if we had websocket and datachannel transport classes which handled their logic and a transport manager or something that handled tracking which one was active or something.

bbaldino: or maybe it's at least worth adding a helper to get the active connection so all that logic is tucked away somewhere else

bbaldino: it seems like there's enough logic around which connection to use for a given message to warrant some sort of 'manager' there.

EDIT: have finished going through it all now...i guess it isn't too bad, but maybe a nice-to-have.  what do you think?

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: The TransportManager needs to advertise the URL which endpoints can connect to. The URL encodes the conference and endpoint ID. And channelBundleId coincides with the endpoint ID.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
bgrozev: Yes, will do.

doc/rest.md
bgrozev: Fixed.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
bgrozev: Good point, I'll add a helper.

src/main/java/org/jitsi/videobridge/Conference.java
bbaldino: 👍

PR #456, opened 2017-04-25
feat: Implements VP8 pictureID rewriting.

gpolitis:


src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -41,9 +46,41 @@
...
+     * The property name of the number of seen frames to keep track of.
+     */
+    public static final String SEEN_FRAME_HISTORY_SIZE_PNAME
+        = "org.jitsi.videobridge.SEEN_FRAME_HISTORY_SIZE";
+
+    /**
+     *
+     */
+    public static final String ENABLE_VP8_PICID_REWRITING_PNAME
+        = "org.jitsi.videobridge.ENABLE_PICID_REWRITING";
```
bgrozev: Is "VP8" intentionally missing from the value?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1099,14 +1210,81 @@ boolean accept(FrameDesc source, byte[] buf, int off, int len)
...
+                                pktOut.getBuffer(),
+                                pktOut.getOffset(),
+                                pktOut.getLength());
+
+                        if (!DePacketizer
+                            .VP8PayloadDescriptor.hasExtendedPictureId(
+                            redBlock.getBuffer(), redBlock.getOffset(),
+                            redBlock.getLength()))
+                        {
+                            // XXX we have observed that using a non-extended
```
bgrozev: Is this meant as a temporary fix? It worries me because it executes on each video packet, allocates new memory and copies the payload. It alo assumes that the VP8 payload descriptor has a certain format (already contains a short PictureID).

bgrozev: Correction (due to George): this does not execute on every packet, only on ones without an extended PictureID. Ignore the comment.

PR #469, opened 2017-05-09
fix: Doesn't accept COLIBRI IQs with focus==null

bgrozev:
when no authorizedSourcePattern is configured. Fixes a unit test failure
reported by Devin Wilson on the mailing list.


dwilson6:
Thanks!

PR #471, opened 2017-05-10
mvn: Updates ice4j.

damencho:
Fixes leak of file descriptors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
-      <version>2.0-20170420.153542-14</version>
+      <version>2.0-20170510.225531-16</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>


dbareiro:
Hi, George/Damian.

This is an additional new fix to those already found in the testing repo for the "too many open files" issue?

Kind regards,
Daniel

damencho:
Yes, that's correct.

dbareiro:
Thanks for confirming, Damian.

Eight days ago I'm using testing and for now I have not had this issue again. Is this a new scenario that can lead to a leak? Is there any other open issue in relation to this new fix? Maybe the issue 1390 for jitsi-meet?

I understand that after several automatic tests, this would have to be automatically integrated into testing repo, right? If so, when you think this could happen?

Kind regards,
Daniel

dbareiro:
Hi, Damian.

> A user reported that he was still seeing issues after we had several fixes on the subject and we 
> managed to reproduce and fix it. Yes, the communication on the subject is in jitsi/jitsi-meet#1390.

Thanks for confirm.

> Not sure what is failing in performance testing and why it is not going to testing. Will check
> later today.

Thank you very much. If you can tell us here when the performance testing have passed to have this in the testing repo, it would be great.

Thanks again.

Kind regards,
Daniel

PR #472, opened 2017-05-11
fix: Takes into account empty conferences for graceful shutdown.

bgrozev:


gpolitis:
Just passing by, we've solved the conferences of size 1 problem, right?

bgrozev:
> Just passing by, we've solved the conferences of size 1 problem, right?

What's that problem?

gpolitis:
Thank you for explaining the context of this PR (offline). LGTM.

bgrozev:
Yeah, I think we've solved the size 1 problem, or mostly solved it. Even if not, conferences should eventually drop to size 0 (even if the conference expire thread stop working, resulting in accumulating conferences with size 0).

PR #474, opened 2017-05-18
High fps allocation for on-stage participants.

gpolitis:
Depends on https://github.com/jitsi/libjitsi/pull/308

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
@@ -163,6 +163,24 @@ public class BandwidthProbing
         if (totalNeededBps < 1)
         {
             // Not much.
+
+            if (logger.isDebugEnabled())
+            {
+                long bweBps = ((VideoMediaStream) destStream)
+                    .getOrCreateBandwidthEstimator().getLatestEstimate();
+
+                logger.debug("probing" +
+                    "," + System.currentTimeMillis() +
+                    "," + destStream.hashCode() +
+                    "," + 0 +
+                    "," + totalOptimalBps +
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/videobridge/osgi/JvbBundleConfig.java
```diff
@@ -264,6 +265,9 @@ private void getCallStatsJavaSDKSystemPropertyDefaults(
         // Trust the bandwidth estimations by default.
         defaults.put(BitrateController.TRUST_BWE_PNAME, true_);
 
+        // Trust the bandwidth estimations by default.
```
bgrozev: Can you please fix the comment (what does this enable exactly)?

src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
```diff
@@ -183,14 +215,16 @@ public void run()
...
         {
-            logger.debug("padding,stream="+ destStream.hashCode()
-                + " padding_bps=" + paddingBps
-                + ",optimal_bps=" + totalOptimalBps
-                + ",current_bps=" + totalCurrentBps
-                + ",target_bps=" + totalTargetBps
-                + ",needed_bps=" + totalNeededBps
-                + ",max_padding_bps=" + maxPaddingBps
-                + ",bwe_bps=" + bweBps);
+            logger.debug("probing" +
```
bgrozev: Why are we going away from the line protocol format?

gpolitis: Thanks for raising this, I will revert these changes and re-introduce them in a later PR so that they're compatible with our log collection infrastructure.

PR #477, opened 2017-05-26
Openfire maven

guusdk:
These commits modify the build process for the Openfire plugin that's part of the JVB project.


guusdk:
Perhaps the entire build.xml file can be deleted?

damencho:
We cannot delete build.xml for now, cause we use it to build the deb package, there is one included ant file from resources/install/build.xml.

If we are removing dist.openfire, should we stop producing a jar file for jvb that people can use as an openfire plugin?
https://download.jitsi.org/jitsi-videobridge/openfire/

openfire/pom.xml
```diff
@@ -12,10 +12,25 @@
...
+            <name>Apache License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
 
     <properties>
         <!-- Override the default plugin name, as the default include illegal characters. -->
         <plugin.name>jitsivideobridge</plugin.name>
+        <videobridge.version>1.0-20170523.204532-356</videobridge.version>
```
damencho: Why there is a particular version here? Is this bound to the version of the plugin?

guusdk: The only alternative that I can think if is to depend on -SNAPSHOT, which on one hand would automatically allow us to follow the changes in videobridge, but on the other hand... well, make use automatically follow the changes in videobridge - even when they introduce breaking changes. :) By pinning the dependency to a specific version, we prevent that.

guusdk:
You should most definitely not stop producing the Openfire plugin! :) (but, you can now build it using Maven, instead of the Ant script.

On a side-note: perhaps the Openfire-plugin source could go into its own repository, instead of be part of the JVB repo. It's completely separated now, as far as I can tell.

PR #481, opened 2017-05-26
feat: Advertise a pre-configured region in statistics.

bgrozev:


src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -259,6 +261,21 @@
      */
     public static final String VIDEOSTREAMS = "videostreams";
 
+    /**
+     * The name of the "region" statistic.
+     */
+    public static final String REGION = "region";
```
paweldomas: you guys do not try to keep alphabetical order here ? just a question

bgrozev: Not that I know of (although I tend to think it would be nice)

PR #483, opened 2017-06-01
feat: Improvements to the bitrate allocation algorithm.

gpolitis:
- don't forward any thumbnails unless the on-stage speaker has 360p.
- reduces the changes in the number of thumbnails that are forwarded

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -438,7 +438,7 @@ public class BitrateController
         }
 
         // Compute the bitrate allocation.
-        List<TrackBitrateAllocation>
+        TrackBitrateAllocation[]
             trackBitrateAllocations = allocate(bweBps, conferenceEndpoints);
 
         // Update the the controllers based on the allocation and send a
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -574,10 +577,12 @@ public void update(List<Endpoint> conferenceEndpoints, long bweBps)
         }
 
         long oldMaxBandwidth = 0;
+        int oldForwardedParticipantsLen = 0;
         while (oldMaxBandwidth != maxBandwidth)
         {
             oldMaxBandwidth = maxBandwidth;
 
+            int forwardedParticipantsLen = 0;
```
bgrozev: I must be missing something -- this seems to be set to 0 and never updated.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -895,6 +926,7 @@ private TrackBitrateAllocation(
             // consider 720p@30fps, 360p@30fps, 180p@30fps, 180p@15fps,
             // 180p@7.5fps while for the thumbnails we consider 180p@30fps,
             // 180p@15fps and 180p@7.5fps
+            int preferredIdx = -1;
```
bgrozev: Why is there an auxiliary local variable? It looks like this.preferredIdx is never accessed in the loop below, and is unconditionally set to the value of preferredIdx in the end.

gpolitis: because the instance field is final and we update the local variable multiple times inside the loop

bgrozev: Ah, I see.

PR #498, opened 2017-07-12
add support for the ReceiverVideoConstraint message from clients

bbaldino:


src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -237,6 +248,7 @@ private static synchronized RecurringRunnableExecutor getRecurringExecutor()
         throws Exception
     {
         super(content, id, channelBundleId, transportNamespace, initiator);
+        this.maxFrameHeight = MAX_FRAME_HEIGHT_DEFAULT;
```
bgrozev: Why not initialize this with the field declaration? If we were to add a second constructor and forget to initialize it there it would default to 0.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -909,12 +917,14 @@ private RateSnapshot(long bps, RTPEncodingDesc encoding)
          */
         private TrackBitrateAllocation(
             Endpoint endpoint, MediaStreamTrackDesc track,
-            boolean fitsInLastN, boolean selected)
+            boolean fitsInLastN, boolean selected, int maxFrameHeight)
         {
+            System.out.println("BB: TrackBitrateAllocation using a maxFrameHeight of " + maxFrameHeight);
```
bgrozev: Use 'logger' if we still need the log?

src/main/java/org/jitsi/videobridge/VideoChannel.java
bbaldino: yup, good idea. will fix.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bbaldino: ack, thought i fixed all of these.  thanks.

bbaldino:
cc @emcho

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -84,6 +84,13 @@
         = "EndpointMessage";
 
     /**
+     * The {@link Videobridge#COLIBRI_CLASS} value indicating a
+     * {@code ReceiverVideoConstraint} message.
+     */
+    public static final String RECEIVER_VIDEO_CONSTRAINT
```
gpolitis: I know it's technically unnecessary, but can we rename this to something like `COLIBRI_CLASS_VIDEO_CONSTRAINT`, since all the other messages have that prefix?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -945,11 +954,15 @@ private TrackBitrateAllocation(
...
+                if (encoding.getHeight() > this.maxFrameHeight)
+                {
+                    continue;
+                }
                 if (selected)
                 {
                     // For the selected participant we favor resolution over
                     // frame rate.
-                    if (encoding.getHeight() < ONSTAGE_PREFERRED_HEIGHT
+                    if (encoding.getHeight() <= ONSTAGE_PREFERRED_HEIGHT
```
gpolitis: This is wrong as it would add 360p at low fps, which is something that we don't want to have.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
bbaldino: definitely...fixed

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bbaldino: ok, will revert back

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -945,6 +954,10 @@ private TrackBitrateAllocation(
             int preferredIdx = 0;
             for (RTPEncodingDesc encoding : encodings)
             {
+                if (encoding.getHeight() > this.maxFrameHeight)
```
saghul: does this work ok with portrait resolutions? an iPhone would send 720x1280, which would not pass the 720 test, but I have no idea if the getHeight method accommodates for this?

gpolitis: Right now it works because of the way the height is discovered/set. But, in the future, where the height is either discovered from the RTP payload or from signaling from the client, it won't. I'm thinking it's more appropriate to change this into a resolution field, which would hold the multiplication result.

Issue #497, opened 2017-07-08
Debian package: use root-owned directories for executable files


Glandos:
When I tried to install and run the Debian packages on top of a grsecurity kernel (the one in the Debian archive), I have a lot of errors like this:

```
grsec: denied untrusted exec (due to not being in trusted group and file in non-root-owned directory) of /usr/share/jitsi-videobridge/jvb.sh
grsec: From 127.0.0.6: denied untrusted exec (due to not being in trusted group and file in non-root-owned directory) of /tmp/jna-105622/jna5088846550930907346.tmp by /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java[java:10856] uid/euid:999/999 gid/egid:1006/1006, parent /usr/bin/sudo[sudo:10684] uid/euid:0/0 gid/egid:1006/1006
grsec: From 127.0.0.6: denied untrusted exec (due to not being in trusted group and file in world-writable directory) of /tmp/jna2712961947937848304.tmp by /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java[java:12409] uid/euid:999/999 gid/egid:1006/1006, parent /usr/bin/sudo[sudo:11993] uid/euid:0/0 gid/egid:1006/1006
```

The explanation is obvious for the first, the start script should be in /usr/bin or any other root-owned directory.
The 2 last were more difficult to track, but they are linked to JNA, and how it tries to load system libraries: it tries first to load the `lib*.so` file in its path, and if it fails, it extracts it in a temporary directory, and run it. JNA itselfs loads `libjnidispatch.so` from `/tmp/jna-105622`, which causes the second error from grsecurity. Then libjitsi tries to load some more, like `libjnopus.so`, and this library is extracted from `libjitsi.jar` into `/tmp` itself, which causes the third error from grsecurity.

`/tmp/jna-105622` is owned by jvb:jitsi, so it is not so important. But `/tmp` is world-writable. Anyone can fool jitsi-videobridge at startup.

I know that grsecurity is not supported, but I suggest to make the following changes for the Debian packaging:
- Put the start script in /usr/bin, not in the home directory
- Extract every native library that jitsi-videobridge uses, and put them in /usr/lib/jitsi/ or something similar, to avoid conflicts with system libraries
- Find a way for JNA to load system libraries from the previous path.

This will also make the package much cleaner, since all executable files will be visible to the administrator.

Glandos:
As a workaround, I have now done the following things:
[...]

PR #502, opened 2017-07-14
[Debian package] Move launcher to /usr/bin

Glandos:
Starts to work on #497.
The launcher is moved from `/usr/share/jitsi-videobridge/jvb.sh` into `/usr/bin/jvb`.
This is *far* from perfect, since it only works for linux-64, but I want to gather comments before going on.
If the way to do it seems wrong, please raise your hand.

(Sorry for other commits, I've rebased too late)

### DIFF
--- a/resources/install/build-debian.xml
+++ b/resources/install/build-debian.xml
@@ -108,10 +108,8 @@
               todir="${debian.dir}/etc/jitsi/videobridge"
               overwrite="true"/>
         <copy file="resources/install/${include-dir}/jvb.sh"
-              todir="${debian.dir}/usr/share/jitsi-videobridge"
+              tofile="${debian.dir}/usr/bin/jvb"
               overwrite="true"/>
-        <chmod file="${debian.dir}/usr/share/jitsi-videobridge/jvb.sh"
-               perm="755"/>
         <copy file="resources/graceful_shutdown.sh"
               todir="${debian.dir}/usr/share/jitsi-videobridge"
               overwrite="true"/>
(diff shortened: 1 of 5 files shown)


resources/install/build-debian.xml
```diff
@@ -123,6 +121,16 @@
...
+        <chown owner="root">
+            <fileset dir="${debian.dir}/usr/bin"></fileset>
+        </chown>
+        <chgrp group="root">
+            <fileset dir="${debian.dir}/usr/bin"></fileset>
+        </chgrp>
+        <chmod perm="755">
+            <fileset dir="${debian.dir}/usr/bin"></fileset>
+        </chmod>
+
```
damencho: Why you need to touch some system folder permissions which are not jvb related.

resources/install/linux-64/jvb.sh
```diff
@@ -15,13 +15,13 @@ if [[ "$1" == "--help"  || $# -lt 1 ]]; then
...
 
-SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+BASE_LIB_DIR="$(echo ~jvb)"
 
 mainClass="org.jitsi.videobridge.Main"
-cp=$(JARS=($SCRIPT_DIR/jitsi-videobridge*.jar $SCRIPT_DIR/lib/*.jar); IFS=:; echo "${JARS[*]}")
-libs="$SCRIPT_DIR/lib/native/linux-64"
-logging_config="$SCRIPT_DIR/lib/logging.properties"
-videobridge_rc="$SCRIPT_DIR/lib/videobridge.rc"
+cp="$BASE_LIB_DIR/*:$BASE_LIB_DIR/lib/*"
```
damencho: I think cp was containing a list of jars and not star paths, not sure whether the later works.

resources/install/build-debian.xml
Glandos: You're right, I just did this for no reason.

resources/install/linux-64/jvb.sh
Glandos: See http://docs.oracle.com/javase/8/docs/technotes/tools/unix/classpath.html#A1100762
and specifically:

> Class path entries can contain the base name wildcard character (*), which is considered equivalent to specifying a list of all of the files in the directory with the extension .jar or .JAR

damencho: I think I've been experiencing this problem and I think there is a reason that current -cp behave like this. Have you tested this?

bgrozev:

PR #504, opened 2017-07-20
Adapts to latest LJ + experimental code that is disabled.

gpolitis:


src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -386,15 +392,37 @@ void setOptimalIndex(int optimalIndex)
 
         RawPacket[] pktsOut = bitstreamController.rtpTransform(pktIn);
 
-        if (!ArrayUtils.isNullOrEmpty(pktsOut)
-            && pktIn.getSSRCAsLong() != targetSSRC)
+        if (REMOVE_VP8_EXTENDED_CONTROL_BITS
+            || (!ArrayUtils.isNullOrEmpty(pktsOut)
```
bgrozev: What is the idea here? Why do we want to execute the block if e.g. pktsOut == null && REMOVE_VP8_EXTENDED_CONTROL_BITS?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -386,15 +392,37 @@ void setOptimalIndex(int optimalIndex)
...
+
+                        if (shiftLen > 0)
+                        {
+                            System.arraycopy(pktOut.getBuffer(),
+                                block.getOffset() + shiftLen,
+                                pktOut.getBuffer(),
+                                block.getOffset(),
+                                block.getLength() - shiftLen);
+
+                            pktOut.setLength(pktOut.getLength() - shiftLen);
```
bgrozev: Doesn't this brake the packet if there is more than one RED block?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -386,15 +392,37 @@ void setOptimalIndex(int optimalIndex)
...
                     pktOut.setSSRC((int) targetSSRC);
+
+                    if (REMOVE_VP8_EXTENDED_CONTROL_BITS)
+                    {
+                        // remove extended control bits
+                        REDBlock block = bitrateController.getVideoChannel()
+                            .getStream().getPrimaryREDBlock(pktOut);
+
+                        int shiftLen = DePacketizer
+                            .VP8PayloadDescriptor.getSize(block) - 1;
```
bgrozev: I assume this is something for Edge. But it looks like it would brake the VP8 RTP payload format: if the X bit is set the length of the PD must be at least 2. Should we clear the X bit after removing the rest of the PD?

gpolitis:
Closing as obsolete.

PR #505, opened 2017-07-25
deb: Log and fail if jvb.sh is not executable.

bgrozev:


bbaldino:
will wait for the test for tradition's sake :)

PR #507, opened 2017-07-26
Changes in preparation for relay support

bgrozev:


src/main/java/org/jitsi/videobridge/Channel.java
```diff
@@ -196,15 +196,21 @@ public Channel(
         throws Exception
     {
         if (content == null)
+        {
```
bbaldino: 👍

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -1205,7 +1205,7 @@ private void describe(
      */
     private void describeDtlsControl(IceUdpTransportPacketExtension transportPE)
     {
-        DtlsControlImpl dtlsControl = getDtlsControl(/* channel */ null);
+        DtlsControlImpl dtlsControl = getSrtpControl(/* channel */ null);
```
bbaldino: i'm surprised this works without a cast, shouldn't this store the result in an `SrtpControl` (and then check for `instanceof DtlsControlImpl`)?

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -1445,7 +1445,7 @@ public BundleContext getBundleContext()
      * {@inheritDoc}
      */
     @Override
-    public DtlsControlImpl getDtlsControl(Channel channel)
+    public DtlsControlImpl getSrtpControl(Channel channel)
```
bbaldino: Ah, so this is why the above worked without the cast.  Shouldn't this return `SrtpControl`?

src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -978,9 +978,10 @@ public void removeChannelListener(WebRtcDataStreamListener listener)
     private void runOnDtlsTransport(StreamConnector connector)
         throws IOException
     {
-        DtlsControlImpl dtlsControl
-            = getTransportManager().getDtlsControl(this);
-        DtlsTransformEngine engine = dtlsControl.getTransformEngine();
+        SrtpControl dtlsControl
```
bbaldino: variable `dtlsControl` -> `srtpControl` ?

src/main/java/org/jitsi/videobridge/octo/OctoRelayImpl.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+package org.jitsi.videobridge.octo;
+
+import org.ice4j.socket.*;
+import org.jitsi.util.*;
+
+import java.net.*;
+
+/**
+ * Implements a bridge-to-bridge (Octo) relay.
```
bbaldino: could we add some more description around what it means to implement that relay?  i.e. what this class actually implements/is responsible for?

src/main/java/org/jitsi/videobridge/octo/OctoRelayImpl.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * The socket used to send and receive Octo packets.
+     */
+    private MultiplexingDatagramSocket socket;
+
+    /**
+     * The Octo relay ID which will be advertised by jitsi-videobridge.
+     */
+    private String relayId;
```
bbaldino: i don't understand what this represents, could you elaborate?

src/main/java/org/jitsi/videobridge/octo/OctoRelayService.java
```diff
@@ -0,0 +1,130 @@
...
+    public void start(BundleContext bundleContext)
+    {
+        cfg
+            = ServiceUtils.getService(
+                    bundleContext, ConfigurationService.class);
+
+        String address = cfg.getString(ADDRESS_PNAME, null);
+        int port = cfg.getInt(PORT_PNAME, -1);
+
+
```
bbaldino: nit: extra newline

src/main/java/org/jitsi/videobridge/octo/OctoRelayImpl.java
```diff
@@ -0,0 +1,94 @@
...
+import org.jitsi.util.*;
+
+import java.net.*;
+
+/**
+ * Implements a bridge-to-bridge (Octo) relay.
+ *
+ * @author Boris Grozev
+ */
+public class OctoRelayImpl
```
bbaldino: should this be named with 'Impl' if it's not an implementation of any interface?  just a question, not sure what our convention is.

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
bgrozev: If it did we would have to cast. IceUdpTransportManager will still use Dtls, another type of SrtpControl is used in OctoTransportManager (not included in this PR)

src/main/java/org/jitsi/videobridge/SctpConnection.java
bgrozev: Well, it really is a DtlsControl (otherwise the cast below wouldn't work). I think the name of the variable could be a useful clue.

bgrozev: But I don't mind changing it if you prefer

src/main/java/org/jitsi/videobridge/octo/OctoRelayImpl.java
bgrozev: Added a description

src/main/java/org/jitsi/videobridge/octo/OctoRelayImpl.java
bgrozev: Renamed to OctoRelay

src/main/java/org/jitsi/videobridge/octo/OctoRelayImpl.java
bgrozev: Added a description and an example

src/main/java/org/jitsi/videobridge/octo/OctoRelayService.java
bgrozev: Removed

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
bbaldino: i think having the method be called `getSrtpControl` but returning `DtlsControlImpl` is confusing, but i get what you're saying.  But, why do we need to cast at all?  this seems to defeat a lot of the purpose of having a class hierarchy (where the behavior differs but the API is the same, so the caller doesn't have to know/care which underlying subclass it's actually dealing with, only that it implements the SrtpControl interface).

src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
```diff
@@ -21,18 +21,21 @@
 import java.net.*;
 
 /**
- * Implements a bridge-to-bridge (Octo) relay.
+ * Implements a bridge-to-bridge (Octo) relay. This class is responsible for
+ * holding the {@link DatagramSocket} which is to be used for bridge-to-bridge
+ * communication, and the "relay ID" which other bridges use to discover the
+ * socket address on which this bridge is accessible.
```
bbaldino: this is great, thank you.

src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
```diff
@@ -41,16 +44,20 @@
 
     /**
      * The Octo relay ID which will be advertised by jitsi-videobridge.
+     * Other bridges can use this ID in order to discover the socket address
+     * that this bridge is accessible on. With the current implementation the
+     * ID just encodes a pre-configured IP address and port,
+     * e.g. "10.0.0.1:20000"
```
bbaldino: thanks!  so, other bridges must be able to use this id directly (without any other mapping) to know how to reach this bridge?  if so, it will have to be an address, won't it? does it make sense to define this as a String as opposed to something that more strictly models an address and would be serialized? (and i'm assuming it would be something like InetAddress here)

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
bgrozev: The base method (in TransportManager) is called getSrtpControl and returns SrtpControl. IceUdpTransportManager uses DTLS specifically (because it has to set fingerprints for example), and as such return the more specific class. The caller can use SrtpControl if it doesn't care what the implementation is, or if it knows that the transport manager is IceUdpTransportManager, it can treat the value as DtlsControlImpl (without a cast). I don't think this defeats the purpose of the hierarchy.

src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
bgrozev: They need to use the ID to find how to reach the bridge, but they might or might not use an extra mapping. For the first implementation I'm planning to use an address:port pair, but I wanted to leave it more abstract in case we need to change it later (e.g. to use a domain name, or something else)

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
bbaldino: that does make sense, but does any external caller of `IceUdpTransportManager::getSrtpControl` need to specifically know it's using a `DtlsControlImpl`?

src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
bbaldino: ok cool, sounds good.

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
bgrozev: No, I don't see anything outside IceUdpTransportManager which uses the return value as DtlsControl.

bgrozev: Changed to return SrtpControl

src/main/java/org/jitsi/videobridge/SctpConnection.java
bbaldino: since we changed the others, might as well change this name as well?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -240,15 +246,21 @@ public Conference(Videobridge videobridge,
                       String id,
                       String focus,
                       String name,
-                      boolean enableLogging)
+                      boolean enableLogging,
+                      String gid)
```
guusdk: The lack of a null-check on `gid`, while the same commit does add such checks for others, implies that `gid`  can be null. Could this be documented explicitly, both in the constructor, as well as in the getter?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1852,6 +1864,16 @@ public static String getLoggingId(Conference conference)
     }
 
     /**
+     * Gets the global ID of the conference.
+     * @return
```
guusdk: It would be helpful to know that the value returned can be `null`

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -107,6 +107,12 @@
     private final String id;
 
     /**
+     * The "global" id of this conference, set by the controller (e.g. jicofo)
+     * as opposed to the bridge.
+     */
```
guusdk: What purpose does this new field have? Why is a 'global' identifier useful to have? In what sense is it "global"?

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -233,9 +233,9 @@ public Videobridge()
      * @return a new <tt>Conference</tt> instance with an ID unique to the
      * <tt>Conference</tt> instances listed by this <tt>Videobridge</tt>
      */
-    public Conference createConference(String focus, String name)
+    public Conference createConference(String focus, String name, String gid)
```
guusdk: If `gid` can indeed be null, changing the method signature here breaks the API (I know it did for me, which is what triggered me to write this in the first place). Perhaps adding overloaded methods (where `gid` is null) would be good (which would restore backwards compatibility). Also, the new argument is not defined in javadoc.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -258,7 +258,7 @@ public Conference createConference(String focus, String name)
      * <tt>Conference</tt> instances listed by this <tt>Videobridge</tt>
      */
     public Conference createConference(
-        String focus, String name, boolean enableLogging)
+        String focus, String name, boolean enableLogging, String gid)
```
guusdk: (same comment as above)

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: It's unique across a set of jicofo instances and a set of jitsi-videobridge instances.

bgrozev:
Thanks, @guusdk! I've opened #520, which adds the missing documentation.

I haven't added the new constructors that you suggested because the code has already been merged, but I'll have that in mind for the next time.

PR #508, opened 2017-07-28
Rename variables/add helper methods in SimulcastController

bbaldino:


bbaldino:
i made sure all these commits were independent, so i think we should **not** squash this pr before merging

bbaldino:
also, this relies on https://github.com/jitsi/libjitsi/pull/346

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1066,8 +1066,6 @@ void setOptimalIndex(int newOptimalIdx)
          */
         void rtcpTransform(ByteArrayBuffer baf)
         {
-            SeenFrame maxSentFrame = this.maxSentFrame;
-
```
bgrozev: I think the local variable is there intentionally -- this.maxSentFrame can be modified by other threads.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1088,23 +1088,23 @@ void rtcpTransform(ByteArrayBuffer baf)
          */
         private int getMaxSeqNum()
         {
-            return mostRecentSentFrame == null ? seqNumOff : mostRecentSentFrame.getMaxSeqNum();
+            return !haveSentFrame() ? seqNumOff : mostRecentSentFrame.getMaxSeqNum();
```
bgrozev: I would invert these

bgrozev:
> also, this relies on jitsi/libjitsi#346

Maybe re-open with the same branch name so that the tests run both?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: didn't realize that...nice catch, thanks.  will restore it and add a comment.

bbaldino: after looking at this more, if multiple threads access this variable then i think we have issues in other parts of this class as well (access to this is never synchronized).  @gpolitis do you know?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: i like that, reads better.  fixed.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -787,23 +787,23 @@ public boolean accept(
...
-                    || TimeUtils.rtpDiff(srcTs, mrKeyFrame.srcTs) > 0;
+                boolean isNewerThanMostRecentKeyFrame = !haveSentKeyFrame()
+                    || RTPUtils.isNewerTimestampThan(srcTs, mostRecentSentKeyFrame.srcTs);
 
-                if (currentIdx > -1
+                if (!isSuspended()
                     // we haven't seen anything yet and this is an independent
                     // frame.
-                    && (maxSentFrame == null && sourceFrameDesc.isIndependent()
+                    && (!haveSentFrame() && sourceFrameDesc.isIndependent()
```
gpolitis: I'm not the english expert here, but shouldn't that be has?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
gpolitis: Like Boris said, the local variable here is intentional. What kind of problems do you see @bbaldino ?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: i think both are correct, it just depends on what perspective you name the variable from...i was thinking of it from the point of the class itself, so it'd be "we" or "i" ("have i/we sent a frame yet?" -> `haveSentFrame`) but it'd also be correct to say "has it/this class sent a frame yet?" -> `hasSentFrame`.  not sure i feel strongly about one way vs the other.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: after talking with @gpolitis it is only this method which needs to have `maxSentFrame` protected, as it can be called from a different thread than `accept`.  it's only written in `accept` itself (and `setTL0Idx`, which is called from `accept`) so the write there are fine.

bbaldino: so i will revert this change (to keep the local variable) and add a comment

bbaldino: @gpolitis after looking even further at this, i'm still a bit confused.  this assignment doesn't actually protect this method from `maxSeenFrame` changing behind its back (as we just create an alias here, not a deep copy of the member), so this will only protect us from the other thread setting `this.maxSeenFrame` to null.  it won't, for example, prevent the other thread from changing `maxSeenFrame.tsTranslation` to null behind our back (which could theoretically cause an NPE here)

bbaldino: @gpolitis is that all you intended to protect against?

gpolitis: tsTranslation is conceptually final, so it can't be changed to null (it's not marked as such because we're reusing/resetting the SeenFrames). However, if we don't store the frame in a local variable, next time we check the frame might have changed to something with tsTranslation == null.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
gpolitis: I don't mind at all, I was just wondering if it's a typo.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: gotcha, ok thanks @gpolitis

bgrozev:
@damencho just said that the failing test fails intermittently, so we can consider this run successful: http://pr-reports.jitsi.org/jitsi-videobridge-PR-test/547

bbaldino:
ok cool, i rolled in the libjitsi update to the commit that depends on it so that part should be addressed now.  just a reminder that, when we're ready, NOT to squash this one.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
virtuacoplenny: In the comment for haveSentFrame you wrote "has sent a frame" but the method is called "haveSentFrame." I think if both George and I felt a little off by the naming then it's worth re-examining to make sure it makes as much sense as you want.

gpolitis:
jenkins please test

damencho:
jenkins test this please

PR #511, opened 2017-08-02
Move the RawPacket API further up the accept/translatorWillWrite chain

bbaldino:
going any further requires changing the RtpTranslator#WriteFilter interface in libjitsi which involves lots of changes, so will save those for another PR


bbaldino:
also, forgot to mark the commits as such but i think we should squash all this

PR #512, opened 2017-08-02
minor/cosmetic refactoring in SimulcastController

bbaldino:
relies on https://github.com/jitsi/libjitsi/pull/349#issuecomment-319764527

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
@@ -103,7 +103,7 @@ public class SimulcastController
      * The SSRC to protect when probing for bandwidth and for RTP/RTCP packet
      * rewritting.
      */
-    private final long targetSSRC;
+    private final long targetSsrc;
 
     /**
      * The running index for the temporal base layer frames, i.e., the frames
@@ -135,11 +135,11 @@ public class SimulcastController
         RTPEncodingDesc[] rtpEncodings = source.getRTPEncodings();
         if (ArrayUtils.isNullOrEmpty(rtpEncodings))
         {
-            targetSSRC = -1;
+            targetSsrc = -1;
         }
         else
         {
-            targetSSRC = rtpEncodings[0].getPrimarySSRC();
+            targetSsrc = rtpEncodings[0].getPrimarySSRC();
         }
 
         bitstreamController = new BitstreamController();
@@ -154,7 +154,7 @@ public class SimulcastController
      */
     long getTargetSSRC()
     {
-        return targetSSRC;
+        return targetSsrc;
     }
 
     /**
@@ -225,8 +225,8 @@ public class SimulcastController
             && (currentTL0IsActive || targetTL0Idx < 0 /* => currentIdx < 0 */))
         {
             // An intra-codec/simulcast switch is NOT pending.
-            long sourceSSRC = sourceFrameDesc.getRTPEncoding().getPrimarySSRC();
-            boolean accept = sourceSSRC == bitstreamController.getTL0SSRC();
+            long sourceSsrc = sourceFrameDesc.getRTPEncoding().getPrimarySSRC();
+            boolean accept = sourceSsrc == bitstreamController.getTL0SSRC();
 
             if (!accept)
             {
@@ -255,8 +255,8 @@ public class SimulcastController
         {
             // An intra-codec switch requires a key frame.
 
-            long sourceSSRC = sourceFrameDesc.getRTPEncoding().getPrimarySSRC();
-            boolean accept = sourceSSRC == bitstreamController.getTL0SSRC();
+            long sourceSsrc = sourceFrameDesc.getRTPEncoding().getPrimarySSRC();
+            boolean accept = sourceSsrc == bitstreamController.getTL0SSRC();
 
             if (!accept)
             {
@@ -359,7 +359,7 @@ public class SimulcastController
 
             ((RTPTranslatorImpl) sourceStream.getRTPTranslator())
                 .getRtcpFeedbackMessageSender().sendFIR(
-                (int) targetSSRC);
+                (int) targetSsrc);
         }
     }
 
@@ -391,14 +391,14 @@ public class SimulcastController
         RawPacket[] pktsOut = bitstreamController.rtpTransform(pktIn);
 
         if (!ArrayUtils.isNullOrEmpty(pktsOut)
-            && pktIn.getSSRCAsLong() != targetSSRC)
+            && pktIn.getSSRCAsLong() != targetSsrc)
         {
-            // Rewrite the SSRC of the output RTP stream.
+            // Rewrite the SSSRC of the output RTP stream.
             for (RawPacket pktOut : pktsOut)
             {
                 if (pktOut != null)
                 {
-                    pktOut.setSSRC((int) targetSSRC);
+                    pktOut.setSSRC((int) targetSsrc);
                 }
             }
         }
@@ -447,7 +447,7 @@ public class SimulcastController
                     bitstreamController.rtcpTransform(baf);
 
                     // Rewrite senderSSRC
-                    RTCPHeaderUtils.setSenderSSRC(baf, (int) targetSSRC);
+                    RTCPHeaderUtils.setSenderSSRC(baf, (int) targetSsrc);
                 }
                 break;
                 case RTCPPacket.BYE:
@@ -551,7 +551,7 @@ public class SimulcastController
          * The special value -1 indicates that we're not accepting (and hence
          * not forwarding) anything.
          */
-        private long tl0SSRC = -1;
+        private long tl0Ssrc = -1;
 
         /**
          * The subjective quality index for of the TL0 of this instance.
@@ -690,13 +690,13 @@ public class SimulcastController
             if (ArrayUtils.isNullOrEmpty(rtpEncodings) || tl0Idx < 0)
             {
                 this.availableQualityIndices = null;
-                this.tl0SSRC = -1;
+                this.tl0Ssrc = -1;
                 this.isAdaptive = false;
             }
             else
             {
                 tl0Idx = rtpEncodings[tl0Idx].getBaseLayer().getIndex();
-                tl0SSRC = rtpEncodings[tl0Idx].getPrimarySSRC();
+                tl0Ssrc = rtpEncodings[tl0Idx].getPrimarySSRC();
 
                 // find the available qualities in this bitstream.
 
@@ -1058,7 +1058,7 @@ public class SimulcastController
          */
         RawPacket[] rtpTransform(RawPacket pktIn)
         {
-            if (pktIn.getSSRCAsLong() != tl0SSRC)
+            if (pktIn.getSSRCAsLong() != tl0Ssrc)
             {
                 return null;
             }
@@ -1133,7 +1133,7 @@ public class SimulcastController
          */
         long getTL0SSRC()
         {
-            return tl0SSRC;
+            return tl0Ssrc;
         }
 
         /**
@@ -1285,7 +1285,7 @@ public class SimulcastController
             RawPacket[] rtpTransform(RawPacket pktIn)
             {
                 RawPacket[] pktsOut;
-                long srcSSRC = pktIn.getSSRCAsLong();
+                long srcSsrc = pktIn.getSSRCAsLong();
 
                 if (maybeFixInitialIndependentFrame)
                 {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -722,6 +722,37 @@ void setTL0Idx(int newTL0Idx)
...
+         * //FIXME(brian): this could be static, but can't add a static method
+         * to the inner class.  should it be put elsewhere and made a static
+         * helper method? (the outer class?)
+         */
+        private int findMaximumQualityIndex(int[] availableQualityIndices, int targetIdx)
+        {
+            int maxIndex = -1;
+            if (availableQualityIndices != null)
+            {
+                for (int i = 0; i < availableQualityIndices.length; ++i)
```
bgrozev: I think this changes the semantics a little. Suppose availableQualityIndices=[100], targetIdx=50. The old code would set currentIdx to 100 (line 767), while the new code leaves it at -1.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1219,37 +1220,53 @@ private boolean isSuspended()
...
+                    if (frame.lastSequenceNumberKnown())
+                    {
+                        srcSeqNumLimit = frame.getEnd();
+                    }
+                    else
+                    {
+                        srcSeqNumLimit = frame.getMaxSeen();
+                    }
+                }
+                if (frame.firstSequenceNumberKnown())
```
bgrozev: Again a tiny change of semantics: with the old code, if srcSeqNumLimit is -1, then srcSeqNumStart is not updated (line 1246). Not saying this is a problem, just asking if it's has been considered.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: i don't quite follow your specific example (maybe there's a typo?) but i think i see what you're referring to...you're right, but i'm confused by this actually.  the old logic effectively means we could choose a quality index that was larger than than our target which doesn't seem correct? e.g.

```
availableQualityIndices = {10, 11, 12, 13}
targetIdx = 4;
```

we would do `currentIdx = availableQualityIndices[0];` without checking if it was `<= targetIdx` before, right?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -48,7 +48,7 @@
     /**
      * The property name of the number of seen frames to keep track of.
      */
-    public static final String SEEN_FRAME_HISTORY_SIZE_PNAME
+    private static final String SEEN_FRAME_HISTORY_SIZE_PNAME
```
bgrozev: Our policy is for all property names to be public.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: yes nice catch!  will fix this.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: darn, this one doesn't get used anywhere externally (nor does it seem like it should be accessible).  but if that's the case i will change it back.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bgrozev: Yes, that's what I mean (by "= [100]" I meant it's an array of length 1). No idea which one is correct. @gpolitis ?

bbaldino:
Jenkins, test this please

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -391,14 +391,14 @@ void setOptimalIndex(int optimalIndex)
         RawPacket[] pktsOut = bitstreamController.rtpTransform(pktIn);
 
         if (!ArrayUtils.isNullOrEmpty(pktsOut)
-            && pktIn.getSSRCAsLong() != targetSSRC)
+            && pktIn.getSSRCAsLong() != targetSsrc)
         {
-            // Rewrite the SSRC of the output RTP stream.
+            // Rewrite the SSSRC of the output RTP stream.
```
virtuacoplenny: Should "SSSRC" have three S's?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: talked with george and clarified a few things...we do want it with the old behavior.  couple other tweaks i want to make to this as well so will work on those.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: how tf did this even happen.  will fix it.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1205,41 +1234,81 @@ private boolean isSuspended()
...
-                    || availableQualityIndices == null
-                    || availableQualityIndices.length < 2)
-                {
-                    int end = source.getEnd();
-                    srcSeqNumLimit
-                        = end != -1 ? source.getEnd() : source.getMaxSeen() ;
-                }
-
-                if (srcSeqNumLimit == -1)
+                if (!isAccepting())
```
virtuacoplenny: You mentioned that you might want to change isAccepting and I agree. I don't think it provides enough information on what accepting means. Maybe a phrase that describes why it is or is not accepting is more appropriate. Normally I'd say it's good enough but you are focusing on renaming so...

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1205,41 +1234,81 @@ private boolean isSuspended()
...
              *
-             * @param source the {@link FrameDesc} that the RTP packet belongs
-             * to.
              * @param pkt the packet for which to decide to accept
              * @return <tt>true</tt> to allow the specified packet/<tt>buffer</tt>
              * to be written into the arbitrary target/receiver that owns this
              * {@link SeenFrame} ; otherwise, <tt>false</tt>
              */
-            boolean accept(FrameDesc source, RawPacket pkt)
+            boolean accept(RawPacket pkt)
```
virtuacoplenny: As someone with really no context here, isAccepting and accept (shouldAccept) existing together kind of confuses me. They seem to do the same thing based on name.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1205,41 +1234,81 @@ private boolean isSuspended()
...
+            {
+                return srcSeqNumLimit != -1;
+            }
+
+            /**
+             * Update this {@link SeenFrame} with information from its
+             * corresponding {@link FrameDesc}
+             * @param frame the {@link FrameDesc} from which to update
+             */
+            void update(FrameDesc frame)
```
virtuacoplenny: I think this could be renamed to something more descriptive than "update." (Ignore if people that would be reading this generally would know what that means.)

bbaldino:
took a shot at another approach here.  rather than independently updating the sequence numbers of a frame in both SeenFrame and FrameDesc (where there is a 1:1 relationship), I changed SeenFrame to just hold a reference to the FrameDesc it corresponds to.  this seems to work in a basic test, but i'm worried there may be a corner case this doesn't handle correctly...let me know what you think @gpolitis

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1191,67 +1216,139 @@ private boolean isSuspended()
...
+                // We'll accept the packet if it's no newer than our srcSeqNumLimit
+                // FIXME(brian): in what scenario will this test fail?  mostRecentSentFrame
+                // will always, at least, be equal to the frame to which this
+                // sequence number belongs, and we update srcSeqNumLimit based
+                // on the highest seen sequence number of that frame.  if seqNum
+                // was from a new frame
+                boolean accept =
+                    RTPUtils.isOlderSequenceNumberThan(
+                        seqNum,
+                        RTPUtils.applySequenceNumberDelta(frameDesc.getMaxSeen(), 1));
```
bbaldino: @gpolitis wondering about this.  given that `pkt` is a packet for this frame, how could it have a sequence number larger than this frame's max sequence number?

bgrozev: "max seen" is not the frame's max sequence number, just the latest that was "seen".

bbaldino: i understand that, but `seqNum` comes from this frame, so `maxSeen` will have been updated based on this packet, meaning it can't possibly be newer than `maxSeen`, no?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -815,7 +843,7 @@ public boolean accept(
                     // TODO ask for independent frame if we're skipping a TL0.
 
                     int sourceIdx = sourceFrameDesc.getRTPEncoding().getIndex();
-                    if (sourceEncodings[currentIdx].requires(sourceIdx))
+                    if (sourceEncodings[this.currentIdx].requires(sourceIdx))
```
gpolitis: This needs to be newCurrentIdx

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -759,40 +804,23 @@ public boolean accept(
...
+                            + ",new_idx=" + newCurrentIdx);
                     }
-
-                    this.currentIdx = currentIdx;
+                    this.currentIdx = newCurrentIdx;
                 }
 
                 boolean isNewerThanMostRecentFrame = !haveSentFrame()
-                    || RTPUtils.isNewerTimestampThan(srcTs, mostRecentSentFrame.srcTs);
+                    || RTPUtils.isNewerTimestampThan(srcTs, mostRecentSentFrame.getTs());
```
gpolitis: This is going to create problems.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -759,40 +804,23 @@ public boolean accept(
...
 
                 boolean isNewerThanMostRecentKeyFrame = !haveSentKeyFrame()
-                    || RTPUtils.isNewerTimestampThan(srcTs, mostRecentSentKeyFrame.srcTs);
+                    || RTPUtils.isNewerTimestampThan(srcTs, mostRecentSentKeyFrame.getTs());
```
gpolitis: Same here.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1177,10 +1205,7 @@ private boolean isSuspended()
              */
             private boolean maybeFixInitialIndependentFrame = true;
 
-            /**
-             * The source timestamp of this frame.
-             */
-            private long srcTs;
+            private FrameDesc frameDesc;
```
gpolitis: I'm not sure what the problem was with the old approach, but I think this may create memory leaks because the frame keeps a reference to the encoding -> track -> stream..

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1191,67 +1216,139 @@ private boolean isSuspended()
...
              * to be written into the arbitrary target/receiver that owns this
              * {@link SeenFrame} ; otherwise, <tt>false</tt>
              */
-            boolean accept(FrameDesc source, RawPacket pkt)
+            boolean accept(RawPacket pkt)
             {
-                if (this == mostRecentSentFrame /* the max frame can expand */
-                    || availableQualityIndices == null
-                    || availableQualityIndices.length < 2)
-                {
```
gpolitis: It's a bit hard for me to understand where all this logic went.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
bbaldino: `this.currentIdx` has already been updated to the new value of `newCurrentIdx` by here, so i think this should be alright?

bbaldino:
i simplified this back down to mostly cosmetic changes.  after talking with george putting FrameDesc in SeenFrame doesn't make sense and getting into those changes was a bit meaty for this PR anyway, so will keep it simple.

PR #516, opened 2017-08-08
Refactor SimulcastController#accept

bbaldino:


bbaldino:
@gpolitis consider this a WIP at this point, but wanted to get some feedback from you on how this looks (still have a few open questions in the code).  This does appear to be (at least, fundamentally) working...I have joined with a couple participants and see movements between layers looking alright.

PR #518, opened 2017-08-10
Break out BitstreamController and SeenFrame, add unit tests for SeenFrame

bbaldino:
depends on https://github.com/jitsi/libjitsi/pull/359


bbaldino:
i started working on unit tests for BitstreamController next, but it's a bit tricky so though it might be good to get this one posted now

src/test/java/org/jitsi/videobridge/cc/SeenFrameTest.java
```diff
@@ -0,0 +1,93 @@
...
+                // Current frame, adaptive stream, sequence number not in range
+                {   5,              10,             12,         true,               true,       false },
+                // Old frame, non-adaptive stream, sequence number in range
+                {   5,              10,             7,          false,              false,      true },
+                // Old frame, non-adaptive stream, sequence number not in range
+                {   5,              10,             12,         false,              false,      false },
+                // Old frame, adaptive stream, sequence number in range
+                {   5,              10,             7,          false,              true,       false },
+                // Old frame, adaptive stream, sequence number not in range
+                {   5,              10,             12,         false,              true,       false },
```
virtuacoplenny: Would it be better to put these values in a variable and then re-use or is that too much? That might reduce the comments.

bbaldino: you mean specifically the values `5`, `10`, `7` and `12`?  i think that'd be a bit weird because i don't know what i'd name them (the only thing interesting about them is their relationship to one another, other than that they're totally arbitrary).  Even if they were named, i'd still keep the comment above each describing the scenario at a higher level.  i did think about:
1) turning the first 3 values into a boolean `packetWithinFrame` which would be true or false.  this then corresponds to the way it's described in the comment, but then the test would need to translate that into a set of 3 sequence numbers to meet that criteria didn't feel all that much better than what it is now
2) putting the 3 seq num values in a mini-class and having one called `packetWithinFrame` and another `packetOutsideFrame` but this is pretty much the same as the first idea, just executed differently.

neither seemed great enough to replace what was there.

jitsi-developers:
You could call 7 and 12 "inRange" and "outOfRange" respectively.  That way
you could drop the "sequence number in range" comment (although I notice
that the final bool also represents "in range").

Martin

jitsi-developers:
I'd also tend to just label the columns - rather describe each scenario.

Martin

PR #519, opened 2017-08-11
fix a downscaling issue

bbaldino:
fix an issue with downscaling where we wouldn't switch to a quality that was lower than the target index even if the target index relied on it

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
@@ -187,7 +187,8 @@ public class SimulcastController
      * a step closer (or all the way) to the targetQualityIndex
      */
     private boolean shouldSwitchToStream(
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -288,7 +299,7 @@ public boolean accept(RawPacket pkt)
         if (currentBaseLayerIndex != targetBaseLayerIndex)
         {
             // We do want to switch to another layer
-            if (shouldSwitchToStream(currentIndex, sourceLayerIndex, targetIndex))
+            if (shouldSwitchToStream(currentIndex, sourceLayerIndex, targetIndex, sourceEncodings))
```
gpolitis: This is why the original code was comparing TL0 indices. Can we change this fix to match the original behavior?

PR #521, opened 2017-08-16
fix: Catch more Exceptions.

bgrozev:


bbaldino:
do we definitely want to catch all exceptions here, and not just IOException and whatever the new one is?

PR #526, opened 2017-08-22
change jetty websocket endpoint send to the async method to avoid the…

bbaldino:
… multi-threaded access issues we were seeing with the synchronous version

### DIFF
--- a/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
+++ b/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
@@ -229,19 +229,12 @@ class EndpointMessageTransport
     private void sendMessage(
         ColibriWebSocket dst, String message, String errorMessage)
     {
-        try
-        {
-            dst.getRemote().sendString(message);
-            endpoint.getConference().getVideobridge().getStatistics()
-                .totalColibriWebSocketMessagesSent.incrementAndGet();
-        }
-        catch (IOException ioe)
-        {
-            logger.error(
-                "Failed to send a message over a WebSocket (endpoint="
-                    + endpoint.getID() + "): " + errorMessage,
-                ioe);
-        }
+        // We'll use the async version of sendString since this may be called
+        // from multiple threads.  It's just fire-and-forget though, so we
+        // don't wait on the result
+        dst.getRemote().sendStringByFuture(message);
+        endpoint.getConference().getVideobridge().getStatistics()
+            .totalColibriWebSocketMessagesSent.incrementAndGet();
     }
 
     /**


bbaldino:
@bgrozev note that this version doesn't throw, so maybe we won't need the other exception change after all.

ibauersachs:
Please shorten the commit title. Best practice is 50 characters or less; a detailed description can follow afterwards (with a max len of 72 chars).

bgrozev:
LGTM

bgrozev:
I think I can fix the commit message in github while merging

ibauersachs:
Don't bother much with this particular PR if it's a hassle, but please, _please_ read [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/), and try to follow it in the future.

bbaldino:
@ibauersachs, from that article:

> 50 characters is not a hard limit, just a rule of thumb. Keeping subject lines at this length ensures that they are readable, and forces the author to think for a moment about the most concise way to explain what’s going on...So shoot for 50 characters, but consider 72 the hard limit.

i try to make my commit messages informative and concise and while this message may be longer than 50 characters, i do think it achieves what's described there and isn't egregiously long.  i always try to provide both what and some sense of why, and i thought this was short enough that it was alright to keep to a single line--but i'll try and be more aware and stick to the 72 character limit in the future.

also, you're not suggesting the entire message be limited to 72 characters, correct?  just the first line?

ibauersachs:
The first line (the subject) should be less than 50, but as that doc says, 72 should be a hard limit. The entire message can be a novel, if you like :-)

For this case, e.g.:
```
Send messages to websocket endpoints async

This prevents multi-threaded access issues we were seeing with the synchronous
version.
```

Btw, I only now looked at the change itself. What about the now unused `errorMessage` parameter? And the now incorrect Javadoc?

bbaldino:
> The first line (the subject) should be less than 50, but as that doc says, 72 should be a hard limit. The entire message can be a novel, if you like :-)
👍 

> Btw, I only now looked at the change itself. What about the now unused errorMessage parameter? 
The error message I left because I thought it might be something worth keeping/potentially used, but I do think there's a valid argument to be made that it should be removed.

> And the now incorrect Javadoc?
What part is incorrect?

ibauersachs:
> Which part is incorrect?

The errorMessage parameter description (saying that this is logged on failure).

jitsi-developers:
On 22 August 2017 at 23:35, Ingo Bauersachs <notifications@github.com>
wrote:

> The first line (the subject) should be less than 50, but as that doc says,
> 72 should be a hard limit.
>
The entire message can be a novel, if you like :-)
>
But a novel broken up into lines of no more than 72 characters if at all
possible.
(Exceptions are for things like quoted error messages and URLs)

> For this case, e.g.:
>
> Send messages to websocket endpoints async
>
> I'd prefer "Send message to websocket endpoints asynchronously"  (which
*just* fits).

-- 
Martin

PR #527, opened 2017-08-28
fix: Update the bitrate controller of new participant.

gpolitis:
Consider the case where there is no audio (so no active speaker events),
and a there's a single video sender, all other participants only
receive video.

When a new participants join, nothing updates the bitrate controller, so the
new participant has no simulcast controller, which results in nothing
being forwarded towards this participant, until the next participant
joins.

So, when a new participant joins the call, his/her bitrate controller needs
to be explicitely updated.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/VideoChannel.java
+++ b/src/main/java/org/jitsi/videobridge/VideoChannel.java
@@ -336,6 +336,8 @@ public class VideoChannel
     {
         super.initialize(rtpLevelRelayType);
 
+        bitrateController.update(null, -1);
+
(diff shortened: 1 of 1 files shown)


bbaldino:
i see a couple places where update gets called on the `BitrateController`, will none of these kick in?

`VideoChannel#setRtpEncodingParameters` this one seems like it would be called every time?
`VideoChannel#setLastN` I can't remember if we always call this by default or not

gpolitis:
We have a report from a user that none of the updates kick in. Not sure this fixes it, so please don't merge this yet.

PR #529, opened 2017-08-30
feat: Implements send-side bandwidth estimations.

gpolitis:
Depends on https://github.com/jitsi/libjitsi/pull/361

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -1412,6 +1412,8 @@ public class RtpChannel
     {
         if ((rtpHeaderExtensions != null) && (rtpHeaderExtensions.size() > 0))
         {
+            stream.clearRTPExtensions();
+
             for (RTPHdrExtPacketExtension rtpHdrExtPacketExtension
                     : rtpHeaderExtensions)
             {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1412,6 +1412,8 @@ public void setRtpHeaderExtensions(
     {
         if ((rtpHeaderExtensions != null) && (rtpHeaderExtensions.size() > 0))
         {
+            stream.clearRTPExtensions();
```
bbaldino: interesting, so not having this before was a bug, ya?

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -336,6 +336,8 @@ void initialize(RTPLevelRelayType rtpLevelRelayType)
     {
         super.initialize(rtpLevelRelayType);
 
+        bitrateController.update(null, -1);
```
bbaldino: so this obsolesces https://github.com/jitsi/jitsi-videobridge/pull/527, ya?

gpolitis: right, just closed it.

src/main/java/org/jitsi/videobridge/RtpChannel.java
gpolitis: Not a bug, it was just not being used.

PR #536, opened 2017-09-18
Fix logging wrong set of endpoints

debugmaster:
@gpolitis small issue in debug messages.


cmoreira-daitan:
I have signed already.

Issue #537, opened 2017-09-19
Expire always set to 60 when SctpConnection is patched


oliverhausler:
When I create a new `SctpConnection` using REST, I can give it an expiration time, which is reflected in the response.

```
"expire": 15
```

But as soon as I modify an existing SctpConnection, "expire" is always set to 60.

This is especially bad, as I cannot set

```
"expire": 0
```

in order to expire a channel immediately. That means, when I expire a complete conference, `sctpConnections` hang until they time out after 60 secs, which probably wastes resources.

Look here:

I have 3 participants and I want to remove one:

```
{
                "id": "eaa06df320b1d370",
                "contents": [{
                                "name": "video",
                                "channels": [{
                                                "id": "a848ceea50555480",
                                                "expire": 0
                                }, {
                                                "id": "43a568c6caef46cd"
                                }, {
                                                "id": "292324e69ee4ad1"
                                }],
                                "sctpconnections": []
                }, {
                                "name": "audio",
                                "channels": [{
                                                "id": "1b3468eb18111083",
                                                "expire": 0
                                }, {
                                                "id": "475c91baf131855e"
                                }, {
                                                "id": "8c4fd3987e7db4b4"
                                }],
                                "sctpconnections": []
                }, {
                                "name": "data",
                                "channels": [],
                                "sctpconnections": [{
                                                "id": "aef15be015747a8a",
                                                "expire": 0
                                }, {
                                                "id": "80a98f8fe6878ab7"
                                }, {
                                                "id": "29b6e2900ca2fba6"
                                }]
                }],
                "channel-bundles": []
}
```

See how I reset expire for the participant I want to remove. The response is this:

```
{
                "id": "eaa06df320b1d370",
                "contents": [{
                                "name": "video",
                                "channels": [{
                                                "id": "43a568c6caef46cd",
                                                "expire": 15,
                                                "initiator": true,
                                                "endpoint": "NA-lTs2sSgimlkjgGybgXQ|||||ALa90V2be0sFgEWDRduuYGI",
                                                "direction": "sendrecv",
                                                "channel-bundle-id": "NA-lTs2sSgimlkjgGybgXQ|||||ALa90V2be0sFgEWDRduuYGI",
                                                "rtp-level-relay-type": "translator",
                                                "last-n": 2,
                                                "sources": [3967858994],
                                                "ssrcs": [2105642583]
                                }, {
                                                "id": "292324e69ee4ad1",
                                                "expire": 15,
[...]

PR #539, opened 2017-09-19
Sctp expire fixes

bgrozev:
Tries to address #537

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -1048,11 +1048,13 @@ public class Videobridge
                                     "Invalid 'expire' value: " + expire);
                         }
 
+                        sctpConn.setExpire(expire);
+
                         // Check if SCTP connection has expired.
                         if (expire == 0 && sctpConn.isExpired())
+                        {
                             continue;
-
-                        sctpConn.setExpire(expire);
+                        }
                     }
 
                     // endpoint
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1037,29 +1035,33 @@ else if (sctpConn == null)
...
+                        return IQUtils.createError(
+                            conferenceIQ,
+                            XMPPError.Condition.bad_request,
+                            "Invalid 'expire' value: " + expire);
+                    }
 
-                        // Check if SCTP connection has expired.
-                        if (expire == 0 && sctpConn.isExpired())
-                            continue;
+                    sctpConn.setExpire(expire);
```
bbaldino: is this the real fix? moving this line above the if check?

bgrozev: This one attempts to fix the problem of an expired (expire=0) sctpconnection being included in the response. I think it's appropriate since we hande channels the same way:
https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/Videobridge.java#L866

oliverhausler: Seems to work fine here. The SctpConnection is removed now.

I am seeing this in the logs: `WARNING: SCTP connection with NA-lTs2sSgimlkjgGybgXQ|||||APMUI_-LJEOiqKtXwQ19WoQ not ready yet.` Might be related.

PR #542, opened 2017-09-27
Moves callstats logic to jitsi-stats module.

damencho:


src/main/java/org/jitsi/videobridge/eventadmin/callstats/ConferencePeriodicRunnable.java
```diff
@@ -0,0 +1,153 @@
...
+                        continue;
+                    }
+
+                    MediaStreamStats2 stats = stream.getMediaStreamStats();
+                    if (stats == null)
+                    {
+                        continue;
+                    }
+
+                    Endpoint endpoint = channel.getEndpoint();
```
bgrozev: Why do we get the channel's endpoint instead of using "e"?

src/main/java/org/jitsi/videobridge/eventadmin/callstats/ConferencePeriodicRunnable.java
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Get stats receive or send from current conference.
+     *
+     * @param receive whether to get receive if <tt>true</tt> or
+     * send statistics otherwise.
+     * @param <T> the type of result stats Collection, ReceiveTrackStats
+     * or SendTrackStats.
+     * @return the result collection of stats grouped by endpointID.
+     */
+    private <T extends Collection> Map<String, T> getTrackStats(boolean receive)
```
bgrozev: Why do we need "T" here? Can't we just return a "Map<String, Collection>"?

src/main/java/org/jitsi/videobridge/eventadmin/callstats/ConferencePeriodicRunnable.java
damencho: Yep, I will change that. I was just moving the previous code here.

src/main/java/org/jitsi/videobridge/eventadmin/callstats/ConferencePeriodicRunnable.java
bgrozev: Nevermind, it seems that it is needed.

damencho: This is how the abstract methods of AbstractStatsPeriodicRunnable are defined, to specify the types expected in the collections. 
https://github.com/jitsi/jitsi-stats/blob/master/src/main/java/org/jitsi/stats/media/AbstractStatsPeriodicRunnable.java#L107

It seems more clear for the expected result for the abstract methods, WDYT?

src/main/java/org/jitsi/videobridge/eventadmin/callstats/ConferencePeriodicRunnable.java
damencho: Just forced pushed the changed.

PR #550, opened 2017-10-11
Uses statsId if available for callstats reporting.

damencho:


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -503,6 +504,28 @@ void describeChannelBundles(ColibriConferenceIQ iq)
...
+     * Adds the endpoint of this <tt>Conference</tt> as
+     * <tt>ColibriConferenceIQ.Endpoint</tt> instances in <tt>iq</tt>.
+     * @param iq the <tt>ColibriConferenceIQ</tt> in which to describe.
+     */
+    void describeEndpoints(ColibriConferenceIQ iq)
+    {
+        synchronized (transportManagers)
+        {
+            for (Map.Entry<String, IceUdpTransportManager> entry
+                : transportManagers.entrySet())
```
bgrozev: Why do we iterate over the transport managers? There might be endpoints that don't use channel bundles. Can we iterate over the endpoints instead?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -503,6 +504,28 @@ void describeChannelBundles(ColibriConferenceIQ iq)
...
+            {
+                ColibriConferenceIQ.Endpoint responseBundleIQ
```
bgrozev: Why "bundle IQ"?

src/main/java/org/jitsi/videobridge/TransportManager.java
```diff
@@ -217,6 +217,36 @@ else if (RawUdpTransportPacketExtension.NAMESPACE.equals(namespace))
...
      * Sets the values of the properties of a specific
+     * <tt>ColibriConferenceIQ.Endpoint</tt>
+     * to the values of the respective properties of this instance. Thus, the
+     * specified <tt>iq</tt> may be thought of as containing a description of
+     * this instance.
+     *
+     * @param iq the <tt>ColibriConferenceIQ.Endpoint</tt> on which to set the
+     * values of the properties of this instance
+     */
+    public void describe(ColibriConferenceIQ.Endpoint iq)
```
bgrozev: Can we avoid the code duplication with public void describe(ColibriConferenceIQ.ChannelBundle iq)?

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -785,7 +785,7 @@ public IQ handleColibriConferenceIQ(
             {
                 String channelID = channelIQ.getID();
                 int channelExpire = channelIQ.getExpire();
-                String channelBundleId = channelIQ.getChannelBundleId();
+                String channelBundleId = channelIQ.getEndpoint();
```
bgrozev: We already have an "endpoint" variable in the same block, with the same value. Can we remove "channelBundleId" and use "endpoint" instead (we can move its initialization to the top of the block).

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -967,7 +967,7 @@ public IQ handleColibriConferenceIQ(
                 String endpointID = sctpConnIq.getEndpoint();
                 SctpConnection sctpConn;
                 int expire = sctpConnIq.getExpire();
-                String channelBundleId = sctpConnIq.getChannelBundleId();
+                String channelBundleId = sctpConnIq.getEndpoint();
```
bgrozev: Same as above.

src/main/java/org/jitsi/videobridge/rest/JSONSerializer.java
```diff
@@ -348,7 +410,7 @@ public static JSONObject serializeChannelCommon(
         else
         {
             String id = channelCommon.getID();
-            String channelBundleId = channelCommon.getChannelBundleId();
+            String channelBundleId = channelCommon.getEndpoint();
             String endpoint = channelCommon.getEndpoint();
```
bgrozev: Same as above

bgrozev:
I'm a little confused about the legacy mode, specifically under what conditions it works. If the IDs of channel bundles coincide with the IDs of endpoints, it seems OK. But if they are different, it seems like there will be a subtle failure -- we would include channel-bundle-id elements with incorrect IDs, which would probably confuse an old colibri implementation. Am I correct? Have we decided to go ahead and stop supporting the case where the IDs don't coincide?

bgrozev:
> Have we decided to go ahead and stop supporting the case where the IDs don't coincide?

To clarify: I'd be fine with this. And I just confirmed with Emil that he's OK, too.

Can we add also update bundle.md with the new scheme, and specifically add a note on how to fix it if your focus implementation is affected (either stop using channel-bundles, OR use them with the same values as endpoint IDs)?

damencho:
Just pushed a commit which I think meet the comments and @bgrozev please check bundle.md, as I'm not sure whether the editions are correct.

src/main/java/org/jitsi/videobridge/rest/JSONSerializer.java
```diff
@@ -425,17 +424,15 @@ public static JSONObject serializeChannelCommon(
...
-            }
             // endpoint
             if (endpoint != null)
             {
+                // channelBundleId
+                jsonObject.put(
+                    ColibriConferenceIQ.ChannelCommon
+                        .CHANNEL_BUNDLE_ID_ATTR_NAME,
+                    endpoint);
+
```
bgrozev: Why do we need this last one? I think the JSON-XML translation should be kept as simple as possible. Any backward-compat stuff should be handled in another layer.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -510,23 +510,27 @@ void describeChannelBundles(ColibriConferenceIQ iq)
...
-                String id = entry.getKey();
-                Endpoint en = getEndpoint(id);
+            Endpoint en = i.next();
 
-                ColibriConferenceIQ.Endpoint responseBundleIQ
-                    = new ColibriConferenceIQ.Endpoint(
-                        id,
-                        en != null ? en.getStatsId() : null,
-                        en != null ? en.getDisplayName() : null);
+            ColibriConferenceIQ.Endpoint responseBundleIQ
```
bgrozev: s/responseBundleIQ/endpointIQ/?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -510,23 +510,27 @@ void describeChannelBundles(ColibriConferenceIQ iq)
...
+                = new ColibriConferenceIQ.Endpoint(
+                id,
```
bgrozev: Whitespace

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -510,23 +510,27 @@ void describeChannelBundles(ColibriConferenceIQ iq)
      */
     void describeEndpoints(ColibriConferenceIQ iq)
     {
-        synchronized (transportManagers)
+        for (Iterator<Endpoint> i = getEndpoints().iterator(); i.hasNext(); )
```
bgrozev: What's the reason for an iterator instead of foreach?

damencho: Cause copy-paste is easier :P, will fix all those

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -503,6 +503,27 @@ void describeChannelBundles(ColibriConferenceIQ iq)
     }
 
     /**
+     * Adds the endpoint of this <tt>Conference</tt> as
+     * <tt>ColibriConferenceIQ.Endpoint</tt> instances in <tt>iq</tt>.
+     * @param iq the <tt>ColibriConferenceIQ</tt> in which to describe.
+     */
+    void describeEndpoints(ColibriConferenceIQ iq)
+    {
+        for (Iterator<Endpoint> i = getEndpoints().iterator(); i.hasNext(); )
```
bgrozev: Why are we using an iterator?

PR #552, opened 2017-10-16
mvn: Set the source and target level to 1.8.

bgrozev:
Woo-hoo


gpolitis:
Capture the flag when this goes hulk!

PR #555, opened 2017-10-17
ref: Simplifies code (avoid the use of iterators).

bgrozev:


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1000,16 +982,11 @@ private Endpoint getEndpoint(String id, boolean create)
...
-                Endpoint e = i.next();
+            endpoints.removeIf(Endpoint::isExpired);
 
-                if (e.isExpired())
-                {
-                    i.remove();
-                    changed = true;
-                }
-                else if (e.getID().equals(id))
+            for (Endpoint e : endpoints)
```
bbaldino: if we want to continue the java 8-ness, this could be:

```
endpoint = endpoints.stream()
    .filter(e -> e.getID().equals(id))
    .findFirst()
    .orElse(null);
```

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1000,16 +982,11 @@ private Endpoint getEndpoint(String id, boolean create)
 
         synchronized (endpoints)
         {
-            for (Iterator<Endpoint> i = endpoints.iterator(); i.hasNext();)
-            {
-                Endpoint e = i.next();
+            endpoints.removeIf(Endpoint::isExpired);
```
ibauersachs: This lacks the assignment of `changed`

src/main/java/org/jitsi/videobridge/Conference.java
ibauersachs: In pure Java 8-ness, the return should be `Optional<Endpoint>` instead of `void`.

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: Thanks, good catch

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: I didn't understand that last part, and I'd rather avoid changing too much. I adopted Brian's suggestion.

ibauersachs: [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) or some [(anti) usage examples](https://dzone.com/articles/optional-ispresent-is-bad-for-you)

bgrozev: This looks consistent with the article. 

Or are you suggesting to change the return type of getEndpoint to Optional<Endpoint>? I'll leave this as it is for now, but I will also think about where it's appropriate to use Optional. Thanks for bringing it up!

ibauersachs: Yes, I meant changing the return type. I'm not really a fan of `Optional`, just wanted to bring it up. It's more important to return empty lists/arrays instead of null at other places.

PR #559, opened 2017-10-23
ref: Cleanup, simplification (using more java 8 features).

bgrozev:


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1164,13 +1144,14 @@ public Content getOrCreateContent(String name)
 
         synchronized (contents)
         {
-            for (Content aContent : contents)
+            content
+                = contents.stream()
+                    .filter(c -> c.getName().equals(name))
+                    .findAny().orElse(null);
```
bbaldino: i think `findFirst` would be more consistent with the original logic here (though i know that it shouldn't really make a difference, since we only expect it to match one)

src/main/java/org/jitsi/videobridge/Content.java
```diff
@@ -157,21 +172,18 @@
...
     public Content(Conference conference, String name)
     {
-        if (conference == null)
-            throw new NullPointerException("conference");
-        if (name == null)
-            throw new NullPointerException("name");
-
-        this.conference = conference;
-        this.name = name;
+        this.conference = Objects.requireNonNull(conference, "conference");
```
bbaldino: same question about `@NonNull` as above

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -266,17 +266,8 @@ public Conference(Videobridge videobridge,
...
-            throw new NullPointerException("videobridge");
-        }
-        if (id == null)
-        {
-            throw new NullPointerException("id");
-        }
-
-        this.videobridge = videobridge;
-        this.id = id;
+        this.videobridge = Objects.requireNonNull(videobridge, "videobridge");
```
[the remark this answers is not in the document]
bbaldino: after looking some more it looks like this was never standardized (though there are some libs that provide this sort of thing).

bgrozev:
On the use of @NonNull: Brian and I discussed it off-list and couldn't find a good way to use it (e.g. javax.validation.constraints.NotNull doesn't give a compile-time error, nor a runtime error). Suggestions are welcome.

PR #561, opened 2017-10-23
Octo

bgrozev:


bgrozev:
This PR is mostly to have the tests run. The new code is not actively used, but it shouldn't break anything, so it should be safe to merge (assuming the tests pass).

svener:
Since the stream sent to the client is sent to another videobridge at the same time, when VB1 and VB2 access two users at the same time, the browser will be stuck

bgrozev:
I don't understand your concern, if you think there's a problem can you please elaborate?

PR #563, opened 2017-10-30
ref(events): introduce MSG_TRANSPORT_READY_TOPIC

paweldomas:
The event will be fired when either WebSocket or SCTP endpoint message transport is ready. This will fix broadcasting of the initial endpoint connection status when WebSockets are used (and random build failures).


gpolitis:
Can somebody please help me understand how this fixes the random test failures because? I'm just curious to know how the fix works.

damencho:
@gpolitis when user joins and is not using datachannles, but websockets is not notified for initial connection statuses of participants.

gpolitis:
makes sense, I actually missed the addition of `notifyTransportChannelConnected();` in the `onWebSocketConnect()` which is the actual fix.

PR #567, opened 2017-11-07
Octo support and refactoring/cleanup

bgrozev:


src/main/java/org/jitsi/videobridge/Content.java
```diff
@@ -239,53 +245,91 @@ public RtpChannel createRtpChannel(String channelBundleId,
...
+                                       String transportNamespace,
+                                       Boolean initiator,
+                                       RTPLevelRelayType rtpLevelRelayType,
+                                       boolean octo)
+        throws Exception
+    {
+        RtpChannel channel;
 
-        do
+        synchronized (channels)
```
bbaldino: does this synchronized block need to be this big?  doesn't it just need to protect the `channels.put(id, channel);` ?  i see that `generateUniqueChannelID` already locks it on its own.

src/main/java/org/jitsi/videobridge/OctoChannel.java
```diff
@@ -0,0 +1,317 @@
...
+     */
+    private RtpChannelDatagramFilter rtpFilter
+        = new OctoDatagramPacketFilter(false);
+
+    /**
+     * The {@link org.ice4j.socket.DatagramPacketFilter} which (only) accepts
+     * RTCP packets for this channel.
+     */
+    private RtpChannelDatagramFilter rtcpFilter
+        = new OctoDatagramPacketFilter(true);
```
bbaldino: comment the boolean args for this and the rtp filter?

src/main/java/org/jitsi/videobridge/OctoChannel.java
```diff
@@ -0,0 +1,317 @@
...
+        // TODO: uncomment after the mvn versions are updated
+        //commonIq.setType(ColibriConferenceIQ.OctoChannel.TYPE);
+    }
+
+    /**
+     * @return The Octo ID of the conference.
+     */
+    public String getConferenceId()
+    {
+        return conferenceId;
```
bbaldino: is this not implemented in any of the channel base classes?  if not, why does octo channel need it and not the others?

src/main/java/org/jitsi/videobridge/OctoChannel.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    /**
+     * TODO: this should be in a utility somewhere.
+     * @param data
+     * @param off
+     * @param len
+     * @return
+     */
+    private boolean isRTCP(byte[] data, int off, int len)
```
bbaldino: should we use `RawPacket` here instead?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Reads the conference ID from an Octo header.
+     * @param buf the buffer which contains the Octo header.
+     * @param off the offset in {@code buf} at which the Octo header begins.
+     * @param len the length of the buffer.
+     * @return the Octo conference ID read from the buffer.
+     */
+    public static String readConferenceId(byte[] buf, int off, int len)
```
bbaldino: use `RawPacket` ?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Reads the {@link MediaType} from an Octo header.
+     * @param buf the buffer which contains the Octo header.
+     * @param off the offset in {@code buf} at which the Octo header begins.
+     * @param len the length of the buffer.
+     * @return the {@link MediaType} from the given Octo header.
+     */
+    public static MediaType readMediaType(byte[] buf, int off, int len)
```
bbaldino: `RawPacket`?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Reads the {@code r} flag from an Octo header.
+     * @param buf the buffer which contains the Octo header.
+     * @param off the offset in {@code buf} at which the Octo header begins.
+     * @param len the length of the buffer.
+     * @return the {@code r} flag from the given Octo header.
+     */
+    public static boolean readRflag(byte[] buf, int off, int len)
```
bbaldino: RawPacket?

src/main/java/org/jitsi/videobridge/Content.java
bgrozev: We do need it. If we release the lock before the call to channels.put() it is possible for another thread to call generateUniqueChannelID() and receive the same ID.

The block has many lines of code, but the only thing it does (apart from generating an ID and channels.put()) is to instantiate a channel. If readability is a concern I can extract that logic to a separate method, but as for reducing the code that executes inside the synchronized block, I don't think there's anything to do.

src/main/java/org/jitsi/videobridge/OctoChannel.java
bgrozev: Oh, nice catch. This is Octo-specific, and is not the same thing as the regular ID of the conference. It wasn't meant to be public. I'll re-work it somehow.

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
bgrozev: All the Octo stuff currently happens before the packets reach libjitsi (or after they leave libjitsi), so there's no RawPacket instance.

src/main/java/org/jitsi/videobridge/Content.java
bbaldino: not sure i follow, why not just remove it from here and change `channels.put(id, channel);` to:
```
synchronized (channels) {
  channels.put(id, channel);
}
```

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
bbaldino: i see...but is there use in changing the buffers to a RawPacket earlier (at this point?)

src/main/java/org/jitsi/videobridge/Content.java
bgrozev: Thread1 calls generateUniqueChannelID(), gets "a" and yields.
Thread2 calls generateUniqueChannelID() and also gets "a" (this is possible because Thread1 yielded before adding an entry for "a").

Both threads instantiate a channel with ID "a".

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
bgrozev: I don't think so. Apart from the check for RTP vs RTCP, I don't think we will need to do any RTP or RTCP specific things with the packets in the Octo layer. Note that OctoPacket only has static methods, we don't create any instances.

src/main/java/org/jitsi/videobridge/Content.java
bbaldino: oops, missed that.  nevermind then :)

bgrozev:
Closing temporarily to allow tests on another PR to run. Will re-open.

PR #576, opened 2017-11-30
Encoding secondary ssrc

bbaldino:
depends on https://github.com/jitsi/libjitsi/pull/385

SQUASH

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
   <properties>
     <assembly.skipAssembly>true</assembly.skipAssembly>
     <exec.mainClass>org.jitsi.videobridge.Main</exec.mainClass>
-    <jitsi-desktop.version>2.13.9aab784</jitsi-desktop.version>
+    <jitsi-desktop.version>2.13.fdf384f</jitsi-desktop.version>
     <smack.version>4.2.1</smack.version>
   </properties>
 
--- a/pom.xml
+++ b/pom.xml
@@ -187,6 +187,28 @@
       <version>3.4</version>
       <scope>test</scope>
     </dependency>
+	<!-- https://mvnrepository.com/artifact/org.powermock/powermock-core -->
+	<dependency>
+		<groupId>org.powermock</groupId>
+		<artifactId>powermock-core</artifactId>
+		<version>1.7.3</version>
+		<scope>test</scope>
+	</dependency>
+	<!-- https://mvnrepository.com/artifact/org.powermock/powermock-api-easymock -->
+	<dependency>
+		<groupId>org.powermock</groupId>
+		<artifactId>powermock-api-easymock</artifactId>
+		<version>1.7.3</version>
+		<scope>test</scope>
+	</dependency>
+	<!-- https://mvnrepository.com/artifact/org.powermock/powermock-module-junit4 -->
+	<dependency>
+		<groupId>org.powermock</groupId>
+		<artifactId>powermock-module-junit4</artifactId>
+		<version>1.7.3</version>
+		<scope>test</scope>
+	</dependency>
+
     <dependency>
       <groupId>org.jitsi</groupId>
       <artifactId>jitsi-stats</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -140,7 +140,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>libjitsi</artifactId>
-      <version>1.0-20171013.160715-320</version>
+      <version>1.0-20171204.173547-328</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
--- a/src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
+++ b/src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
@@ -22,7 +22,6 @@ import org.jitsi.service.configuration.*;
 import org.jitsi.service.libjitsi.*;
 
 import java.util.*;
-import java.util.stream.*;
 
 /**
  * A factory of {@link MediaStreamTrackDesc}s from jingle signaling.
@@ -109,7 +108,7 @@ public class MediaStreamTrackFactory
      * @return an array that holds the simulcast encodings.
      */
     private static RTPEncodingDesc[] createRTPEncodings(
-        MediaStreamTrackDesc track, Long[] primary,
+        MediaStreamTrackDesc track, long[] primary,
         int spatialLen, int temporalLen, Map<Long, SecondarySsrcs> secondarySsrcs)
     {
         RTPEncodingDesc[] rtpEncodings
@@ -197,69 +196,11 @@ public class MediaStreamTrackFactory
 
     }
 
-    private static SourceGroupPacketExtension getGroup(String semantics, List<SourceGroupPacketExtension> groups)
-    {
-        return groups.stream()
-            .filter(sg -> sg.getSemantics().equalsIgnoreCase(semantics))
-            .findFirst()
-            .orElse(null);
-    }
-
-    /**
-     * Given a list of sources and groups, find the primary ssrc.
-     * NOTE: this assumes there is only a single 'video source' described
-     * by the given sources/groups (the rest must be things like rtx, fec,
-     * simulcast sub-streams, etc.)
-     * @param sources
-     * @param sourceGroups
-     * @return
-     */
-    private static long getPrimarySsrc(List<SourcePacketExtension> sources,
-                                       List<SourceGroupPacketExtension> sourceGroups)
-    {
-        // Based on there only being a single video source, we can discover
-        // the primary ssrc by doing the following:
-        // 1) check if there is a simulcast group; if there is, then the first
-        //    ssrc in that group will be the overall primary
-        // 2) if no sim group, check if there is an fec group; if there is, then
-        //    the first ssrc in that group will be the overall primary
-        // 3) if no fec group, check if there is an rtx group; if there is, then
-        //    the first ssrc in that group will be the overall primary
-        // 4) if no fec group, then there should only be a single source, total,
-        //    which will be the primary
-        SourceGroupPacketExtension simGroup =
-            getGroup(SourceGroupPacketExtension.SEMANTICS_SIMULCAST, sourceGroups);
-        if (simGroup != null)
-        {
-            return simGroup.getSources().get(0).getSSRC();
-        }
-
-        SourceGroupPacketExtension fecGroup =
-            getGroup(SourceGroupPacketExtension.SEMANTICS_FEC, sourceGroups);
-        if (fecGroup != null)
-        {
-            return fecGroup.getSources().get(0).getSSRC();
-        }
-
-        SourceGroupPacketExtension rtxGroup =
-            getGroup(SourceGroupPacketExtension.SEMANTICS_FID, sourceGroups);
-        if (rtxGroup != null)
-        {
-            return rtxGroup.getSources().get(0).getSSRC();
-        }
-
-        if (!sources.isEmpty())
-        {
-            return sources.get(0).getSSRC();
-        }
-        return -1;
-    }
-
     /**
      * Get the 'secondary' ssrcs for the given primary ssrc.  'Secondary' here
      * is defined as things like rtx or fec ssrcs.
-     * @param ssrc
-     * @param sourceGroups
+     * @param ssrc the primay ssrc for which to get the secondary ssrcs
+     * @param sourceGroups the source groups
      * @return a map of secondary ssrc -> type (rtx, fec, etc.)
      */
     private static List<SecondarySsrc> getSecondarySsrcs(long ssrc, List<SourceGroupPacketExtension> sourceGroups)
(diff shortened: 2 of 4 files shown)


src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -380,4 +377,27 @@ private static int qid(int streamIdx, int spatialIdx, int temporalIdx,
         return streamIdx * spatialLen * temporalLen
             + spatialIdx * temporalLen + temporalIdx;
     }
+
+
+    static class Pair<K, V>
```
bickelj: If you already have commons-lang3 on the classpath, could use Pair provided by it.

bbaldino: i was originally using the one in javafx, which i thought was included with java 8 but i guess not...i searched around for a bit and was surprised not to find something more 'built-in'.  this code probably isn't quite ready to be merged yet, i needed to put it up so a libjitsi PR could have the tests run correctly.  hopefully we can replace this hack with something better.  i assume we don't already include commons-lang3 (didn't see it come up in auto-complete).

bbaldino:
@gpolitis and/or @bgrozev this is more fec pre-work and I'm ready for some feedback on it.  I have a note in there i'd be interested in some feedback on, as well as if you guys know of any existing lib we have to pull in `Pair` (right now I wrote a little one on my own) or maybe a better approach to that piece in general.

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -95,19 +96,9 @@
...
-    /**
-     * Creates encodings.
-     *
-     * @param track the track that will own the temporal encodings.
-     * @param primary the array of the primary SSRCs for the simulcast streams.
-     * @param rtx the array of the RTX SSRCs for the simulcast streams.
-     * @param spatialLen the number of spatial encodings per simulcast stream.
-     * @param temporalLen the number of temporal encodings per simulcast stream.
-     * @return an array that holds the simulcast encodings.
-     */
```
bgrozev: Why did you remove the javadoc?

bbaldino: fixed

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -95,19 +96,9 @@
...
     private static RTPEncodingDesc[] createRTPEncodings(
-        MediaStreamTrackDesc track, long[] primary, long[] rtx,
-        int spatialLen, int temporalLen)
+        MediaStreamTrackDesc track, Long[] primary,
+        int spatialLen, int temporalLen, Map<Long, List<Pair<Long, String>>> secondarySsrcs)
```
bgrozev: This gets confusing quickly. How about an abstraction at some level, e.g. List<Pair<Long, String>> becomes SecondarySsrcs, which would leave us with a Map<Long, SecondarySsrcs>

bbaldino: yeah i agree.  you're talking about making a SecondarySsrcs inner class, right?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -174,8 +166,14 @@ else if (temporalIdx > 0)
                     rtpEncodings[idx]
                         = new RTPEncodingDesc(track, idx,
                         primary[streamIdx],
-                        rtx == null || rtx.length == 0 ? -1 : rtx[streamIdx],
                         temporalId, spatialId, height, frameRate, dependencies);
+                    List<Pair<Long, String>> ssrcSecondarySsrcs = secondarySsrcs.get(primary[streamIdx]);
+                    if (ssrcSecondarySsrcs != null)
+                    {
+                        ssrcSecondarySsrcs.forEach(ssrcSecondarySsrc -> {
+                            rtpEncodings[idx].addSecondarySsrc(ssrcSecondarySsrc.getKey(), ssrcSecondarySsrc.getValue());
```
bgrozev: Wouldn't it be nice if the API took a Pair (or even a SecondarySsrc class)?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
bgrozev: Now that I've read the rest of this I tend to like the following better: replace the abstract Pair class with a SecondarySsrc class in libjitsi. Here we'd have Map<Long, List<SecondarySsrc>> which isn't too bad, and we can pass SecondarySsrc as a single parameter to methods. But I don't have any strong preference either way.

bgrozev: But I think it makes sense to define one structure in libjitsi and use it everywhere.

bbaldino: i just implemented it with both `SecondarySsrc` and `SecondarySsrcs` classes and it isn't bad.  but after looking at it, i'm not sure i like passing `SecondarySsrc` into `RTPEncodingDesc` (as opposed to the individual arguments) because then we'd have to make `SecondarySsrc` accessible elsewhere and it doesn't seem worth it to have it be more than an internal helper and leaking it into a broader scope.  I'll push this change and you can let me know what you think.

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -174,8 +178,14 @@ else if (temporalIdx > 0)
                     rtpEncodings[idx]
                         = new RTPEncodingDesc(track, idx,
                         primary[streamIdx],
-                        rtx == null || rtx.length == 0 ? -1 : rtx[streamIdx],
                         temporalId, spatialId, height, frameRate, dependencies);
+                    SecondarySsrcs ssrcSecondarySsrcs = secondarySsrcs.get(primary[streamIdx]);
```
gpolitis: There are some hidden NPEs here as the encoding expects a `long` and we have a `Long`. In the current flows I don't think we will ever get an NPE, but then again I've seen a impossible exception in the bridge logs..

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -174,8 +178,14 @@ else if (temporalIdx > 0)
...
+                    if (ssrcSecondarySsrcs != null)
+                    {
+                        ssrcSecondarySsrcs.forEach(ssrcSecondarySsrc -> {
+                            rtpEncodings[idx].addSecondarySsrc(ssrcSecondarySsrc.ssrc, ssrcSecondarySsrc.type);
```
gpolitis: same as above. If we don't check for null, I would love to at least have a comment here.

gpolitis:
Does this assume we have a single track? If so, I think we should try to be a bit more general and support multiple tracks.

bbaldino:
> Does this assume we have a single track? If so, I think we should try to be a bit more general and support multiple tracks.

This does assume a single track...my impression is that a fair bit of code (throughout the codebase) would have to change to fully support more than 1 track, so the extra logic to handle it in one place didn't seem worth it.  Do you think it's worth it to handle that here?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
bbaldino: Yeah, it seems like we shouldn't ever hit that in this flow (I think `SourcePacketExtension#getSSRC` would have to return an invalid `Long` for that to happen?)  Is there another scenario where it could?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
bbaldino: what sort of comment were you thinking?

gpolitis:
> This does assume a single track...my impression is that a fair bit of code (throughout the codebase) 

Do you mean in the jitsi-meet codebase? Sure. The JVB should be able to handle multiple tracks internally, and I think the previous factory method exposed that. Given the new abstractions that you have introduced, I think it would be a shame to not support multiple tracks.

bbaldino:
hold off on merging this for a little bit, i want to add some unit tests too.

pom.xml
```diff
@@ -187,6 +187,28 @@
       <version>3.4</version>
       <scope>test</scope>
     </dependency>
+	<!-- https://mvnrepository.com/artifact/org.powermock/powermock-core -->
```
gpolitis: Did you use powermock because of some specific feature that you're missing in easymock?

bbaldino: yeah, it's needed to mock the static method.  PS don't merge this yet, working on another take at creating the tracks.

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -384,4 +429,111 @@ private static int qid(int streamIdx, int spatialIdx, int temporalIdx,
...
+        {
+            return secondarySsrcs.iterator();
+        }
+    }
+
+    /**
+     * Describes one or more primary video ssrcs for a single
+     * {@link MediaStreamTrackDesc}
+     */
+    private static class TrackSsrcs
```
gpolitis: What is the purpose of this class? Couldn't we just use a `List`?

bbaldino: this makes things a lot more readable, since we use it in a list. so this avoids having List<List<Long>> and instead gives List<TrackSsrcs> which is a lot clearer.

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -86,31 +96,51 @@
...
+     * ssrcs, so we'll translate them into constants defined in libjitsi
+     */
+    private static Map<String, String> secondarySsrcTypeMap = null;
+
+    private static synchronized Map<String, String> getSecondarySsrcTypeMap()
+    {
+        if (secondarySsrcTypeMap == null)
+        {
+            secondarySsrcTypeMap = new HashMap<>();
+            secondarySsrcTypeMap.put(SourceGroupPacketExtension.SEMANTICS_FID, Constants.RTX);
```
gpolitis: Do we have support for flex-fec SSRCs?

bbaldino: not yet, this PR doesn't have the flexfec work (just one of the base constants) that's coming in later PR

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -384,4 +429,111 @@ private static int qid(int streamIdx, int spatialIdx, int temporalIdx,
...
+    {
+        int numEncodings = primarySsrcs.size() * numSpatialLayersPerStream * numTemporalLayersPerStream;
+        boolean isSimulcast = primarySsrcs.size() > 1;
+        RTPEncodingDesc[] rtpEncodings = new RTPEncodingDesc[numEncodings];
+        MediaStreamTrackDesc track =
+            new MediaStreamTrackDesc(receiver, rtpEncodings, isSimulcast);
+
+        RTPEncodingDesc[] encodings = createRTPEncodings(track, primarySsrcs,
+            numSpatialLayersPerStream, numTemporalLayersPerStream, allSecondarySsrcs);
+        assert(encodings.length <= numEncodings);
```
gpolitis: Shouldn't this be equal?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
gpolitis: I see, how about extending a `List` then?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
bbaldino: i was just trying to prevent a buffer overflow with this check...i didn't want to necessarily also enforce that the number of encodings had to be exactly the same (it didn't seem unreasonable that perhaps that logic might fail to create a certain encoding, for example)

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
bbaldino: i gave this a shot but ran into a couple problems:

1) it has to derive from a list implementation, not just `List` (or else i'd have to implement the abstract `List` methods, which doesn't really seem worth it), which doesn't feel right to me: nothing about this requires it being an `ArrayList`, for example.
2) It specifically would derive from `ArrayList<Long>` (since we want it to only hold ssrcs), which breaks some of the ways I'm currently constructing it.  `Arrays.asList(group.getSources().get(0).getSSRC())`, for example, evaluates to List<T> and doesn't allow assigning to a `TrackSsrcs` object which derives from `ArrayList<Long>`

PR #579, opened 2017-12-03
Add documentation for using non-standard ports

nbartos:


bgrozev:
Thanks for the contribution! But I think that's something more suitable for the jitsi-meet repo, i.e. the [quick install document](https://github.com/jitsi/jitsi-meet/blob/master/doc/quick-install.md) and/or the [manual install document](https://github.com/jitsi/jitsi-meet/blob/master/doc/manual-install.md).

PR #581, opened 2017-12-05
Fix dominant speaker initial event

bgrozev:
Sends the initial dominant speaker changed event when a colibri web-socket connects. Refactors the handling of an Endpoint's SctpConnection to make it simpler.


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -606,30 +572,11 @@ private void endpointPropertyChange(
             PropertyChangeEvent ev)
     {
         String propertyName = ev.getPropertyName();
-        boolean maybeRemoveEndpoint;
```
bbaldino: so this event is now gone?

bgrozev: Correct.

bgrozev:
The fix is really just the addition of this line in EndpointMessageTransport:
```
endpoint.getConference().endpointMessageTransportConnected(endpoint);
```

The rest is refactoring. The way it used to work:
SctpConnection sets itself as the SctpConnection of an Endpoint. The Endpoint fires a PropertyChangeEvent. The Conference sees the PropertyChangeEvent, and waits for the SctpConnection to become ready (adding itself as a listener if necessary). The Conference notifies Endpoint that the Endpoint's SctpConnection is ready.

The way it works now:
SctpConnection sets itself as the SctpConnection of an Endpoint. The Endpoint waits for it to become ready. When it's ready it notifies notifies the Conference (with the code I cited above, which is generic and also executes when a web-socket is ready).

PR #584, opened 2017-12-13
fix: Fixes expiration of endpoints.

bgrozev:
Commit 2a15aa43c40 introduces a bug which fails to expire endpoints,
which results in leaks. This attempts to resolve this problem.


bgrozev:
Please squash

PR #585, opened 2017-12-14
feat: Passes the logging id to the media stream.

gpolitis:
This PR enables logs correlation on the bridge. It also updates libjitsi@1.0-20180117.223511-333.


gpolitis:
@bbaldino or @bgrozev could you please take a look?

PR #589, opened 2017-12-22
WIP: Media stream track factory rewrite

gpolitis:
depends on https://github.com/jitsi/libjitsi/pull/392. Fixes a problem with video coming from firefox being dropped.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -1925,9 +1925,11 @@ public class RtpChannel
 
         if (mediaStreamTrackReceiver != null)
         {
+            MediaStreamTrackFactory mediaStreamTrackFactory
+                = new MediaStreamTrackFactory
+                (mediaStreamTrackReceiver, sources, sourceGroups);
             MediaStreamTrackDesc[] newTracks
-                = MediaStreamTrackFactory.createMediaStreamTracks(
-                    mediaStreamTrackReceiver, sources, sourceGroups);
+                = mediaStreamTrackFactory.createMediaStreamTracks();
 
             return mediaStreamTrackReceiver.setMediaStreamTracks(newTracks);
         }
--- a/src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
+++ b/src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
@@ -24,6 +24,7 @@ import org.jitsi.service.neomedia.codec.*;
 import org.jitsi.util.*;
 
 import java.util.*;
+import java.util.concurrent.*;
 import java.util.stream.*;
 
 /**
@@ -34,9 +35,21 @@ import java.util.stream.*;
 public class MediaStreamTrackFactory
 {
     /**
-     * The {@link ConfigurationService} to pull configuration options from.
+     * The system property name that for a boolean that's controlling whether
+     * or not to enable temporal scalability filtering for VP8.
      */
-    private static ConfigurationService cfg = LibJitsi.getConfigurationService();
+    public static final String ENABLE_VP8_SVC_PNAME = "org.jitsi" +
+       ".videobridge.ENABLE_SVC" ;
+
+    /**
+     * The default number of temporal layers to use for VP8 simulcast.
+     */
+    private static final int VP8_SIMULCAST_TEMPORAL_LAYERS = 3;
+
+    /**
+     * The resolution of the base stream when activating simulcast for VP8.
+     */
+    private static final int VP8_SIMULCAST_BASE_LAYER_HEIGHT = 180;
 
     /**
      * The {@link Logger} used by the {@link MediaStreamTrackDesc} class and its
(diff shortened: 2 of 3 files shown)


src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -1925,9 +1925,11 @@ public boolean setRtpEncodingParameters(
 
         if (mediaStreamTrackReceiver != null)
         {
+            MediaStreamTrackFactory mediaStreamTrackFactory
+                = new MediaStreamTrackFactory
+                (mediaStreamTrackReceiver, sources, sourceGroups);
```
bgrozev: Formatting

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -34,347 +35,69 @@
...
-        ".videobridge.ENABLE_SVC";
-
-    /**
-     * The system property name that for a boolean that's controlling whether or
-     * not to enable temporal scalability filtering for VP8.
-     */
-    public static final String ENABLE_VP9_SVC_PNAME = "org.jitsi" +
-        ".videobridge.ENABLE_VP9_SVC";
+    public static final String ENABLE_VP8_SVC_PNAME = "org.jitsi" +
+       ".videobridge.ENABLE_SVC" ;
```
bgrozev: Formatting

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
-    private static int qid(int streamIdx, int spatialIdx, int temporalIdx,
-                           int spatialLen, int temporalLen)
+    public MediaStreamTrackDesc[] createMediaStreamTracks()
     {
-        return streamIdx * spatialLen * temporalLen
-            + spatialIdx * temporalLen + temporalIdx;
-    }
+        tracksMap.clear();
+
+	if (sourceGroups != null && !sourceGroups.isEmpty())
```
bgrozev: There's a mix of spaces and tabs below. Have you stopped using an IDE again? :)

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
 
-        public Long get(int index)
-        {
-            return trackSsrcs.get(index);
-        }
+        List<RTPEncodingDesc> encodings = new ArrayList<>();
 
-        @Override
-        public Iterator<Long> iterator()
+        int qot = 0;
```
bgrozev: qot?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
+        int height = VP8_SIMULCAST_BASE_LAYER_HEIGHT;
+        for (SourcePacketExtension source : simGroup.getSources())
         {
-            return trackSsrcs.iterator();
-        }
-    }
+            double frameRate = 7.5;
+            long primarySSRC = source.getSSRC();
+            tracksMap.put(primarySSRC, track);
```
bgrozev: Is this intentionally inside the loop? I don't think it's a problem, but seems strange.

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
+            List<SourceGroupPacketExtension> groupsCopy
+                = new ArrayList<>(sourceGroups);
+
+            // We first process the special SIM group. The SIM group represents
+            // a single track (or video source) it has 3 spatial encodings and
+            // each spatial encoding has 3 temporal encodings. The temporal
+            // sub-encodings share the SSRCs of the parent spatial encoding. 
+	    Iterator<SourceGroupPacketExtension> it = groupsCopy.iterator();
+
+	    while (it.hasNext())
```
bgrozev: I find iterators hard to follow. How about checking the semantics in addVP8SimulcastTrack and returning a boolean? Then this iterator block will become just this:
```
groupsCopy.removeIf(this::addVP8SimulcastTrack);
```

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
+                // Set the secondary SSRC for the matching encodings.
+                long secondarySSRC = groupSources.get(1).getSSRC();
+                tracksMap.put(secondarySSRC, track);
+                for (RTPEncodingDesc encoding : track.getRTPEncodings())
+                {
+                    if (encoding.getPrimarySSRC() == primarySSRC)
+                    {
+                        encoding.addSecondarySsrc(
+                                secondarySSRC,
+                                semanticsMap.get(sourceGroup.getSemantics()));
```
bgrozev: In other places we compare the semantics with equalsIgnoreCase. So I think it is either not necessary to use equalsIgnoreCase, or we need to take the case into account here as well

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
+            }
         }
 
-        @Override
-        public Iterator<SecondarySsrc> iterator()
-        {
-            return secondarySsrcs.iterator();
-        }
+        Collection<MediaStreamTrackDesc> tracks = tracksMap.values();
+	return tracks.toArray(new MediaStreamTrackDesc[tracks.size()]);
```
bgrozev: Above we map the primary SSRC and the secondary SSRC to their track. So tracksMap's values might contain duplicates. Is this intentional?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
     }
 
-    /**
-     * Describes one or more primary video ssrcs for a single
-     * {@link MediaStreamTrackDesc}
-     */
-    private static class TrackSsrcs
-        implements Iterable<Long>
+    private MediaStreamTrackDesc addPlainTrack(long primarySSRC)
```
bgrozev: Can you please add docs for this? It returns the existing track (even if it's a simulcast track) if there is one, which is not obvious

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
bgrozev: We should also probably handle the case where we were signaled a group with unknown semantics

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
bbaldino: i agree that it'd be nice to make the loop more readable, but not sure i like calling `addVP8SimulcastTrack` for every track and it just returning true for the ones that actually were simulcast tracks.  i think it'd be clearer to do these sorts of things more procedurally, like:

```
  // sim group processing
  simGroups = getSimGroups(sourceGroups)
  (do the work with the sim groups)
  sourceGroups.remove(simGroups)
  
  // rtx/fec processing
  ...
```

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
-        public List<SecondarySsrc> secondarySsrcs;
+                long primarySSRC = groupSources.get(0).getSSRC();
+    
+                MediaStreamTrackDesc track = addPlainTrack(primarySSRC);
 
-        public SecondarySsrcs(List<SecondarySsrc> secondarySsrcs)
+                // Set the secondary SSRC for the matching encodings.
+                long secondarySSRC = groupSources.get(1).getSSRC();
+                tracksMap.put(secondarySSRC, track);
+                for (RTPEncodingDesc encoding : track.getRTPEncodings())
```
bbaldino: what about having a second form of `addPlainTrack` that took a secondary ssrc and did this work there?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
-        private List<Long> trackSsrcs;
+        MediaStreamTrackDesc track = tracksMap.get(primarySSRC);
 
-        public TrackSsrcs(List<Long> trackSsrcs)
+        // Check if we already have a track for this SSRC.
+        if (track == null)
         {
-            this.trackSsrcs = trackSsrcs;
+            // It looks like this is the first time we see this SSRC,
+            // so we create a new track for it.
```
bbaldino: i liked having it where an ssrc was removed from source groups/sources completely once it was processed rather than dealing with the possibility of adding the same track twice, any reason to go back to this?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -383,157 +106,159 @@ private static void removeReferences(TrackSsrcs trackSsrcs, List<SourcePacketExt
...
-        @Override
-        public Iterator<Long> iterator()
+        int qot = 0;
+        int height = VP8_SIMULCAST_BASE_LAYER_HEIGHT;
+        for (SourcePacketExtension source : simGroup.getSources())
         {
-            return trackSsrcs.iterator();
-        }
-    }
+            double frameRate = 7.5;
```
bbaldino: can we use a constant for this?

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -34,347 +35,69 @@
...
      */
     private static final int VP8_SIMULCAST_TEMPORAL_LAYERS = 3;
 
     /**
      * The resolution of the base stream when activating simulcast for VP8.
-     *
-     * FIXME: hardcoded ugh.. this should be either signaled or somehow included
-     * in the RTP stream.
      */
     private static final int VP8_SIMULCAST_BASE_LAYER_HEIGHT = 180;
```
bbaldino: i think keeping this FIXME is worthwhile

gpolitis:
https://github.com/jitsi/jitsi-videobridge/pull/591 fixes the problem with video coming from firefox being dropped. When/if further changes to signaling are required, we can revive this PR.

PR #590, opened 2018-01-05
Fixes signaling of ungrouped SSRCs

bgrozev:
Fixes a problem with video coming from firefox being dropped.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -1904,12 +1904,12 @@ public class RtpChannel
         // so that downstreams do not need to take care of nulls.
         if (sources == null)
         {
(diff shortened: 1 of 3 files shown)


gpolitis:
Does this fix 1-1 calls between FF-Chrome that go through the JVB?

bgrozev:
Yes. I haven't tested, but I think it isn't limited to 1-1 (we would never forward anything coming from firefox)

bbaldino:
thanks for finding my mistake, @bgrozev :) i'll merge this...but want to add a unit test for this scenario first so i'll probably copy your branch and add that to merge it.

bbaldino:
boris is out and i don't have perms on his repo, so cloned this branch and added a unit test here: https://github.com/jitsi/jitsi-videobridge/pull/591.

PR #591, opened 2018-01-05
Fix groupless ssrcs

bbaldino:
All credit goes to @bgrozev here, I just added a unit test on top of his fix for the issue.  original PR was here: https://github.com/jitsi/jitsi-videobridge/pull/590/

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -358,8 +358,13 @@ public class BitrateController
         SimulcastController simulcastController
             = ssrcToBitrateController.get(ssrc);
 
-        return simulcastController != null
-            && simulcastController.accept(pkt);
+        if (simulcastController == null) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -358,8 +358,13 @@ public boolean accept(RawPacket pkt)
         SimulcastController simulcastController
             = ssrcToBitrateController.get(ssrc);
 
-        return simulcastController != null
-            && simulcastController.accept(pkt);
+        if (simulcastController == null) {
+            logger.warn("Dropping an RTP packet, because the SSRC has not " +
```
bbaldino: @bgrozev i do worry a bit that this log might get too verbose.  i wonder if we should try and only log this every X packets or something.

bgrozev:
Yeah makes sense. But I don't expect to see this unless something went
wrong.

Boris

PR #596, opened 2018-01-17
Fix media stream track factory

bgrozev:


bgrozev:
Jenkins test this please

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -323,19 +323,37 @@ private static void removeReferences(
...
-            for (SourcePacketExtension source : group.getSources())
-            {
-                if (trackSsrcs.contains(source.getSSRC()))
-                {
-                    return true;
-                }
-            }
-            return false;
-        });
+        List<SourceGroupPacketExtension> groupsToRemove
```
bbaldino: add `// Remove any groups to which any of the ssrcs of this track belongs`

src/main/java/org/jitsi/videobridge/xmpp/MediaStreamTrackFactory.java
```diff
@@ -323,19 +323,37 @@ private static void removeReferences(
...
+            = sourceGroups.stream()
+                .filter(
+                    group -> group.getSources().stream().anyMatch(
+                        source -> trackSsrcs.contains(source.getSSRC())))
+                .collect(Collectors.toList());
+
+        sourceGroups.removeAll(groupsToRemove);
+
+        Set<Long> ssrcsToRemove = extractSsrcs(groupsToRemove);
+        sources.removeIf(
```
bbaldino:
```
/** 
  * Remove not only the ssrcs in the track itself, but any ssrcs that were in groups along with ssrcs from the track. E.g. if we have:`
  * SIM 1 2 3
  * RTX 1 10
  * RTX 2 20
  * RTX 3 30
  * then we need to make sure ssrcs 10, 20 and 30 don't create tracks of their own and are removed along with the processing of the track with ssrcs 1, 2 and 3
  */

PR #600, opened 2018-01-22
feat: Adds a tmpfs backed directory dedicated for the JVB.

gpolitis:


bbaldino:
> The purpose of this is to provide a tmpfs backed directory to the bridge

To what end?  What's it trying to achieve?

gpolitis:
It's for temporarily storing the packet traces, which could reach 20 mbytes/s. I just want to store a few seconds of data, so the trace file should never go above a certain size, the traces will be read by an agent and transferred elsewhere where they can be stored in a more compact format.

bbaldino:
Ah gotcha.  Yeah that'd be nice

bgrozev:
I'd prefer to keep the task of creating directories and setting their permissions a part of the installation process (together with the code which does the same for /var/log/jitsi, /etc/jitsi/videobridge, etc):
https://github.com/jitsi/jitsi-videobridge/blob/master/resources/install/debian/postinst#L99

bgrozev:
Ah, right

gpolitis:
test this please

gpolitis:
test this please

gpolitis:
test this please

gpolitis:
test this please

gpolitis:
test this please

gpolitis:
test this please

gpolitis:
@bgrozev @bbaldino could you please merge this without waiting for the tests?

PR #604, opened 2018-01-26
fix: Uses dynamic payload type mapping for VP8 when probing for bandwidth.

gpolitis:


gpolitis:
@bgrozev or @bbaldino could you please take a look?

src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
```diff
@@ -79,6 +80,11 @@
      */
     private final VideoChannel dest;
 
+    /**
+     * The VP8 payload type to use when probing with the SSRC of the bridge.
+     */
+    private int vp8PT = -1;
```
bbaldino: technically this should match whatever video PT that's being sent, is that right?  we're just hard-coding to vp8 for now?

gpolitis: Even if the conference runs over VP9, we can still send padding packets with VP8 payload type; as long as it's signaled, we're good. We're not sending any real media anyway, just padding (empty packets).

bbaldino:
lgtm

PR #606, opened 2018-01-29
logs: Converts some logs into time series logs.

gpolitis:


gpolitis:
Nothing really interesting to here, just a some logging related changes. I'll merge this later if nobody objects.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -511,17 +518,28 @@ public void update(List<Endpoint> conferenceEndpoints, long bweBps)
                     ctrl.setOptimalIndex(optimalIdx);
 
                     MediaStreamTrackDesc sourceTrack = ctrl.getSource();
-                    if (sourceTrack != null && logger.isDebugEnabled())
+                    if (sourceTrack != null
+                            && timeSeriesLogger.isTraceEnabled())
```
bbaldino: is the check for `isTraceEnabled` still required here, since we're not doing any string concatenation at the call?  for the time series logger it looks like we could handle these in the logger itself.

gpolitis: I would argue it is still necessary because creating the time series point has some cost (creating a hash-map, adding elements to the new hash-map, etc). I haven't quantified this, but checking if the trace is enabled or not doesn't seem particularly ugly to me.

bbaldino: right, i'm just saying that this check could be inside the timeserieslogger `.trace()` method itself, right?  the only reason we can't do this with the normal logger for normal logging calls is that the string concatenation happens at the call site, so the library has no control over preventing it, but here we should be able to?  just a thought...i hate seeing these checks for `isXXXEnabled` everywhere and how easy it is to forget them, so wondering if we can make the timeserieslogger better

gpolitis: Maybe I'm missing your point, but creating the time series point -which allocates a new hash-map, adds elements to the new hash-map, etc- happens at the call site.

bbaldino: ok, i see...i didn't notice that the time series point was being created off of the diagnostics context, i thought it was being created off of the time series logger.  my bad.

Issue #610, opened 2018-02-01
Please clarify the need and usage of WebSockets (/colibri-ws/)


vi:
https://github.com/jitsi/jitsi-videobridge/blob/master/doc/rest.md

> A WebSocket API for communication with conference endpoints (`/colibri-ws/`)

Where can I find the API documentation of what can be sent to and received from this websocket?  
How does this interact with the usual REST interface (`/colibri/`)?

bgrozev:
This is independent from the `/colibri/` interface. It is meant for clients to use instead of WebRTC data channels (the main motivation being the lack of data channel support in Edge). It is implemented as just another transport for the same messages that are supported over the data channel (i.e. last-n notifications, selecting endpoints, endpoint-to-endpoint messages).

vi:
And is the data channel usage itself documented somewhere? ([datachannel-communication.md](https://github.com/jitsi/jitsi-videobridge/blob/master/doc/datachannel-communication.md) is short-esque and with no actual session examples)

Can this (or data channels or something else) be used to monitor such events from infrastructure signaling server, not from the peer? Or should a dummy peer just for monitoring events be used for this? Like a private API version of WebSocket interface.

bgrozev:
> And is the data channel usage itself documented somewhere?

Unfortunately not. There's fragments [here](https://github.com/jitsi/jitsi-videobridge/blob/master/doc/last-n.md)


> Can this (or data channels or something else) be used to monitor such events from infrastructure signaling server, not from a peer? Or should a dummy peer just for monitoring events be used for this?

The web sockets are only available for Endpoints in a conference. You could create a dummy endpoint through colibri.

vi:
What is the best way of maintaining a dummy peer just for receiving events?  
Can it be established using REST-Colibri+WebSocket alone (without any WebRTC video, audio or data channel)? Will connected WebSocket keep that peer from expiring?

bgrozev:
> What is the best way of maintaining a dummy peer just for receiving events?
> Can it be established using REST-Colibri+WebSocket alone (without any WebRTC video, audio or data channel)? Will connected WebSocket keep that dummy peer from expiring?

I'm not sure if and Endpoint without any channels will work. You might need to create a dummy channel for it. The WebSocket *should* keep it alive, but we haven't used it in this way, so you might run into some issues.

vi:
[...]

PR #613, opened 2018-02-05
Channel stats dealer implementation

sidorovis:
Work in progress, just a pre-pull-request for review, comments and implementation wants.

https://github.com/jitsi/jitsi-videobridge/issues/610

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -164,6 +164,10 @@ public class Videobridge
     public static final String XMPP_API_PNAME
         = "org.jitsi.videobridge." + XMPP_API;
 
+    public static final String EXPIRE_SLEEP_TIME_SEC
+            = "org.jitsi.videobridge.EXPIRE_SLEEP_TIME_SEC";
+
+
     public static Collection<Videobridge> getVideobridges(
             BundleContext bundleContext)
     {
@@ -215,7 +219,6 @@ public class Videobridge
      */
     public Videobridge()
     {
-        new VideobridgeExpireThread(this).start();
     }
 
     /**
@@ -1389,6 +1392,11 @@ public class Videobridge
             logger.warn("No authorized source regexp configured. Will accept "
                             + "requests from any source.");
         }
+
+        Integer expireSleepTimeSec = (cfg == null) ? Channel.DEFAULT_EXPIRE :
+                cfg.getInt(EXPIRE_SLEEP_TIME_SEC, Channel.DEFAULT_EXPIRE);
+
+        new VideobridgeExpireThread(this, expireSleepTimeSec).start();
 
         // <conference>
         ProviderManager.addIQProvider(
--- a/src/main/java/org/jitsi/videobridge/VideobridgeExpireThread.java
+++ b/src/main/java/org/jitsi/videobridge/VideobridgeExpireThread.java
@@ -38,6 +38,11 @@ class VideobridgeExpireThread
         = Logger.getLogger(VideobridgeExpireThread.class);
 
     /**
+     * Sleep time for expire check cycle (in seconds)
+     */
+    private final int expireSleepTimeSec;
+
+    /**
      * The <tt>Videobridge</tt> which has its {@link Channel}s expired by this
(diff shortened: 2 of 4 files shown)


bgrozev:
Thanks for sharing this, we do appreciate it.

But it's not something that we can accept in this state. We would like to have an abstract implementation that we can reuse in both jitsi-videobridge and jicofo. I am thinking about something like:
* A RestEventHandler class in jicoco, which takes the HTTP endpoint to send to as a parameter (what you have as endpointOrigin). This implements the EventHandler interface.
* Code in jitsi-videobridge somewhere (perhaps Videobridge.java) which reads the endpoint from a configuration property (e.g org.jitsi.videobridge.RestEventHandler.ENDPOINT), and instantiates and registers the RestEventHandler with the existing EventAdmin. It might have to specify a list of event topics to accept, I'm not quite sure.

This way we can easily add the same functionality to jicofo later on. You can see an example of how something registers to the EventAdmin as an EventHandler in BridgeSelector in jicofo. And see Videobridge.getEventAdmin() to see how to access the EventAdmin instance for videobridge.

gpolitis:
This code solves a problem and I can think of many use cases where it can be useful, but I don't think it belongs in mainline, as the code may end-up unmaintained, just like with the metric service and the influx service in the past.

@sidorovis have you thought about mining the logs? The bridge outputs this information and you can write a fluentd rule to push it pretty much anywhere you wish. If you dislike the mining approach, I believe we still have code that can override the bundles that we load on startup. So you can maintain a separate bundle in a private repo (jitsi-videobridge-contib or jitsi-videobridge-rest-dealer sound nice). If you decide to go that route, we'll be more than happy to include a reference to it in some text file in the mainline repo or in the wiki pages.

These are just my suggestions and my -1 vote, if the other developers think this is useful and that it should be included in mainline so be it.

sidorovis:
I don't think that reading logs for this is a good solution, the goal is to get information from JVB that client "A" was disconnected / lost connection / unavailable as soon as possible, so another controlling server (for example jicofo) will get that info from JVB and follow necessary procedure (let all other clients know that "A" is timely unavailable or something else).
I believe that JVB should do that because JVB has a direct connection with clients and actually know when any of them was disconnected.
Also I don't think that while this code is so simple it could became unmaintained. There is no way to kick it out to "unmaintained" territory without deleting 'channel-created', 'channel-expired' event.
Probably as another one bundle it would be also a good solution. 
But we still need to update VideobridgeExpireThread with expireSleepTimeSec parameter. Otherwise this check is not so often and there is no another way to understand that connection was expired.

Probably you have another way how to define that connection was "paused", "reconnecting", "expired", "disconnected", "lost"? Any of that would be great!

bgrozev:
Sorry about the confusion on this one, I should have consulted with the others before posting.

> Probably you have another way how to define that connection was "paused", "reconnecting", "expired", "disconnected", "lost"?

Take a look at EndpointConnectionStatus. Currently it notifies other endpoints in a conference when an endpoint's connection times out (without having expired yet, or being expired by the controller).

> But we still need to update VideobridgeExpireThread with expireSleepTimeSec parameter.

This change looks OK to me (I'd prefer if the property was local to VideobridgeExpireThread).

sidorovis:
EndpointConnectionStatus is used to send messages to connected clients (via data-channel or WebSocket). But in my case we are looking for something that will not require endpoint_id and pwd to receive such information from JVB. So EndpointConnectionStatus 

I will open another Pull Request with only expireSleepTimeSec parameter (with "to property was local to VideobridgeExpireThread" improvement).

sidorovis:
So from perspectives of newbie in this world, I added https://github.com/sidorovis/jitsi-videobridge-rest-dealer repo with example of implementation for rest-dealer.

I'm not sure that you mentioned that, but I'm not sure how to add necessary additional bundle.
Probably adding possibility to load addional list of bundles from some "additional_bundles.txt" OR what is much better - from some folder where jar files with bundles would be situated.

But I'm not sure how to make OSGi to read such folder and load jars. Probably adding some "spring-osgi" will solve this issue.

Anyway looking for your comments to understand what do you think is the best way?

sidorovis:
Could be easily replaced by https://github.com/jitsi/jicoco/pull/30 and https://github.com/sidorovis/jitsi-videobridge-rest-dealer

Issue #598, opened 2018-01-18
Since SimulcastController has been introduced the last connected peer does not receive video for a while


nojz:
How it worked before:
1. There is a client connected in and sending video.
2. Second (last) client connects, it receives delta frames so it sends Picture Loss Indication (PLI) message.
3. First client receives the PLI and sends a keyframe.
4. Second client receives the video.

How it fails after SimulcastController has been introduced:
1. There is a client connected in and sending video.
2. Second (last) client connects and here comes the problem, it does not receive delta frames because SimulcastController drops them (see: [https://github.com/jitsi/jitsi-videobridge/blob/f2eba1aa25d77f57a827bd1acd6013b912c81cf0/src/main/java/org/jitsi/videobridge/cc/SimulcastController.java#L267](SimulcastController.java:267)) so it won't send PLI and it won't receive video until a keyframe is sent randomly. (I don't know the logic here, but it is usually 1 min later.)

I also noticed that there is a keyframe sent when a peer is added to the conference (jitsi-videobridge requests it I guess), but that's too early because the client is not connected at that time only a few seconds later.

Do you have a recommendation how I should solve this problem and/or could you check whether your logic is right?

gpolitis:
@nojz are you by any chance using the REST interface?

dwilson6:
I have been seeing the same issue. It tends to happen when the second client is only receiving a stream, not sending. The clients are not sending simulcast streams. Your diagnosis sounds correct. After about a minute and a half the stream comes through when the chrome webrtc stack sends a keyframe to the bridge.

We are seeing this using the xmpp api

gpolitis:
@dwilson6 is using your own client? We've had complains from other people with the same issue. This might be related https://github.com/jitsi/jitsi-videobridge/issues/553.

dwilson6:
Yes, using a custom client. We tried adding in a fake ssrc for recvonly connections but were seeing the streams come through for a few seconds and then they would freeze indefinitely but we might not have been doing it correctly.

nojz:
@gpolitis indeed, I am using the REST interface.

gpolitis:
Could any of you please post somewhere a sample SDP from a client? I'm wondering whether an SSRC for the bridge gets advertised.

nojz:
Here is a sample remote description (IPs replaced with 0.0.0.0) of the second client:

> type: offer, sdp: v=0
[...]

Issue #553, opened 2017-10-16
Keyframe is requested from peers too early when using REST interface.


vi:
According to [rest-colibri.md](../blob/master/doc/rest-colibri.md), one should send two PATCH requests to the video bridge's REST interface: the first gets bridge's credentials, the second sends peer's credentials. Actual ICE connection gets established after the second one.

In order to provide video streams to the newly connected peer, Videobridge requests keyframes from already connected peers (FIR). But happens on the first PATCH requests, when new peer is not yet fully joined and not on the second PATCH.

How do I properly join a peer and immediately receive incoming video, without waiting for eventual another keyframe?

vi:
Looks like setting the `"last-n"` at the very end (using a third PATCH request) resolves the problem.

jitsi-developers:
I actually thought this was a scenario the bridge was handling (by caching
some of the incoming data, I think?).  But Boris or George would know for
sure.

bgrozev:
I don't know off the top of my head. I'll try to take a look and let you know in the following days.

vi:
If LastN is expected to be set by REST user at the very end of the joining process then the documentation (rest-colibri.md) should be fixed to include the third PATCH.

bgrozev:
> If LastN is expected to be set by REST user at the very end of the joining process then the documentation (rest-colibri.md) should be fixed to include the third PATCH.

No, I don't think that's the expectation. It should default to last-n=-1 (send all streams), which should result in correct behavior for keyframe requests.

gpolitis:
> Looks like setting the "last-n" at the very end (using a third PATCH request) resolves the problem.

@bbaldino is right, this is a workaround for some other underlying issue. Normally the bridge should take care of this automatically.

gpolitis:
> If LastN is expected to be set by REST user at the very end of the joining process then the documentation (rest-colibri.md) should be fixed to include the third PATCH.

That's not the expectation. The reason why this workaround solves the problem for you is because setting the last-n value causes an update to the bitrate allocator which results in a FIR.

There's a bunch of places in VideoChannel.java where we update the bitrate controller, for example when the pinned/selected endpoints change, when the streams are signaled, when the channel is initialized, when last-n is set, etc.

In your call flow the update isn't triggered, for some reason. I would expect it to be called when the channel is initialized or when you signal the streams.

gpolitis:
[...]

PR #618, opened 2018-03-01
Update the bitrate controller when the stream starts

dwilson6:
This fixes an issue with keyframes being requested too early when a stream is starting (#553, #598). Specifically sometimes when a user enters after a stream is being shared, the FIR is sent and the keyframe comes through before its ICE connection is finished setting up.

Related PR in libjtisi: https://github.com/jitsi/libjitsi/pull/404

### DIFF
--- a/src/main/java/org/jitsi/videobridge/VideoChannel.java
+++ b/src/main/java/org/jitsi/videobridge/VideoChannel.java
@@ -288,6 +288,10 @@ public class VideoChannel
         boolean previouslyStarted = getStream().isStarted();
         super.maybeStartStream();
 
+        // If a recvonly channel is created, existing streams won't be forwarded to it until the next
+        // keyframe comes in. bitrateController.update gets called before ICE has completed so the
+        // keyframe comes in before this channel is actually started. So update the bitrate controller when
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -278,6 +278,22 @@ private static synchronized RecurringRunnableExecutor getRecurringExecutor()
...
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void maybeStartStream()
+        throws IOException
+    {
+        boolean previouslyStarted = getStream().isStarted();
```
gpolitis: Could you please add some comments here that explain why this is necessary?

dwilson6: Yep, done

dwilson6:
@gpolitis anything else you'd want changed here or in the corresponding librjitsi PR?

gpolitis:
LGTM. I'll test it later today and then merge it.

gpolitis:
Closing in favor of https://github.com/jitsi/jitsi-videobridge/pull/682 which contains some related improvements.

PR #620, opened 2018-03-01
Another iteration of the Octo support

bgrozev:


src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -1301,38 +1319,29 @@ private synchronized void doStartConnectivityEstablishment(
...
-                for (IceMediaStream stream : iceAgent.getStreams())
-                {
-                    for (Component component : stream.getComponents())
-                    {
-                        component.updateRemoteCandidates();
-                    }
-                }
+                iceAgent.getStreams()
+                    .forEach(
+                        stream -> stream.getComponents()
```
bbaldino: why put this on a new line?

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
```diff
@@ -164,54 +165,85 @@ public static void deserializeCandidates(
...
-                Integer i;
-
-                if (lastN instanceof Integer)
-                    i = (Integer) lastN;
-                else if (lastN instanceof Number)
-                    i = Integer.valueOf(((Number) lastN).intValue());
-                else
-                    i = Integer.valueOf(lastN.toString());
+                Integer i = objectToInteger(lastN);
                 channelIQ.setLastN(i);
```
bbaldino: since you have the helper now, maybe it should just be `channelIQ.setLastN(objectToInteger(lastN))`?

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
```diff
@@ -164,54 +165,85 @@ public static void deserializeCandidates(
...
-
-                if (receivingSimulcastStream instanceof Integer)
-                    i = (Integer) receivingSimulcastStream;
-                else if (receivingSimulcastStream instanceof Number)
-                    i = Integer.valueOf(
-                            ((Number) receivingSimulcastStream).intValue());
-                else
-                    i = Integer.valueOf(receivingSimulcastStream.toString());
+                Integer i = objectToInteger(receivingSimulcastStream);
                 channelIQ.setReceivingSimulcastLayer(i);
```
bbaldino: same as above

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
```diff
@@ -164,54 +165,85 @@ public static void deserializeCandidates(
...
+        {
+            i = (Integer) o;
+        }
+        else if (o instanceof Number)
+        {
+            i = ((Number) o).intValue();
+        }
+        else
+        {
+            i = Integer.valueOf(o.toString());
```
bbaldino: do we want to swallow the possible exception/give a default value here? or is letting it bubble up intentional?

bbaldino: i'm surprised the method didn't need a `throws ...`

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
```diff
@@ -164,54 +165,85 @@ public static void deserializeCandidates(
...
+
+    private static Boolean objectToBoolean(Object o)
+    {
+        if (o instanceof Boolean)
+        {
+            return (Boolean) o;
+        }
+        else
+        {
+            return Boolean.valueOf(o.toString());
```
bbaldino: same as above

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
```diff
@@ -326,39 +360,40 @@ public static void deserializeChannelCommon(
...
         // endpoint
         if (endpoint != null)
+        {
             channelIQ.setEndpoint(endpoint.toString());
+        }
         // expire
         if (expire != null)
         {
-            int i;
+            int i = objectToInteger(expire);
```
bbaldino: wish this was named something more useful, i see it was existing code but it'd be nice to clean it up.

src/main/java/org/jitsi/videobridge/Content.java
```diff
@@ -250,36 +250,76 @@ public RtpChannel createRtpChannel(String channelBundleId,
...
         RtpChannel channel;
 
         synchronized (channels)
         {
             String id = generateUniqueChannelID();
 
-            switch (getMediaType())
+            if (octo)
+            {
+                channel = new OctoChannel(this, id);
```
bbaldino: we can't use `getMediaType()` for `OctoChannel` because it contains all media types, is that right?  If so, how does it end up being created within a single `Content`? Won't it need to be in both?

src/main/java/org/jitsi/videobridge/OctoChannel.java
```diff
@@ -0,0 +1,317 @@
...
+     * The {@link TransportManager} if this channel.
+     */
+    private OctoTransportManager transportManager;
+
+    /**
+     * The {@link org.ice4j.socket.DatagramPacketFilter} which (only) accepts
+     * RTP packets for this channel.
+     */
+    private RtpChannelDatagramFilter rtpFilter
+        = new OctoDatagramPacketFilter(false);
```
bbaldino: I'm wishing we had sub-types `RtpChannelRtpFilter` and `RtcpChannelRtcpFilter` instead of having to pass true/false into these all the time (even if the sub-types did nothing but pass the true/false to the parent type)

src/main/java/org/jitsi/videobridge/OctoChannel.java
```diff
@@ -0,0 +1,317 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void describe(ColibriConferenceIQ.ChannelCommon commonIq)
+    {
+        super.describe(commonIq);
+
+        // TODO: uncomment after the mvn versions are updated
+        //commonIq.setType(ColibriConferenceIQ.OctoChannel.TYPE);
```
bbaldino: should this be uncommented now?

src/main/java/org/jitsi/videobridge/OctoChannel.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    /**
+     * TODO: this should be in a utility somewhere.
+     * @param data
+     * @param off
+     * @param len
+     * @return
+     */
+    private boolean isRTCP(byte[] data, int off, int len)
```
bbaldino: agreed, i actually thought we had one already.  what about putting it in `RTPUtils`?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -0,0 +1,242 @@
...
+     * Writes an Octo header to the specified buffer at the specified offset.
+     * @param buf the buffer to write to.
+     * @param off the offset to write at.
+     * @param r the value of the {@code r} flag.
+     * @param mediaType the media type.
+     * @param s the value of the {@code s} flag.
+     * @param conferenceId the Octo ID of the conference.
+     * @param endpointId the Octo ID of the endpoint.
+     */
+    public static void writeHeaders(byte[] buf, int off,
```
bbaldino: i thought in general we were favoring `RawPacket` (instead of buf, off, len) for these sorts of APIs?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -0,0 +1,242 @@
...
+ * R: Source-is-a-relay flag. 1 if the source of the packet is a relay, and
+ * 0 if it is an endpoint (bridge).
+ * <p/>
+ * M: media type (audio, video, or data).
+ * <p/>
+ * S: Simulcast layer ID.
+ *
+ * @author Boris Grozev
+ */
+public class OctoPacket
```
bbaldino: i'm wondering why go with the static reader methods as opposed to `OctoPacket` deriving from `DatagramPacket` (or one of the other packet types) and having the members parse the internal buffer for the values?

not clear to me that it would necessarily better, but was wondering if you considered it.

src/main/java/org/jitsi/videobridge/octo/OctoTransportManager.java
```diff
@@ -0,0 +1,459 @@
...
+     */
+    @Override
+    public MediaStreamTarget getStreamTarget(Channel channel)
+    {
+        // We'll just hack in the local address so that we have something
+        // non-null. It should never be used anyway.
+        DatagramSocket socket = octoRelay.getSocket();
+        InetAddress inetAddress = socket.getLocalAddress();
+        int port = socket.getLocalPort();
+        return new MediaStreamTarget(inetAddress, port, inetAddress, port);
```
bbaldino: should it re-create this every call?

src/main/java/org/jitsi/videobridge/octo/OctoTransportManager.java
```diff
@@ -0,0 +1,459 @@
...
+    {
+        byte[] buf = p.getData();
+        int off = p.getOffset();
+        int len = p.getLength();
+
+
+        if (off >= OctoPacket.OCTO_HEADER_LENGTH)
+        {
+            // If there is space before the offset, just use it.
+            off -= OctoPacket.OCTO_HEADER_LENGTH;
```
bbaldino: this seems a bit weird to me: is the offset not basically communicating "as far as you're concerned, the buffer starts here"?  should we really be writing over data before the offset?  i think this should trust/uphold the contract of the buf/len/offset fields.

src/main/java/org/jitsi/videobridge/Content.java
```diff
@@ -552,6 +546,25 @@ private String generateChannelID()
                     System.currentTimeMillis() + Videobridge.RANDOM.nextLong());
     }
 
+    /**
+     * @return a new channel ID, unique in the list of this {@link Content}'s
+     * channels.
+     */
+    private String generateUniqueChannelID()
```
bbaldino: the old code was calling this as well, but i only see it added here, not removed from anywhere else?

src/main/java/org/jitsi/videobridge/octo/OctoTransportManager.java
bgrozev: Yeah, MediaStreamTarget is just a container for a pair of InetSocketAddress. The method is only called once when the media stream starts, so it's not worth caching

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
bgrozev: I haven't considered it before. We operate on the DatagramPacket instances coming from libjitsi, so we would have to create a new wrapping instance if we wanted to do this (or am I missing your point?).

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
bgrozev: Yes, but Octo is implemented in the bridge only. It intercepts the DatagramPackets that libjitsi sends, so there isn't an existing RawPacket instance that I can use.

src/main/java/org/jitsi/videobridge/OctoChannel.java
bgrozev: Yes! Will fix

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
bgrozev: Do you mean "i" or "objectToInteger"? I've removed the "i" as you suggested above.

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
bgrozev: It's a RuntimeException. I don't think we want to swallow it -- if we get invalid JSON parsing should fail.

src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
bgrozev: Fixed (all instances)

src/main/java/org/jitsi/videobridge/Content.java
bgrozev: The old code calls generateChannelID (without "Unique")

src/main/java/org/jitsi/videobridge/octo/OctoTransportManager.java
bgrozev: Hmm. This is interesting. It would be unfortunate if can't use this, as it might affect performance. I'm pretty sure we already do something like this in other places, so I think in our use-case it's safe. I'll do some digging and perhaps we can talk about this offline too

bgrozev: It looks like DatagramSocket#receive will respect the offset and not change it or override the data at smaller indexes. However in our code, we generally use a given byte[] only for one packet at a time. That is, there are never two packets sharing a byte[], but a byte[] can be recycled and reused for another packet. I think in this case it's a reasonable assumption that the whole byte[] is available to the packet (it just happens to be that there are some unused bytes if offset>0), and therefore the offset can be reset. The code above works with DatagramPackets which come from RTPConnectorOutputStream, and the assumption holds.

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
bbaldino: ok, my comment was based on the assumption that, wherever we created the `DatagramPacket` (or where we would normally create a RawPacket or whatever) we would instead just create an `OctoPacket`, but are you saying that there is no place where that would happen?

src/main/java/org/jitsi/videobridge/octo/OctoTransportManager.java
bbaldino: to me this seems like a "gotcha" waiting to happen...we pass buf, off, len in lots of places under the premise that they will be used as documented (what would the purpose of passing `off` be otherwise) but now we have a place that ignores the `off` field

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
bgrozev: Correct. The DatagramPacket instance comes from libjitsi, which doesn't know about Octo. We intercept the packet and add/remove Octo headers in jitsi-videobridge

src/main/java/org/jitsi/videobridge/octo/OctoTransportManager.java
bgrozev: `off` isn't ignored, it is updated. In this scheme, its purpose is to indicate where the packet's data starts (and not to protect the memory before `off`).

bbaldino: ok, i think after re-reading this some more (and also noticing that it's actually a `DatagramPacket` that's passed and not individual buf/off/len fields) i think i agree that it feels ok.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -381,6 +381,33 @@ public boolean accept(RawPacket pkt)
...
+
+    /**
+     * Computes a new bitrate allocation for every endpoint in the conference,
+     * and updates the state of this instance so that bitrate allocation is
+     * eventually met.
+     * The list of endpoints will be fetched from the
+     * {@link ConferenceSpeechActivity} of the conference, and the estimated
+     * available bandwidth will be fetched from the {@link BandwidthEstimator}.
+     */
+    public void update()
```
bbaldino: this is nice!

src/main/java/org/jitsi/videobridge/EndpointBase.java
```diff
@@ -0,0 +1,378 @@
+/*
```
bbaldino: normally i prefer this naming scheme (with `Base` for the common base object) but we seem to typically use `Abstract` as in `AbstractEndpoint`, so maybe it'd be better to do that for consistency.

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
```diff
@@ -210,41 +210,40 @@ private void doMonitor()
...
 
                 cleanupExpiredEndpointsStatus();
             }
         }
     }
 
-    private void monitorEndpointActivity(Endpoint endpoint)
+    private void monitorEndpointActivity(EndpointBase endpointBase)
     {
+        if (!(endpointBase instanceof Endpoint))
```
bbaldino: this makes me wonder if it's worth renaming `Endpoint` to something that conveys a directly-connected endpoint vs a remote one?

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
```diff
@@ -210,41 +210,40 @@ private void doMonitor()
...
+            return;
+        }
+
+        Endpoint endpoint = (Endpoint) endpointBase;
         String endpointId = endpoint.getID();
-        long lastActivity = 0;
-        long mostRecentChannelCreated = 0;
 
         // Go over all RTP channels to get the latest timestamp
         List<RtpChannel> rtpChannels = endpoint.getChannels(null);
```
bbaldino: this is another place where a method that takes no arguments (and conveys what that means) would be useful.

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -867,4 +867,23 @@ public void run()
...
+    public Endpoint getEndpoint()
+    {
+        EndpointBase endpointBase = super.getEndpoint();
+        if (endpointBase instanceof Endpoint)
+        {
+            return (Endpoint) endpointBase;
+        }
+        else
+        {
+            throw new IllegalStateException(
```
bbaldino: should this really throw here?  it seems like we have to check `instanceof` in a lot of places to check for `Endpoint` which defeats some of the advantages of the polymorphism there...could the octo/remote endpoints just have no-ops for the things we don't want them to do?

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -876,7 +876,7 @@ public void run()
     public Endpoint getEndpoint()
     {
         EndpointBase endpointBase = super.getEndpoint();
-        if (endpointBase instanceof Endpoint)
+        if (endpointBase == null || endpointBase instanceof Endpoint)
```
bbaldino: what's the scenario where this happens? (a videochannel doesn't have an endpoint) is there a time where we create a videochannel but don't set its endpoint yet?

src/main/java/org/jitsi/videobridge/EndpointBase.java
```diff
@@ -334,6 +341,7 @@ public String toString()
      */
     public void expire()
     {
+        mediaStreamTracks = new LinkedList<>();
```
bbaldino: can this just be `mediaStreamTracks.clear()`?

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
bgrozev: I wanted to avoid renaming Endpoint because it is used in so many places (including in property names)

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
bgrozev: I actually added the method already, but I didn't see this usage. Will update

src/main/java/org/jitsi/videobridge/VideoChannel.java
bgrozev: Jicofo always sets endpoints to channels, but that's not necessarily a requirement. I'll remove the whole methods as you suggested above.

src/main/java/org/jitsi/videobridge/VideoChannel.java
bgrozev: I added getSelectedEndpoints() to EndpointBase (which allowed me to remove this method from VideoChannel), but I think get/set SctpConnection() should stay in Endpoint only.

sproot:
Hi guys, thanks for moving on videobridge scaling! That's awesome!

I have a few questions.

1) Does it somehow related to the Cascaded SFUs from this video?
https://www.youtube.com/watch?v=o2SVegc0NNk
2) Do you have some approximate estimates or plans on this feature to complete?
3) Does it works already in beta? alpha?

Thank you! Great job! 👍

bgrozev:
Hi @sproot 
We don't have any estimates about when this will be completed (there isn't a alpha yet)

src/main/java/org/jitsi/videobridge/LipSyncHack.java
```diff
@@ -255,16 +255,15 @@ void onRTPTranslatorWillWriteAudio(
             return;
         }
 
-        List<MediaStreamTrackDesc> sourceTracks
-            = source.getEndpoint().getMediaStreamTracks();
-        if (sourceTracks.isEmpty())
+        MediaStreamTrackDesc[] sourceTracks
```
bbaldino: ah bummer, we can't keep this as a `List`?

src/main/java/org/jitsi/videobridge/OctoChannel.java
```diff
@@ -295,20 +320,47 @@ public boolean accept(DatagramPacket p)
...
+            }
+
+            return false;
         }
     }
+
+    /**
+     * TODO
+     */
+    private void handleDataPacket(DatagramPacket p)
```
bbaldino: what does a `data` packet refer to here?  data from a client's datachannel?

src/main/java/org/jitsi/videobridge/OctoEndpoints.java
```diff
@@ -0,0 +1,291 @@
...
+    private OctoChannel videoChannel;
+
+    /**
+     * Used to synchronize access in this class. Note that some of the
+     * operations access the {@link Conference}'s list of endpoints which
+     * uses a separate lock. Therefore, to avoid possible deadlocks this
+     * class' API MUST NOT be called while holding a lock on the conference
+     * endpoints list.
+     */
+    private final Object syncRoot = new Object();
```
bbaldino: if this is on the class level, couldn't `this` be used?

bbaldino: if not, can we rename it to something that makes it clear the intended scope of what this lock covers? i.e. if i'm adding new code, how can we make it clear whether or not i should acquire this lock when modifying data.

bbaldino: also, if we use `this`, then some of the methods (which are entirely synchronized) can change from:

```
private final Foo method(...) {
    synchronized(syncRoot) {
        // code
    }
}
```
to
```
synchronized private final Foo method(...) {
    // code
}
```

src/main/java/org/jitsi/videobridge/OctoChannel.java
bgrozev: Yes. I'll probably rename this when I get it done, because it's confusing (suggestions are welcome)

src/main/java/org/jitsi/videobridge/OctoEndpoints.java
bgrozev: As a rule of thumb I avoid synchronizing on `this` and use a private object (to make it impossible for someone use the same lock in another class). I'll clarify the doc.

src/main/java/org/jitsi/videobridge/LipSyncHack.java
bgrozev: What I did in the reverted commit was cache a copy of the array as a `List`. Now I avoided the extra copy. If we want `List` though (which I would prefer), I think we should make MediaStreamTrackReceiver hold and/or return a list.

bgrozev:
Jenkins, test this again please

bgrozev:
Jenkins, test this again please

PR #623, opened 2018-03-07
logs: Record when the bridge is oversending.

gpolitis:
This is a supplementary commit to https://github.com/jitsi/jitsi-videobridge/pull/622.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -578,6 +578,14 @@ public class BitrateController
                                 .addField("optimal_idx", trackOptimalIdx)
                                 .addField("current_bps", trackCurrentBps)
                                 .addField("target_bps", trackTargetBps)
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1109,6 +1123,7 @@ void improve(long maxBps)
                 if (!ENABLE_ONSTAGE_VIDEO_SUSPEND)
                 {
                     ratesIdx = 0;
+                    oversending = rates[0].bps > maxBps;
```
bgrozev: Is the flag always cleared? I.e. if an endpoint is selected and oversending, and then gets unselected, the oversending flag should be reset to false, right?

gpolitis: right, the TrackBitrateAllocation is ephemeral and gets destroyed after the bandwidth distribution decision.

PR #624, opened 2018-03-11
debian: add missing dependency on procps

saghul:
The postinstall script calls out to sysctl: https://github.com/jitsi/jitsi-videobridge/blob/master/resources/install/debian/postinst#L150
which is foung in the procps package.

As unusual as it may sound, there are systems which do not have the procps package installed,
such as the debian-slim Docker images.


gpolitis:
LGTM. @damencho any objections?

PR #629, opened 2018-03-23
Fix npe

bgrozev:


gpolitis:
test this please

src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -657,7 +657,7 @@ private boolean addReceiveSSRC(int receiveSSRC, boolean checkLimit)
     @Override
     protected void closeStream()
     {
-        if (!streamClosed)
+        if (!streamClosed && stream != null)
         {
```
damencho: What if stream is not closed, but stream is null. Should we always set `streamClosed = true;`

bgrozev: I don't think that it matters. The `streamClosed` flag is only used to prevent calling `close` twice.

PR #630, opened 2018-03-23
fix: Tries to fix a problem with VideobridgeExpireThread

bgrozev:
blocking for a prolonged period of time and resulting in conferences
failing to be expired.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/VideobridgeExpireThread.java
+++ b/src/main/java/org/jitsi/videobridge/VideobridgeExpireThread.java
@@ -16,6 +16,7 @@
 package org.jitsi.videobridge;
 
 import java.util.*;
+import java.util.concurrent.*;
 
 import org.jitsi.osgi.*;
 import org.jitsi.service.configuration.*;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/VideobridgeExpireThread.java
```diff
@@ -47,6 +48,13 @@
         = new RecurringRunnableExecutor(
             VideobridgeExpireThread.class.getSimpleName());
 
+    /**
+     * The executor used to expire individual {@link Channel}s.
+     */
+    private static final Executor CHANNEL_EXPIRE_EXECUTOR
```
bbaldino: i thought the idea would be to have a multi-threaded executor which handled the various expirations, is it being channel-specific here just an intermediary step?

bgrozev: One executor (with multiple threads) is used to expire all Channel-s (but not Content-s and Conference-s).

bbaldino: right, but is the plan to have a single (multi-threaded) executor handling all (channel, content, conference) expirations?

bgrozev: Oh, I misread your comment.

No, I was going to keep it like this, because expiring a Content and Conference seems simple and safe to do.

bgrozev: I don't mind changing it, though.

bbaldino: yeah i'd say while we're in here we might as well cover potential issues in all the areas and we'll be consistent with how things are expired.

PR #632, opened 2018-03-26
Adds class loader for bundles

sidorovis:


bgrozev:
Jenkins, add to whitelist

sidorovis:
https://github.com/jitsi/jicoco/pull/31

sidorovis:
Je_nkins was started a few minutes too early. https://github.com/jitsi/jicofo/pull/268 was not pushed yet.

sidorovis:
Strange, that test for firefox passed and the same for chrome not. Is it possible that OneOnOneTest is unrelated to that issue?
Still jicofo version was not pushed yet and jicofo has very strange jicoco dependency version.

damencho:
I've restarted the test and will take a look at the results. Mind that if you have a jicoco change, as jicofo depends on jicoco, the scripts will change jicofo's pom to use latest snapshot (the one build before that from your PR).

sidorovis:
@damencho jicoco version updated according to result artifact 1.1-20180327.183532-3

sidorovis:
@damencho @bgrozev when we should expect to see this updates at the jitsi-videobridge release? (the best of course through apt-get), but any released stable jar - would be good enough.

damencho:
This is already in unstable repo. Stable will come in few weeks I suppose.

PR #637, opened 2018-04-02
Octo data

bgrozev:


src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -0,0 +1,436 @@
...
+    /**
+     * Notifies this {@code Endpoint} that a {@code PinnedEndpointChangedEvent}
+     * has been received.
+     *
+     * @param src the transport channel by which {@code jsonObject} has
+     * been received.
+     * @param jsonObject the JSON object with {@link Videobridge#COLIBRI_CLASS}
+     * {@code PinnedEndpointChangedEvent} which has been received.
+     */
+    abstract protected void onPinnedEndpointChangedEvent(
```
bbaldino: i noticed a method above we just had a no-op implementation instead of making it abstract, is there a reason we force this to be implemented but `notifyTransportChannelConnected` has a default no-op?

bbaldino: or maybe it's better to ask, why do some of the message handlers below have default implementations and some don't?

src/main/java/org/jitsi/videobridge/Content.java
```diff
@@ -31,6 +31,7 @@
 import org.jitsi.util.*;
 import org.jitsi.util.event.*;
 import org.jitsi.videobridge.util.*;
+import org.jitsi.videobridge.octo.*;
```
bbaldino: import added but the file didn't change otherwise...is this needed?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -105,7 +100,24 @@ public Endpoint(String id, Conference conference)
...
+        return new EndpointMessageTransport(this);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public EndpointMessageTransport getMessageTransport()
+    {
+        return (EndpointMessageTransport) super.getMessageTransport();
```
bbaldino: what's the need for the cast here?

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
```diff
@@ -328,15 +328,14 @@ private void sendEndpointConnectionStatus(
             if (msgReceiver == null)
             {
                 // We broadcast the message also to the endpoint itself for
-                // debugging purposes
-                conference.broadcastMessage(msg);
+                // debugging purposes.
+                conference.broadcastMessage(msg, true);
```
bbaldino: could we add an inline comment as to what the `true` does

src/main/java/org/jitsi/videobridge/Endpoint.java
bgrozev: Because super returns the abstract class

bbaldino: but that type will declare all the methods, right?  the instance will be the subclass, but shouldn't having it typed as the abstract class be fine?

bgrozev: Endpoint.java uses methods specific to the subclass.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
bgrozev: If I could use a single implementation I just placed it in the base class. If I ended up with one actual implementation and a no-op, I added the no-op to the base class and overrode it. If I had two different implementations I made the base method abstract.

Many of these fell to the third category only because I wanted to log a warning in case of Octo (I don't expect them to go through Octo, and I'd like to know if something happens and they end up going). I could move these to the base implementation if you this that's a better approach.

src/main/java/org/jitsi/videobridge/Endpoint.java
bbaldino: what are the methods in `EndpointMessageTransport` that `Endpoint` uses which `AbstractEndpointMessageTransport` doesn't define?

bgrozev: `getSctpConnection`, `setSctpConnection`, `onWebSocketClose`, `onWebSocketConnect`, `onWebSocketText`,

bbaldino: those all seem like generic methods though, why not define those in the top class/interface?

bgrozev: They are not generic, SCTP and web sockets are only used by the Endpoint class (and not by the other AbstractEndpoint implementation)

PR #638, opened 2018-04-03
Configurable relayId

webintoit:
Many cloud providers, like `Scaleway` do not allows to bind directly to public ip, for those relayId may be configurable and can be different from bindAddress. 
E.g. bind to 0.0.0.0:42000 but relayId will be `11.11.11.11:42000`.


webintoit:
Jenkins CLA done

PR #639, opened 2018-04-03
ref: Simplifies the mvn-exec.sh script.

gpolitis:
The mvn-exec.sh script can be used to run from source. This PR removes most of the old options but that shouldn't break anything as the script is mostly used for debugging.


gpolitis:
@bgrozev or @bbaldino could you PTAL?

PR #652, opened 2018-04-19
fix: Prevents random RTP stream corruption.

gpolitis:


src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1090,14 +1157,38 @@ void setOptimalIndex(int newOptimalIdx)
          */
         RawPacket[] rtpTransform(RawPacket pktIn)
         {
+            long tl0Ssrc = this.tl0Ssrc;
             if (pktIn.getSSRCAsLong() != tl0Ssrc)
```
bbaldino: hmmm, what's going on here?  this used to be accessing the member variable, but now we shadow that with a local variable of the same name/value? why?

gpolitis: It's so that we print the correct value in the debug logs (the field is written by the translator thread)

bbaldino: you mean the call to `timeSeriesLogger.trace(...)` will be handled in another thread, where `tl0Ssrc` may have been updated before it reads it and logs?

gpolitis: right, there are several threads accessing the simulcast controller, the translator thread (in which the accept method runs) and the transform threads which are driving the packets in/out of the simulcast controller. The tl0Ssrc is written by the translator thread, whereas it's read by on of the transformer threads

bbaldino: right, but just so i understand, this is at least somewhat synchronized, no?  are you saying that--ignoring the trace call--`tl0ssrc` could change within the context of execution of this one method?

i believe the answer to that is no, but that the logging (where the `tl0ssrc` reference is passed) is run on yet _another_ thread, and therefore `tl0ssrc` can change before that log takes place, is that right?

gpolitis: > right, but just so i understand, this is at least somewhat synchronized, no?

No, it's not synchronized (it doesn't have to be).

> are you saying that--ignoring the trace call--tl0ssrc could change within the context of execution of this one method?

Yes.

> i believe the answer to that is no, but that the logging (where the tl0ssrc reference is passed) is run on yet another thread, and therefore tl0ssrc can change before that log takes place, is that right?

`tl0Ssrc` is passed by value. The call to the logger happens in one of the packet transform threads, which only reads the field.

bbaldino: Ok, so is it that previously `tl0Ssrc` was only read once, so even though it wasn't synchronized consistency wasn't a problem, but now we read it twice...once in the initial check, and then later in the log message, and now there's a chance it changes between those 2 lines so we want to save it off to prevent that?

gpolitis: Yes!

gpolitis: We could definitely come up with better names for a few fields, but I think doing so for the tl0Ssrc alone would leave the code in an incoherent state.

bbaldino: I'm not talking about the member variable, just the new local variable you introduced.

gpolitis: 👍

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -1090,14 +1157,38 @@ void setOptimalIndex(int newOptimalIdx)
          */
         RawPacket[] rtpTransform(RawPacket pktIn)
         {
-            if (pktIn.getSSRCAsLong() != tl0Ssrc)
+            long localTl0Ssrc = this.tl0Ssrc;
```
bbaldino: perhaps nitpicky, but can we name this something else?  either `currTl0Ssrc` or `tl0SsrcCopy`?  'local' brings up notions of 'local vs remote', as in local and remote ssrcs.

gpolitis: There is `localMaxSentFrameCopy` a few lines bellow, so.. `localTl0SsrcCopy`?

gpolitis: This is now in. Let me know if there's anything else that should change or if not feel free to merge or just approve.

PR #658, opened 2018-05-10
fix: Handles the case of an existing SctpConnection gracefully

bgrozev:


gpolitis:
The failure has something to do with torture and it's fine to merge this according to @paweldomas.

PR #660, opened 2018-05-14
Exclude SLF4J multiple bindings

realnumber:
This patch resolves the multiple SLF4J bindings error for current version (https://github.com/jitsi/jitsi-videobridge/commit/82d5c7a25d215de59360b2584c033bf3e700ac42) of JVB upon compile. Tested on Ubuntu 16.04.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,6 +137,16 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-stats</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-simple</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-jdk14</artifactId>
+        </exclusion>
+      </exclusions>
       <version>1.0-20171010.224039-4</version>
     </dependency>
     <dependency>


gpolitis:
@bgrozev @damencho do you guys have any recollection why we introduced this dependency in the first place?

damencho:
slf4j in jvb is because of Openfire integration. And in jitsi-stats it comes from callstats dependency.

gpolitis:
@damencho what would be the consequence if we don't include it in the JVB pom file?

damencho:
@gpolitis I'm not sure. We need to test: executing ant copy-runtime-dependencies-from-maven with both versions of the pom file and compare lib folder content.

guusdk:
Would it be feasible to have a jitsi-stats artifacts that does not provide the slf4j bindings (only the API) at all? Other than that, excluding the bindings from the dependency manually (like this PR does), seems like a reasonable approach.

damencho:
@guusdk Well that comes from our dependency which we do not control, callstats. https://github.com/jitsi/jitsi-stats/blob/master/pom.xml
And jitsi-stats should provide all needed dependencies, and those that have a problem with that can fix it ... this is my understanding.

guusdk:
I'm unfamiliar with these jitsi projects, but, generically, you'd want to define in projects that are libraries (intended to be re-used by other projects) only the slf4j API, and not specific slf4j bindings. You'd define the slf4j bindings (which are deployment specific) only in the 'end' project - the bit that executes, stand-alone.

Following that logic, I'd exclude the bindings from the jitsi-stats project (apply the change that's suggested in this PR to the jitsi-stats project, instead of the videobridge) if and only if, the jitsi-stats project is not supposed to be a stand-alone project. If jitsi-stats is also an executable in its own right, then this PR on videobridge makes sense to me (or, alternatively, you'd have to split up the jitsi-stats project in parts that are intended to be re-used as dependencies, and a part that is the executable bit - which sounds like way to much work for the issue at hand).

Again, this is generally speaking. We'd have to test how such a change would affect the Jitsi stack.

PR #663, opened 2018-05-17
fix: release resources when shutting down.

guusdk:
When the bridge is being closed, resources should be released explicitly. This commit closes down harvesters, which in turn unbinds from the network interface.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
+++ b/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
@@ -299,6 +299,37 @@ public class IceUdpTransportManager
     }
 
     /**
+     * Stops the static <tt>Harvester</tt> instances used by all
+     * <tt>IceUdpTransportManager</tt> instances, that is
+     * {@link #tcpHarvester} and {@link #singlePortHarvesters}.
+     *
+     * @param cfg the {@link ConfigurationService} which provides values to
+     * configurable properties of the behavior/logic of the method
+     * implementation
+     */
+    public static void closeStaticConfiguration(ConfigurationService cfg)
+    {
+        synchronized (IceUdpTransportManager.class)
+        {
+            if (!staticConfigurationInitialized)
+            {
+                return;
+            }
+            staticConfigurationInitialized = false;
+
+            if (singlePortHarvesters != null)
+            {
+                singlePortHarvesters.forEach(AbstractUdpListener::close);
+            }
+
+            if (tcpHarvester != null)
+            {
+                tcpHarvester.close();
+            }
+        }
+    }
+
+    /**
      * The single (if any) <tt>Channel</tt> instance, whose sockets are
      * currently configured to accept DTLS packets.
      */
--- a/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
+++ b/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
@@ -320,12 +320,17 @@ public class IceUdpTransportManager
             if (singlePortHarvesters != null)
             {
                 singlePortHarvesters.forEach(AbstractUdpListener::close);
+                singlePortHarvesters = null;
             }
 
             if (tcpHarvester != null)
             {
                 tcpHarvester.close();
+                tcpHarvester = null;
             }
+
+            // Reset the flag to initial state.
+            healthy = true;
         }
     }
 
--- a/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -1629,7 +1629,11 @@ public class Videobridge
         // Shut down harvesters.
         IceUdpTransportManager.closeStaticConfiguration(cfg);
 
-        // Clear all system properties that were ice4j properties.
+        // Clear all system properties that were ice4j properties. This is done
+        // to deal with any properties that are conditionally set during
+        // initialization. If the conditions have changed upon restart (of the
+        // component, rather than the JVM), it would not be enough to "not set"
+        // the system property (as it would have survived the restart).
         if (cfg != null)
         {
             List<String> ice4jPropertyNames
--- a/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -1598,8 +1598,36 @@ public class Videobridge
     void stop(BundleContext bundleContext)
         throws Exception
     {
-        videobridgeExpireThread.stop(bundleContext);
-        this.bundleContext = null;
+        try
+        {
+            ConfigurationService cfg
+                = ServiceUtils2.getService(
+                    bundleContext,
+                    ConfigurationService.class);
+
+            stopIce4j(bundleContext, cfg);
+        }
+        finally
+        {
+            videobridgeExpireThread.stop(bundleContext);
+            this.bundleContext = null;
+        }
+    }
+
+    /**
+     * Implements the ice4j-related portion of {@link #stop(BundleContext)}.
+     *
+     * @param bundleContext the {@code BundleContext} in which this
+     * {@code Videobridge} is to start
+     * @param cfg the {@code ConfigurationService} registered in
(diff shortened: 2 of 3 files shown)


guusdk:
I've updated the ice4j dependency, as this change is needed to be able to release the UDP-based harvester: https://github.com/jitsi/ice4j/pull/140

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -811,6 +811,14 @@ public synchronized void close()
             // is removed.
             closed = true;
 
+            if (singlePortHarvesters != null) {
```
bgrozev: This will close the harvesters which are shared for the whole jitsi-videobridge instance (effectively stopping the bridge) once a single transport manager instance is closed (e.g. one participant exits a conference).

guusdk: That definitely would be a bug (not a feature).

guusdk:
I have opted to create `org.jitsi.videobridge.IceUdpTransportManager#closeStaticConfiguration`. This is a counterpart of the pre-existing `org.jitsi.videobridge.IceUdpTransportManager#initializeStaticConfiguration`

Unlike its initialize counterpart, the close implementation is called directly from `Videobridge`. I am assuming that this is desirable for the initialize method as well, as soon as the issue described in the 'TODO' comment on line 758 is fixed (which is somewhat out of scope for this PR).

In `Videobridge`, a `stopIce4j` method was added to mimic the pre-existing `startIce4j`.

I've additionally added a bit of cleanup of system properties. This aims to prevent issues where settings are changed, but not effectuated after a reload of the component.

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -298,6 +298,37 @@ else if (strategy != null)
...
+            staticConfigurationInitialized = false;
+
+            if (singlePortHarvesters != null)
+            {
+                singlePortHarvesters.forEach(AbstractUdpListener::close);
+            }
+
+            if (tcpHarvester != null)
+            {
+                tcpHarvester.close();
```
bgrozev: Can you also set `tcpHarvester` and `singlePortHarvesters` to null?

I wonder if we should also reset the `healthy` flag.

bgrozev: I think we should reset the `healthy` flag. Its purpose is to trigger (be false) only when the single port harvester is configured but failed to bind.

LGTO apart from this. The failed tests are likely unrelated.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1598,8 +1598,76 @@ private void startIce4j(
...
+                if (propertyValue != null)
+                {
+                    String newPropertyName
+                        = newPrefix
+                        + propertyName.substring(oldPrefix.length());
+                    System.clearProperty(newPropertyName);
+                }
+            }
+
+            System.clearProperty(VideoChannel.ENABLE_LIPSYNC_HACK_PNAME);
```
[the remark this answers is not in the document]
guusdk: The context of this PR is working with a bridge that is restarted, while the JVM is not. By clearing / "unsetting" properties, I'm trying to deal with any properties that are conditionally set during initialization. If the conditions have changed upon restart, it would not be enough to "not set"  the system property (as it would have survived the restart).

jitsi-developers:
That's probably worth a comment somewhere in the code - it's the sort of
thing that could easily get forgotten.

Martin

bgrozev:
@guusdk I repeated the tests at least twice and they keep failing.

guusdk:
As Martin suggested, I added a bit of comment on the clearing of system properties.

guusdk:
@bgrozev About the failing tests - I'm a bit confused. The test lately have not been very stable, but your last comment seems to indicate that you at least would have expected them to succeed at least once, by now. I'm unsure how to interpret the failing of these tests.

guusdk:
I know you guys are busy, and that I've been asking a lot lately, but, given that @bgrozev approved this almost two weeks ago, and the only failing test case is the IOS test that fails all the time on various PRs: can this be merged please?

guusdk:
I"ve rebased this on the current master, in an effort to get these tests to re-execute, and hopefully, pass.

damencho:
I will look at the tests and will re-run it when I think I had fixed them, cause there are some issues with one of the machines at the moment.

damencho:
jenkins test this please

bgrozev:
I don't see what the problem is exactly, but it's not only the iOS tests that are failing.

guusdk:
Boris is obviously right (I had already removed my last comment, The '1 failing check' meta-message threw me off). There are more tests consistently failing in this PR.

I've been trying to reproduce these test failures locally without success. I'm getting a couple of other tests that fail (that mainly have to do with differences in the setup, I think - I'm running this on Openfire).

guusdk:
Maybe this is a hint: Screenshots made for the failing tests all give me a black screen that has this text:

> Sorry, you're not allowed to be here :(

org jitsi meet test chatpaneltest-web participant1

guusdk:
As the changes in this PR only change behavior when the bridge is closing, I'm wondering if the failing tests are a result of a failure to properly close things down after a previous iteration.

guusdk:
jenkins test this please

guusdk:
I'm adding some commits to selectively disable parts of the new code, in an attempt to identify what's causing the test failures.

Please don't merge this PR until I've removed all commits that have DONTMERGE in their subject.

guusdk:
Not calling stopIce4j makes the test go green.

guusdk:
Calling stopIce4j, but not invoking IceUdpTransportManager.closeStaticConfiguration from within it did cause two tests to fail - but one failure is a simple timeout. I suspect that these failures are unrelated.

guusdk:
Calling IceUdpTransportManager.closeStaticConfiguration, but not closing every harvester from within it again results in mostly green tests (mobile devices timed out, which are assumed unrelated).

guusdk:
Again, no significant problems when the TCP harvester is being closed, while commenting out the closing of all single-port harvesters. I'm deducing that the problem is caused by closing the single-port harvesters, which I'll test next.

bgrozev:
> Not calling stopIce4j makes the test go green.

I am surprised `stopIce4j` ever gets executed as part of the tests

guusdk:
> I am surprised stopIce4j ever gets executed as part of the tests

As am I. When I'm running the tests locally, I don't get these test failures. I suspect it might be related to how the test-environment is being setup. At some point, I think I'll need someones help to look into that (as I don't have access there).

guusdk:
I am now reproducing the test failures, by trying to close the single-port harvesters.

The test failures are triggered by this bit of code, it appears:

    if (singlePortHarvesters != null)
    {
        singlePortHarvesters.forEach(AbstractUdpListener::close);
        singlePortHarvesters = null;
    }

guusdk:
This surely isn't caused by the Java 8 specific syntax and API?

guusdk:
I've now tested with the original code, replacing Java 8 specific code constructs with something that's compatible with earlier versions. That didn't make a difference: the tests still fail.

guusdk:
I'm out of ideas on how to drill down further here. Closing the (single-port) harvesters appears to cause some tests to fail, but I'm unsure why, and I'm unable to debug this further. @bgrozev @damencho thoughts?

bgrozev:
`ShutdownTest.tearDown:98 » NoSuchMethod org.ice4j.ice.harvest.SinglePortUdpHar...`

I think part of the problem is jicofo's mvn tests failing because it's running an older version of ice4j. Can you create a PR in jicofo to update ice4j in pom.xml (and use the same branch name as this PR)?

guusdk:
Updated Jicofo's ice4j dependency in https://github.com/jitsi/jicofo/pull/294

guusdk:
That had good results! Thanks @bgrozev!

guusdk:
I've rebased this PR to get rid of the commits I made to test things. Jenkins test this please.

PR #665, opened 2018-05-18
Add config option to disable filtering tracks by owner

npapp:
If you do not specify an owner for a SourcePacketExtension there is no video and log is full of these warnings:
JVB 2018-05-01 21:22:16.504 WARNING: [11116] org.jitsi.videobridge.cc.BitrateController.log() Dropping an RTP packet, because the SSRC has not been signaled:1944969955

I think there are many of us who does not need this feature and you cannot set the owner via REST API by the way and the whole implementation is based on XMPP (owner is stored as org.jxmpp.jid.Jid) so I think it makes sense to make it configurable.

Let me know what you think and feel free to make any modifications or tell me if formatting, naming, variable order, etc. needs to be changed.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -16,6 +16,8 @@
 package org.jitsi.videobridge;
 
 import org.jitsi.impl.neomedia.rtp.*;
+import org.jitsi.service.configuration.ConfigurationService;
+import org.jitsi.service.libjitsi.LibJitsi;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.util.event.*;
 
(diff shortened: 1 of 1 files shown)


npapp:
I have signed your CLA as part of LogMeIn, Inc.

bgrozev:
Can you see if this fixes your problem? I'd prefer to avoid adding extra properties and logic unless necessary
https://github.com/bgrozev/jitsi-videobridge/tree/match-null-owner

nojz:
Yes, it fixes our problem. We were thinking about that change as well but we were not sure if it does not break your logic. We would be happy to see it getting merged.

npapp:
Yes, it fixes our problem. We were thinking about that change as well but we were not sure if it does not break your logic. We would be happy to see it getting merged.

bgrozev:
I changed #667 a little after discussions with Brian, but it should still solve the problem (and it's merged now)

nojz:
@bgrozev oh it is only used for "octo", then this merged solution looks like the best, yeah!
By the way, may I ask you when is your next stable debian build release date?

npapp:
@bgrozev oh it is only used for "octo", then this merged solution looks like the best, yeah!
By the way, may I ask you when is your next stable debian build release date?

bgrozev:
> By the way, may I ask you when is your next stable debian build release date?

I have no idea. We're using 1056 in production and consider it stable.

Issue #601, opened 2018-01-22
What happened to RTCP termination strategies?


nojz:
After upgrading from 879 to 953 stable build I realized that RTCP termination strategies are gone. I have an own implementation that I used (it was set in the sip-communicator.properties) but it does not work anymore.
As the [documentation](https://github.com/jitsi/jitsi-videobridge/blob/master/doc/rtcp-termination.md) seems kinda outdated (and it is noted as well), could you tell me what happened and how I should adapt to these new changes?

gpolitis:
Hi @nojz, there's been a rewrite of significant parts of the bridge about a year ago and the pluggable RTCP termination strategies architecture got removed. If the current implementation doesn't suite your needs, you can either patch the code to replace the current implementation with your own or propose a new pluggable RTCP termination architecture that's compatible with the new code.

nojz:
Okay, I got it, thanks for the quick reply @gpolitis!

PR #666, opened 2018-05-18
Add video chain manipulator OSGi service

lluczo:
This PR implements pluggable (OSGi) video chain manipulator to insert custom TransformEngines to transform chain.

https://github.com/jitsi/jitsi-videobridge/issues/601

### DIFF
--- a/src/main/java/org/jitsi/videobridge/RtpChannel.java
+++ b/src/main/java/org/jitsi/videobridge/RtpChannel.java
@@ -47,6 +47,9 @@ import org.jitsi.util.event.*;
 import org.jitsi.videobridge.transform.*;
 import org.jitsi.videobridge.xmpp.*;
 import org.jxmpp.jid.*;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceReference;
 
 /**
  * Represents channel in the terms of Jitsi Videobridge.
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/videobridge/RtpChannel.java
```diff
@@ -47,6 +47,9 @@
 import org.jitsi.videobridge.transform.*;
 import org.jitsi.videobridge.xmpp.*;
 import org.jxmpp.jid.*;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceReference;
```
gpolitis: Could you please use a wildcard import here?

src/main/java/org/jitsi/videobridge/VideoTransformChainManipulator.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * An interface for video transform chain manipulation
+ *
+ * You can register VideoTransformChainManipulator through OSGi.
+ *
+ * @author Laszlo Luczo
+ * @author Norbert Papp
+ */
+public interface VideoTransformChainManipulator
```
gpolitis: Can we rename this to TransformChainManipulator? I don't see any video-specific code.

src/main/java/org/jitsi/videobridge/transform/RtpChannelTransformEngine.java
```diff
@@ -67,16 +67,28 @@
...
                     channel.getContent().getConference().getLogger());
 
-        engineChain = createChain();
+        TransformEngine[] chain = createChain();
+
+        if (manipulators != null)
+        {
+            for (VideoTransformChainManipulator manipulator : manipulators)
+            {
+                chain = manipulator.manipulate(chain);
```
gpolitis: If I'm reading this correctly, it looks like the manipulator can't distinguish between a video and an audio stream. Would it make sense to pass the stream (this) as an argument to the manipulate method?

lluczo: Yes, the manipulator works with rtp chains, I will give the channel to manipulate function, so it can determinate the chain (channel) type. 

e.g.:
```boolean video = (channel instanceof VideoChannel);```

File: src/main/java/org/jitsi/videobridge/TransformChainManipulator.java [hunk 4]
```diff
@@ -0,0 +1,27 @@
+package org.jitsi.videobridge;
```
gpolitis: Could you please add the copyright notice here?

npapp: I added the copyright notice.

npapp:
@gpolitis is there anything else we should do?

gpolitis:
jenkins test this please

PR #667, opened 2018-05-22
fix: Match tracks with no owner to all endpoints.

bgrozev:


src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -37,7 +37,8 @@
 {
     /**
      * Filters a list of {@code tracks}, and returns the list consisting of the
-     * tracks which have a given {@code owner}.
+     * tracks which have a given {@code owner} or do not have an owner
+     * specified.
```
bbaldino: so what's the use case for this?  a method `filterTracksByOwner` which also returns tracks which have _no_ owner feels a bit weird.

bbaldino: i see this is probably for https://github.com/jitsi/jitsi-videobridge/pull/665 but this still feels like a non-intuitive behavior for this method, i'd be surprised to pass an owner to this method and get back tracks which didn't have an owner set.  would it make sense to have an optional boolean to include empty-owner tracks or something maybe?

bgrozev: The aim is to to support the two cases (setting or not setting the owner) without extra machinery (e.g. properties that have to be set (and break things if they are set incorrectly)).

I agree that the method implementation is a little non-intuitive (though note that it is clearly documented), but this makes the higher level logic cleaner: your controller (jicofo) can either include the "owner" or not, and things will just work.

bbaldino: But this means that I could call this method with 2 different owners and get a set of potentially overlapping tracks, is there a potential problem with that?

bgrozev: True, but it is quite an edge case: it could only happen if the controller used the "owner" tags inconsistently (i.e. some tracks were tagged and some weren't), and only in the case where there are tracks for more than one endpoint in a channel (i.e. Octo is used).

bbaldino: So is there no way (perhaps involving a bigger change) that we could address this more cleanly?  I know this is minor, but I feel gotchas like this that add up and lead to a frustrating experience developing in the bridge.  I'm not super familiar with this bit, so maybe this is the best option, just wanted to ask.

src/main/java/org/jitsi/videobridge/octo/OctoChannel.java
```diff
@@ -177,6 +178,14 @@ public boolean setRtpEncodingParameters(
                         .getMediaStreamTracks())
                     .map(MediaStreamTrackDesc::getOwner)
                     .collect(Collectors.toSet()));
+
+            for (SourcePacketExtension s : sources)
+            {
+                if (MediaStreamTrackFactory.getOwner(s) == null)
+                {
+                    logger.warn("Received a source without an owner tag.");
```
bbaldino: should we print some information about the s to be able to better debug if this happens?

bgrozev: It should be pretty clear from the colibri xml that we log

PR #668, opened 2018-05-23
fix: Chrome 67 simulcast compatibility.

gpolitis:


src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -193,24 +204,36 @@ protected long getTargetSSRC()
...
             return true;
         }
-        else if ((currentBaseLayerIndex > incomingFrameBaseLayerIndex) &&
-            (incomingFrameBaseLayerIndex >= targetBaseLayerIndex))
+        else
         {
-            // downscale case
-            return true;
+            // Upscale/downscale happens in a window of 300ms since the
+            // reception of the first key frame of a key frame group.
```
bbaldino: this comment seems incomplete?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -336,8 +360,12 @@ public boolean accept(RawPacket pkt)
     protected void setTargetIndex(int newTargetIdx)
     {
         bitstreamController.setTargetIndex(newTargetIdx);
+    }
 
-        if (newTargetIdx < 0)
+    private void maybeRequestKeyFrame(long nowMs)
+    {
+        int targetIdx = bitstreamController.getTargetIndex();
+        if (targetIdx < 0)
```
bbaldino: think we can use `targetIdx == SUSPEND_INDEX`, yeah?

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -349,65 +377,105 @@ protected void setTargetIndex(int newTargetIdx)
...
+                = sourceEncodings[currentTL0Idx].isActive(nowMs);
+        }
+        else
+        {
+            currentTL0IsActive = false;
         }
 
         int targetTL0Idx
-            = sourceEncodings[newTargetIdx].getBaseLayer().getIndex();
+            = sourceEncodings[targetIdx].getBaseLayer().getIndex();
```
bbaldino: maybe it's hard to tell because of the diff, but this was intentional to change this from `newTargetIdx` to `targetIdx`?

gpolitis: The code in `maybeRequestKeyFrame()` was part of the `setTargetIndex(int newTargetIndex)` method. The new method doesn't have access to the `newTargetIndex` variable, so we use a new variable called `targetIndex` (and we get the value from the bitstreamController).

gpolitis: tl;dr; it's the same value, only the name changed

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
```diff
@@ -461,12 +461,12 @@ else if (targetTL0Idx != currentTL0Idx
...
         }
 
-        if (timeSeriesLogger.isTraceEnabled())
+        if (timeSeriesLogger.isInfoEnabled())
         {
             DiagnosticContext diagnosticContext
                 = getDiagnosticContext();
 
-            timeSeriesLogger.trace(diagnosticContext
+            timeSeriesLogger.info(diagnosticContext
```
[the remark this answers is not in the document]
gpolitis: The time series logger works just like any other logger that we have; I wanted this message printed the the simulcast time series logger INFO level is enabled. This particular log message is not as chatty as others that are meant for TRACE level.

src/main/java/org/jitsi/videobridge/cc/SimulcastController.java
gpolitis: I reworded the comment, let me know if it needs more attention.

gpolitis:
The test failure is unrelated (it's a device disconnect problem on Android) so this is clear for merging.

PR #676, opened 2018-06-05
fix: Prevents RTP stream corruption.

gpolitis:


bbaldino:
interested to hear about what the bug was here.  something with the seen frame allocator?

gpolitis:
The packet rewrite thread uses SeenFrames to rewrite RTP packets. The translator thread creates/updates/recycles SeenFrames. It is possible for a SeenFrame to be returned (by the translator thread) to the allocator for re-use, while it's being used by the packet rewrite thread. This happens extremely rarely, but it can happen.

gpolitis:
The "major" fix is disabling the lipsync hack by default because it occasionally injects packets that are not re-written..

gpolitis:
And also bumping the seen frame history size.

PR #678, opened 2018-06-06
Notify endpoints if they are currently selected

bbaldino:
Start tracking which endpoints are 'selected' and notify the endpoints
if they just transitioned from being selected by no participants
to being selected by at least one or vice versa.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -1435,6 +1435,26 @@ public class Conference
         {
             speechActivityPropertyChange(ev);
         }
+        else if (Endpoint.SELECTED_ENDPOINTS_PROPERTY_NAME.equals(ev.getPropertyName()))
+        {
+            Set<String> oldSelectedEndpoints = (Set<String>)ev.getOldValue();
+            Set<String> newSelectedEndpoints = (Set<String>)ev.getNewValue();
+            // Any endpoints in the oldSelectedEndpoints list which AREN'T
+            // in the newSelectedEndpoints list should have their count decremented
+            oldSelectedEndpoints.stream()
+                .filter(oldSelectedEp -> !newSelectedEndpoints.contains(oldSelectedEp))
+                .map(this::getEndpoint)
+                .filter(Objects::nonNull)
+                .forEach(AbstractEndpoint::decrementSelectedCount);
+
+            // Any endpoints in the newSelectedEndpoints list which AREN'T
+            // in the oldSelectedEndpoints list should have their count incremented
+            newSelectedEndpoints.stream()
+                .filter(newSelectedEp -> !oldSelectedEndpoints.contains(newSelectedEp))
+                .map(this::getEndpoint)
+                .filter(Objects::nonNull)
+                .forEach(AbstractEndpoint::incrementSelectedCount);
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bbaldino:
I didn't make this configurable on the bridge...I figured we'd do that on the endpoint side (to turn the processing of the messages on/off).  If you guys think we should have a switch on the bridge side too I can add it.

paweldomas:
> I didn't make this configurable on the bridge...I figured we'd do that on the endpoint side (to turn the processing of the messages on/off). 

I can't find the logic for turning on/off in the lib-jitsi-meet PR:
https://github.com/jitsi/lib-jitsi-meet/pull/756
Am I missing something ?

bbaldino:
> > I didn't make this configurable on the bridge...I figured we'd do that on the endpoint side (to turn the processing of the messages on/off).
> 
> 
> I can't find the logic for turning on/off in the lib-jitsi-meet PR:
> jitsi/lib-jitsi-meet#756
> Am I missing something ?

[This line](https://github.com/jitsi/lib-jitsi-meet/pull/756/files#diff-f1e63fae69e63cb1f07f9120cd842149R1876) checks a value in options, by default this won't be set in config.js

paweldomas:
> This line checks a value in options, by default this won't be set in config.js

I see, so it's not like it will be disabled on the bridge, but the client will just ignore it.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -17,8 +17,13 @@
 
 import java.io.*;
 import java.util.*;
+import java.util.concurrent.atomic.*;
+
 import org.jitsi.util.*;
 import org.jitsi.videobridge.rest.*;
+import org.json.simple.*;
+
+import static org.jitsi.videobridge.EndpointMessageBuilder.createSelectedUpdateMessage;
```
bgrozev: The org.json.simple package seems unused, use * for the static import.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1435,6 +1435,26 @@ else if (source == speechActivity)
...
         }
+        else if (Endpoint.SELECTED_ENDPOINTS_PROPERTY_NAME.equals(ev.getPropertyName()))
+        {
+            Set<String> oldSelectedEndpoints = (Set<String>)ev.getOldValue();
+            Set<String> newSelectedEndpoints = (Set<String>)ev.getNewValue();
+            // Any endpoints in the oldSelectedEndpoints list which AREN'T
+            // in the newSelectedEndpoints list should have their count decremented
+            oldSelectedEndpoints.stream()
+                .filter(oldSelectedEp -> !newSelectedEndpoints.contains(oldSelectedEp))
+                .map(this::getEndpoint)
```
bgrozev: I wonder if calling `getEndpoint` (which has to synchronize and clean the list) each time is ideal, but I guess it's not a problem since the lists should have a single element most of the time.

bbaldino:
> > This line checks a value in options, by default this won't be set in config.js
> 
> I see, so it's not like it will be disabled on the bridge, but the client will just ignore it.

Right, that's what I meant about not being configurable on the bridge...the bridge will always send it and it's just whether or not the client will look at it.  It seemed cheap enough to me to not be a big deal to do that, but it's something we could do.

src/main/java/org/jitsi/videobridge/Endpoint.java
bbaldino: fixed

PR #682, opened 2018-06-08
Update the bitrate controller when the stream starts v2

gpolitis:
This augments and supersedes https://github.com/jitsi/jitsi-videobridge/pull/618.


gpolitis:
Safari and Android fail for unrelated reasons. @bgrozev could you please take another look?

PR #683, opened 2018-06-11
Prevent forcefully enabling the dynamic host harvester

guusdk:
When the single-port UDP harvester is available, there's no need to offer the dynamic-port UDP harvester to the client, as there's little chance that the client would have a different result using either.

However, this does not mean that when _not_ using single-port UDP harvester, the dynamic-port UDP harvester should be forcefully _enabled_. The dynamic-port harvester is enabled by default. By forcefully enabling it, its configuration option (`org.ice4j.ice.harvest.USE_DYNAMIC_HOST_HARVESTER`) is overridden, which is undesirable (as it prevents one from disabling this harvester).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
+++ b/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
@@ -756,7 +756,7 @@ public class IceUdpTransportManager
             = ServiceUtils.getService(
                     getBundleContext(),
                     ConfigurationService.class);
-        boolean enableDynamicHostHarvester = true;
+        boolean disableDynamicHostHarvester = false;
 
         if (rtcpmux)
         {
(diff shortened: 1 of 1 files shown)


guusdk:
I don't think that the Safari test failures are related, are they?

guusdk:
jenkins test this please

guusdk:
Retesting took care of the Safari tests. They're no longer failing. The one test that is failing is the IOS-based test. Earlier today, I was told that this was a hardware issue.

guusdk:
Jenkins test this please.

guusdk:
Jenkins test this please.

guusdk:
Jenkins test this please.

bbaldino:
I'm sure there's context in discussions you and boris have had that I'm missing here, so maybe I just need a bit of catchup: but why are these dynamic at all and not just obeying configuration? (i.e. if single port is enabled in config: start it; if dynamic is enabled: start it).  The only reason I can think of it dynamically enabling/disabling another is for a sort of error-fallback if one fails, but I wonder if it's more confusing to allow that sort of behavior (basically you won't know which ports it'll end up using until you start it up/try a call).

bbaldino:
> My issue with the code as-is, is that is explicitly disobeys configuration.

Is this because you do not enable the single port harvester in config, but the code is enabling it anyway because of the fallback?

bbaldino:
@bgrozev would you be opposed to making the bridge do whatever it is configured to do with regards to harvesters (no more, no less)?

guusdk:
Although I'm not opposed to what @bbaldino is proposing, would it be acceptable to d that out of scope of this PR?

bbaldino:
hey @guusdk i think we need a little more time on this.  i've been chatting with boris during his free moments while at a conference, so hopefully we can come to some decision soon.

bgrozev:
> @bgrozev would you be opposed to making the bridge do whatever it is configured to do with regards to harvesters (no more, no less)?

I don't think that's practical. We don't want to use single port only (in order to avoid extra candidates) for clients, but we need the dynamic ports for health checks. I think what we have now that the PR is merged is as good as it gets. The way I think of it: the config properties describe what the bridge is allowed to use, and then it applies some extra runtime logic in order to not always use everything  (dependent on rtcpmux). But it wouldn't override your config and use something that you've disabled.

PR #689, opened 2018-06-27
feat: Adds new stats to track bwe limited-seconds and participant-seconds.

gpolitis:
This PR adds three new stats:
- total participant-seconds
- total loss-limited participant-seconds (which is the total number of
  seconds that the bandwidth is limited due to loss.
- total loss-degraded participant seconds (which is the total number of
  seconds that the bandwidth is degraded due to loss.

Depends on https://github.com/jitsi/libjitsi/pull/426

### DIFF
--- a/src/main/java/org/jitsi/videobridge/VideoChannel.java
+++ b/src/main/java/org/jitsi/videobridge/VideoChannel.java
@@ -479,24 +479,27 @@ public class VideoChannel
             if (bwe != null)
             {
                 BandwidthEstimator.Statistics bweStats = bwe.getStatistics();
-                bweStats.update(System.currentTimeMillis());
+                if (bweStats != null)
+                {
+                    bweStats.update(System.currentTimeMillis());
 
-                Videobridge.Statistics videobridgeStats
-                    = getContent().getConference().getVideobridge()
-                    .getStatistics();
+                    Videobridge.Statistics videobridgeStats
+                        = getContent().getConference().getVideobridge()
+                        .getStatistics();
 
-                long lossLimitedMs = bweStats.getLossLimitedMs();
-                long lossDegradedMs = bweStats.getLossDegradedMs();
-                long participantMs = bweStats.getLossFreeMs()
-                    + lossDegradedMs + lossLimitedMs;
+                    long lossLimitedMs = bweStats.getLossLimitedMs();
+                    long lossDegradedMs = bweStats.getLossDegradedMs();
+                    long participantMs = bweStats.getLossFreeMs()
+                        + lossDegradedMs + lossLimitedMs;
 
-                videobridgeStats.totalLossControlledParticipantMs
-                    .addAndGet(participantMs);
-                videobridgeStats.totalLossLimitedParticipantMs
-                    .addAndGet(lossLimitedMs);
+                    videobridgeStats.totalLossControlledParticipantMs
+                        .addAndGet(participantMs);
+                    videobridgeStats.totalLossLimitedParticipantMs
+                        .addAndGet(lossLimitedMs);
 
-                videobridgeStats.totalLossDegradedParticipantMs
-                    .addAndGet(lossDegradedMs);
+                    videobridgeStats.totalLossDegradedParticipantMs
+                        .addAndGet(lossDegradedMs);
+                }
             }
         }
 
--- a/src/main/java/org/jitsi/videobridge/VideoChannel.java
+++ b/src/main/java/org/jitsi/videobridge/VideoChannel.java
@@ -281,8 +281,7 @@ public class VideoChannel
         throws IOException
     {
         MediaStream stream = getStream();
-        boolean previouslyStarted
-            = stream != null ? stream.isStarted() : false;
+        boolean previouslyStarted = stream != null && stream.isStarted();
 
         super.maybeStartStream();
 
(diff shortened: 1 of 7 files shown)


gpolitis:
I've also added a link-conditioner script in the resources directory. I can take it out of the PR if anybody disagrees with having it.

gpolitis:
The iOS build seems to be in a pretty bad shape, the build failure is not related to this PR.

src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -445,6 +469,11 @@ protected void generate0()
             totalConferencesCompleted
                 += jvbStats.totalConferencesCompleted.get();
             totalConferenceSeconds += jvbStats.totalConferenceSeconds.get();
+            totalParticipantSeconds += jvbStats.totalParticipantSeconds.get();
```
bbaldino: this is confusing to me.  we use `+=` but isn't this really just trying to achieve `totalParticipantSeconds = jvbStats.totalParticipantSeconds.get()`?  we don't actually accumulate multiple values in `totalParticipantSeconds`, right?  it should just be set to the value of `jvbStats.totalParticipantSeconds`?

gpolitis: In theory there can be multiple Videobridge instances, so totalX is an accumulator.

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -611,7 +634,7 @@ else if ("goog-remb".equals(rtcpFb.getAttribute("type")))
         }
 
         MediaStream mediaStream = getStream();
-        if (mediaStream != null && mediaStream instanceof VideoMediaStreamImpl)
+        if (mediaStream instanceof VideoMediaStreamImpl)
```
damencho: Can mediaStream be null here, as you are getting rid of the check?

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -472,6 +470,32 @@ public boolean expire()
...
+            BandwidthEstimator.Statistics bweStats
+                = ((VideoMediaStream) mediaStream)
+                .getOrCreateBandwidthEstimator().getStatistics();
+            bweStats.update(System.currentTimeMillis());
+
+            Videobridge.Statistics videobridgeStats
+                = getContent().getConference().getVideobridge()
+                .getStatistics();
+
+            long lossLimitedSeconds = bweStats.getLossLimitedMs() / 1000;
```
bgrozev: We are converting from milliseconds to seconds on a per-channel basis. Is this good enough? I expect that a lot of our calls spend very few ms (<500) in the lossDegraded state for example, and they might get cut to 0. I suggest we store these at milliseconds in Videobridge.Statistics and only convert to seconds the totals (in VideobridgeStatistics.java), or not convert to seconds at all.

src/main/java/org/jitsi/videobridge/VideoChannel.java
```diff
@@ -472,6 +470,32 @@ public boolean expire()
...
+            long lossDegradedSeconds = bweStats.getLossDegradedMs() / 1000;
+            long participantSeconds = bweStats.getLossFreeMs() / 1000
+                + lossDegradedSeconds + lossLimitedSeconds;
+
+            videobridgeStats.totalParticipantSeconds
```
bgrozev: This will fail to count audio only participants (e.g. jigasi)

src/main/java/org/jitsi/videobridge/VideoChannel.java
gpolitis: A null check is not needed before using instanceof because the expression `x instanceof SomeClass` is `false` if x is `null`.

src/main/java/org/jitsi/videobridge/VideoChannel.java
gpolitis: Ah, right :( in this PR I care about the loss-controlled seconds so I renamed the variable (we'll have to wait for the true participant-seconds).

src/main/java/org/jitsi/videobridge/VideoChannel.java
gpolitis: 👍

gpolitis:
test this please

src/main/java/org/jitsi/videobridge/VideoChannel.java
bgrozev: Aww, I was excited about it :)

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1834,22 +1834,24 @@ private int getContentStreamCount(
         public AtomicLong totalConferenceSeconds = new AtomicLong();
 
         /**
-         * The total duration in seconds of all participants on this
-         * {@link Videobridge}.
+         * The total number of participant-milliseconds that are loss-controlled
+         * (i.e. the sum of the lengths is seconds) on this {@link Videobridge}.
```
bgrozev: "the sum of the lengths is seconds"? Should this say "in milliseconds"?

PR #704, opened 2018-08-13
Health-check cache

bgrozev:


paweldomas:
There's similar logic in Jicof already: https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/rest/Health.java#L275
Would it be possible to have something shareable in Jicoco ?

bgrozev:
> Would it be possible to have something shareable in Jicoco ?

Jicofo uses a boolean and the bridge uses an exception. I thought the logic was simple enough not to need to be generalized.

PR #706, opened 2018-08-15
fix: Fixes the health check cache.

bgrozev:


src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -182,7 +182,7 @@ public static void check(Videobridge videobridge)
 
         if (cachedExceptionTimestamp > 0 &&
```
paweldomas: will the code be simpler if you include `cachedException != null` as the first `if` condition ?

paweldomas: No, actually not. It would not be correct

bgrozev: No, I want to finish execution (either throw or return) regardless of the value of cachedException

paweldomas: right

PR #708, opened 2018-08-17
Run health checks in a separate thread

bgrozev:


src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -475,4 +476,110 @@ private static void interconnect(Endpoint[] endpoints)
...
+                exception = e;
+            }
+
+            long duration = System.currentTimeMillis() - start;
+            lastResult = exception;
+            lastResultMs = start + duration;
+
+            if (exception == null)
+            {
+                logger.info(
```
damencho: Wouldn't that be too chatty?

bgrozev: It's just one line every 10 seconds, and I think it might be useful to have.

PR #720, opened 2018-10-16
fix: Fixes the order in which TCP/443 is bound.

bgrozev:
Loads the REST bundle before the Videobridge bundle, because Jetty needs
to bind on port 443 before ice4j's TCP harvester does. When Videobridge
loads it starts its health checks, which results in the TCP harvester
binding.


bgrozev:
jenkins, please test this again please

saghul:
🚔  THis could have used a comment so someone is not tempted to reorder things in the future, IMHO.

PR #722, opened 2018-10-16
Redirects all requests to jetty port 80 to the secure port(default 443).

damencho:


src/main/java/org/jitsi/videobridge/rest/PublicRESTBundleActivator.java
```diff
@@ -64,6 +64,8 @@
     public static final String JETTY_CORS_ALLOWED_ORIGINS
         = ".jetty.cors.allowedOrigins";
 
+    public static final String JETTY_TLS_PORT_PNAME = ".jetty.tls.port";
```
bgrozev: Why are we not re-using the existing constant for this (same name in super)?

bgrozev:
What happens if the bridge fails to bind on 80? Just a warning in the logs?

src/main/java/org/jitsi/videobridge/rest/PublicRESTBundleActivator.java
damencho: Ups, I missed it. Fixing it.

damencho:
> What happens if the bridge fails to bind on 80? Just a warning in the logs?

And two stack traces, but they are coming from the abstract and we cannot skip them. But other than that everything continues as usual.

src/main/java/org/jitsi/videobridge/rest/PublicClearPortRedirectBundleActivator.java
```diff
@@ -0,0 +1,173 @@
+/*
+ * Copyright @ 2015 Atlassian Pty Ltd
```
saghul: 2018

damencho: eee copy paste should automatically replace this :)

PR #723, opened 2018-10-19
Use WebRTC data channels opened by the client

paweldomas:
Starting in Chrome M69 it will no longer release WebRTC Data Channels when the "data" content is [rejected] on ICE restart. I suspect it's this [commit in WebRTC] which waits for the channels to be gracefully closed (but haven't bisected that to confirm). Because the conference is moved to new bridge it will never happen. So the stack correctly resets the SCTP transport layer, but keeps the SID = 0 used by the bridge and does not allow it to open it's data channel on SID = 0 again.

As a workaround (because it's to late to file a bug and have a fix in the stable) I was thinking about moving opening of the data channel to the client, so that it know which next SID is free. This also makes the JVB a bit more elastic for other clients, because should accommodate both cases where JVB opens channel or the client.

SID is a SCTP stream identifier which values range goes from 0 - 1024 in WebRTC (up to 65k allowed by SCTP). Each WebRTC data channel has a SID assigned and the rule there is that the DTLS initiator uses even SID number and the client odd ones to avoid conflicts.

[rejected]: https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/JingleSessionPC.js#L989
[commit in WebRTC]: https://webrtc.googlesource.com/src.git/+/cdd05f0cc1edd36ca5c61ab7154beab63fc3bbe7

### DIFF
--- a/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
+++ b/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
@@ -82,6 +82,12 @@ class EndpointMessageTransport
         = new WeakReference<>(null);
 
     /**
+     * The {@link WebRtcDataStream} currently used by this
+     * {@link EndpointMessageTransport} instance for writing.
+     */
+    private WebRtcDataStream webrtcDataStream;
+
+    /**
      * The listener which will be added to {@link SctpConnection}s associated
      * with this endpoint, which will be notified when {@link SctpConnection}s
      * become ready.
@@ -95,26 +101,14 @@ class EndpointMessageTransport
         = new WebRtcDataStreamListener()
     {
         @Override
-        public void onSctpConnectionReady(
-            SctpConnection source)
+        public void onChannelOpened(
+                SctpConnection source, WebRtcDataStream channel)
         {
-            if (source != null)
+            SctpConnection currentConnection = getSctpConnection();
+
+            if (source.equals(currentConnection))
             {
-                if (source.equals(getSctpConnection()))
-                {
-                    try
-                    {
-                        WebRtcDataStream dataStream
-                            = source.getDefaultDataStream();
-                        dataStream.setDataCallback(EndpointMessageTransport.this);
-                        notifyTransportChannelConnected();
-                    }
-                    catch (IOException e)
-                    {
-                        logger.error("Could not get the default data stream.", e);
-                    }
-                }
-                source.removeChannelListener(this);
+                hookUpDefaultWebRtcDataChannel(currentConnection);
             }
         }
     };
@@ -130,6 +124,41 @@ class EndpointMessageTransport
         this.logger
             = Logger.getLogger(
                     classLogger, endpoint.getConference().getLogger());
+    }
+
+    /**
+     * At the time when new data chanel is opened or {@link SctpConnection}
+     * instance is replaced, this method will reelect the default WebRTC data
+     * channel (the one that's used for writing).
+     *
+     * @param connection - The currently used {@link SctpConnection} instance
+     * from which the default data channel will be obtained.
+     */
+    private void hookUpDefaultWebRtcDataChannel(SctpConnection connection)
+    {
+        // FIXME make "data stream" vs "data channel" naming consistent in both
+        // code and comments
+        WebRtcDataStream _defaultStream
+            = connection != null ? connection.getDefaultDataStream() : null;
+
+        if (_defaultStream != null)
+        {
+            WebRtcDataStream oldDataStream = this.webrtcDataStream;
+
+            this.webrtcDataStream = _defaultStream;
+
+            _defaultStream.setDataCallback(EndpointMessageTransport.this);
+
+            if (oldDataStream == null)
+            {
+                logger.info(
+                    String.format(
+                        "WebRTC data channel established for %s",
+                        connection.getLoggingId()));
+
+                notifyTransportChannelConnected();
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


gpolitis:
What is the context of this work? Has Chrome implemented some new behavior that is not already handled by the bridge?

mstyura:
I can confirm that I has an issue which this pull request seem to solve.
I work on project which uses standalone instance of videobridge (without meet). 
Data channel opening is initiated by client joining conference and I've seen that sending open data channel acknowledgment is failed because sctp socked was in listening state. The mention of the problem in this [pull request](https://github.com/jitsi/jitsi-videobridge/pull/723/files#diff-64572911432161b3e51a7e7ea578e6bdR178)

paweldomas:
@gpolitis updated the description

paweldomas:
Please note that I was thinking about it over the weekend and I plan to split "delaying" and accepting client's channels in to two commits. I also plan to rework the packet delaying to the async executor (if possible) to use it for execution ordering and avoid having the delayed packets queue at all.

paweldomas:
Updated to not have the delayed packets queue. It was kind of hard to split the commit into two, so I'll keep it this way unless someone insists.

src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -382,41 +411,51 @@ public boolean expire()
...
-//                    def = null;
+                // Tries to elect a stream with the highest client SID
+                WebRtcDataStream highestClientSid
+                    = channels.values()
+                        .stream()
+                        .filter(
+                            s -> isInitiator()
+                                ? s.getSid() % 2 == 1
+                                : s.getSid() % 2 == 0)
+                        .reduce((s1, s2) -> s1.getSid() > s2.getSid() ? s1 : s2)
```
mstyura: @paweldomas looks like [`.max(Comparator.comparingInt(x -> x.getSid())`](https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html#comparingInt-java.util.function.ToIntFunction-) would be more explicit way to express lookup channel with biggest sid.

paweldomas: yeah, thanks ! that looks useful

mstyura:
Is there any estimates when this PR could be merged?

saghul:
@mstyura Hi there! @paweldomas is out until late December IRC, so it may take a bit.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -81,6 +81,12 @@
     private WeakReference<SctpConnection> sctpConnection
         = new WeakReference<>(null);
 
+    /**
+     * The {@link WebRtcDataStream} currently used by this
+     * {@link EndpointMessageTransport} instance for writing.
+     */
+    private WebRtcDataStream webrtcDataStream;
```
gpolitis: Non-blocking: Maybe add a prefix to indicate that this is the "writable" WebRTC data stream?

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -557,18 +585,32 @@ void setSctpConnection(SctpConnection sctpConnection)
...
             {
-                if (sctpConnection.isReady())
-                {
-                    notifyTransportChannelConnected();
-                }
-                else
-                {
-                    sctpConnection.addChannelListener(webRtcDataStreamListener);
-                }
+                hookUpDefaultWebRtcDataChannel(sctpConnection);
```
gpolitis: Why do we no longer need to check if the sctpConnection is ready?

paweldomas: The behavior of `SctpConnection.getDefaultDataStream()` called from `hookUpDefaultWebRtcDataChannel` has changed to only select the stream and return `null` if there isn't any. Previously this method would try to open a new one which involves writing on the sctp socket and for that it had to be ready.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
paweldomas: sure

src/main/java/org/jitsi/videobridge/SctpConnection.java
```diff
@@ -776,61 +834,12 @@ public void onSctpPacket(
             byte[] data, int sid, int ssn, int tsn, long ppid, int context,
             int flags)
     {
-        if (ppid == WEB_RTC_PPID_CTRL)
-        {
-            // Channel control PPID
-            try
+        sctpDispatcher.execute(() -> {
```
gpolitis: Is there a chance for the data to be altered from some other thread? The buffer is passed by reference, so in theory it's possible' but, in practice, that shouldn't happen in our case, right?

src/main/java/org/jitsi/videobridge/WebRtcDataStream.java
```diff
@@ -160,7 +189,19 @@ public void sendString(String strMsg)
...
     {
         if(dataCallback != null)
+        {
             dataCallback.onBinaryData(this, binMsg);
+        }
+        else
+        {
+            // NOTE consider adding analytics for the missed out data to detect
+            // bugs ?
+            logger.error(
```
gpolitis: Just out of curiosity, has this ever occurred during your trials? If so, have you been able to identify the conditions under which this may occur?

PR #728, opened 2018-12-03
systemd unit file

damencho:


resources/install/debian/control
```diff
@@ -3,15 +3,14 @@ Section: net
...
+Build-Depends: debhelper (>= 9), dh-systemd
 Standards-Version: 3.9.3
 Homepage: https://jitsi.org/videobridge
 
 Package: jitsi-videobridge
 Architecture: any
 Pre-Depends: openjdk-8-jre-headless
-Depends: ${misc:Depends}, procps
-Recommends: authbind
+Depends: ${misc:Depends}, procps, systemd (>= 29.1)
```
saghul: The current version (upstream) is v239. Isn't this dependency ancient?

resources/install/debian/jitsi-videobridge.service
```diff
@@ -0,0 +1,24 @@
...
+RuntimeDirectory=jitsi-videobridge
+RuntimeDirectoryMode=0750
+PIDFile=/var/run/jitsi-videobridge/jitsi-videobridge.pid
+# more threads for this process
+TasksMax=65000
+# allow more open files for this process
+LimitNPROC=65000
+LimitNOFILE=65000
+ExecStart=/bin/bash -c "exec /usr/share/jitsi-videobridge/jvb.sh --host=${JVB_HOST:-localhost} --domain=${JVB_HOSTNAME} --port=${JVB_PORT} --secret=${JVB_SECRET} ${JVB_OPTS} < /dev/null >> ${LOGFILE} 2>&1"
+ExecStartPost=/bin/bash -c "systemctl show -p MainPID jitsi-videobridge.service 2>/dev/null | cut -d= -f2 > /var/run/jitsi-videobridge/jitsi-videobridge.pid"
```
saghul: Doesn't systemd automatically get the right PID since we exec twice?

resources/collect-dump-logs.sh
```diff
@@ -11,6 +11,11 @@ JVB_UID=`id -u jvb`
 RUNNING=""
 unset PID
 
+# for systemd we use different pid file
```
saghul: why a different file

resources/install/debian/control
damencho: Well, I saw it at some place with examples ... I started working on this maybe two months ago and I do not remember what was the reason for this ... there was something not supported before this version. But I think to just remove it then, and leave systemd requirement without version, WDYT?

resources/install/debian/jitsi-videobridge.service
damencho: Yes, the pid is correct.

resources/collect-dump-logs.sh
damencho: Cause jvb user does not have the write to create files under /var/run. But specifying RuntimeDirectory it is created there before we create the pid file, and there the pid file is successfully created.

resources/install/debian/jitsi-videobridge.service
saghul: So, why do we need to do this dance to put it into the file then?

resources/install/debian/control
saghul: Sounds good!

resources/collect-dump-logs.sh
saghul: But doesn't systemd writ the PID file?

resources/install/debian/jitsi-videobridge.service
damencho: Cause systemd does not write pid files, it is up to the underlying scripts/software to do that.

resources/collect-dump-logs.sh
damencho: Nope. Having a specified pid file in the unit file is just for helping systemd the find the process or delete it ... but it is not creating pid files.

resources/install/debian/jitsi-videobridge.service
saghul: Doh. I see. At any rate, I found about `$MAINPID`: https://medium.com/@jbriggs_24705/you-should-be-using-pidfile-and-mainpid-instead-of-pkill-1935e4531931

resources/collect-dump-logs.sh
saghul: Gotcha.

damencho:
I think I fixed the tests with a jicofo PR, let's see.

Issue #649, opened 2018-04-16
VP9 (plain) is not working with recent JVB


Toumassa:
Hi,

As mentioned in #573, if the codec VP9 (not SVC) is set as preferred or the only codec, no video is received by the participants. In p2p it works.

This issue exists in any stable release that comes after jitsi-videobridge_879 2017-01-05. I tested the master today, still the has the issue.

Please, can you look at this?

Toumassa:
Any news on this?

Toumassa:
The code to blame is inside libjitsi. The problem appears after merged pull request `Merge pull request #238 from jitsi/lastn-sim-rtcp-ref `  jitsi/libjitsi#238

PR #735, opened 2019-01-04
simulcast/svc rewriting re-architecture

gpolitis:
What started as an investigation of video artifacts when VP8 simulcast/svc was used ended up in the complete overhaul of the rewriting logic. This PR depends on https://github.com/jitsi/libjitsi/pull/456. This PR provides:

* A fix for the video artifacts
* Re-architecture of the VP8 simulcast/temporal scalability code in order to improve its robustness, resiliency, clarity and readability.
* Improve performance and efficiency of rewriting code (e.g. I've removed keeping track of seen frames in the ingress chain).
* Adds test coverage for VP8 simulcast/temporal scalability rewriting code that covers 100% of the basic use cases (commit pending)
* Adds documentation of VP8 simulcast/svc rewriting code
* Properly support adaptive last-n with H264 and VP9 (#649).


src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -0,0 +1,285 @@
+/*
+ * Copyright @ 2018 Atlassian Pty Ltd
```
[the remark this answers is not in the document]
gpolitis: Thanks @saghul 👍

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * @author George Politis
+ */
+public interface AdaptiveTrackProjectionContext
+{
+    /**
+     * An empty {@link RawPacket} array that is used as a return value and
+     * indicates that the input packet needs to be dropped.
+     */
+    RawPacket[] DROP_PACKET_ARR = new RawPacket[0];
```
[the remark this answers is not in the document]
gpolitis: Static and final modifiers are redundant for interface fields, but I can add it for added clarity?

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    boolean accept(RawPacket rtpPacket, int targetIndex);
+
+    /**
+     * @return true if this instance needs a keyframe, false otherwise.
+     */
+    boolean needsKeyframe();
+
+    /**
+     * Rewrites the RTP packet that is specified as an argument.
```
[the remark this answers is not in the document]
gpolitis: > can you add more detail? rewrites in what way(s)? i assume something like `rewrite the timestamp, sequence number and ssrc to keep the rtp stream consistent` ?

👍 



> also, i'm a bit confused by the return value: why is this named around rewriting but then also returns whether or not it should be accepted? it seems like those 2 bits of logic could be separated

I know it's not pretty and it looks confusing; here I need a way to return an error value, just in case something goes wrong, for example, because of an exotic case we've accepted a packet, so we've created a VP8 frame projection, but when it's time to rewrite the packet we can't find a frame projection. Can you think of a better way to deal with this sort of handling?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
```diff
@@ -0,0 +1,468 @@
...
+     */
+    private int nextExtendedPictureId()
+    {
+        return (extendedPictureId + 1)
+            & DePacketizer.VP8PayloadDescriptor.EXTENDED_PICTURE_ID_MASK;
+    }
+
+    /**
+     * Rewrites an RTP packet. If this is the first packet of a frame and if a
+     * packet cache has been specified, we piggy-backing any missed packets.
```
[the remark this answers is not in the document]
gpolitis: In other words, once a frame has has been fully projected (<=> all the packets of a frame have been projected) it's removed from the map to prevent the map from growing too big. I'll add that in the comment and try to clarify this part a bit.

This is actually implemented in a very poor way and not as I was initially planning (if you look at the implementation of VP8FrameProjection#isFullyProjected() method).

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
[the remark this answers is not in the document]
bbaldino: oops, totally missed that this was an interface.  nevermind!

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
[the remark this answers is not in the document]
bbaldino: sounds like a place where an exception might be appropriate?  i think it would just be caught and the error logged by the layer above.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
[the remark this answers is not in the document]
bbaldino: what is the scenario where we have to piggyback packets?

doc/rtcp-termination.md
```diff
@@ -1,221 +1,94 @@
...
+particular receiver) and it distributes it [[BWD]] among the several video
+tracks that the several senders of a group-call are sharing. This results in a
+bitrate allocation for each track of each sender for each receiver in a call.
+Each video track is _projected_ [[RFC7667], Section-3.7] differently onto each
+receiver, depending on its bitrate allocation and its adaptivity features such
+as simulcast [[SIMULCAST]], scalable video coding [[SVC]] and last-N [[LASTN]].
+
+While RTCP termination and bandwidth distribution are essential group
+conferencing features they are not desirable and they may degrade the quality of
+service in one-on-one calls. Unfortunately, there is no way to them at the
```
bbaldino: think you're missing a word here:
`...there is no way to them at the...

doc/rtcp-termination.md
```diff
@@ -1,221 +1,94 @@
...
+In the next few paragraphs, we'll describe how RTCP termination works exactly
+and which components are responsible for which task. For an overview of the
+media transport aspects of the WebRTC framework and RTP usage
+[[I-D.ietf-rtcweb-rtp-usage]] is highly recommended.
+
+## RTCP for bandwidth estimations
+
+RTCP Sender Reports (SRs) [[RFC3550], Section 6.4.1] are used by media senders
+to compute RTT and by media receivers for A/V syncing and for computing packet
+loss. In its media sender role the JVB projects SSRCs for bandwidth adaptivity,
```
bbaldino: what does `projects` mean in this context?

doc/rtcp-termination.md
```diff
@@ -1,221 +1,94 @@
...
+it needs to produce its own SRs. RTCP SR generation takes place in the various
+`AdaptiveTrackProjectionContext` implementations (see, for example,
+`BasicAdaptiveTrackProjectionContext#rewriteRtcp`).
+
+For estimating the available up-link bandwidth towards a particular receiver
+(or, symmetrically, the available down-link bandwidth of a particular receiver)
```
bbaldino: i'm glad you also phrase these as from the receiver's perspective.  to me, the link which involves packets flowing from the bridge to the receiver is best described as 'the receiver's downlink'--i think calling it the bridge's uplink, though correct, is confusing.  my vote would be to exclusively use the 'receiver's downlink' terminology and not use the other one at all, but just my opinion...i'm thankful you use both at least!

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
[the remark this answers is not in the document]
gpolitis: What about https://github.com/jitsi/jitsi-videobridge/pull/735/commits/7b387bcd611e81f1b98622bf959491bb16bcd25d?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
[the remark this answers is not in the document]
gpolitis: The decision whether to accept/drop a frame takes place when the first packet of a frame is received. 

This is to keep things simple (i.e. make it easy to compute the starting sequence number of the projection of an accepted frame).

This, however, means that we may miss ignore some re-ordered packets of a frame that we end up accepting. In that case, we dive into the ingress cache, fetch the ignored packet and piggy-back it to the packet that we currently process.

Does that make sense? Also have a look in `VP8AdaptiveTrackProjectionContext#createVP8FrameProjection`.

doc/rtcp-termination.md
gpolitis: I added text and references to cover this, let me know if it makes more sense now or if it needs to be further illuminated.

doc/rtcp-termination.md
gpolitis: I have made the receivers downlink terminology the primary terminology and kept the bridge uplink terminology in the parenthesis. We can remove it if it feels meh

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -327,7 +327,8 @@ boolean decodes(@NotNull VP8Frame vp8Frame)
 
             // XXX this actually makes it possible to miss a reference frame :(
             // It would require significant effort to prevent such a case.
-            return isNextTL0(vp8Frame) || matchesTL0PICIDX(vp8Frame);
+            return isNextTemporalBaseLayerFrame(vp8Frame)
+                || matchesTL0PICIDX(vp8Frame);
```
bbaldino: maybe this one should be `isCurrentTemporalBaseLayerFrame` to match?  then it will read very clearly:

`return isNextTemporalBaseLayerFrame(vp8Frame) || isCurrentTemporalBaseLayerFrame(vp8Frame)`

which would be awesome

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -242,62 +235,66 @@ private static AdaptiveTrackProjectionContext makeContext(
...
-            if (source != null)
+            MediaStreamImpl stream
+                = source.getMediaStreamTrackReceiver().getStream();
+            if (stream != null)
             {
-                MediaStreamImpl stream
-                    = source.getMediaStreamTrackReceiver().getStream();
-                if (stream != null)
+                CachingTransformer cachingTransformer
+                    = stream.getCachingTransformer();
```
bbaldino: some of these newlines look like they wouldn't go past 80 characters if they were on the same line?  or maybe it's just hard to tell on github

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -242,62 +235,66 @@ private static AdaptiveTrackProjectionContext makeContext(
...
+                    logger.warn("incoming packet cache is null.");
                 }
             }
+            else
+            {
+                logger.warn("stream is null.");
+            }
+        }
+
+        return contextCopy.rewriteRtp(rtpPacket, incomingRawPacketCache);
```
bbaldino: is there a point in this being called if the stream or packet cache were null?

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -0,0 +1,307 @@
...
+     * @param rtpPacket the RTP packet to determine whether to accept or not.
+     * @return true if the packet is accepted, false otherwise.
+     */
+    public boolean accept(@NotNull RawPacket rtpPacket)
+    {
+        AdaptiveTrackProjectionContext
+            contextCopy = getContext(RawPacket.getPayloadType(rtpPacket));
+
+        int targetIndexCopy = targetIndex;
+        boolean accept = contextCopy.accept(rtpPacket, targetIndexCopy);
```
bbaldino: could we check if `targetIndexCopy == SUSPENDED_INDEX` and, if so, return false immediately?

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -0,0 +1,307 @@
...
+
+        return accept;
+    }
+
+    /**
+     * Gets or creates the adaptive track projection context that corresponds to
+     * the payload type that is specified as a parameter. If the payload type
+     * is different from {@link #contextPayloadType}, then a new adaptive track
+     * projection context is created that is appropriate for the new payload
+     * type.
```
bbaldino: so this will only track a context for one payload type at a time?  does it make sense to have this change/reset between payload types dynamically? for example, should we enforce only a single payload type per `AdaptiveTrackProjection`? Or have `AdaptiveTrackProjection` manage a context per payload type?

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -0,0 +1,307 @@
...
+     * @param track the ssrc of the track projection.
+     * @param format the media format.
+     * @return an adaptive track projection context that corresponds to the
+     * media format that is specified as a parameter.
+     */
+    private static AdaptiveTrackProjectionContext makeContext(
+        @NotNull MediaStreamTrackDesc track, @NotNull MediaFormat format)
+    {
+        if (Constants.VP8.equalsIgnoreCase(format.getEncoding())
+            && track.getRTPEncodings().length > 1)
```
bbaldino: can use `!track.getRTPEncodings().isEmpty())`

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -0,0 +1,307 @@
...
+        {
+            long ssrc = track.getRTPEncodings()[0].getPrimarySSRC();
+            return new VP8AdaptiveTrackProjectionContext(ssrc);
+        }
+        else
+        {
+            return new BasicAdaptiveTrackProjectionContext(format);
```
bbaldino: seeing this code, is `GenericAdaptiveTrackProjectionContext` maybe a better name for this?

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -0,0 +1,307 @@
...
+    /**
+     * Rewrites an RTP packet and it returns any additional RTP packets that
+     * need to be piggy-backed.
+     *
+     * @param rtpPacket the RTP packet to rewrite.
+     * @return any piggy-backed packets to include with the packet.
+     */
+    RawPacket[] rewriteRtp(@NotNull RawPacket rtpPacket)
+        throws
+        RewriteException
```
bbaldino: nit: this seems like odd formatting

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
```diff
@@ -0,0 +1,84 @@
...
+     * is to make the resulting rtp stream continuous.
+     *
+     * @param rtpPacket the RTP packet to rewrite.
+     * @param incomingRawPacketCache the packet cache to pull piggy-backed
+     * packets from.
+     * @return any RTP packets to piggy-back, or {@link #EMPTY_PACKET_ARR}.
+     * @throws RewriteException the underlying code has failed to rewrite the
+     * RTP packet that is specified as an argument.
+     */
+    RawPacket[]
```
bbaldino: nit: formatting. i don't think we use this style? (new line after return type)

src/main/java/org/jitsi/videobridge/cc/BasicAdaptiveTrackProjectionContext.java
```diff
@@ -0,0 +1,204 @@
...
+            rtpPacket.setSequenceNumber(destinationSequenceNumber);
+        }
+
+        synchronized (transmittedSyncRoot)
+        {
+            transmittedBytes += rtpPacket.getLength();
+            transmittedPackets++;
+        }
+
+        return EMPTY_PACKET_ARR;
```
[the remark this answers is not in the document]
bbaldino: so, we only ever return `EMPTY_PACKET_ARR` from this?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
```diff
@@ -0,0 +1,461 @@
...
+            return AdaptiveTrackProjection.EMPTY_PACKET_ARR;
+        }
+
+        // We piggy-back any re-ordered packets of this frame.
+        long vp8FrameSSRC = vp8Frame.getSSRCAsLong()
+            , vp8FrameTimestamp = vp8Frame.getTimestamp();
+
+        List<RawPacket> piggyBackedPackets = new ArrayList<>();
+        boolean done = cache == null;
+        while (!done)
```
bbaldino: so, this is basically finding every sequence number after the packet we're processing (which we've detected as being the first packet of the frame) which is present in the cache, right?  at least, until we hit one of a different frame (which we check by comparing the timestamps)?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
```diff
@@ -0,0 +1,461 @@
...
+            if (lastPacket == null
+                || vp8FrameTimestamp != lastPacket.getTimestamp())
+            {
+                done = true;
+            }
+            else
+            {
+                // the call to accept (synchronized) may update the
+                // maxSequenceNumber.
+                if (accept(lastPacket))
```
bbaldino: is it weird to have this method call accept on its own?  i assume this accept is normally called by a layer above?  should that be done by a layer above this after all the piggy-backed packets are returned?  we don't seem to access `maxSequenceNumber` here, so wondering why we need to call this to update it

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
```diff
@@ -0,0 +1,461 @@
...
+    /**
+     * @return true if we have projected all the packets from starting sequence
+     * number to max sequence number.
+     */
+    boolean isFullyProjected(long nowMs)
+    {
+        // Assume a frame is fully projected after WAIT_MS. A more predictable
+        // approach would be to check if all the packets have been projected,
+        // i.e. keep a bitmap of received packets of a frame.
+        return nowMs - createdMs > WAIT_MS;
```
bbaldino: i'm surprised to find this method return a value based on time, but maybe it's because i don't understand the concept of 'projection' fully.  but your comment does make sense to me (that it would be based on having all the packets of a frame) but i could also see some concept of a 'timeout' being useful (as i see this is used for cleaning out the map)

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -0,0 +1,376 @@
...
+        // We have successfully projected the incoming frame and we've allocated
+        // a starting sequence number for it. Any previous frames can no longer
+        // grow.
+        vp8FrameProjectionMap.put(rtpPacket.getTimestamp(), nextVP8FrameProjection);
+        // The frame attached to the "last" projection is no longer the "last".
+        lastVP8FrameProjection = nextVP8FrameProjection;
+
+        // Cleanup the frame projection map.
+        vp8FrameProjectionMap.entrySet().removeIf(
+            e -> e.getValue().isFullyProjected(nowMs));
```
bbaldino: is this only useful for cleaning the map?  if so and we want this based on time, maybe it should be called `isExpired`? which could imply it either 'timed out' or was completed successfully.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -368,9 +368,9 @@ public PacketTransformer getRTCPTransformer()
      *
      * @return the current padding parameters list for {@link #dest}.
      */
-    List<SimulcastController> getSimulcastControllers()
+    List<AdaptiveTrackProjection> getAdaptiveTrackProjections()
     {
-        return simulcastControllers;
+        return adaptiveTrackProjections;
```
bbaldino: we return the reference here but also modify it locally, so we could run into concurrent modification exception it seems?  maybe return a copy or an unmodifiable version so we can catch callers of this who try to change it?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
gpolitis: isCurrentTemporalBaseLayerFrame sounds a little bit misleading, can I suggest dependsOnSameTemporalBaseLayerFrame? My suggestion is too long, so I'm not too happy about it either.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
gpolitis: Surprisingly they do exceed the 80 chars limit, but I've reformatted them in an effort to make them more readable.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
gpolitis: We want to let the context know that the stream has been suspended so that it can raise the needsKeyframe flag and also allow it to compute a sequence number delta when the target becomes > -1.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
gpolitis: The more general scheme and more correct approach would be to have the `AdaptiveTrackProjection` manage a context per payload type.. I wanted to implement the more general approach but we have no use case for that right now, and I don't have an easy way to test this dynamic payload type changes. So practically what happens is that we create the context once, and it remains the same throughout the life of the `AdaptiveTrackProjection` instance. It's not really possible to simplify and make the context final (=> create it during construction) because we need a payload type (could be VP9, could be H264, could be VP8) so it has to be created on packet receipt.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
gpolitis: With non-adaptive tracks we have a single encoding/quality (the array is non-empty). Here we want to distinguish VP8 simulcast+svc (the array has 9 encodings/qualities) vs anything else.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: Yeah, nicely put :)

gpolitis: I included your description in the code, so future generations won't have to interpret the code in their heads, but rather be illuminated  by your words :)

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: This happens in the transform thread, so by the time we get to rewrite the definitively accepted first packet of a frame, the first packet of another frame may have already been accepted, which means there's no longer space to piggyback anything.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: The idea is that after we've fully projected all the packets of a frame, we no longer need the frame projection instance and it can be expired/discarded to prevent the map that stores these projections from growing too big. Any re-transmissions of that frame can be safely dropped and, basically, the frame projection instance becomes completely useless.

To implement this correctly we need 1) a check that determines if all the packets of the projected frame have been projected and 2nd) an upper bound of the time we're willing to wait for the packets to arrive and be projected (because the sender may decide to never send some packets).

Due to lack of time in this method I've only implemented the 2nd, easier part. Does that make more sense?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
bbaldino: I'm fine with `dependsOnSameTemporalBaseLayerFrame`, I'd rather have it be long and descriptive than somewhat cryptic, so that's fine with me.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
bbaldino: i see, so this is how a context 'learns' it has been suspended?  would it be easier to notify it only when targetIndex changes, rather than for every packet?

gpolitis: Sure, so we could, for example, introduce an `AdaptiveTrackProjectionContext.targetIndex` field in the context classes. This would effectively make the `AdaptiveTrackProjection.targetIndex` field redundant and it would have to be removed.

The only annoyance then would be that the track projection owns the `idealIndex` and the projection context owns the `targetIndex`. I can implement this change, but I also kind of like how we're storing the two indices in the same place now.

Unless you have something else in mind?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: Calling the method `isExpired` would describe what we want to do when the frame is fully projected and calling the method `isFullyProjected` would describe what the method checks. I can rename the method if you still think it's better (in light of this short explanation).

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
bbaldino: Ok, I see what you mean.  I think it's fine as it is.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -0,0 +1,291 @@
...
+     */
+    public boolean accept(@NotNull RawPacket rtpPacket)
+    {
+        AdaptiveTrackProjectionContext
+            contextCopy = getContext(RawPacket.getPayloadType(rtpPacket));
+
+        int targetIndexCopy = targetIndex;
+        boolean accept = contextCopy.accept(rtpPacket, targetIndexCopy);
+
+        if (contextCopy.needsKeyframe() && targetIndexCopy > -1)
```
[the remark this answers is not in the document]
bbaldino: nudge @gpolitis

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
bbaldino: ah, and I see now that the generic context doesn't use the cache.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param incomingRawPacketCache the packet cache to pull piggy-backed
+     * packets from.
+     * @return any RTP packets to piggy-back, or {@link #DROP_PACKET_ARR} if the
+     * packet needs to be dropped.
+     */
+    RawPacket[]
+    rewriteRtp(RawPacket rtpPacket, RawPacketCache incomingRawPacketCache);
+
+    /**
+     * Rewrites the RTCP packet that is specified as an argument.
```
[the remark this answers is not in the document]
bbaldino: (i believe this was in reference to more detail about what 'rewrite' entails)

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: I asked because I don't see `isFullyProjected` used anywhere but here (to clean the map), maybe I'm missing something in the diff?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -0,0 +1,368 @@
...
+    /**
+     * A boolean that indicates whether the incoming VP8 frame that this
+     * instance refers to is a keyframe (RFC7741).
+     */
+    private final boolean isKeyframe;
+
+    /**
+     * A boolean that indicates whether the incoming VP8 frame that this
+     * instance refers to is a reference frame (RFC7741).
+     */
```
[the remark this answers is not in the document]
bbaldino: nude @gpolitis

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -0,0 +1,368 @@
...
+    private final boolean isReference;
+
+    /**
+     * Becomes true if we have skipped a TL0 frame. In such an event, we need a
+     * keyframe to become un-stuck.
+     */
+    private boolean needsKeyframe = false;
```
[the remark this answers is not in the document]
bbaldino: nude @gpolitis

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -0,0 +1,368 @@
...
+     * The max sequence number that will be forwarded for this frame. Note that
+     * this is not necessarily the ending sequence number.
+     */
+    private int maxSequenceNumber = -1;
+
+    /**
+     * Ctor.
+     *
+     * @param firstPacketOfFrame The first VP8 packet of the frame that the
+     * packet refers to.
```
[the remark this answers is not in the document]
bbaldino: nude @gpolitis

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /**
+     * Checks whether the specified RTP packet is part of this frame.
+     *
+     * @param pkt the RTP packet to check whether it's part of this frame.
+     * @return true if the specified RTP packet is part of this frame, false
+     * otherwise.
+     */
+    boolean matchesFrame(@NotNull RawPacket pkt)
```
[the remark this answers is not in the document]
bbaldino: we can leave this as is

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -306,7 +306,7 @@ public Conference createConference(
         // The method Videobridge.getConferenceCountString() should better
         // be executed outside synchronized blocks in order to reduce the
         // risks of causing deadlocks.
-        if (logger.isInfoEnabled())
+        if (logger.isInfoEnabled() && enableLogging)
```
bbaldino: are we sure we don't want to keep these for health checks?  since the health check confs already disable logging, i worry there won't be much evidence in the logs that they're happening at all, which may be useful?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -58,6 +58,13 @@
     private final Map<Long, VP8FrameProjection>
         vp8FrameProjectionMap = new ConcurrentHashMap<>();
 
+    /**
+     * A map that stores the maximum sequence number of frames that are not
+     * (yet) accepted/projected.
```
bbaldino: can you add something about the key, value of the map? i.e. `Maps ssrc to sequence number` or whatever it is.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -228,6 +238,47 @@ VP8FrameProjection createVP8FrameProjection(
         return nextVP8FrameProjection;
     }
 
+    private int getMaxSequenceNumberOfFrame(@NotNull RawPacket rtpPacket)
```
bbaldino: so this is:
```
given an RTP packet, lookup the frame to which it belongs and then find the highest sequence number we've received from that frame
```
?
adding that as a comment would be helpful i think.  also, i'm in favor of passing only what is needed (just the ssrc and timestamp of the packet in this case) as it reduces the 'footprint' of what this method really needs, but don't have to change that.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -228,6 +238,47 @@ VP8FrameProjection createVP8FrameProjection(
...
+            .computeIfAbsent(ssrc, k -> new LRUCache<>(5));
+
+        long timestamp = rtpPacket.getTimestamp();
+        int newMaxSequenceNumber = rtpPacket.getSequenceNumber();
+        if (frameToMaxSequenceNumberMap.containsKey(timestamp))
+        {
+            int previousMaxSequenceNumber
+                = frameToMaxSequenceNumberMap.get(timestamp);
+
+            if (RTPUtils.isOlderSequenceNumberThan(
```
bbaldino: does this method work correctly if one of the given sequence numbers is -1?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -93,6 +93,13 @@
      */
     private boolean endingSequenceNumberIsKnown = false;
 
+    /**
+     * The max sequence number that was seen before the arrival of the first
+     * packet of this frame. This is useful for piggybacking any mis-ordered
+     * packets.
+     */
+    private final int maxSequenceNumberSeenBeforeFirstPacket;
```
bbaldino: so this is `the highest sequence number we've seen from a packet from this stream before seeing a packet that belonged to this frame`?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -192,23 +200,25 @@ VP8FrameProjection createVP8FrameProjection(
...
         // we're interested in the quality of this frame.
-        if (!vp8QualityFilter.acceptFrame(rtpPacket, targetIndex))
+        if (!vp8QualityFilter.acceptFrame(rtpPacket, targetIndex, nowMs))
         {
             return null;
         }
 
-        long nowMs = System.currentTimeMillis();
-
         // We know we want to forward this frame, but we need to make sure it's
```
bbaldino: does `this frame` here refer to lastVP8FrameProject? or nextVP8FrameProject?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
```diff
@@ -283,50 +290,50 @@ private int nextExtendedPictureId()
 
         rewriteRtpInternal(rtpPacket);
 
-        if (originalSequenceNumber != vp8Frame.getStartingSequenceNumber())
+        int piggyBackUntilSequenceNumber
+            = vp8Frame.getMaxSequenceNumberSeenBeforeFirstPacket();
+
+        if (piggyBackUntilSequenceNumber < 0
```
bbaldino: i'm a little confused by what's happening here.  we're making a new frame projection from the current one (what are the triggers for doing that, btw?) and we're signaling piggy-backed packets as part of creating that new frame projection?

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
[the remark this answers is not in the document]
gpolitis: No because we may be accepting packets of low-quality, while we wait for high. I'll add a comment.

src/main/java/org/jitsi/videobridge/cc/BasicAdaptiveTrackProjectionContext.java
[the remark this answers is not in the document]
gpolitis: yeah

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -73,7 +73,7 @@
  * This hierarchy allows for fine-grained filtering up to the {@link FrameDesc}
  * level.
  *
- * The decision of whether to accept or drop a specific RTP packet of a
+ * The decision of whether to project or drop a specific RTP packet of a
```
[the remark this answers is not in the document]
gpolitis: A projection is a "function" of taking an element of one space and somehow putting it on another space. This is what we're doing here, we're taking a track of the ingress and we're projecting it to the egress, depending on the encodings of the input track and how these encodings are implemented (which depends on the codec that is used).

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1297,39 +1305,37 @@ public void close()
...
+                if (ArrayUtils.isNullOrEmpty(ret))
+                {
+                    continue;
+                }
 
-                    if (extrasLen > 0)
-                    {
-                        System.arraycopy(extras, 0, newExtras, 0, extrasLen);
-                    }
+                int extrasLen
```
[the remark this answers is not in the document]
gpolitis: The VP8 adaptive track projection context is returning non-empty array for piggybacking any re-ordered packets of frames that we decide to project.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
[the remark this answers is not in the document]
gpolitis: It's the frame that knows whether a next frame is decodable and it also knows that it's stuck if a TL0 is missed for example. I can move it around if you wish.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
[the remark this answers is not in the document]
gpolitis: it is definitively the first packet of a frame

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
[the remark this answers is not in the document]
gpolitis: it is referenced by other frames

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
gpolitis: I missed that actually!

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: No you're not missing anything, the method is only used here. To me it's clearer the way it is because it makes it obvious that we cleanup the map because the frame is fully projected (i.e. I don't have to check the implementation). I don't really have a strong preference about the name and I'll rename it if you think it makes more sense.

src/main/java/org/jitsi/videobridge/Videobridge.java
gpolitis: Maybe you're right, I'll remove this.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: I don't know, but that (one of the given sequence numbers being -1 in this particular spot) shouldn't happen, right?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
gpolitis: No, this is the highest sequence number of _this_ frame that was seen before the arrival of its first packet. For example, if a frame comprised of packets 1,2,3 is received as 2,1,3 then this field would be equal to 2.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: This frame can only refer to nextVP8FrameProject because the lastVP8FrameProjection is already (in the process of being) forwarded.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: > i'm a little confused by what's happening here. we're making a new frame projection from the current one (what are the triggers for doing that, btw?) and we're signaling piggy-backed packets as part of creating that new frame projection?

A new frame has been accepted (hence, now it has to be projected). But right at this place we're not making a new frame projection, that's done in the makeNext method. Here we're merely projecting the RTP packets of the frame that refers to this frame projection.

Additionally, if this is the first packet of a frame, we piggyback any re-ordered packets that we've ignored.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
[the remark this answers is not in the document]
bbaldino: using `extras`? i see it defined on line 1299, but don't see where it gets modified?  but maybe the diff view is hiding something

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: ok, so this is saying "since we are in `createVP8FrameProjection`, it's a given we'll want to forward the frame (projection) we're creating here, but we need to make sure it will produce a decodable stream (which is basically saying, we need this frame to be 'complete', I guess?)  so we'll need to check that creating a projection from the last frame returns a valid result" ?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: ok, i don't feel too strongly so can leave it as it is.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: ah, well i noticed that `getMaxSequenceNumberOfFrame` returns that by default, but i notice now that this doesn't use that helper.  Why not? also, it looks like other code does use it, which would insert a `-1` value by default if it wasn't found?  so we could get `-1` here?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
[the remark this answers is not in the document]
bbaldino: i can see how frames would be able to tell if they depend on one another, for example, but i was thinking something like this, needing a key frame, was more of a 'stream state' thing than a 'frame state' thing and would make more sense in the projection context?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
bbaldino: when we need to figure out how many to piggy-back, can we not just start at the first sequence number of the frame and increment up, checking each time if `vp8frame.matches(packet sequence number)`? and avoid this extra state?  this way may be simpler, i dunno, just a thought

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
```diff
@@ -0,0 +1,493 @@
...
+     * A timestamp of when this instance was created. It's used to compute how
+     * long this instance has lived. That's used to determines when it's time to
+     * forget about this frame projection (see {@link #isFullyProjected(long)}).
+     */
+    private final long createdMs;
+
+    /**
+     * The projected {@link VP8Frame}.
+     */
+    private final VP8Frame vp8Frame;
```
bbaldino: this is the 'currently' projected frame i guess, yeah? or currently the frame in progress of being projected?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
bbaldino: > Due to lack of time in this method I've only implemented the 2nd, easier part

Lack of time?  We _do_ have time here, don't we?  Did you mean a lack of packet knowledge?

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
bbaldino: gotcha, so the thinking is that: although we technically change this based on payload types, this is only because we need a payload type and don't get it until runtime and, practically, we only expect to get one per session so we don't really expect it to change. is that right?

(also, for questions like these, if you don't mind, let me resolve them--I ended up missing a bunch of these because they were collapsed in github and didn't see the answer because they were marked as resolved)

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
bbaldino: ah sorry!  i misread the `>1`, you're right. my bad.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
bbaldino: > A new frame has been accepted (hence, now it has to be projected).

this is because only packets which have been accepted (i.e. belong to frames which have been accepted) are passed to `rewriteRtp`?  What preconditions are true about a packet before it is passed to `rewriteRtp`?

> Here we're merely projecting the RTP packets of the frame that refers to this frame projection.
> 
> Additionally, if this is the first packet of a frame, we piggyback any re-ordered packets that we've ignored.

Ok, so the piggy-backing is used just in the case where we received the first packet of a frame and had already received some other packets so we return all of them at once.  If we later receive another old packet, we still return it I guess?  I'm thinking of a scenario where, say, a frame has 5 packets (seq num 1, 2, 3, 4, 5)

We receive packet 3 -> save it
We receive packet 4 -> save it
We receive packet 1 -> return packets 1, 3 and 4
We receive packet 2 -> we return it I assume?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: Yeah

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: > What preconditions are true about a packet before it is passed to rewriteRtp?

There exists a frame projection (created in accept) that is able to project the packet; and the answer is yes to all the other questions.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: > it's a given we'll want to forward the frame (projection) we're creating here
we're at least interested to forward it, it has passed for example the quality test.

As you say it is possible that the `createVP8FrameProjection` returns null, in which case we don't forward the frame. A reason for the createVP8FrameProjection to return null is that it believes the new frame won't be decodable at the receiver.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
gpolitis: That's almost how it worked in one of your previous reviews but for some reason some queue in the bridge started to fill up and drop packets..

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: I'm really not sure I follow here :) Other code uses the return value of -1 in different ways, but I don't see the problem with that?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: Actually, it's the frame that this projection is projecting. I'm not sure what you mean by "currently" projected frame, I don't think there's such notion. Maybe you mean the "last" projected frame? In any case, this doesn't really matter at this point, this is just the frame that this projection is supposed to project.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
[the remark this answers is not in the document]
bbaldino: nudge @gpolitis

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: well, it's possible one of the sequence numbers we pass here isn't a sequence number at all, but the value `-1` which represents 'we don't have/know the sequence number'.  `RTPUtils.isOlderSequenceNumberThan`, from what I can tell, doesn't support the case where one of the sequence numbers it is given is invalid.  So, for example, if `previousMaxSequenceNumber` was never set (and is therefore -1) we may get inconsistent/incorrect results when trying to figure out if `sequenceNumber` is newer than `-1`

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -0,0 +1,458 @@
...
+     * @param ssrc the SSRC of the frame
+     * @param timestamp the timestamp of the frame
+     * @param sequenceNumber the sequence number of the RTP packet,
+     * potentially the highest sequence number that we've received from that
+     * frame.
+     */
+    private void storeMaxSequenceNumberOfFrame(
+        long ssrc, long timestamp, int sequenceNumber)
+    {
+        Map<Long, Integer> frameToMaxSequenceNumberMap
```
bbaldino: i think it'd be good to type this as an `LRUCache` instead of a map

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: @gpolitis i think i'm up for letting the other discussions go, but i worry about the stuff here because it may be a bug or error-prone for someone to accidentally add a bug in the future.  i think we should:

1) use the getMaxSequenceNumberOfFrame helper here which you define and use elsewhere
2) only check `RTPUtils.isOlderSequenceNumberThan` if we definitely get a sequence number (and not -1)
i.e.:
```
int previousMaxSequenceNumber = getMaxSequenceNumberOfFrame(ssrc, timestamp);
if (previousMaxSequenceNumber == -1 || RTPUtils.isOlderSequenceNumberThan(previousMaxSequenceNumber, sequenceNumber) 
{
  updateMaxSequenceNumberOfFrame(ssrc, timestamp, sequenceNumber);
}
```
(`updateMaxSequenceNumberOfFrame` does not currently exist, but seems like it should if we have its getter counterpart `getMaxSequenceNumberOfFrame`?).

gpolitis: Thanks for insisting on this, there was a hidden null pointer exception in there (implicitly casting to int from Integer).

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: And I'm not sure we want to forget past values (that's what the LRU does)

gpolitis:
The force push was due to a local rebase on top of current master. No conflicts were fixed.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: oh, i was just saying that this type _is_ actually an LRUCache, no?   

```
ssrcToFrameToMaxSequenceNumberMap
            .computeIfAbsent(ssrc, k -> new LRUCache<>(5));
```

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java
gpolitis: oh, I mean I have personally have run out of bandwidth :)

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: I misunderstood your initial comment, my last commit should address this now.

PR #740, opened 2019-01-10
fix: Re-exposes ComponentImpl#getVideobridge. Fixes #727.

### LINKED ISSUE #727: Public method getVideobridge is no more available in ComponentImpl.java
## Description

In commit https://github.com/jitsi/jitsi-videobridge/commit/58f508343e3692c8d8ee0e458b5093958a741575

```
public Videobridge getVideobridge() {}
```
was removed from org/jitsi/videobridge/xmpp/ComponentImpl.java

## Current behavior
The UI of my admin web page that is used to monitor conferences and perform remote support is broken
image

## Expected Behavior

Can you please add back this public method

## Possible Solution
This is my current work-around
```
    /**
     * Returns the {@link Videobridge} instance that is managing conferences
     * for this component. Returns <tt>null</tt> if no instance is running.
     *
     * @return the videobridge instance, <tt>null</tt> when none is running.
     */
    public Videobridge getVideobridge() // BAO
    {
        BundleContext bundleContext = common.getBundleContext();
        return bundleContext != null ? ServiceUtils2.getService(bundleContext, Videobridge.class) : null;
    }
```
I have to manually replace the class in the jar file with my patch.

bgrozev:


src/main/java/org/jitsi/videobridge/xmpp/ComponentImpl.java
```diff
@@ -477,4 +478,15 @@ public void stop(BundleContext bundleContext)
...
         }
     }
+
+    /**
+     * Returns the {@link Videobridge} instance that is managing conferences
+     * for this component. Returns <tt>null</tt> if no instance is running.
+     *
+     * @return the videobridge instance, <tt>null</tt> when none is running.
+     */
+    public Videobridge getVideobridge()
```
paweldomas: I haven't checked if the methods are ordered alphabetically in this class, but if they are then it would be good to put this one in the order as well.

bgrozev: They are indeed. Fixed. Please squash :)

PR #741, opened 2019-01-10
DO NOT MERGE use StripedExecutorService

paweldomas:
Uses StripedExecutorService to fix the order in which the packets are
processed by the XMPP component. Packets related to one conference must
be processed in the order they were received. That is achieved by
producing a stripe for a conference ID extracted from a Colibri packet.
See docs of the ComponentImpl.getStripeForPacket method for more info.
See also: https://www.javaspecialists.eu/archive/Issue206.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
     <dependency>
       <groupId>org.igniterealtime</groupId>
       <artifactId>tinder</artifactId>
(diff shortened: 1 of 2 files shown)


paweldomas:
Depends on https://github.com/jitsi/jicoco/pull/40

PR #745, opened 2019-01-22
Verify dtls certs

bgrozev:


src/main/java/org/jitsi/videobridge/IceDtlsTransportManager.java
```diff
@@ -500,13 +507,13 @@ private void installIncomingPacketReader(MultiplexingDatagramSocket s) {
...
         iceConnected = true;
         if (iceConnectedProcessed) {
-            logger.info("TransportManager " + id + " already processed ice connected, ignoring new event");
             return;
         }
         // The sctp connection start is triggered by this, but otherwise i don't think we need it (the other channel
         // types no longer do anything needed in there)
         logger.info("BRIAN: iceConnected for transport manager " + id);
         transportConnectedSubscribers.forEach(Runnable::run);
+        //bbb todo look at subscribers
```
bbaldino: what's this?

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -873,8 +838,6 @@ public ColibriShim getColibriShim()
                 }
                 if (endpointId == null)
                 {
-                    //TODO: is it reasonable to enforce this?
-                    // If we're creating a channel, we need to know which endpoint it belongs
```
bbaldino: so do you think this will be alright?  when i wrote the shims i wasn't sure if this would be acceptable (basically that we have to know the endpoint id for a channel)

bgrozev: Yeah, I think that's reasonable for the new bridge, and the error explains what's going on.

PR #747, opened 2019-01-24
Replace media format, remove RAW UDP

bgrozev:


src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -291,6 +291,7 @@ private static AdaptiveTrackProjectionContext makeContext(
      *
      * @param rtpPacket the RTP packet to rewrite.
      * @return any piggy-backed packets to include with the packet.
+     * XXX unused?
```
bbaldino: this does get used

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -338,6 +339,7 @@ private static AdaptiveTrackProjectionContext makeContext(
      *
      * @param rtcpPacket the RTCP packet to rewrite.
      * @return true to let the RTCP packet through, false to drop.
+     * XXX unused???
      */
```
bbaldino: this is not currently used, but i suppose it should (maybe with some tweaks)

bbaldino:
added some tweaks on top of this and pushed to nlj-jvb

PR #752, opened 2019-02-05
Implements adaptive track context switching with some other fixes.

gpolitis:
In theory the media engine can switch from sending simulcast -> non-simulcast on the fly and without additional signaling (either intentionally, or unintentionally due to a bug for example). The JS code can also activate/de-activate simulcast on the fly without signaling that back to the bridge (that would definitely be due to a bug). We recently observed for some scrambling video issues with the latest bridge and suspected that a bug in the JS code was causing the media engine to switch from doing simulcast -> not simulcast without signaling that to the bridge. This PR implements adaptive track projection context switches and treats the change from simulcast -> non-simulcast stream as a codec switch. Further analysis showed that there this wasn't an issue in the client and that there's no such simulcast -> non-simulcast switch, but it's good to be able to cope with such switches on the fly, so I'm opening a PR for that functionality.

Also this PR includes fixes for a few minor issues and it also brings back logging of the COLIBRI messages received/sent by the bridge.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/RtpState.java
+++ b/src/main/java/org/jitsi/videobridge/cc/RtpState.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright @ 2019 8x8, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


src/main/java/org/jitsi/videobridge/cc/RtpState.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * The highest sent sequence number.
+     */
+    public final int maxSequenceNumber;
+
+    /**
+     * The number of transmitted bytes so far.
+     */
+    public final long transmittedBytes;
```
bbaldino: what do we use this and transmittedPackets for?

bbaldino: where?

gpolitis: It's the responsibility of the adaptive track context to update these values in the SRs, so you will find these updates in GenericAdaptiveTrackProjectionContext and VP8AdaptiveTrackProjectionContext, along with timestamp updates in order for the receiving endpoint to be able to correctly map RTP timestamps to NTP timestamps.

PR #753, opened 2019-02-08
fix: Does not expire Octo channels due to lack of transport activity.

bgrozev:


gpolitis:
some non-blocking questions.. Is it possible to somehow "leak" Octo channels? Is there a way to detect such leaks, if they exist? Are they always expired by signaling?

bgrozev:
Good point, @gpolitis . I tested and under normal circumstances the channels are indeed expired by jicofo. But if this fails it would be a leak in jitsi-videobridge. As a compromise I've set the timeout to 2h, which is long enough to avoid the bug and short enough that the leak isn't too bad. As I said when we spoke, this is temporary and the code will change when I port Octo to the new bridge.

Please squash

PR #755, opened 2019-02-11
fix: Partially handles screen-sharing simulcast.

gpolitis:
This is not the complete implementation for adding simulcast support for screensharing, but it's an improvement over what we have today and can be merged on its own. There will be a supplemental PR that completes this work.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
+++ b/src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
@@ -184,8 +184,21 @@ public class AdaptiveTrackProjection
         // suspended so that it can raise the needsKeyframe flag and also allow
         // it to compute a sequence number delta when the target becomes > -1.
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -184,8 +184,12 @@ public boolean accept(@NotNull RawPacket rtpPacket)
         // suspended so that it can raise the needsKeyframe flag and also allow
         // it to compute a sequence number delta when the target becomes > -1.
 
+        RTPEncodingDesc encoding = getSource()
+            .getMediaStreamTrackReceiver().findRTPEncodingDesc(rtpPacket);
```
bbaldino: should we handle an encoding not being found? (maybe signaling hasn't come through yet?)

gpolitis: yeah, I'll make this more graceful.

PR #757, opened 2019-02-11
Refactor shim

bgrozev:


src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -121,7 +121,7 @@ private static void check(ColibriShim.ConferenceShim conferenceShim)
             }
 
             // Create and install the transport manager
-            conferenceShim.getOrCreateChannelBundle(endpoint.getID());
+            conference.getShim().getOrCreateChannelBundle(endpoint.getID());
```
bbaldino: since we hide the shim for `getOrCreateEndpoint`, should we hide it here as well? that is, just call `conference.getOrCreateChannelBundle` here?

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
```diff
@@ -0,0 +1,196 @@
...
+{
+    public final Conference conference;
+
+    private final Map<MediaType, ContentShim> contents = new HashMap<>();
+    private final Map<String, ChannelBundleShim> channelBundles = new HashMap<>();
+
+    public ConferenceShim(Conference conference)
+    {
+        this.conference = conference;
+        System.out.println("ConferenceShim created conference " + conference.getID());
```
bbaldino: guess we should remove this now

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public Collection<ContentShim> getContents()
+    {
+        synchronized (contents)
+        {
+            return new ArrayList<>(contents.values());
+        }
+    }
+
+    public AbstractEndpoint getOrCreateEndpoint(String endpointId)
```
bbaldino: i thought the idea with these changes was to change the shims from being the primary part (e.g. responsible for creating their 'real life' counterparts conference, endpoints, etc.) to being more 'ornamental' (just used for the signaling) and therefore we'd enforce all creation happened through conference, etc. and the shim was just updated when it needed to be (or would query the real thing when it needed to).  how do you see the responsibilities for the shim vs. the real classes at this point?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -478,4 +476,45 @@ public void decrementSelectedCount()
...
+     *
+     * @param mediaType the media type of the channel.
+     *
+     * @return the channel.
+     */
+    private ChannelShim getChannelOfMediaType(MediaType mediaType)
+    {
+        return
+            channelShims.stream()
+                    .filter(Objects::nonNull)
```
bbaldino: i think this filter be after the weak reference `get`?

bbaldino: EDIT: nevermind, i see you do the null check differently there, but could just do another `Objects::nonNull` filter

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
bbaldino: sorry, i wrote this before checking out the other commit.  after seeing the rest, i think that's what you did there...am i understanding it right? things like content and channel (which don't exist in the new scheme) will be created by the shims, but other things (conference, endpoint) will be created directly via the real thing and they will create their own shims? is that right?

src/main/java/org/jitsi/videobridge/health/Health.java
bgrozev: I think `getOrCreateEndpoint` should be in `Conference` because it's "native" (deals with `Endpoint`s that are part of `Conference`), while `getOrCreateChannelBundle` deals with channel bundles that are shim-only. I'd push everything that is shim-only to the shims.

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
bgrozev: Yes, exactly

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
bgrozev: You mean two filters instead of one filter with an "&&" in the condition?

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
bgrozev: Looks removed in the second commit

src/main/java/org/jitsi/videobridge/health/Health.java
bbaldino: i considered TransportManagers the 'native' side of a channel bundle, is that not accurate?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
bbaldino: yeah but don't care either way actually

src/main/java/org/jitsi/videobridge/health/Health.java
bgrozev: I think as them as part of an `Endpoint`

bgrozev: I agree there's more work there -- actually moving the transport manager to `Endpoint`, but I'd leave that for a later PR.

PR #760, opened 2019-02-14
Add support for VP8 simulcast stream with no temporal scalability

tezet:
This commit brings back support for VP8 simulcast streams with no temporal scalability. JVB used to support it with ENABLE_SVC flag disabled. It was broken by recent simulcast re-architecture preventing any simulcast layer switching for streams with no temporal scalability.
VP8 simulcast with no temporal scalability may be produced by HW encoders with no SVC features (majority of Android devices).
Ideally there shouldn't be need for this flag at all and we should base the logic entirely on the payload.

As a side effect it was necessary to remove use of GenericAdaptiveTrackProjectionContext for VP8 stream with no temporal scalability. I think that single streams may be served by VP8AdaptiveTrackProjectionContext anyway.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
+++ b/src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
@@ -264,32 +264,17 @@ public class AdaptiveTrackProjection
 
         if (Constants.VP8.equalsIgnoreCase(format.getEncoding()))
         {
-            // Context switch between VP8 simulcast and VP8 non-simulcast (sort
-            // of pretend that they're different codecs).
-            //
-            // HACK: When simulcast is activated, we also get temporal
-            // scalability, conversely if temporal scalability is disabled
-            // then simulcast is disabled.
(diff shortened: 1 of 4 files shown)


gpolitis:
Hi, thank you for your PR.

Routing everything through the complex path in the bridge (the path that deals with svc) where we skip/drop frames and manage the sequence numbers/timestamps is not optimal for non-scalable streams. With simple simulcast or single non-scalable streams you don't have to do that. A better approach would be to augment the GenericAdaptiveTrackProjectionContext to handle simulcast for non-scalable streams. That would give us simulcast handling for h264 as well which would be nice.

I'll leave this open for a while so that people can comment and argue, but I don't feel confortable merging it.

saghul:
+1 to what @gpolitis said. Having H.264 simulcast in mind would be nice since Safari TP (and soon Chrome apparently) have it.

tezet:
Hi @gpolitis, 

Thanks for review. I agree, that would be the cleanest solution. I will look into it.

PR #761, opened 2019-02-14
Clean up of the transport managers

bgrozev:


src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -199,8 +178,8 @@ public IceUdpTransportManager(Conference conference,
         conference.appendDiagnosticInformation(diagnosticContext);
         diagnosticContext.put("transport", hashCode());
 
-        iceAgent = createIceAgent(controlling, iceStreamName, true /* rtcpmux */);
-        iceStream = iceAgent.getStream(iceStreamName);
+        iceAgent = createIceAgent(controlling, DEFAULT_ICE_STREAM_NAME);
```
bbaldino: the thing i didn't like about using the same name for all of them is it makes debugging a bit harder.  do you not think so?

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -46,7 +46,7 @@
      * The name default of the single <tt>IceStream</tt> that this
      * <tt>TransportManager</tt> will create/use.
      */
-    private static final String DEFAULT_ICE_STREAM_NAME = "stream";
+    protected static final String DEFAULT_ICE_STREAM_NAME = "stream";
```
bbaldino: it doesn't look like we use this in the subclass, do we?  we just use the stream directly

src/main/java/org/jitsi/videobridge/IceDtlsTransportManager.java
```diff
@@ -100,12 +78,16 @@ public IceDtlsTransportManager(String id, Conference conference)
...
         iceAgent.addStateChangeListener(this::iceAgentStateChange);
 
-        outgoingPacketQueue = new PacketInfoQueue(id, TaskPools.IO_POOL, this::handleOutgoingPacket);
-        logger.info("BRIAN: finished IceDtlsTransportManager ctor");
+        outgoingPacketQueue
+                = new PacketInfoQueue(
+                        "TM-outgoing-" + id,
+                        TaskPools.IO_POOL,
+                        this::handleOutgoingPacket);
+        logger.debug("Constructor finished, id=" + id);
```
bbaldino: this we can probably get rid of, but we can do it later.  not blocking.

src/main/java/org/jitsi/videobridge/IceDtlsTransportManager.java
```diff
@@ -136,9 +134,14 @@ public void setEndpoint(Endpoint endpoint)
     @Override
     public boolean isConnected()
     {
-        return iceAgent.getState().isEstablished();
+        // TODO: do we consider this TM connected when ICE completes, or when
+        // ICE and DTLS both complete?
```
bbaldino: good question, i'd have to check what all relies on this.  does seem like there's a case to be made for it waiting for DTLS?

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
```diff
@@ -474,27 +457,320 @@ private void iceStreamPairChange(PropertyChangeEvent ev)
...
+        }
+        else if (IceProcessingState.FAILED.equals(newState)
+            || (IceProcessingState.RUNNING.equals(oldState)
+                    && IceProcessingState.TERMINATED.equals(newState)))
+        {
+            onIceFailed();
+        }
+    }
+
+    protected void onIceConnected()
```
bbaldino: looks like there is logic in `IceDtlsTransportManager#onIceConnected` that could be here.  saving that for later?

src/main/java/org/jitsi/videobridge/IceDtlsTransportManager.java
bgrozev: It looks like the only thing that waits for ICE, but not DTLS is IceDtlsTM (I even removed the listeners for "ice connected" in one of the commits). So I think we should refactor and simplify further at some point.

src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java
bgrozev: The iceConnectedProcessed check? I don't mind it in IceDtls (IceUdp is getting fat).

PR #762, opened 2019-02-15
Move transport managers from Conference to Endpoint.

bgrozev:
Don't merge: I broke something while cleaning up, so there is a bug somewhere out there that should be a small fix.


bgrozev:
I hadn't actually broken anything, I was just testing in a different environment. ICE fails when I run the bridge locally. I don't know why, but it also happens on the `nlj-jvb` branch, so I think this PR is safe.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -690,18 +706,83 @@ public void decrementSelectedCount()
...
+     */
+    public IceDtlsTransportManager getTransportManager()
+    {
+        if (transportManager != null)
+        {
+            return transportManager;
+        }
+        else if (transportManagerException != null)
+        {
+            // We've already tried and failed to initialize the TM.
```
bbaldino: should we just rethrow it here? (and throw it down below)?  we could still save it, but re-throwing looks like it'd make the logic below in `setTransportInfo` easier (instead of checking for a null return and then checking if the exception member is set, we'd just catch the exception)

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
```diff
@@ -96,38 +87,32 @@ public Conference getConference()
...
+        for (AbstractEndpoint endpoint : getEndpoints())
         {
-            channelBundleIdsToDescribe.forEach(bundleId -> {
-                ChannelBundleShim channelBundle = channelBundles.get(bundleId);
-                if (channelBundle != null)
+            String endpointId = endpoint.getID();
+            if (endpointIds.contains(endpointId))
+            {
+                TransportManager transportManager
+                        = ((Endpoint)endpoint).getTransportManager();
```
bbaldino: do we need to expose the transport manager here?  can we just ask the endpoint to describe it's transport manager? i.e. `endpoint.describeTransportManager(responseBundleIQ)`

src/main/java/org/jitsi/videobridge/shim/VideobridgeShim.java
```diff
@@ -135,6 +136,9 @@
                     = new ColibriConferenceIQ.Channel();
             channelShim.describe(responseChannelIQ);
             createdOrUpdatedChannels.add(responseChannelIQ);
+
+            // TODO: do we still want to support the legacy mode of putting
+            // <transport> inside <channel>?
```
bbaldino: this implies non-rtcp mux, doesn't it?  so if anything we should error on that maybe?  otherwise we'd have to do a lot of error checking to make sure the multiple transport descriptions matched

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -99,8 +98,16 @@
      */
     private boolean expired = false;
 
-    //Public for now since the channel needs to reach in and grab it
+     /**
+     * The string used to identify this endpoint for the purposes of logging.
+     */
+    protected final String logPrefix;
```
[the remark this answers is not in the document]
bgrozev: I had the same thought but didn't go all the way. It would be good if we can also integrate the Diagnostic context stuff there.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -410,25 +421,34 @@ public long getLastActivity()
...
+                = this.transceiver.getPacketIOActivity();
 
         int maxExpireTimeSecsFromChannelShims = channelShims.stream()
                 .map(WeakReference::get)
                 .filter(Objects::nonNull)
                 .map(ChannelShim::getExpire)
                 .mapToInt(exp -> exp)
                 .max()
-                .orElse(0);
+                .orElse(60);
```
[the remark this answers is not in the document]
bgrozev: Nice catch. I think I was hoping to solve the problem of the expire thread running before ICE has finished, but I see now that it's not a solution. Will revert the change (I have it on my list to fix the expire problem)

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
bgrozev: I don't know. My thinking was that I want to avoid adding "describe" code to Endpoint. I'll take a look again.

src/main/java/org/jitsi/videobridge/shim/VideobridgeShim.java
bgrozev: It does not imply non-rtcpmux necessarily. For a while at least jicofo used to include the transport in the video channel (but not in the audio channel) and the current bridge code just routes that to the channel's channle-bundle. I'm just wondering if other colibri implementations still do something like this. I'm fine not doing it, though. Maybe I will just log a warning to make it easy to diagnose.

src/main/java/org/jitsi/videobridge/Endpoint.java
bgrozev: Yeah, I'll do that. I'm not a fan of getters that throw exceptions, but with the current usage it will be better this way

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
bgrozev: Yeah, I did what you suggested

PR #776, opened 2019-03-18
Rawpacket changes, buffer pool

bbaldino:
depends on https://github.com/jitsi/jitsi-media-transform/pull/16


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -58,7 +58,7 @@
      */
     private static final Predicate<Packet> DTLS_PREDICATE
         = packet -> {
-                int b = packet.getBuffer().get(0) & 0xFF;
+                int b = packet.getBuffer()[0] & 0xFF;
```
bgrozev: Should this take `offset` into account?

bgrozev: I'll add a fix and merge. There's a lot of things we need to fix and it would be easier to fix them incrementally

PR #787, opened 2019-03-21
Removes systemd dependency.

damencho:


damencho:
I successfully installed it on ubuntu 14.04 (need to find some package updates for init.d system and java8, but other than that it installed). 
It wasn't started after installation was the only thing I spotted, and you need /etc/init.d/jitsi-videobridge start

PR #790, opened 2019-03-26
Don't clone packets for the last handler

bbaldino:


src/main/java/org/jitsi/videobridge/octo/OctoEndpoint.java
```diff
@@ -91,6 +92,12 @@ public void onNewSsrcAssociation(
 
     }
 
+    @Override
+    public void sendRtp(PacketInfo packet, String sourceEpId)
```
bgrozev: Should we move the PotentialPacketHandler implementation to Endpoint?

bbaldino: so then we'd need to split it into 2 interfaces, right?  one for `wants`, which would be at `AbstractEndpoint` and another for `sendRtp` which `Endpoint` and `OctoTentacle` would implement?

PR #791, opened 2019-03-26
Port BitrateController

bbaldino:
I did not get as far as I wanted to with this (we still do not preserve the PacketInfo instance, meaning we lose the receivedTime/timeline), but it is a step in the right direction (we avoid packet conversion) and I think the next step will be a pretty big change.

I think we can actually change this flow a bit to work such that, rather than caching all packets, it holds only those that it may want in incomplete frames and can then release them.  So instead of copying and caching all packets of which we may only use some, we can just hold some pending packets until we discover we do (or don't) need them.  This is along the lines of the 'jitter buffer scheme' idea I had for this class before.  We'll probably need to talk with george about that change, but we can discuss it more tomorrow.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -1133,9 +1133,8 @@ public class BitrateController
         });
     }
 
-    public RawPacket[] transformRtp(@NotNull PacketInfo packetInfo)
+    public VideoRtpPacket[] transformRtp(@NotNull PacketInfo packetInfo)
     {
-        RawPacket pkt = RawPacketExtensionsKt.toLegacyRawPacket(packetInfo.getPacket());
         VideoRtpPacket videoPacket = (VideoRtpPacket)packetInfo.getPacket();
         if (firstMediaMs == -1)
         {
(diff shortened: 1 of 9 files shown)


bbaldino:
I'll add that it's tricky to move it all the way to PacketInfo as it is now as doing a cache of PacketInfo in the same spirit as `RtpPacketCache` is a bit awkward.  Also, after looking at it I realized I think we can avoid a lot of copies (every time we put something in the cache) because I don't _think_ we actually need one at all here.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1453,143 +1482,4 @@ int getIdealIndex()
...
-        RTCPTransformer()
-        {
-            super(RTCPPacketPredicate.INSTANCE);
-        }
-
-        /**
-         * {@inheritDoc}
-         */
-        @Override
-        public RawPacket transform(RawPacket pkt)
```
bgrozev: Was this unused?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1136,12 +1136,13 @@ public void addPayloadType(PayloadType payloadType)
...
     {
         RawPacket pkt = RawPacketExtensionsKt.toLegacyRawPacket(packetInfo.getPacket());
+        VideoRtpPacket videoPacket = (VideoRtpPacket)packetInfo.getPacket();
         if (firstMediaMs == -1)
         {
             firstMediaMs = System.currentTimeMillis();
         }
 
-        long ssrc = pkt.getSSRCAsLong();
+        Long ssrc = videoPacket.getSsrc();
```
bgrozev: Why capital-L Long?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bgrozev: I don't see it used anywhere in our branch, so I'll merge. We should see if it is actually needed, if it isn't there's a bunch of rtcp code in cc that we can remove

PR #811, opened 2019-04-15
fixes rtcp sr handling

gpolitis:


gpolitis:
Depends on https://github.com/jitsi/jitsi-media-transform/pull/53 and https://github.com/jitsi/jitsi-rtp/pull/21

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -380,19 +381,19 @@ private RtpState getRtpState()
     /**
      * Rewrites an RTCP packet.
      *
-     * @param rtcpPacket the RTCP packet to rewrite.
+     * @param rtcpSrPacket the RTCP SR packet to rewrite.
      * @return true to let the RTCP packet through, false to drop.
      * XXX unused???
      */
-    public boolean rewriteRtcp(@NotNull RawPacket rtcpPacket)
+    public boolean rewriteRtcp(@NotNull RtcpSrPacket rtcpSrPacket)
```
bbaldino: are RTCP SRs the only packets we'll ever send here?  if so, should we name it as such (since we're no longer bound to a generic rewriteRtp/rewriteRtcp interface)?  if not (and we'll send other rtcp through here) then i think we should leave the generic `rtcpPacket` param name and test for whether or not it's an sr in the method

gpolitis: eventually we would like to bring back the old processing for bye and sdes, so this code would have a compound rtcp packet as a parameter (there's some details about that in the trello board); but sdes/bye handling is not guarantee to make its way in this pr. i don't particularly like rewriteRcp either

src/main/java/org/jitsi/videobridge/PotentialPacketHandler.java
```diff
@@ -30,9 +30,16 @@
     boolean wants(PacketInfo packet, String sourceEpId);
 
     /**
-     * Send the given 'packet' (which came from 'source')
+     * Send the given rtp 'packet' (which came from 'source')
```
bgrozev: I wonder  if we need separate methods for RTP and RTCP here, since they both take a PacketInfo. Presumably we'd also call `wants` for RTCP to avoid unnecessary copying, so one method for both makes sense to me

bbaldino:
But we don't route rtcp to multiple destinations so we'd at least want to
stop after we found where it should go to avoid the rest of the looping. I
don't think it seems bad to have separate methods for rtp and rtcp.

src/main/java/org/jitsi/videobridge/PotentialPacketHandler.java
gpolitis: shall I rename sendRtp to sendRtpRtcp (or something else) and remove the sendRtcp method?

bbaldino: I think we should keep the separate methods. The way we forward them is pretty different and I think it's clearer to have the two handlers distinct from one another, in the new code we treat rtcp and rtp differently.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -411,6 +406,19 @@ public boolean accept(RtpPacket rtpPacket)
         return adaptiveTrackProjection.accept(rtpPacket);
     }
 
+    public boolean transformRtcp(RtcpSrPacket rtcpSrPacket)
+    {
+        long ssrc = rtcpSrPacket.getSenderSsrc();
+        if (ssrc < 0)
+        {
+            return false;
```
bgrozev: `getSenderSsrc` will now always return a positive value :)

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -411,6 +406,19 @@ public boolean accept(RtpPacket rtpPacket)
...
+        }
+
+        AdaptiveTrackProjection adaptiveTrackProjection = adaptiveTrackProjectionMap.get(ssrc);
+
+        return adaptiveTrackProjection != null && adaptiveTrackProjection.rewriteRtcp(rtcpSrPacket);
```
bgrozev: Line length

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -396,69 +396,42 @@ public boolean accept(
     /**
      * Rewrites the RTCP packet that is specified as an argument.
      *
-     * @param rtcpPacket the RTCP packet to transform.
+     * @param rtcpSrPacket the RTCP packet to transform.
```
bgrozev: Nice. Can we clean up the unused imports in this file?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -396,69 +396,42 @@ public boolean accept(
...
-        while (it.hasNext())
+        VP8FrameProjection lastVP8FrameProjectionCopy = lastVP8FrameProjection;
+        if (lastVP8FrameProjectionCopy.getVP8Frame() == null
+            || rtcpSrPacket.getSenderSsrc() != lastVP8FrameProjectionCopy.getSSRC())
         {
-            ByteArrayBuffer baf = it.next();
-            switch (RTCPUtils.getPacketType(baf))
+            return false;
+        }
+        else
```
bgrozev: Can we drop the `else` and avoid the extra indentation?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -396,69 +396,42 @@ public boolean accept(
...
+        {
+            long srcTs = rtcpSrPacket.getSenderInfo().getRtpTimestamp();
+            long delta = RTPUtils.rtpTimestampDiff(
```
bgrozev: Can we use jitsi-rtp here?
`            long delta = RtpUtils.Companion.getTimestampDiff(`

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -396,69 +396,42 @@ public boolean accept(
...
+                lastVP8FrameProjectionCopy.getTimestamp(),
+                lastVP8FrameProjectionCopy.getVP8Frame().getTimestamp());
+
+            long dstTs = RTPUtils.as32Bits(srcTs + delta);
```
bgrozev: Can we add a utility or use `&0xffff_ffff` directly? We should eventually stop using the libjitsi utilities

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -396,69 +396,42 @@ public boolean accept(
...
-        boolean removed = false;
-        RTCPIterator it = new RTCPIterator(rtcpPacket);
-        while (it.hasNext())
+        VP8FrameProjection lastVP8FrameProjectionCopy = lastVP8FrameProjection;
+        if (lastVP8FrameProjectionCopy.getVP8Frame() == null
+            || rtcpSrPacket.getSenderSsrc() != lastVP8FrameProjectionCopy.getSSRC())
         {
-            ByteArrayBuffer baf = it.next();
-            switch (RTCPUtils.getPacketType(baf))
+            return false;
```
bgrozev: I think the RTCP SRs for audio will now also travel this path (they will hit `conference.handleIncomingRtcp`, not sure if they come this deep). Are we going to drop them here? I'm okay working on this when I get to handling RTCP for audio as we discussed

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -961,6 +987,42 @@ else if (transportManagerException != null)
...
+    /**
+     * Sends a message to this {@link Endpoint} in order to notify it that the list/set of {@code lastN} has changed.
+     *
+     * @param forwardedEndpoints the collection of forwarded endpoints.
+     * @param endpointsEnteringLastN the <tt>Endpoint</tt>s which are entering
+     * the list of <tt>Endpoint</tt>s defined by <tt>lastN</tt>
+     * @param conferenceEndpoints the collection of all endpoints in the
+     * conference.
+     */
+    public void sendLastNEndpointsChangeEvent(
```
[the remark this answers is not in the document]
gpolitis: yeah

src/main/java/org/jitsi/videobridge/PotentialPacketHandler.java
gpolitis: not sure what to implement, i'm personally fine with either @bgrozev @bbaldino can you please guys reach some consensus?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1132,15 +1132,10 @@ public OctoTentacle getTentacle()
      */
     void handleIncomingRtcp(PacketInfo packetInfo, AbstractEndpoint source)
     {
+        String sourceEpId = source != null ? source.getID() : null;
```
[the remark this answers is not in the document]
gpolitis: > Is there anything different for RTCP at this level? Can we merge the logic for handleIncomingRtp and handleIncomingRtcp (perhaps rename to handleIncomingPacket)? This would simplify the code and avoid the unnecessary clone().

i'm fine with that as long as it doesn't get vetoed /cc @bbaldino

src/main/java/org/jitsi/videobridge/PotentialPacketHandler.java
bbaldino: I guess I don't see the commonality between the 2 in terms of implementation. I don't see "wants" even as applying to rtcp...rtcp isn't about wanting, it's about routing it to the single place it must go, not asking everywhere if they want it like rtp

bbaldino: Because of this, for example, that clone logic doesn't even apply to rtcp so why do we care about reusing it?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: if they reach this deep they'll probably get dropped, yeah.. :/

src/main/java/org/jitsi/videobridge/Conference.java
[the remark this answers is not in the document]
bbaldino: See my other comment, it applies here as well

src/main/java/org/jitsi/videobridge/PotentialPacketHandler.java
bgrozev: "it's about routing it to the single place it must go"

No, SRs are being forwarded to everyone and the BitrateController "wants" only the SRs for the SSRC that it is currently forwarding (out of the simulcast set)

bbaldino: Ok, but it's not true for pli/fir. It just seems like we're going to have a method that has to handle rtp and rtcp differently anyway, so why not just have the two methods/paths since they do differ.

bgrozev: Which of the 3 points above do you disagree with?

I agree that we have to handle them differently *somewhere*, but I think we should do it after we re-use the common logic, and without complicating the wants/send interface

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1115,7 +1115,11 @@ public void setMaxRxFrameHeightPx(int maxRxFrameHeightPx)
      */
     public void setSelectedEndpointIds(Set<String> selectedEndpointIds)
     {
-        this.selectedEndpointIds = new HashSet<>(selectedEndpointIds);
+        if (!this.selectedEndpointIds.equals(selectedEndpointIds))
+        {
+            this.selectedEndpointIds = new HashSet<>(selectedEndpointIds);
+            update(lastEndpointOrdering, getAvailableBandwidth());
```
bgrozev: Nice catch! Was that causing the failures to switch?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: line length is consistent with the new rule we voted a while back (120 for all projects)

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: i think so, yeah. there may be more calls missing, i'm going through all this now

src/main/java/org/jitsi/videobridge/PotentialPacketHandler.java
bbaldino: I think 1 and 3. I don't think reusing the clone logic will apply directly, pli/fir will only go to 1 place and we don't want to clone and will route totally differently. I assume we need to rewrite the srs so we do need to clone there? Or will we just rebuild a new one?

bbaldino: I think 1 and 3. I don't think reusing the clone logic will apply directly, pli/fir will only go to 1 place and we don't want to clone and will route totally differently. I assume we need to rewrite the srs so we do need to clone there? Or will we just rebuild a new one?

bbaldino: If we did combine, we'd have a when that basically treated rtp and rtcp srs one way (assuming srs are the same?), Pli/fir another way...is that right?

bgrozev: 1. By inconsisntent I mean the following: the `wants` path takes a `PacketInfo` and does the RTP/RTCP check internally. The `sendPath` also takes a `PacketInfo`, but the RTP/RTCP check happens prior. 
3. Yes, we `clone()` RTCP SRs. And while not critical it would be nice to avoid the last one (we already have code to do it!).

To clarify, here is what I'm suggesting (updated to handle the FIR/PLI case optimally)
```
interface PotentialPacketHandler{
boolean wants(PacketInfo)
void send(PacketInfo)
}

Conference.sendOut(PacketInfo, Source) {
  <<<existing logic for fanning out without an extra clone(), calling the single send() of PPH>>>
}

Conference.handleIncomingRtp(PacketInfo, Source){
  sendOut(PacketInfo, Source)
}

Conference.handleIncomingRtcp(PacketInfo, Source){
  if (PacketInfo is RTCP FB) {
     endpoint = find(PacketInfo)
     if(endpoint.wants(PacketInfo, Source) endpoint.send(PacketInfo, Source)
  }
  else {
    sendOut(PacketInfo, Source)
  }
}
```

bbaldino: I'm good with that

gpolitis: i've updated the code

src/main/java/org/jitsi/videobridge/Conference.java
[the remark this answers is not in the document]
gpolitis: this is now taken care of

PR #814, opened 2019-04-18
Adds a debug interface

bgrozev:


src/main/java/org/jitsi/videobridge/rest/DebugRequestHandler.java
```diff
@@ -0,0 +1,145 @@
...
+import javax.servlet.http.*;
+import java.io.*;
+
+import static org.jitsi.videobridge.rest.HandlerImpl.DEBUG;
+
+/**
+ * Handles requests for "/colibri/debug*".
+ *
+ * Note that using this interface MAY disrupt running conferences or even
+ * cause a deadlock. It is really meant only for debugging, which is why it is
```
bbaldino: The conference interruption I get but potentially causing a deadlock seems bad. Can it be avoided?

bgrozev: I've tried to avoid acquiring locks, or holding any objects (I just push primitives and JSON/NodeStatsBlock objects), but it traverses a big chunk of the classes down from Conference and it's hard to guarantee. I don't have a specific deadlock in mind.

bgrozev:
> Exciting!

Yeah, I wish we'd done that a long time ago! I always thought that doing it this way would be too much to maintain, but it turned out to be reasonable.

src/main/java/org/jitsi/videobridge/rest/DebugRequestHandler.java
bbaldino: Ah ok, I thought you meant you were seeing one occur.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1161,7 +1161,7 @@ public JSONObject getDebugState(boolean full, String endpointId)
     {
         JSONObject debugState = new JSONObject();
         debugState.put("id", id);
-        debugState.put("name", name.toString());
+        debugState.put("name", name == null ? null : name.toString());
```
bbaldino: Should we use an empty string or `no name` or something instead of null?

bgrozev: I don't think it makes a difference. A null value in json will make it clear that it was not set, as opposed to someone making a conference named `no name` or `null` (I'm just rationalizing my reluctance to make an update)

PR #815, opened 2019-04-23
wip: fixes various adaptivity regressions in the new bridge

gpolitis:
depends on https://github.com/jitsi/jitsi-media-transform/pull/65

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -341,16 +341,13 @@ public class Endpoint
     }
 
     /**
-     * {@inheritDoc}
+     * Notifies this {@code Endpoint} that the list of {@code Endpoint}s ordered
+     * by speech activity (i.e. the dominant speaker history) has changed.
      */
-    @Override
-    public void propertyChange(PropertyChangeEvent evt)
+    void speechActivityEndpointsChanged(List<String> endpoints)
     {
-        if (Conference.ENDPOINTS_PROPERTY_NAME.equals(evt.getPropertyName()))
-        {
-            bitrateController.endpointOrderingChanged(
-                    getConference().getSpeechActivity().getEndpoints());
-        }
+        super.speechActivityEndpointsChanged(endpoints);
+        bitrateController.endpointOrderingChanged(endpoints);
     }
 
     /**
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -390,7 +390,6 @@ public class Endpoint
     {
         super.setMaxReceiveFrameHeightPx(maxReceiveFrameHeightPx);
         bitrateController.setMaxRxFrameHeightPx(maxReceiveFrameHeightPx);
-        bitrateController.constraintsChanged();
     }
 
     /**
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -45,7 +45,6 @@ import org.jitsi_modified.sctp4j.*;
 import org.jitsi_modified.service.neomedia.rtp.*;
 import org.json.simple.*;
 
-import java.beans.*;
 import java.io.*;
 import java.nio.*;
 import java.time.*;
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -143,6 +143,11 @@ public class Endpoint
     private AtomicInteger selectedCount = new AtomicInteger(0);
 
     /**
+     * The diagnostic context of this instance.
+     */
+    private final DiagnosticContext diagnosticContext = new DiagnosticContext();
+
+    /**
      * The bitrate controller.
      */
     private final BitrateController bitrateController;
@@ -224,6 +229,7 @@ public class Endpoint
                 TaskPools.CPU_POOL,
                 TaskPools.CPU_POOL,
                 TaskPools.SCHEDULED_POOL,
+                diagnosticContext,
                 logger);
         transceiver.setIncomingRtpHandler(
                 new ConsumerNode("RTP receiver chain handler")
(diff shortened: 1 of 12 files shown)


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -390,7 +387,6 @@ public void setMaxReceiveFrameHeightPx(int maxReceiveFrameHeightPx)
     {
         super.setMaxReceiveFrameHeightPx(maxReceiveFrameHeightPx);
         bitrateController.setMaxRxFrameHeightPx(maxReceiveFrameHeightPx);
-        bitrateController.constraintsChanged();
```
bbaldino: How is it this can go away, isn't it needed?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1133,7 +1136,7 @@ public void setPinnedEndpointIds(Set<String> pinnedEndpointIds)
         if (!this.pinnedEndpointIds.equals(pinnedEndpointIds))
         {
             this.pinnedEndpointIds = new HashSet<>(pinnedEndpointIds);
-            update(lastEndpointOrdering, getAvailableBandwidth());
+            update();
```
bbaldino: I moved this away from the single blanket update function because I found it confusing less explicit. Is there an issue with the updated scheme? Why go back?

gpolitis: i found it confusing that an update() caller can set the lastendpointordering and then update() itself sets the lastendpointordering0; my idea is that update() computes a result based on the internal state, but it doesn't alter the state (other than storing the result of course)

bbaldino: Ok, give me a chance to look on non-mobile and I can probably get a better picture. I'm thinking maybe we can at least rename update to something that reflects that better.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -599,20 +587,29 @@ public void constraintsChanged()
...
      * @param bweBps the current bandwidth estimation (in bps)
      */
-    private synchronized void update(List<AbstractEndpoint> conferenceEndpoints, long bweBps)
+    private synchronized void update()
     {
-        lastEndpointOrdering = conferenceEndpoints;
-        lastBwe = bweBps;
+        long nowMs = System.currentTimeMillis();
+
+        long bweBps = getAvailableBandwidth(nowMs);
```
bbaldino: i don't see the advantage of having this method always re-get everything rather than having the other methods (`endpointOrderingChanged`, etc.)--which know exactly what changed--pass them into this method, so this method can focus just on dealing with the current state and not worrying about acquiring the current state. is there something i'm missing?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -599,20 +587,29 @@ public void constraintsChanged()
...
 
         logger.debug("TEMP: BitrateController " + hashCode() + " updating. " +
                 "bwebps is " + bweBps);
 
         // Create a copy as we may modify the list in the prioritize method.
-        conferenceEndpoints = new ArrayList<>(conferenceEndpoints);
+        List<AbstractEndpoint> sortedEndpoints = new ArrayList<>(sortedEndpointIds.size());
+        for (String endpointId : sortedEndpointIds)
+        {
+            AbstractEndpoint abstractEndpoint = destinationEndpoint.getConference().getEndpoint(endpointId);
```
bbaldino: why did we change the member to hold string endpoint ids instead of endpoints if this method is just going to get the endpoints?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -599,20 +587,29 @@ public void constraintsChanged()
...
+            {
+                sortedEndpoints.add(abstractEndpoint);
+            }
+        }
 
         // Compute the bitrate allocation.
         TrackBitrateAllocation[]
-            trackBitrateAllocations = allocate(bweBps, conferenceEndpoints);
+            trackBitrateAllocations = allocate(bweBps, sortedEndpoints);
         logger.debug("TEMP: BitrateController#update got " + trackBitrateAllocations.length +
```
bbaldino: maybe wrap this in an `isDebugEnabled` check and remove the TEMP line from it while you're in there?

src/main/java/org/jitsi/videobridge/Endpoint.java
bbaldino: ok, i see it's been moved to `setMaxRxFrameHeightPx`

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: can't holding the endpoints lead to memory leaks?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: but the `endpointOrderingChanged` doesn't know anything about bandwidth, nor `setMaxRxFrameHeightPx`. the last bandwidth estimation is part of the internal state that `update()` operates on and it decides to apply a hack to that bandwidth estimation that we may decide to remove in the future. i think the name `getAvailableBandwidth` is not representative but, conceptually, that method _is_ part of the update step

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bbaldino: so maybe weakreference is the right way to go?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bbaldino: that's true, but the thinking was that, for example `endpointOrderingChanged` gets the new endpoint ordering and knows to grab the existing/cached value for bandwidth, and `bandwidthChanged` would know to pass in the new bandwidth but use the cached endpoint ordering.  imagine calculating either of those things is costly--this method has to acquire both of them every time, since it doesn't know what changed.  having the other methods do that work ensures we only re-get the thing that actually changed.  that was the thinking, at least (as well as giving a more specific API for users to call--explicitly stating endpointOrderingChanged, etc. but we can achieve that with either of these schemes)

bbaldino: i think i feel most strongly about `update` not being public (which it isn't) so i'm ok with this, i guess i just didn't see the need to change it from how it already was, as this way doesn't seem like an improvement.

PR #821, opened 2019-04-27
Payload verification

bgrozev:


src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjectionContext.java
```diff
@@ -98,4 +99,10 @@
      * processes.
      */
     PayloadType getPayloadType();
+
+    /**
+     * Gets a JSON representation of the parts of this object's state that
+     * are deemed useful for debugging.
+     */
+    JSONObject getDebugState();
```
bbaldino: Maybe we should have an interface for this?

bgrozev: I actually had an interface, but I then I needed a slightly different signature in Videobridge and in Conference so I didn't think it was worth keeping (also the jmt code mostly uses a NodeStatsBlock so it's also different).

bgrozev:
Screenshot 2019-04-29 at 14 30 00

PR #822, opened 2019-04-30
Some changes needed to make RTCP to work with Octo

bgrozev:
Don't merge just yet.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -1124,7 +1124,7 @@ public class Conference
             RtcpFbPacket rtcpFbPacket = (RtcpFbPacket) packet;
 
             // XXX we could make this faster with a map
-            AbstractEndpoint endpoint
+            AbstractEndpoint target
                 = findEndpointByReceiveSSRC(
                     rtcpFbPacket.getMediaSourceSsrc());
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1066,45 +1034,43 @@ public ConferenceShim getShim()
      * Broadcasts the packet to all endpoints and tentacles that want it.
      *
      * @param packetInfo the packet
-     * @param source the source endpoint, or {@code null} to indicate that the
-     * packet comes from another bridge via Octo.
+     * @param source the ID of the source endpoint.
      */
-    private void sendOut(PacketInfo packetInfo, AbstractEndpoint source)
+    private void sendOut(PacketInfo packetInfo, String source)
```
bbaldino: if `source` is going to be a string, can we rename it to something less ambiguous? `sourceEpId`?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -489,16 +494,30 @@ public void send(PacketInfo packetInfo, String sourceEpId)
         else if (packet instanceof RtcpSrPacket)
         {
             RtcpSrPacket rtcpSrPacket = (RtcpSrPacket) packet;
-            if (bitrateController.transformRtcp(rtcpSrPacket))
+            AbstractEndpoint sourceEndpoint = getConference().getEndpoint(source);
```
bbaldino: so we have to look up the endpoint and the media type of the ssrc foreach packet foreach sender?

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: I specifically renamed all of them to `source` because it's clear from the type what it is, and I don't think the longer names add anything here.

src/main/java/org/jitsi/videobridge/Endpoint.java
bgrozev: Well, it's limited to SRs, but you're right that's not great. I didn't like any of the alternatives, but something you mentioned in the jmt PR gave an idea. If nothing else I'll move this logic a step back to Conference so that it executes only once per packet.

PR #825, opened 2019-05-01
Rtcp octo

bgrozev:


src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -149,7 +147,7 @@ private static void resolveSSRCAsEndpoint(
      * The <tt>DominantSpeakerIdentification</tt> instance which
      * detects/identifies the active/dominant speaker in {@link #conference}.
      */
-    private final DominantSpeakerIdentification dominantSpeakerIdentification
+    private DominantSpeakerIdentification dominantSpeakerIdentification
```
gpolitis: not opposing, just curious why?

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -198,9 +184,6 @@ public ConferenceSpeechActivity(Conference conference)
         this.conference = Objects.requireNonNull(conference, "conference");
         logger = Logger.getLogger(classLogger, conference.getLogger());
 
-         // The PropertyChangeListener will weakly reference this instance and
-         // will unregister itself from the conference sooner or later.
-        conference.addPropertyChangeListener(propertyChangeListener);
         dominantSpeakerIdentification
```
gpolitis: i think it's safe to leave as, but i'm surprised we check for null elsewhere

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -829,12 +862,13 @@ private synchronized void update()
...
             {
-                logger.debug("TEMP: adding a new track projection for ssrc" +
-                        rtpEncoding.getPrimarySSRC());
                 adaptiveTrackProjectionMap.put(
                     rtpEncoding.getPrimarySSRC(),
                     adaptiveTrackProjection);
 
+                // TODO: RTX packets should never reach here (they get de-RTXed
+                // in the receiver chain). Do we still need this mapping?
+                // Perhaps we need it in order to block SR for the RTX streams?
```
gpolitis: that's a really good point, in theory it should be safe to remove

bgrozev: Do you know whether RTX streams have SRs?

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
bgrozev: I must have missed something while rebasing. I want to set it to null in `expire`.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: not in practice at least

PR #828, opened 2019-05-13
Merge rtcp

bgrozev:


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -473,13 +473,21 @@ public void send(PacketInfo packetInfo)
...
+            if (extras == null)
+            {
+                logger.warn(
+                    "Dropping a packet which was supposed to be accepted:"
+                        + packet);
+                return;
+            }
+
+            // The original packet was transformed in place.
+            // TODO: should we send this *after* the extras?
```
gpolitis: it shouldn't make a difference

bgrozev: Doesn't that affect the order in which the packets are sent?

gpolitis: yes, and that shouldn't make a difference

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1112,13 +1097,19 @@ public OctoTentacle getTentacle()
...
-     * Handles an RTCP packet coming from a specific endpoint.
+     * Handles an RTP/RTCP packet coming from a specific endpoint.
      * @param packetInfo
      */
-    public void handleIncomingRtcp(PacketInfo packetInfo)
+    public void handleIncomingPacket(PacketInfo packetInfo)
     {
         Packet packet = packetInfo.getPacket();
-        if (packet instanceof RtcpFbPliPacket
+        if (packet instanceof RtpPacket)
```
bbaldino: why not just add conditions to this if that exclude the tests in the `else if` to avoid the duplicate `else` thing?  or add a helper so we could have something like:

```
if (packet instanceof RtpPacket && notSingleDestRtcp(packetInfo) {
   // fan out logic
} else {
  // Single-dest rtcp logic
}

File: src/main/java/org/jitsi/videobridge/Endpoint.java [hunk 3]
```diff
@@ -473,13 +473,21 @@ public void send(PacketInfo packetInfo)
...
+            transceiver.sendPacket(packetInfo);
```
bbaldino: wait, does the returned array not already include the original packet?  given that the 'extras' may contain packets _before_ this packetInfo, this feels weird to me.  the bitratecontroller is the best to determine the order in which the packets it returns should be sent (since it's aware of things being piggybacked/older packets now being forwarded) so it seems like we should trust its return value rather than sending the original in place and then the extras

bbaldino: (ok, i see you changed the behavior below about not including the original packet, but i still don't get the point of this change...does it bring us closer to how we want things to end up?)

src/main/java/org/jitsi/videobridge/util/PartitionedByteBufferPool.java
```diff
@@ -260,7 +260,7 @@ void addStats(JSONObject stats)
             byte[] buf = pool.poll();
             if (buf == null)
             {
-                buf = new byte[DEFAULT_BUFFER_SIZE];
+                buf = new byte[Math.max(DEFAULT_BUFFER_SIZE, requiredSize)];
```
bbaldino: was this actually happening?  isn't DEFAULT_BUFFER_SIZE 1500?

bgrozev: Yeah, it's 1520, I'm not sure it was happening. We are observing some strange exceptions that may be explained by this, but I haven't been able to confirm.

src/main/java/org/jitsi/videobridge/Endpoint.java
bgrozev: "does it bring us closer to how we want things to end up?" 
Yes, the PacketInfo instance is preserved, except for piggybacked packets. So preserved for the vast majority of packets.

PR #829, opened 2019-05-13
Simplify, adapt to JMT,

bgrozev:
log when the role is not set at start time

### DIFF
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -212,8 +212,8 @@ public class DtlsTransport extends IceTransport
             fingerprintPE = new DtlsFingerprintPacketExtension();
             pe.addChildExtension(fingerprintPE);
         }
-        fingerprintPE.setFingerprint(dtlsStack.getCertificateInfo().getLocalFingerprint());
-        fingerprintPE.setHash(dtlsStack.getCertificateInfo().getLocalFingerprintHashFunction());
+        fingerprintPE.setFingerprint(dtlsStack.getLocalFingerprint());
+        fingerprintPE.setHash(dtlsStack.getLocalFingerprintHashFunction());
 
         // TODO: don't we only support ACTIVE right now?
         fingerprintPE.setSetup("ACTPASS");
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -158,22 +158,23 @@ public class DtlsTransport extends IceTransport
         if (dtlsStack.getRole() == null && !fingerprintExtensions.isEmpty())
         {
             String setup = fingerprintExtensions.get(0).getSetup();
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -211,8 +212,8 @@ protected void describe(IceUdpTransportPacketExtension pe)
             fingerprintPE = new DtlsFingerprintPacketExtension();
             pe.addChildExtension(fingerprintPE);
         }
-        fingerprintPE.setFingerprint(dtlsStack.getLocalFingerprint());
-        fingerprintPE.setHash(dtlsStack.getLocalFingerprintHashFunction());
+        fingerprintPE.setFingerprint(dtlsStack.getCertificateInfo().getLocalFingerprint());
```
gpolitis: non blocking but, would it make sense to store the certificate info in a local var to avoid the admittedly  improbable case where a different certificate info is returned?

bgrozev: The cert in a DtlsStack instance never changes (it's a Kotlin 'val'), that's part of what the JMT PR did.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -211,8 +212,8 @@ protected void describe(IceUdpTransportPacketExtension pe)
...
+        fingerprintPE.setHash(dtlsStack.getCertificateInfo().getLocalFingerprintHashFunction());
```
bbaldino: why expose certificate info?

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -380,6 +381,11 @@ protected void onIceConnected()
         TaskPools.IO_POOL.submit(() -> {
             try
             {
+                if (dtlsStack.getRole() == null)
+                {
+                    logger.warn(logPrefix +
```
bbaldino: do things not throw when this happens?  i think they should, as nothing will work

bgrozev: There's explicit code to handle the case (which fails silently if the role is never set): https://github.com/jitsi/jitsi-media-transform/blob/master/src/main/kotlin/org/jitsi/nlj/dtls/DtlsStack.kt#L171

bbaldino: ah right, i think that does happen at times.  so why the warning then?

maybe we just want to add a timeout on that future and then log an error if the role isn't set within the timeout (probably better than waiting indefinitely)

bgrozev: I don't think it should normally happen: clients should provide the necessarty info together with the ICE ufrag/password, and this executes after ICE has completed hence we should know the state. I see it with jigasi, which doesn't work right now, so I prefer to see a warning if it happens.

Agreed about having a timeout, will add a card

PR #832, opened 2019-05-21
minor bug fixes, refactoring and style

gpolitis:


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -496,7 +496,18 @@ else if (packet instanceof RtcpSrPacket)
...
+            {
+                if (logger.isDebugEnabled())
+                {
+                    logger.debug(
+                        "dropping an sr from ssrc="
+                            + rtcpSrPacket.getSenderSsrc()
+                            + ", timestamp="
+                            + rtcpSrPacket.getSenderInfo().getRtpTimestamp());
+                }
+                return;
```
bgrozev: Doesn't this drop all SRs for audio?

bgrozev: Note that we're already filtering out the SRs for video streams that the BC doesn't accept (in `Endpoint.wants()`/`BitrateController.accept()`.

gpolitis: right, i forgot that, i'll revert this

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1545,6 +1523,15 @@ int getTargetIndex()
                 ? ratedIndices[ratedTargetIdx].encoding.getIndex() : -1;
         }
 
+
+        int getPreferredIndex()
```
bbaldino: can this be private or protected?

gpolitis: all these methods (gettargetindex, gettargetbitrate, getpreferredindex, etc) already are package-private and intellij idea doesn't raise a warning. If we change them to be protected then intellij idea gives a warning. We can make them private (i don't see the reason why but i don't really mind either), but i think they should all be the same for concistency

bbaldino: It's always good to scope methods as tightly as they can be. This better isolates the class, makes testing and future refactoring easier and better communicates how methods are supposed to be used (or not be used externally).

bgrozev: Note that all these are members of a private class, so they are already only accessible from `BitrateController`

PR #837, opened 2019-05-31
Increase queue sizes to 1024.

bgrozev:
Keep track of dropped packets and expose it at /colibri/stats/queue

### DIFF
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -32,6 +32,7 @@ import org.jitsi.rtp.*;
 import org.jitsi.rtp.extensions.*;
 import org.jitsi.utils.*;
 import org.jitsi.utils.logging.*;
+import org.jitsi.utils.queue.*;
 import org.jitsi.videobridge.util.*;
 import org.jitsi.xmpp.extensions.jingle.*;
 import org.json.simple.*;
@@ -69,6 +70,12 @@ public class DtlsTransport extends IceTransport
     private static final Predicate<Packet> NON_DTLS_PREDICATE
             = DTLS_PREDICATE.negate();
 
+    /**
+     * Count the number of dropped packets and exceptions.
+     */
+    static final CountingErrorHandler queueErrorCounter
+            = new CountingErrorHandler();
+
     private final Logger logger;
     private final DtlsStack dtlsStack;
     private final ProtocolReceiver dtlsReceiver;
@@ -103,7 +110,9 @@ public class DtlsTransport extends IceTransport
                 = new PacketInfoQueue(
                         getClass().getSimpleName() + "-outgoing-packet-queue",
                         TaskPools.IO_POOL,
-                        this::handleOutgoingPacket);
+                        this::handleOutgoingPacket,
+                        1024);
+        outgoingPacketQueue.setErrorHandler(queueErrorCounter);
 
         dtlsStack = new DtlsStack(endpoint.getID());
         dtlsReceiver = new ProtocolReceiver(dtlsStack);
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -103,7 +110,9 @@ public DtlsTransport(Endpoint endpoint)
                 = new PacketInfoQueue(
                         getClass().getSimpleName() + "-outgoing-packet-queue",
                         TaskPools.IO_POOL,
-                        this::handleOutgoingPacket);
+                        this::handleOutgoingPacket,
+                        1024);
+        outgoingPacketQueue.setErrorHandler(queueErrorCounter);
```
bbaldino: can we just create the `CountingErrorHandler` in place? `outgoingPacketQueue.setErrorHandler(new CountingErrorHandler());`

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -496,7 +505,7 @@ public JSONObject getDebugState()
      * A terminating {@link Node}, which sends the packets through a
      * datagram socket.
      */
-    class SocketSenderNode extends ConsumerNode
+    private class SocketSenderNode extends ConsumerNode
```
bbaldino: 👍

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1174,6 +1177,42 @@ public JSONObject getDebugState(String conferenceId, String endpointId)
         return debugState;
     }
 
+    /**
+     * Gets statistics for the different {@code PacketQueue}s that this bridge
+     * uses.
+     * TODO: is there a better place for this?
+     */
+    public JSONObject getQueueStats()
+    {
```
bbaldino: since the string creation can be costly i'm tempted to say we should always do constants for any un-changing strings like this just as a practice.  what do you think?

src/main/java/org/jitsi/videobridge/octo/OctoTransceiver.java
```diff
@@ -78,7 +85,9 @@
         incomingPacketQueue = new PacketInfoQueue(
                 "octo-tranceiver-incoming-packet-queue",
                 TaskPools.CPU_POOL,
-                this::processPacket);
+                this::processPacket,
+                1024);
+        incomingPacketQueue.setErrorHandler(queueErrorCounter);
```
bbaldino: same as above

src/main/java/org/jitsi/videobridge/DtlsTransport.java
bgrozev: It's static, I want one instance to count all dropped packets

src/main/java/org/jitsi/videobridge/Videobridge.java
bgrozev: Okay, will fix

src/main/java/org/jitsi/videobridge/octo/OctoTransceiver.java
bbaldino: nevermind, see this is static as well.

src/main/java/org/jitsi/videobridge/Videobridge.java
bgrozev: Looking at it again, I think this code is better without constants. It's the only place where these strings are used, and having if they are defined in place everything is in the method body and easier to understand. As for performance, this doesn't create new string instances. The compiler "interns" literals, so it should be equivalent to defining a static final object in terms of performance.

PR #842, opened 2019-06-04
Memory pool updates

bgrozev:


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -357,19 +358,38 @@ private void installIncomingPacketReader(DatagramSocket socket)
...
+                                buf,
+                                RtpPacket.BYTES_TO_LEAVE_AT_START_OF_PACKET,
+                                len);
                     PacketInfo pktInfo = new PacketInfo(pkt);
                     pktInfo.setReceivedTime(System.currentTimeMillis());
                     incomingPipelineRoot.processPacket(pktInfo);
-                    bbuf = ByteBufferPool.getBuffer(1500);
-                    p.setData(bbuf, 0, 1500);
+
+                    p.setData(receiveBuf, 0, receiveBuf.length);
```
bbaldino: shouldn't this be using `buf`?

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -357,19 +358,38 @@ private void installIncomingPacketReader(DatagramSocket socket)
         // enough (it'll only be a bit of the DTLS path) that running it in the
         // IO pool is fine
         TaskPools.IO_POOL.submit(() -> {
-            byte[] bbuf = ByteBufferPool.getBuffer(1500);
-            DatagramPacket p = new DatagramPacket(bbuf, 0, 1500);
+
+            // We need this buffer to be 1500 bytes because we don't know how
+            // big the received packet will be. But we don't want to allocate
+            // large buffers for all packets.
+            byte[] receiveBuf = new byte[1500];
```
bbaldino: it doesn't seem to me like this precludes using the pool for this buffer.  we will just grab it once (and return it once when this entire thing exits) but we can still get a 1500 byte buffer from the pool, that way we can maintain the invariant that all packet buffers come from the pool, which would be useful.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -357,19 +358,38 @@ private void installIncomingPacketReader(DatagramSocket socket)
...
                 {
                     socket.receive(p);
-                    Packet pkt = new UnparsedPacket(bbuf, 0, p.getLength());
+                    int len = p.getLength();
+                    byte[] buf
+                        = ByteBufferPool.getBuffer(
+                                len +
+                                RtpPacket.BYTES_TO_LEAVE_AT_START_OF_PACKET +
+                                RtpPacket.BYTES_TO_LEAVE_AT_END_OF_PACKET);
+                    System.arraycopy(
```
bbaldino: is getting rid of the buffer retrieval/return from/to the pool in need of this change considering we add an arraycopy? 
 we're not `allocating` a big buffer for every one of these packets--they're already allocated, presumably--we're just retrieving and returning them to/from the pool.

src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
```diff
@@ -123,13 +169,31 @@ private static String getStackTraceAsString()
...
             numReturns.incrementAndGet();
         }
-        poolImpl.returnBuffer(buf);
+
+        int len = buf.length;
+        if (len <= T1)
+        {
+            pool1.returnBuffer(buf);
+        }
+        else if (len <= T2)
```
bbaldino: you use `<=` here but `<` above in `getBuffer`

src/main/java/org/jitsi/videobridge/util/PartitionedByteBufferPool.java
```diff
@@ -44,12 +44,7 @@
     /**
      * How many buffers to pre-allocate in each partition.
      */
-    private static final int INITIAL_SIZE = 100;
-
-    /**
-     * The size of buffers to allocate.
-     */
-    private static final int DEFAULT_BUFFER_SIZE = 1500 + RtpPacket.BYTES_TO_LEAVE_AT_END_OF_PACKET;
+    private static final int INITIAL_SIZE = 10;
```
bbaldino: 10 seems really low?

src/main/java/org/jitsi/videobridge/util/PartitionedByteBufferPool.java
```diff
@@ -79,12 +74,15 @@
...
     private boolean enableStatistics = false;
 
+    private final int defaultSize;
+
     /**
      * Initializes a new {@link PartitionedByteBufferPool} instance with
      * a given initial size for each partition.
      */
-    PartitionedByteBufferPool()
+    PartitionedByteBufferPool(int defaultSize)
```
bbaldino: had to look for the usage to figure out this is the default buffer size, it's unclear from the name and the comment.  can you change it to `defaultBufferSize`?

src/main/java/org/jitsi/videobridge/DtlsTransport.java
bgrozev: No, `buf` is the buffer which we pass through the pipeline. We want to reset the `DatagramPacket` with the single instance that we receive into.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
bgrozev: We were allocating a big buffer, because we need a big buffer with `DatagramSocket.receive` since we don't know how big a packet we'll receive.

src/main/java/org/jitsi/videobridge/util/PartitionedByteBufferPool.java
bgrozev: We now have 24 partitions in total. Allocating 2400 initial buffers seemed excessive. In fact, I don't think pre-allocating achieves anything

src/main/java/org/jitsi/videobridge/DtlsTransport.java
bbaldino: ok, i see now.  do we need to re-set it since it's the same buffer?  does the offset/length of the underlying buffer change after calling receive?

bbaldino: ok, i see at least `receive` does alter the underlying packet's length

bkempe:
@bgrozev I'm curious what the reasoning was behind adding the flag `UseConcMarkSweepGC`, since it's deprecated since Java 9. Did tests for JVB2 show advantages?

bgrozev:
@bkempe we use jdk8 in most deployments, which defaults to Parallel GC. In an experiment we ran CMS performed much better than the default, and better than untweaked G1. If I understand correctly newer systems without CMS default to G1, which is what we want.

It would be nice to run further tests to find an optimal G1 configuration.

pause-cms

bkempe:
@bgrozev thanks for the detailed info. One thing to note is that even Java 11 does not automatically default to G1: https://dzone.com/articles/pitfalls-in-jvm-and-docker-defaults

PR #844, opened 2019-06-04
implements keyframe requests to foreign endpoints

gpolitis:
depends on: https://github.com/jitsi/jitsi-media-transform/pull/87

implements keyframe requests for foreign endpoints; keyframes can be requested by the bitrate controller

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -1132,11 +1132,13 @@ public class Conference
         else if (packet instanceof RtcpFbPliPacket
                 || packet instanceof RtcpFbFirPacket)
         {
-            RtcpFbPacket rtcpFbPacket = (RtcpFbPacket) packet;
+            long mediaSsrc = (packet instanceof RtcpFbPliPacket)
+                ? ((RtcpFbPliPacket) packet).getMediaSourceSsrc()
+                : ((RtcpFbFirPacket) packet).getMediaSenderSsrc();
 
             // XXX we could make this faster with a map
             AbstractEndpoint targetEndpoint
-                = findEndpointByReceiveSSRC(rtcpFbPacket.getMediaSourceSsrc());
+                = findEndpointByReceiveSSRC(mediaSsrc);
 
             PotentialPacketHandler pph = null;
             if (targetEndpoint instanceof Endpoint)
@@ -1155,8 +1157,7 @@ public class Conference
             {
                 if (logger.isDebugEnabled())
                 {
-                    logger.debug("Dropping FIR/PLI for media ssrc "
-                        + rtcpFbPacket.getMediaSourceSsrc());
+                    logger.debug("Dropping FIR/PLI for media ssrc " + mediaSsrc);
                 }
             }
             else if (pph.wants(packetInfo))
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -387,6 +387,11 @@ public class Conference
         {
             remoteEndpoint.requestKeyframe(mediaSsrc);
         }
+        else if (logger.isDebugEnabled())
+        {
+            logger.debug(
+                "Cannot request keyframe because the endpoint was not found.");
+        }
     }
     /**
      * Sets the values of the properties of a specific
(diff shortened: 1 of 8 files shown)


src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
```diff
@@ -93,7 +99,33 @@ public OctoTentacle(Conference conference)
             = bundleContext == null ? null :
                 ServiceUtils2.getService(bundleContext, OctoRelayService.class);
 
-        relay = octoRelayService == null ? null : octoRelayService.getRelay();
+
+        if (octoRelayService != null)
+        {
+            relay = octoRelayService.getRelay();
+            keyframeRequester = new KeyframeRequester();
+            keyframeRequester.attach(new Node("octo keyframe relay node")
```
bgrozev: Use a ConsumerNode

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -243,7 +244,7 @@ public boolean accept(@NotNull VideoRtpPacket videoRtpPacket)
                 // but that's how the interface is defined and it's the only
                 // built-in single argument functional interface.  In the future
                 // we can make our own interface here to clean it up a bit
-                keyframeRequester.accept(targetSsrc);
+                keyframeRequester.run();
```
bbaldino: please update (remove, in this case) the comment

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1199,19 +1199,11 @@ public void describe(ColibriConferenceIQ.ChannelBundle channelBundle)
...
+    @Override
+    public void requestKeyframe(long mediaSsrc)
     {
-        AbstractEndpoint endpoint
-                = getConference().findEndpointByReceiveSSRC(ssrc);
-        if (endpoint instanceof Endpoint)
-        {
-            ((Endpoint) endpoint).transceiver.requestKeyFrame(ssrc);
-        }
+        transceiver.requestKeyFrame(mediaSsrc);
```
bbaldino: i like this version of `requestKeyFrame` much better, didn't remember what a hack the old way was.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -836,8 +836,11 @@ private synchronized void update()
             adaptiveTrackProjection
                 = new AdaptiveTrackProjection(
                     diagnosticContext,
-                    trackBitrateAllocation.track,
-                    destinationEndpoint::requestKeyframe);
+                    trackBitrateAllocation.track, () ->
+                        destinationEndpoint.getConference().requestKeyframe(
+                            trackBitrateAllocation.endpointID,
+                            trackBitrateAllocation.targetSSRC));
+
```
bbaldino: does removing the flexibility of passing the ssrc buy us anything here?  the old call syntax here looked much cleaner, and allowed us to choose which ssrc we requested the keyframe on (which maybe doesn't _happen_ to make a difference since chrome will send on all, but logically seems more correct)

gpolitis: if we know the endpoint we don't have to iterate over the endpoints map in conference but i can bring back the old signature if you find it significantly more readable

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1112,11 +1132,13 @@ public void handleIncomingPacket(PacketInfo packetInfo)
         else if (packet instanceof RtcpFbPliPacket
                 || packet instanceof RtcpFbFirPacket)
         {
-            RtcpFbPacket rtcpFbPacket = (RtcpFbPacket) packet;
+            long mediaSsrc = (packet instanceof RtcpFbPliPacket)
```
bgrozev: Ugh....nice catch!

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bbaldino: realizing now that i misread this line.  i read it that we were requesting a keyframe from `destinationEndpoint` (which i know realize doesn't make sense), but i see we're just using it to get to the conference.  probably a helper we can write here to help readability, but don't worry about it for this.

PR #845, opened 2019-06-05
fix: Do not accept packets until ICE and DTLS connect.

bgrozev:
ICE/DTLS may take a long time (>1s) and if we keep caching packets
coming from other endpoint in the conference we:
1. Might fill the SRTP cache
2. Will cause a spike in bitrate once we connect (while videos are probably
not decodable because yet because there's no keyframe)
3. Might cause problems with the timing of the audio playback

### DIFF
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -89,7 +89,7 @@ public class DtlsTransport extends IceTransport
     private final Node incomingPipelineRoot;
     private final Node outgoingDtlsPipelineRoot;
     private final Node outgoingSrtpPipelineRoot;
-    protected boolean dtlsHandshakeComplete = false;
+    private boolean dtlsHandshakeComplete = false;
 
     /**
      * Initializes a new {@link DtlsTransport} instance for a specific endpoint.
@@ -210,12 +210,16 @@ public class DtlsTransport extends IceTransport
         super.startConnectivityEstablishment(transportPacketExtension);
     }
 
+    /**
+     * Returns {@code true} if this {@link DtlsTransport} is connected. It is
+     * considered connected if the underlying ICE connection has been
+     * established and the DTLS session has been established.
+     * @return
+     */
     @Override
     public boolean isConnected()
     {
-        // TODO: do we consider this TM connected when ICE completes, or when
-        // ICE and DTLS both complete?
-        return super.isConnected();
+        return super.isConnected() && dtlsHandshakeComplete;
     }
 
     /**
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -407,11 +407,33 @@ public class Endpoint
     }
 
     /**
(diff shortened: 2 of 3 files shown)


bbaldino:
I'm confused by some of the reasoning here:

> Will cause a spike in bitrate once we connect (while videos are probably not decodable because yet because there's no keyframe)

The first frame is likely a keyframe here, no?  That's one of the reasons we wanted to cache all these packets in the first place.

> Might cause problems with the timing of the audio playback
Won't audio will be delayed by the same amount?  And this seems like a decision the client-side jitter buffer is best to make.

Also, if we were to do this change, is there value in the SRTP cache at all?  It's main purpose is to catch the initial keyframe and now we'd be dropping it before it ever reached SRTP.

bgrozev:
> The first frame is likely a keyframe here, no?

I am not sure. I'd have to look at what the bitrate controller does initially. But see below.

> Also, if we were to do this change, is there value in the SRTP cache at all? It's main purpose is to catch the initial keyframe and now we'd be dropping it before it ever reached SRTP.

This does not change anything on the receive side. We still cache the keyframe coming from an endpoing (if it arrives before SRTP is ready).

On the send side

bgrozev:
Here's more on the problem. An endpoint (E) joins a running conference. We create an Endpoint when receive the colibri request for new channels, and Conference starts to route packets to it (via wants/send) before we've even started ICE. At a later stage, say X ms later, ICE and DTLS complete and we send all the packets that we've accumulated in a burst.

I am not quite sure how the bitrate controller behaves. I think that it will request a keyframe for all streams before it starts to forward anything. But and the keyframes will likely arrive before our endpoint finishes ICE and DTLS (since that’s multiple RTTs). When we’re ready with SRTP the cache will have:
1. Xms of audio for all other endpoints
2. Some video packets, with varying length <Xms, depending on when the keyframes arrived

That is, unless we completely filled up the cache (it's limited to 1024) and some packets were dropped. We then send the contents in a burst, likely causing packet loss, likely losing some of these keyframe packets (causing the endpoint to NACK before it can render). When I join a 4-way conference the SRTP Encrypt cache grow to over 700-800 and over 1024 in one of the tests that I ran (note that there's a bug in JMT master and the Node reports the current size of the cache and not size that it reached). There is also some visual freezing and drops to lower quality in the first few seconds. This is what joining a 4-way call looks like in terms of traffic:

Total received bytes:
Screenshot 2019-06-06 at 13 16 28

Bytes sent in RTCP:
Screenshot 2019-06-06 at 13 19 35



With this PR we wait until DTLS has completed before accepting packets from other endpoints. Until DTLS has completed be block the packets from reaching the bitrate controller. When DTLS completes, we start sending audio immediately and request keyframes. The cache is rarely used in practice at all, so there is no initial burst. There is a slight delay before rendering, but it is smooth and we start with high quality right away. The delay is similar to what we see when the selected endpoint changes, and is acceptable IMO. We could optimize the delay with preemptive kf requests, but I think that’s something for later. Recording (the server is in the US, all endpoints are in EU): [before](http://boris.jitsi.net/before.mov) and [after](http://boris.jitsi.net/after.mov)

bgrozev:
Example received stream before:
Screenshot 2019-06-06 at 14 03 27

PR #851, opened 2019-06-11
Merge master

bgrozev:


bgrozev:
This should make the PR tests actually run, since Jenkins tries to make a merge and fails due to conflicts.

PR #852, opened 2019-06-11
Removes unused code and imports.

bgrozev:


src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -118,10 +116,10 @@ public VP8AdaptiveTrackProjectionContext(
         // Compute the starting sequence number and the timestamp of the initial
         // frame based on the RTP state.
         int startingSequenceNumber =
-            (rtpState.maxSequenceNumber + 1) & RawPacket.SEQUENCE_NUMBER_MASK;
+            (rtpState.maxSequenceNumber + 1) & 0xffff;
```
bbaldino: looks like you added new constants for these?

bbaldino: oh, i see they're private.  i guess we should put these in RTP.

PR #853, opened 2019-06-11
Fixes moving endpoints between bridges with Octo

bgrozev:
Expires OctoEndpoints if a request for a new local endpoint with the
same ID is received.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -775,6 +775,18 @@ public class Conference
     public Endpoint getOrCreateLocalEndpoint(String id)
     {
         AbstractEndpoint endpoint = getEndpoint(id, /* create */ true);
+        if (endpoint instanceof OctoEndpoint)
+        {
+            // It is possible that an Endpoint was migrated from another bridge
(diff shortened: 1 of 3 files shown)


bgrozev:
I don't think I understand your suggestion. Are you saying that instead of calling `expire` on the endpoint, the code in `Conference` should check that is has an `OctoEndpoint` instance, find the associated `OctoEndpoints` (which is currently not exposed outside the octo package) and expire it this way? Or that `OctoEndpoint` finds the `OctoEndpoints` instance in another way instead of having a link to the parent (it would have to be through some sort of parent, either `Conference` or `Tentacle`.

bbaldino:
Maybe I don't understand the design of `OctoEndpoints`, I thought there was one instance per conference that modeled the remote endpoints and that was used as a proxy for interactions with all `OctoEndpoint` instances.  But now I'm having memories of us treating `OctoEndpoint` more like a normal one and holding them directly in the conference, is that right?  But yeah, I was thinking it'd cast which doesn't feel great either, I just hate the parent pointer scheme and it feels like an antipattern.  I guess let's not worry about it for now, in the future we could limit the expire call to an interface which OctoEndpoints implements and just pass that down or something.  /shrug

PR #854, opened 2019-06-11
various fixes for bandwidth estimations/distribution

gpolitis:
Depends on https://github.com/jitsi/jitsi-utils/pull/5

### DIFF
--- a/src/main/java/org/jitsi/videobridge/EncodingsManager.java
+++ b/src/main/java/org/jitsi/videobridge/EncodingsManager.java
@@ -16,10 +16,10 @@
 
 package org.jitsi.videobridge;
 
-import org.eclipse.jetty.util.*;
 import org.jitsi.nlj.rtp.*;
 
 import java.util.*;
+import java.util.concurrent.*;
 
 /**
  * Process information signaled about encodings (payload types, SSRCs, SSRC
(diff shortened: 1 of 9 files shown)


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1195,6 +1187,11 @@ public void addPayloadType(PayloadType payloadType)
     {
         payloadTypes.put(payloadType.getPt(), payloadType);
         adaptiveTrackProjections.forEach(atp -> atp.addPayloadType(payloadType));
+
+        if (payloadType.getEncoding() == PayloadTypeEncoding.RTX)
```
bgrozev: Do we not signal an RTX payload type with firefox?

gpolitis: no

src/main/java/org/jitsi/videobridge/EncodingsManager.java
```diff
@@ -58,7 +57,7 @@
      * @param secondarySsrc the secondary SSRC in the SSRC association.
      * @param type the association type.
      */
-    public void addSsrcAssociation(
+    public synchronized void addSsrcAssociation(
```
bgrozev: I think we should release the lock before calling into the listeners (since we don't know what the code they run and they could be aqcuiring other locks). Also, as a general good practice synchronize on another object which is private to the instance.

gpolitis: i'll take care of that, thanks

gpolitis: i provided an alternative, please check

PR #860, opened 2019-06-18
Packet delay

bbaldino:
relies on https://github.com/jitsi/jitsi-media-transform/pull/95

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -21,6 +21,8 @@ import org.ice4j.stack.*;
 import org.jitsi.eventadmin.*;
 import org.jitsi.impl.neomedia.transform.*;
 import org.jitsi.nlj.*;
+import org.jitsi.nlj.stats.*;
+import org.jitsi.nlj.util.*;
 import org.jitsi.osgi.*;
 import org.jitsi.rtp.util.*;
 import org.jitsi.service.configuration.*;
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1131,9 +1132,9 @@ public void setShutdownRunnable(Runnable shutdownRunnable)
      * to include. If not specified, all of the conference's endpoints will be
      * included.
      */
-    public JSONObject getDebugState(String conferenceId, String endpointId)
+    public OrderedJsonObject getDebugState(String conferenceId, String endpointId)
```
bgrozev: Is there a reason to change the signature too? Looks like the return type can remain the more abstract one.

bgrozev:
That's cool! I meant for these stats to be exposed at `/colibri/stats/transit` (so we can grab them and throw the to dd): https://github.com/jitsi/jitsi-videobridge/blob/nlj-jvb/src/main/java/org/jitsi/videobridge/rest/StatisticsRequestHandler.java#L122

bbaldino:
A couple notes on the changes I pushed here:

1. Because you mentioned adding more in the future, I added the `PacketTransitStats` class to act as a place to combine the transit stats (the other ones already have a single places that makes to gather them).  I named it `PacketTransitStats` because, given we're using "transit", I couldn't think of something else other than packets that made sense with "transit".  "transit" can still work as shorthand for the REST endpoint, I think, but let me know if I'm missing something on the use case.

1. `Videobridge.java` still accesses the delay fields directly from `DtlsTransport` because, after our discussion, I imagined that `PacketTransitStats` would potentially carry more information and could be enabled/disabled (like the others) due to a runtime cost.  The end-to-end stats, however, are always on, so `Videobridge.java` should always be able to print them.  I can imagine an implementation of `PacketDelayStats` which would support this (something almost like 'log levels' which are turned on and off within the top-level stat: i.e. `transit` stats are on a level of 'info' by default--which includes the end-to-end delay--but could be configured to be more 'verbose' and include other more costly stats), but I think it's something we should be consistent on and apply to the other stats as well.

1. I did not plumb through enabling/disabling of the transit stats at this time since there is currently nothing to turn on/off.

PR #861, opened 2019-06-21
Track jitter introduced by the bridge

bbaldino:
relies on https://github.com/jitsi/jitsi-media-transform/pull/99

### DIFF
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -34,6 +34,7 @@ import org.jitsi.rtp.rtp.*;
 import org.jitsi.utils.*;
 import org.jitsi.utils.logging.*;
 import org.jitsi.utils.queue.*;
+import org.jitsi.videobridge.stats.*;
 import org.jitsi.videobridge.util.*;
 import org.jitsi.xmpp.extensions.jingle.*;
 import org.json.simple.*;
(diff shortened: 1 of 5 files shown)


bbaldino:
I added a `DoubleAverage` class here, and I imagine we'd eventually build out a set of common utilities like this (maybe in jitsi-utils?) with base/subclasses to cover more cases; but for now I just added what I needed.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -72,6 +72,7 @@
             = DTLS_PREDICATE.negate();
 
     public static final PacketDelayStats packetDelayStats = new PacketDelayStats();
+    private final BridgeJitterStats bridgeJitterStats = new BridgeJitterStats();
```
bgrozev: Can we go back to following the convention for java? Add docs on the fields (I mention it now because I think a doc on this one will be actually helpful), and keep the statics on top.

bgrozev:
This is awesome!

I don't know if keeping the global average makes sense. Over time spikes or general trends to e.g. increase with load will be diluted. The same goes for packet delay. I've been thinking about this and I have two ideas: 
1. Use RateStatistics or a similar approach. The problem is that is uses a lock, and the interval is quite long.
2. Reset the counter every N seconds, saving the accumulated value (to be returned by the getters). The disadvantage is that the value it return is going to be a bit stale. But I think this is acceptable. We should be able to do it without locking if we can accept that some values might not be counted.

That's a separate task, though, I'm just sharing my thoughts.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
bbaldino: yeah will fix this

bbaldino:
I was thinking instead of resetting it every N seconds, how about we just reset it after each access (just the global one).  If the idea is to poll this regularly for datadog, maybe this is fine?

bgrozev:
I don't think it's a good idea to depend on the stats being accessed from only one place at a certain period.

PR #863, opened 2019-06-24
Request preemptive keyframes when active speaker changes

bbaldino:
relies on https://github.com/jitsi/jitsi-media-transform/pull/100

@bgrozev let me know if the Octo stuff doesn't look right

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -435,6 +435,7 @@ public class Conference
             broadcastMessage(
                     createDominantSpeakerEndpointChangeEvent(
                         dominantSpeaker.getID()));
+            dominantSpeaker.requestKeyframe();
         }
     }
 
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -18,6 +18,7 @@ package org.jitsi.videobridge;
 import org.jetbrains.annotations.*;
 import org.jitsi.eventadmin.*;
 import org.jitsi.nlj.*;
+import org.jitsi.nlj.stats.*;
 import org.jitsi.rtp.*;
 import org.jitsi.rtp.rtcp.rtcpfb.payload_specific_fb.*;
 import org.jitsi.rtp.rtp.*;
@@ -435,8 +436,48 @@ public class Conference
             broadcastMessage(
                     createDominantSpeakerEndpointChangeEvent(
                         dominantSpeaker.getID()));
-            dominantSpeaker.requestKeyframe();
+            if (getEndpointCount() >= 3)
+            {
+                double senderRtt = getRtt(dominantSpeaker);
+                double maxReceiverRtt = getMaxReceiverRtt(dominantSpeaker.getID());
+                // We add an additional 10ms delay to reduce the risk of the keyframe arriving
+                // too early
+                double keyframeDelay = maxReceiverRtt + senderRtt + 10;
+                if (logger.isDebugEnabled())
+                {
+                    logger.debug("Scheduling keyframe request from " + dominantSpeaker.getID() + " after a delay" +
+                            "of " + keyframeDelay + "ms");
+                }
+                TaskPools.SCHEDULED_POOL.schedule((Runnable) dominantSpeaker::requestKeyframe, (long)keyframeDelay, TimeUnit.MILLISECONDS);
+            }
         }
+    }
+
+    private double getRtt(AbstractEndpoint endpoint)
+    {
+        if (endpoint instanceof Endpoint)
+        {
+            Endpoint localDominantSpeaker = (Endpoint)endpoint;
+            return localDominantSpeaker.getTransceiver().getTransceiverStats().getEndpointConnectionStats().getRtt();
+        }
+        else
+        {
+            // Octo endpoint
+            return 100;
+        }
+    }
+
+    private double getMaxReceiverRtt(String excludedEndpointId)
+    {
+        return endpointsCache.stream()
+                .filter(ep -> !ep.getID().equalsIgnoreCase(excludedEndpointId))
+                .map(Endpoint::getTransceiver)
+                .map(Transceiver::getTransceiverStats)
+                .map(TransceiverStats::getEndpointConnectionStats)
+                .map(EndpointConnectionStats.Snapshot::getRtt)
+                .mapToDouble(Double::valueOf)
+                .max()
+                .orElse(0);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/jitsi/videobridge/octo/OctoEndpoint.java
```diff
@@ -78,6 +78,25 @@ public void requestKeyframe(long mediaSsrc)
...
+        if (tentacle != null)
+        {
+            long firstVideoSsrc = receiveSsrcs.entrySet().stream()
+                    .filter(entry -> entry.getValue().equals(MediaType.VIDEO))
+                    .findFirst()
+                    .map(Map.Entry::getKey)
+                    .orElse(-1L);
+            if (firstVideoSsrc >= 0)
+            {
+                requestKeyframe(firstVideoSsrc);
```
bbaldino: i could just call `tentacle.requestKeyframe(firstVideoSsrc);` here, but i thought it would be better to call the existing code.  it's unfortunate that then we have to look up the tentacle twice, though, so definitely open to changing this if you think it'd be better.

src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
```diff
@@ -244,7 +245,14 @@ public void setSources(
                 return;
             }
 
-            endpoint.addReceiveSsrc(source.getSSRC());
+            if (audioSources.contains(source))
```
bbaldino: i considered using the original audio and video lists, but since there's some filtering that didn't seem great (and would also result in some duplicated code) so did it this way.  open to another idea though.

src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
```diff
@@ -248,12 +248,12 @@ public void describe(ColibriConferenceIQ.ChannelCommon commonIq)
     /**
      * Sets the list of sources signaled for this channel.
      */
-    public void setSources(@NotNull List<SourcePacketExtension> sources)
+    public void setSources(@NotNull List<SourcePacketExtension> sources, MediaType mediaType)
```
bgrozev: The `ChannelShim` already has an associated `ContentShim`. I think it would be better to call `getMediaType` in the method instead of taking it as a param. Or is there a reason this won't work?

src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
bgrozev: I think this is fine.

src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
bbaldino: I was looking for something like this and guess I just missed it.  Will change it.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -435,9 +436,50 @@ void dominantSpeakerChanged()
...
+            }
+        }
+    }
+
+    private double getRtt(AbstractEndpoint endpoint)
+    {
+        if (endpoint instanceof Endpoint)
+        {
+            Endpoint localDominantSpeaker = (Endpoint)endpoint;
+            return localDominantSpeaker.getTransceiver().getTransceiverStats().getEndpointConnectionStats().getRtt();
```
bbaldino: this (and below) call to get the endpoint's RTT is a bit gnarly.  i considered adding a `getRtt` function to `Endpoint`, but it felt a little weird to bubble up that one stat.  now that i'm seeing it again after a day i think it's probably better to add the `getRtt` function to `Endpoint`, so i'll do that.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1230,7 +1230,13 @@ public void describe(ColibriConferenceIQ.ChannelBundle channelBundle)
     @Override
     public void requestKeyframe(long mediaSsrc)
```
bbaldino: i kept this one around because i think it still makes sense for Octo, but for a normal endpoint it maps to the same call.  (actually, in `Endpoint` this should probably call the other `requestKeyframe` method, will tweak that).

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -435,9 +436,50 @@ void dominantSpeakerChanged()
...
+        }
+        else
+        {
+            // Octo endpoint
+            return 100;
```
bbaldino: this is a hack.  thoughts on what to do here?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -435,9 +436,47 @@ void dominantSpeakerChanged()
             broadcastMessage(
                     createDominantSpeakerEndpointChangeEvent(
                         dominantSpeaker.getID()));
+            if (getEndpointCount() >= 3)
+            {
+                double senderRtt = getRtt(dominantSpeaker);
+                double maxReceiverRtt = getMaxReceiverRtt(dominantSpeaker.getID());
+                // We add an additional 10ms delay to reduce the risk of the keyframe arriving
+                // too early
+                double keyframeDelay = maxReceiverRtt + senderRtt + 10;
```
bgrozev: This should subtract the sender RTT: https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/VideoChannel.java#L640

src/main/java/org/jitsi/videobridge/octo/OctoEndpoint.java
```diff
@@ -78,6 +78,25 @@ public void requestKeyframe(long mediaSsrc)
...
     }
 
+    @Override
+    public void requestKeyframe()
+    {
+        // just making sure the tentacle hasn't expired
+        OctoTentacle tentacle = getConference().getTentacle();
+        if (tentacle != null)
+        {
+            long firstVideoSsrc = receiveSsrcs.entrySet().stream()
```
bgrozev: Do we have the same problem with RTX SSRCs as in JMT?

src/main/java/org/jitsi/videobridge/Conference.java
bbaldino: nice catch

src/main/java/org/jitsi/videobridge/octo/OctoEndpoint.java
bbaldino: hm, i suspect we will.  will fix that.

bbaldino:
replaced by https://github.com/jitsi/jitsi-videobridge/pull/898

PR #865, opened 2019-07-04
Octo fixes

bgrozev:


src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
```diff
@@ -436,16 +436,35 @@ else if (buf.length >= octoPacketLength)
...
      * @param conferenceId the conference ID.
      */
-    void removeHandler(String conferenceId)
+    void removeHandler(String conferenceId, PacketHandler handler)
     {
-        packetHandlers.remove(conferenceId);
+        synchronized (packetHandlers)
+        {
+            // If the Colibri conference for this GID was re-created, and the
+            // original and the original Conference object is expired after a
```
gpolitis: there's a small typo here

gpolitis:
apologies for the naive question, but when and why does jicofo send an octo channel request without the type=octo attribute set correctly?

bgrozev:
I haven't looked into it, my guess is because the code to expire channel and/or update sources is generic and the type is just not set. I think on the protocol level it makes sense to not mandate the attribute, so I fixed it here in the bridge.

src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
```diff
@@ -216,10 +216,16 @@ public void processOctoChannels(
         {
             tentacle.expire();
         }
-        else
+        else if (audioChannel instanceof ColibriConferenceIQ.OctoChannel
```
bbaldino: should this re-use `VideobridgeShim#isOctoChannel`?

bgrozev: No, this is intentional. The methods to get the relays are only available in the OctoChannel class.

PR #866, opened 2019-07-05
adds a check subcommand to the time series cli tool

gpolitis:


gpolitis:
@bbaldino, about what this PR does: it makes the `trackBitrateAllocation.improve()` step greedy. today, if we we call improve, it will move the target to the next rated quality (if there's enough available bandwidth, of course). with this pr, the improve step will move the target the next rated quality with the least bitrate.

About the issue that this PR addresses: it's possible that two encodings have the same bitrate (there's a relevant comment in the `trackBitrateAllocation.improve()` method that illuminates this a bit further). in that case, the `bitrateController.improve()` step may stop pre-maturely (the stop condition is: the target bitrate hasn't changed between two improve iterations). With a greedy `trackBitrateAllocation.improve()` step, this issue is avoided.

does this make sense? do you think it'd be useful if we included this description either in the commit description or somewhere in the code?

PR #871, opened 2019-07-19
Unthrottle bitrate allocation updates when the bandwidth increases

gpolitis:
Right now we require a 15% (configurable) change in the bwe in order to
update the bitrate allocation. This is an ugly hack to prevent too many
resolution/UI changes in case the bridge produces too low bandwdith
estimate, at the risk of clogging the receiver's pipe.

The code prevented not only downscaling, but also upscaling, which is
problematic: Suppose that the target bitrate is 2.5Mbps, and that the
last bitrate allocation was performed with a 2.4Mbps bandwidth esitmate.
The bridge keeps probing and, suppose that, eventually the bandwidth
estimate reaches 2.6Mbps, which is plenty to accomodate the target
bitrat; but the minimum bandwidth estimate that would trigger a new
bitrate allocation is 2.4Mbps + 2.4Mbps * 15% = 2.76Mbps.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -159,7 +159,7 @@ public class BitrateController
      * The default value of the bandwidth change threshold above which we react
      * with a new bandwidth allocation.
      */
-    private static final int BWE_CHANGE_THRESHOLD_PCT_DEFAULT = 15;
+    private static final int BWE_CHANGE_THRESHOLD_PCT_DEFAULT = -15;
 
     /**
      * The ConfigurationService to get config values from.
@@ -363,8 +363,8 @@ public class BitrateController
         {
             return true;
         }
-        return Math.abs(previousBwe - currentBwe)
-            >= previousBwe * BWE_CHANGE_THRESHOLD_PCT / 100;
+        return previousBwe - currentBwe
+            < previousBwe * BWE_CHANGE_THRESHOLD_PCT / 100;
     }
 
     /**
@@ -605,13 +605,14 @@ public class BitrateController
             // do not update the bitrate allocation. The goal is to limit
             // the resolution changes due to bandwidth estimation changes,
             // as often resolution changes can negatively impact user
-            // experience.
+            // experience, at the risk of clogging the receiver pipe.
         }
         else
         {
             if (logger.isDebugEnabled())
             {
-                logger.debug(destinationEndpoint.getID() + " bandwidth has changed, updating");
+                logger.debug(destinationEndpoint.getID() +
+                        " bandwidth has changed, updating");
             }
 
             lastBwe = newBandwidthBps;
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -159,7 +159,7 @@ public class BitrateController
      * The default value of the bandwidth change threshold above which we react
      * with a new bandwidth allocation.
      */
-    private static final int BWE_CHANGE_THRESHOLD_PCT_DEFAULT = -15;
+    private static final int BWE_CHANGE_THRESHOLD_PCT_DEFAULT = 15;
 
     /**
      * The ConfigurationService to get config values from.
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -363,8 +363,8 @@ private static boolean isLargerThanBweThreshold(
         {
             return true;
         }
-        return Math.abs(previousBwe - currentBwe)
-            >= previousBwe * BWE_CHANGE_THRESHOLD_PCT / 100;
+        return previousBwe - currentBwe
```
bgrozev: This is really confusing. 
`p - c < -0.15p` is the same as `c - p > 0.15p`. Which is true only if the bwe grew and it grew by more than 15%. Therefore we completely ignore drops in bwe, and we only react once the increase is over 15% (which is the issue the commit message claims to solve).

Can you please normalize the expressions (remove the extra minus signs and use "current - previous" as that's easier to interpret), and add a comment for the intention?

gpolitis: i don't know what i've been thinking nor how it worked in my tests

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -356,15 +356,33 @@ public BitrateController(
...
+        // probing and, suppose that, eventually the bandwidth estimate reaches
+        // 2.6Mbps, which is plenty to accomodate the target bitrat; but the
+        // minimum bandwidth estimate that would trigger a new bitrate
+        // allocation is 2.4Mbps + 2.4Mbps * 15% = 2.76Mbps.
+        //
+        // if, on the other hand, the bwe has decreased, we require a 15%
+        // (configurable) drop at last in order to update the bitrate
+        // allocation. This is an ugly hack to prevent too many resolution/UI
+        // changes in case the bridge produces too low bandwdith estimate, at
+        // the risk of clogging the receiver's pipe.
```
bbaldino: this seems counterintuitive to me, isn't it usually the opposite? (drop bitrate down quickly so as not to oversend, and cautiously ramp up?)

bbaldino: (not that the original bug  doesn't  still need to be fixed, just wondering if this tradeoff makes sense. did we find before that not  having this threshold made things too sensitive?)

gpolitis: > this seems counterintuitive to me, isn't it usually the opposite? (drop bitrate down quickly so as not to oversend, and cautiously ramp up?)

you're right brian, and this is how the aimd works in our gcc implementation.

some context first, back when we were working on this (2-3 years ago) we had noticed "false positives" of low bandwidth during our standups on beta (it seemed that the bandwidth estimator was complaining that the bandwidth had dropped when in reality it hadn't, i don't remember if the investigation was conclusive) causing unnecessary fluctuations in the user experience (i.e. the video resolution as well as in the number of participants that the bridge is forwarding).

We introduced this hack in an effort to mitigate the issue: the bitrate controller doesn't "trust" the bandwidth estimation if it's not much lower than the last bwe that triggered an update, at the risk of clogging the receiver's link.

tl;dr; this is an ugly hack, and the bitrate controller shouldn't contain logic that is related to producing or altering the bandwidth estimation because its responsibility is to distribute the bandwidth and not to produce a bandwidth estimate.

back to present day, It turns out that, apart from the risk of clogging the receiver's link, this has another negative side effect: it may block ramp-up to HD and cause a perpetual bandwidth probing situation  because the bandwidth estimate hasn't changed more than 15% of the last bandwidth estimate that triggered an update.

bgrozev: I wonder how often the bandwidth estimations change. I think we still want to avoid calling `update()` often, because even without any bwe change it might make a different decision because of streams' bitrates changing.

bbaldino: talked with george a bit about this, summarizing some thoughts here:

1) it'd be interesting to experiment with this on the new bridge--removing the throttling altogether just to see what we learn, and/or perhaps one of these options:
1a) Changing the throttle threshold to include an absolute value (i.e. something like: if the bandwidth has changed by 15% or XXXkbps, whichever is smaller; this addresses the issue of 15% being a large number at high bitrates)
1b) Changing the throttle threshold to include a sense of clock time (i.e. "if the bandwidth change is more than 15% OR I haven't updated the bandwidth in XXXms)

gpolitis: ditto what brian said; i'll merge this later today because it fixes a bug in the hack, unless anybody opposes to the idea, and we can have a more elaborate discussion on which of the above suggestions we tackle next

PR #873, opened 2019-07-19
Merge master

bgrozev:


gpolitis:
why do we need the new dependency? (jitsi-xmpp-extensions)

bgrozev:
Ops, it's already there. I removed it.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -215,6 +215,19 @@ public boolean accept(@NotNull VideoRtpPacket videoRtpPacket)
             logger.warn(
                 "Dropping an RTP packet, no quality index (SSRC=" +
                     videoRtpPacket.getSsrc() + ").");
+            MediaStreamTrackDesc sourceTrack = getSource();
+            if (sourceTrack != null)
+            {
+                logger.warn(
```
bbaldino: confused by this, if we find the source track is not null we drop it?  and if it is null we  drop it?

bbaldino: oh, ok...after expanding i think i see now. nevermind.

PR #876, opened 2019-07-24
Updates debian packaging and changing the name of the package 

damencho:
Do not merge, please. The package and everything seems fine, but I just want to test installation and upgrade to the new package before proceeding with the merge.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -200,10 +200,50 @@
 
     <!-- Jitsi Videobridge Version -->
     <target name="version">
+        <!-- extracts major version -->
+        <loadresource property="version.major">
+            <propertyresource name="label"/>
+            <filterchain>
+                <tokenfilter>
+                    <filetokenizer/>
+                    <replaceregex pattern="\..*\-.*" replace="" />
+                </tokenfilter>
+            </filterchain>
+        </loadresource>
+        <!-- extracts minor version -->
+        <loadresource property="version.minor">
+            <propertyresource name="label"/>
+            <filterchain>
+                <tokenfilter>
+                    <filetokenizer/>
+                    <replaceregex pattern="\d*\." replace="" />
+                    <replaceregex pattern="-\d*" replace="" />
+                </tokenfilter>
+            </filterchain>
+        </loadresource>
+        <!-- extracts build id -->
+        <loadresource property="version.buildid">
+            <propertyresource name="label"/>
+            <filterchain>
+                <tokenfilter>
+                    <filetokenizer/>
+                    <replaceregex pattern=".*\-" replace="" />
+                </tokenfilter>
+            </filterchain>
+        </loadresource>
+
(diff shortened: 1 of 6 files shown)


damencho:
Temporary close it.

build.xml
```diff
@@ -200,10 +200,50 @@
 
     <!-- Jitsi Videobridge Version -->
     <target name="version">
+        <!-- extracts major version -->
```
bbaldino: so this is still using ant, is that right? is there a maven method to achieve this? or even leverage a bash script or something maybe?  it'd be nice not to add new dependencies on ant.

resources/install/debian/postinst
```diff
@@ -152,6 +152,12 @@ case "$1" in
         # OpenVZ containers do not typically allow permission to modify the
         # kernel with sysctl config values.
         sysctl --system || true
+
+        # Make sure the service is enabled, on upgrade from old to new bridge
+        # we see the service as masked (failing to stop on remove)
+        if [ -d /run/systemd/system ]; then
+            systemctl unmask jitsi-videobridge
```
bbaldino: should this be `jitsi-videobridge2`?

build.xml
damencho: There is no new dependency added. We can do that but it is a separate project, not part of this one. All our java packages use ant. Ant and packaging is used even before maven was introduced for the java projects.

resources/install/debian/postinst
damencho: This was removed in later commits.

damencho: It is automatically handled by debhelper.

damencho: I was trying to find out what was wrong with upgrade till I found that debhelper was not adding the stuff for systemd because the unit file was still with the old name, but it needs to match the package. Upgrades now works, but clean install still have some problems I'm looking into. I closed the PR so we do not do milion of PR tests while I'm experimenting in my fork branch.

build.xml
bbaldino: i get that, but we'd prefer to be off of ant, no?  by 'new dependency' i didn't mean adding a dependency on ant entirely, but adding another 'feature' which relies on ant.  and i agree we shouldn't try and get off of ant now (can be a separate project, as you said) i just wondered if there was an approach for this that didn't use ant that wouldn't be so hard to do now.

damencho: I don't want to do that now as it brings a lot of changes. And we will slow down stuff for a week or two which we cannot afford at the moment.  
Yes sure we can get rid of ant at some point, but this was not discussed and not planned at this point. This is the case for jicofo, jigasi and videobridge.
Half of the build process currently that is available for jvb can be ripped off as it is for building a Debian source package, which no longer works and is not something we will do in the future. 
But still, there is much more other stuff that is handled while building the package.

PR #882, opened 2019-07-31
Fix audio only octo endpoints

bgrozev:


bbaldino:
so i see we set an absolute set of the sources each time rather than adding them one at a time, is that right? but it looks like we never removed a signaled ssrc, so how is this different?  (i don't doubt that something about it must be, just don't follow what)

src/main/java/org/jitsi/videobridge/octo/OctoEndpoint.java
```diff
@@ -81,7 +81,7 @@ public void requestKeyframe(long mediaSsrc)
     @Override
     public boolean shouldExpire()
     {
-        return ArrayUtils.isNullOrEmpty(getMediaStreamTracks());
```
bgrozev: @bbaldino this line is the actual fix. It takes into account all signaled SSRCs (which are now removed), instead of using the tracks (which are video only).

bgrozev:
See my inline comment

src/main/java/org/jitsi/videobridge/octo/OctoEndpoint.java
bbaldino: ah gotcha, good to know.  what is the other change for?

bgrozev: For setting the SSRCs globally so that they are cleaned up properly.

PR #892, opened 2019-08-21
pass media type when adding rx ssrcs

bbaldino:
this will need to be updated once https://github.com/jitsi/jitsi-media-transform/pull/114 is in.


src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.jitsi.videobridge;
 
+import org.jitsi.utils.MediaType;
```
bgrozev: *?

PR #895, opened 2019-08-28
Restored audio channels count statistics from JVB 1.0.

mstyura:
Hi, @bgrozev 

I'm working on project which uses JVB 1.0 as a java library. 
Now I'm trying to migrate to JVB 2.0 and found out that you had removed audio channels statistics during rewrite.
Is there any chance it could be restored? 
Here are proposed changes with restore.

Thanks,
Yura.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/stats/CallStatsIOTransport.java
+++ b/src/main/java/org/jitsi/videobridge/stats/CallStatsIOTransport.java
@@ -198,7 +198,7 @@ public class CallStatsIOTransport
             long measurementInterval)
     {
         bsib.audioFabricCount(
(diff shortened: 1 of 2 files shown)


mstyura:
Hi, @bbaldino,
I've used this field to report number of audio-enable peers to data dog (part internal to my project).
The presence of video channels statistics in JVB 2.0 confused me, that's why I added audio channels back.
Have no objection if both fields completely removed.

mstyura:
Thanks, @bbaldino.
I don't strictly need this statistics now. I'm actually have much bigger issue with migration, this one with statistics looks good point to start discussion :)

PR #896, opened 2019-08-28
Logging v2

bbaldino:
jvb changes to adapt to the new logger in https://github.com/jitsi/jitsi-utils/pull/11

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -22,8 +22,10 @@ import org.jitsi.rtp.*;
 import org.jitsi.rtp.rtcp.rtcpfb.payload_specific_fb.*;
 import org.jitsi.rtp.rtp.*;
 import org.jitsi.utils.event.*;
-import org.jitsi.utils.logging.*;
-import org.jitsi.utils.logging.Logger;
+import org.jitsi.utils.logging.DiagnosticContext;
+import org.jitsi.utils.logging2.*;
+import org.jitsi.utils.logging2.Logger;
+import org.jitsi.utils.logging2.LoggerImpl;
 import org.jitsi.videobridge.octo.*;
 import org.jitsi.videobridge.shim.*;
 import org.jitsi.videobridge.util.*;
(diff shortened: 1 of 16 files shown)


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -231,21 +235,28 @@ public Conference(Videobridge videobridge,
...
+        Map<String, String> context = new HashMap<>();
+        context.put("confId", id);
+        context.put("gId", gid);
+        if (name != null)
         {
-            logger.setLevel(Level.WARNING);
+            context.put("name", name.toString());
         }
 
+        Level minLevel = enableLogging ? Level.ALL : Level.OFF;
```
bgrozev: I think this should be WARN or ERROR for health check conferences

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -538,13 +549,14 @@ private void updateStatisticsOnExpire()
...
             StringBuilder sb = new StringBuilder("expire_conf,");
-            sb.append(logPrefix)
-                .append("duration=").append(durationSeconds)
+            sb.append("duration=").append(durationSeconds)
                 .append(",conf_completed=")
                     .append(videobridgeStatistics.totalConferencesCompleted)
                 .append(",has_failed=").append(hasFailed)
                 .append(",has_partially_failed=").append(hasPartiallyFailed);
-            logger.info(Logger.Category.STATISTICS, sb.toString());
+            //TODO(brian): bring back cateogry
```
bgrozev: I don't know if the category makes sense anymore. It was briefly used with a fluentd setup a long time ago.

src/main/java/org/jitsi/videobridge/Conference.java
bbaldino: yup you're right, will fix.

src/main/java/org/jitsi/videobridge/Conference.java
bbaldino: so do you think this should be removed completely? or just put on debug?

bgrozev: No, I think it's fine in INFO, I'm just saying we can remove the TODO and not worry about it (I don't think the conf_completed part is needed either).

PR #900, opened 2019-09-04
Logging msg supplier changes

bbaldino:


bbaldino:
hold off on this, as if the map builder helpers go in to jitsi-utils (https://github.com/jitsi/jitsi-utils/pull/14) i'll update this to use them

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -297,15 +285,40 @@ public Conference createConference(
...
         // The method Videobridge.getConferenceCountString() should better
         // be executed outside synchronized blocks in order to reduce the
         // risks of causing deadlocks.
-        if (logger.isInfoEnabled())
-        {
-            logger.info("create_conf, id=" + conference.getID()
-                    + " gid=" + conference.getGid()
-                    + " logging=" + enableLogging);
-        }
+        logger.info(() -> "create_conf, id=" + conference.getID()
```
bgrozev: I wonder if we should care about using suppliers for anything INFO and above. WARN is probably going to get logged, so the performance is the same. If we have enough WARN messages that are not being logged to affect performance we have another problem.

I think the convention so far has been to always require a level check for DEBUG and TRACE (even when it’s a single string), but never require it for INFO and above. But obviously here we're not using this convention. Is it worth all the extra code just to avoid either an `if`, or a few string concatenations when a conference is created?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -116,6 +116,7 @@ public VP8AdaptiveTrackProjectionContext(
         logContext.put("id", Integer.toString(hashCode()));
         this.logger = parentLogger.createChildLogger(VP8AdaptiveTrackProjectionContext.class.getName(), logContext);
         this.payloadType = payloadType;
+        this.vp8QualityFilter = new VP8QualityFilter(parentLogger);
```
bgrozev: Do we not want to pass `logger` here to keep the `id` we just added? Also `id` is generic and might clash with something from the parent's context, maybe rename it? I can't think of a good name though

src/main/java/org/jitsi/videobridge/Videobridge.java
bgrozev: I phrased the above generally, but this is the only place in this PR where it applies.

bbaldino: yeah, i wondered that too, but, for health checks, for example, we do set the level to warn (not info) so there is maybe some difference there (assuming we actually used this form everywhere, which i don't think we do).

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
bbaldino: i'm not sure we actually even need the added 'id' here.  i added it a while back when i found it useful during debugging, but we'll have more context now.  i'm going to remove it and we can re-add it in the future if we need it.

src/main/java/org/jitsi/videobridge/datachannel/RemotelyOpenedDataChannel.java
```diff
@@ -32,10 +33,11 @@
      */
     public RemotelyOpenedDataChannel(
             DataChannelStack.DataChannelDataSender dataChannelDataSender,
+            Logger parentLogger,
```
JonathanLennox: Usually parentLogger is added as the last argument, but here you have it as the second. Any reason?

bbaldino: i haven't necessarily sought to always set it as the last arg (in kotlin especially, where it's best  to put args with default values last) but i can't necessarily  explain why i put it second here. i  don't feel strongly about it.

JonathanLennox: Ok, as you like. As things move to Kotlin the arg order matters less anyway.
LGTM.

src/main/java/org/jitsi/videobridge/Videobridge.java
bgrozev: So why did you leave this in? It goes against the usual convention, it adds an extra method, it has to use a `final` local variable, and you seem to agree that its only use is for health checks, i.e. to avoid 6 string concatenations once every 10 second.

bgrozev: And it was an unresolved comment on the PR.

bbaldino: sorry, i didn't mean to steam roll this in or anything, but i didn't know you had an issue with the helper method (as you didn't mention that) and there was no response to my comment after multiple days.  i don't feel strongly about it so fine with changing it.

PR #902, opened 2019-09-06
Migrate debug interface to jersey

bbaldino:
this is the debug handler ported over to jersey.  i did make a couple tweaks, but i can change them back if you want:

1) explicitly enabling/disabling the debug interface  seemed unnecessary, since we don't do anything other than prevent calls, but i'll change that back if you think it should remain as it was

2) i changed the paths for enabling/disabling packet verification to something that seemed like a slight improvement, but i don't feel strongly about the change


src/main/java/org/jitsi/videobridge/rest2/Conferences.java
```diff
@@ -0,0 +1,89 @@
+/*
```
JonathanLennox: This commit just says "rename variable" but it's also adding this file...

src/main/java/org/jitsi/videobridge/rest/HandlerImpl.java
```diff
@@ -1023,10 +1017,6 @@ else if (target.startsWith(STATISTICS))
             statisticsRequestHandler.handleStatsRequest(
                     target, request, response);
         }
-        else if (target.startsWith(DEBUG))
```
JonathanLennox: With this code gone and DebugRequestHandler gone, is HandlerImpl.DEBUG now no longer needed?

src/main/java/org/jitsi/videobridge/rest2/Conferences.java
bbaldino: good catch, fixed

src/main/java/org/jitsi/videobridge/rest/HandlerImpl.java
bbaldino: yup, fixed

PR #905, opened 2019-09-09
port shutdown handler to jersey

bbaldino:


src/main/java/org/jitsi/videobridge/rest/shutdown/Shutdown.java
```diff
@@ -0,0 +1,79 @@
...
+
+        String ipAddress = request.getHeader("X-FORWARDED-FOR");
+        if (ipAddress == null)
+        {
+            ipAddress = request.getRemoteAddr();
+        }
+        shutdownIq.setFrom(JidCreate.from(ipAddress));
+        videobridgeProvider.get().handleShutdownIQ(shutdownIq);
+
+        return Response.ok().build();
```
JonathanLennox: The previous code had:
```  
              IQ responseIQ	
                    = videobridge.handleShutdownIQ(	
                            requestShutdownIQ);	

                if (IQ.Type.result.equals(responseIQ.getType()))	
                {	
                    status = HttpServletResponse.SC_OK;	
                }	
                else	
                {	
                    status = getHttpStatusCodeForResultIq(responseIQ);	
                }
        ...
        response.setStatus(status);	
```

Should we be handling IQ error responses similarly here?

bbaldino: yep, i think we should...missed this and thought it just threw.  i'll change it.

JonathanLennox:
LEBTM now.

PR #906, opened 2019-09-11
Import codec depacketizers to jitsi-media-transform.

JonathanLennox:


src/main/java/org/jitsi/videobridge/cc/GenericAdaptiveTrackProjectionContext.java
```diff
@@ -73,17 +73,17 @@ private static boolean isKeyframe(
 
         if (payloadType instanceof Vp8PayloadType)
         {
-            return org.jitsi.impl.neomedia.codec.video.vp8.DePacketizer
+            return org.jitsi_modified.impl.neomedia.codec.video.vp8.DePacketizer
```
bbaldino: do you know if we actually need the fully-qualified name here?

JonathanLennox: Yes we do, because the classes are all called DePacketizer, in separate packages.

Arguably this is stupid and should be fixed, but that's the way it was arranged in libjitsi.

bbaldino: as we talked about, we can get rid of this whole thing and just use `VideoRtpPacket#isKeyframe` in the code that's calling this.

PR #909, opened 2019-09-13
Adjust shim's member visibility.

mstyura:
I'm migrating to `JVB 2.0` and I think default visibility of changed members was introduced non-intentionally.
If default visibility was design decision please let me know.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
+++ b/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
@@ -383,7 +383,7 @@ public class ChannelShim
     /**
      * @return This {@link ChannelShim}'s endpoint.
      */
-    public Endpoint getEndpoint()
+    Endpoint getEndpoint()
     {
         return endpoint;
     }
@@ -391,7 +391,7 @@ public class ChannelShim
     /**
      * @return the ID of this channel.
      */
-    public String getId()
+    String getId()
     {
         return id;
     }
--- a/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
+++ b/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
@@ -114,7 +114,7 @@ public class ChannelShim
     /**
      * The source groups that were signaled for this channel.
      */
-    Collection<SourceGroupPacketExtension> sourceGroups;
+    private Collection<SourceGroupPacketExtension> sourceGroups;
 
     /**
      * The expire value for this channel.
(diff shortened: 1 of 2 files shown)


bgrozev:
Jenkins, add to whitelist

src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
```diff
@@ -383,15 +383,15 @@ public void setDirection(String direction)
     /**
      * @return This {@link ChannelShim}'s endpoint.
      */
-    Endpoint getEndpoint()
+    public Endpoint getEndpoint()
```
bgrozev: I think these are intentionally package-private. We don't want to expose any of the inner workings of the shim layer.

Do you need this anywhere? Why?

bgrozev: Channels and Contents only exist in the shim layer. You can probably acomplish what you're trying to do by getting the Conference's endpoints and checking the `acceptAudio` and `acceptVideo` fields which are equivalent to the existance of an audio or video channel for the endpoint.

mstyura: > I think these are intentionally package-private.

@bgrozev thanks for clarification, it was not obvious from code if this intentional or not, because some members annotated public, some - default.

> Do you need this anywhere? Why?

That's due to the way I've used `JVB1.0` in project I'm working on. I had tiny layer on server which translated `SDP` (offer) to Colibri, then Colibri was processed via `Videobridge.handleConferenceIQ`, after that new `SDP` (answer) was generated from the updated state of conference on Videobridge. To obtain state of conference I've used to read public properties of Confent/Channel, which are now moved to shim layer and became implementation details, as I understood.
That's completely fine, I'll try to workaround this API changes. Just was not sure if this was intentional change in `JVB 2.0`. Thanks for help!

BTW, what is the most appropriate and **quick** way to ask for design changes in `JVB 2.0`? So far I've used to create PRs, but that's probably abuse and mostly produces PR which are then rejected after explanation of changes in `JVB 2.0` :)

bgrozev: Hi @mstyura we appreciate your contributions, and it's OK to open PRs even if they don't get merged (I actually find it handy to see the code you're working on this way). But you can also open issues on github or post to community.jitsi.org. I think they're all equally quick.

The part of this PR which makes things `private` is okay.

mstyura: @bgrozev `public` part of PR was just undone.

PR #911, opened 2019-09-13
fix: Fixes the media stream tracks factory tests

bgrozev:
Always sets the SVC flags as intended.


bbaldino:
i think this can be cleaner, but it's a side effect of config stuff which we'll be changing soon anyway, so we can address it then.

PR #913, opened 2019-09-16
Fix unit tests

bbaldino:
noticed that we were specifying a specific test suite, so other tests weren't being run


bbaldino:
jenkins test this please

PR #915, opened 2019-09-17
Adapts to ice4j using jitsi-utils logger

bbaldino:
relies on https://github.com/jitsi/ice4j/pull/186 (pom will need to be updated after that is merged)

### DIFF
--- a/src/main/java/org/jitsi/videobridge/IceTransport.java
+++ b/src/main/java/org/jitsi/videobridge/IceTransport.java
@@ -316,7 +316,7 @@ public class IceTransport
             boolean controlling, String streamName, ConfigurationService cfg)
             throws IOException
     {
-        Agent iceAgent = new Agent(logger.getLevel(), iceUfragPrefix);
+        Agent iceAgent = new Agent(iceUfragPrefix, logger);
 
         //add videobridge specific harvesters such as a mapping and an Amazon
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/IceTransport.java
```diff
@@ -196,10 +196,13 @@
         iceComponent = iceStream.getComponent(Component.RTP);
         iceStream.addPairChangeListener(iceStreamPairChangeListener);
 
+        //TODO(brian): this should be replaced by passing a log context
+        // when creating the child logger, but we don't have this
+        // value yet when creating the child logger, so we need to
+        // change the logger to allow adding context after creation,
+        // then we can get rid of the need for logPrefix here
         this.logPrefix
```
bgrozev: Would it be terrible if we re-initialized the Logger instance when we create the Agent?

bbaldino: yeah i don't think that would work.  i don't think it'll be too bad to enable adding context after a logger's initial creation though

bbaldino:
sorry, have to close this because the branch name didn't match that used in the ice4j PR so the tests don't pass. jitsi/jitsi-videobridge#918 is the new one.

PR #917, opened 2019-09-18
Fill in SSRCs field in channel IQ.

mstyura:
In `JVB 1.0` `conference.describeDeep` produce channel's description with `SSRC` information. 
Example:
```
<channel endpoint="#1" expire="15" id="e352c15176eab2c0" channel-bundle-id="#1" rtp-level-relay-type="translator">
...
    <source xmlns="urn:xmpp:jingle:apps:rtp:ssma:0" ssrc="2481496264" />
    <ssrc>4228755975</ssrc>
...
</channel>
```
In `JVB 2.0` `<ssrc>` attribute is missing.
I don't know if it's intentional or not. Since it might be accidental missing information here is PR which adds it back.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
+++ b/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
@@ -241,6 +241,26 @@ public class ChannelShim
                 bridgeSource.setSSRC(localSsrc);
                 iq.addSource(bridgeSource);
             }
+
+            if (sources != null && !sources.isEmpty())
+            {
+                int[] ssrcs = new int[sources.size()];
+                int i = 0;
+                for (SourcePacketExtension source : sources)
+                {
+                    long ssrc = source.getSSRC();
+                    if (ssrc != -1)
+                    {
+                        ssrcs[i] = (int) source.getSSRC();
+                        i++;
+                    }
+                }
+                if (i < ssrcs.length)
+                {
+                    ssrcs = Arrays.copyOf(ssrcs, i);
+                }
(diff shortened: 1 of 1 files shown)


mstyura:
@bgrozev could you please hit merge button if PR is ok for you. Thanks a lot!

src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
```diff
@@ -241,6 +241,26 @@ public void describe(ColibriConferenceIQ.ChannelCommon commonIq)
                 bridgeSource.setSSRC(localSsrc);
                 iq.addSource(bridgeSource);
             }
+
+            if (sources != null && !sources.isEmpty())
```
bbaldino: could this not have been:
```
if (sources != null) {
    int[] ssrcs = sources.stream()
                    .map(SourcePacketExtension::getSSRC)
                    .mapToInt(Long::intValue)
                    .toArray();

    iq.setSSRCs(ssrcs);
}
```
?
I think that's a clearer

mstyura: @bbaldino Yes, it might be written in that manner, with extra `filter` invocation to ensure there is no `-1`. But I've decided here not to use stream API for that, if you insist I could do the same with stream API in different PR as long as this already merged. Thanks for review, Brian!

bbaldino: Oops, yeah forgot the filter.  Ok, if I make some other changes in the bridge I may tweak that.  Thanks @mstyura

PR #919, opened 2019-09-19
Added endpoint expire event.

mstyura:
In `JVB 1.0` there was an `CHANNEL_EXPIRED_TOPIC`. 
I've used it to find out when endpoint is expired, by handler like this:
```
final void onChannelExpired(Object eventSource, String eventTopic, Event event) {
    Channel channel = (Channel) eventSource;
    if (channel.getEndpoint().isExpired()) {
        // endpoint expiration handling here
    }
}
```
The `CHANNEL_EXPIRED_TOPIC` event was removed from `JVB 2.0` as there is no more the notion of `CHANNEL` in `JVB 2.0` (shim layer is not counted).

AFAIK, `Endpont` is now the primary entity in a conference, so it make sense to have dedicated event for endpoint expiration.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -74,7 +74,7 @@ public class Conference
      * (we iterate over it for each RTP packet) and the Octo endpoints are not
      * needed.
      */
-    private List<Endpoint> endpointsCache = Collections.EMPTY_LIST;
+    private List<Endpoint> endpointsCache = Collections.emptyList();
 
     /**
      * The {@link EventAdmin} instance (to be) used by this {@code Conference}
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -939,8 +939,20 @@ public class Conference
      */
     void endpointExpired(AbstractEndpoint endpoint)
     {
-        if (endpoints.remove(endpoint.getID()) != null)
+        final AbstractEndpoint removedEndpoint;
+        synchronized (endpoints)
         {
+            removedEndpoint = endpoints.remove(endpoint.getID());
+        }
+
+        if (removedEndpoint != null)
+        {
+            final EventAdmin eventAdmin = getEventAdmin();
+            if (eventAdmin != null)
+            {
+                eventAdmin.sendEvent(
+                    EventFactory.endpointExpired(removedEndpoint));
+            }
             updateEndpointsCache();
             endpointsChanged();
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -939,8 +939,20 @@ else if (Endpoint.SELECTED_ENDPOINTS_PROPERTY_NAME
      */
     void endpointExpired(AbstractEndpoint endpoint)
     {
-        if (endpoints.remove(endpoint.getID()) != null)
+        final AbstractEndpoint removedEndpoint;
+        synchronized (endpoints)
```
bbaldino: `endpoints` is already a `ConcurrentHashMap`, so i don't think the synchronized is needed here?

mstyura: I also think it is not necessary, but it is wrapped into `syncronized` in other places, so I thought it's probably missed here :\ Do you think it is not necessary here actually and I can remove it?

bbaldino: hm, i see it's synchronized in some places to allow a broader operation to be atomic with the modification of the map.  yes, i think it should be synchronized here and, after looking, i see at least one bug elsewhere due to operations happening outside of a synchronized block.

bgrozev:
@mstyura broader question: Can you summarize what you use `EventAdmin` events for? The code is broken because we don't use these events for anything AFAIK, and it would be good to understand how other projects use them.

mstyura:
@bgrozev I'm not using event admin instance directly. I have a bundle inherited from `org.jitsi.osgi.EventHandlerActivator`, which subscribes to the following events (at least they were triggered in JVB 1.0):
* EventFactory.ENDPOINT_EXPIRED_TOPIC - new as replacement to expired channel event;
* EventFactory.CONFERENCE_CREATED_TOPIC - don't actually need this now;
* EventFactory.CONFERENCE_EXPIRED_TOPIC
* EventFactory.MSG_TRANSPORT_READY_TOPIC

If there is other way to subscribe to these events let me please know.

mstyura:
To give more context of how JVB is used in project I'm working on, here is a copy of short description from [one of my questions](https://community.jitsi.org/t/webrtc-client-to-jvb-connection-recovery/15418):

> I’m working on native mobile client app which uses WebRTC SDK for mobile 3 under the hood.
I’m also have a tiny server app which translates client requests to create/join/hold/leave conference to corresponding requests to JVB instance.
Tiny server app runs JVB instance within same process, so server has direct access to all public interface of JVB.
Basically what tiny server does - it transforms incoming request with SDP from client to join call to corresponding request to JVB.

bgrozev:
Okay, this makes sense. I think plugging into the API with an external bundle is something we should keep. We will probably clean up more of these events though.

PR #921, opened 2019-09-23
Updates jicoco to latest.

damencho:
Checks whether the incoming iq is from the control room we have.


damencho:
Still cannot figure out, why it fails ... all tests pass, but overall job result is failed :(

PR #922, opened 2019-09-23
Comment fix & tiny sha1 lookup refactoring.

mstyura:


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -194,9 +195,13 @@ else if (!StringUtils.isNullOrEmpty(setup))
         {
             dtlsStack.setRemoteFingerprints(remoteFingerprints);
 
-            String hash = remoteFingerprints.entrySet().iterator().next().getKey();
+            final boolean hasSha1Hash = remoteFingerprints
+                .keySet()
+                .stream()
+                .anyMatch(hash -> hash.equalsIgnoreCase("sha1"));
```
damencho: It used to check for 'sha-1' string and now this is changed to 'sha1', is this intentional?

mstyura: Thanks a lot @damencho, nice catch! It was completely accidental change.

JonathanLennox: To be clear, this changes from checking whether the first fingerprint is sha-1, to checking whether any fingerprint is sha-1?

That's probably the right thing to do, but I wanted to make sure it was intentional.

mstyura: @JonathanLennox yes, it was intentional change. I understand that previous implementation was slightly different, but I suppose it actually meant to check presence of sha-1. BTW, I don't know if it make sense to have more than one fingerprint actually, I suppose there is only 0 or 1 such attribute possible.

JonathanLennox: Yes, you can have multiple fingerprints -- see [RFC 8122](https://tools.ietf.org/html/rfc8122).

The reason is that either you might support multiple hash functions (e.g. both sha-1 and sha-256) or you might not yet know at offer time which certificate you will be using.

mstyura: @JonathanLennox as one of the authors listed in this RFC, you are definitely know better how it supposed to work :) Let me know if this implementation comply to the spec from your point of you.

JonathanLennox: This part seems right.

(A quick inspection indicates that `DtlsStack.kt` isn't correctly handling multiple fingerprints, but that's a separate issue.)

mstyura: @JonathanLennox could you please then approve and merge? Or I need an approval from Brian or Boris as well?

JonathanLennox: It LGTM, but I don't yet have approval rights on the videobridge repo.  We'll need @bbaldino or @bgrozev.

PR #923, opened 2019-09-23
Updates jicoco to latest.

damencho:
Checks whether the incoming iq is from the control room we have.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jicoco</artifactId>
-      <version>1.1-20190509.130302-15</version>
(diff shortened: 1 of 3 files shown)


damencho:
I may see something strange ... here

damencho:
So the thing is that we put by default `org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth.${JVB_HOSTNAME}/.*` in jvb config. That works with the current PR with default components setup, but needs adjusting if mucs are configured. We may consider adjusting it for new configs when we move the use mucs by default and leave it as it is for now.
@bgrozev WDYT?

gpolitis:
@damencho and I went through this and we decided that it's fine to merge this.

PR #924, opened 2019-09-24
Fill proper DTLS setup value when transport is described.

mstyura:
The `setup` field was properly described in `JVB 1.0`, but not in `JVB 2.0`:
https://github.com/jitsi/jitsi-videobridge/blob/838e190c4dcf0ca5b94a4065780a59a82792a744/src/main/java/org/jitsi/videobridge/IceUdpTransportManager.java#L1321-L1327

According to [this document](https://xmpp.org/extensions/xep-0320.html) role spelling must be lower case:
```
<xs:attribute name='setup' use='required'/>
    <xs:simpleType>
        <xs:restriction base='xs:NCName'>
        <xs:enumeration value='active'/>
        <xs:enumeration value='passive'/>
        <xs:enumeration value='actpass'/>
        <xs:enumeration value='holdconn'/>
        <xs:annotation>
            <xs:documentation>
            the 'holdconn' value is not used and included only for completeness.
            </xs:documentation>
        </xs:annotation>
        </xs:restriction>
    </xs:simpleType>
</xs:attribute>
```

### DIFF
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -248,7 +248,11 @@ public class DtlsTransport extends IceTransport
 
         final DtlsRole role = dtlsStack.getRole();
         final String setupRole;
-        if (role instanceof DtlsServer)
+        if (role == null)
+        {
+            setupRole = "actpass";
+        }
+        else if (role instanceof DtlsServer)
         {
             setupRole = "passive";
         }
@@ -258,7 +262,7 @@ public class DtlsTransport extends IceTransport
         }
         else
         {
-            setupRole = "actpass";
+            throw new IllegalStateException("Can not describe role " + role);
         }
         fingerprintPE.setSetup(setupRole);
     }
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -250,6 +250,7 @@ public class DtlsTransport extends IceTransport
         final String setupRole;
         if (role == null)
         {
+            // We've not chosen a role yet, so we can be either
             setupRole = "actpass";
         }
         else if (role instanceof DtlsServer)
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -246,8 +246,21 @@ protected void describe(IceUdpTransportPacketExtension pe)
         fingerprintPE.setFingerprint(dtlsStack.getLocalFingerprint());
         fingerprintPE.setHash(dtlsStack.getLocalFingerprintHashFunction());
 
-        // TODO: don't we only support ACTIVE right now?
-        fingerprintPE.setSetup("ACTPASS");
+        final DtlsRole role = dtlsStack.getRole();
+        final String setupRole;
+        if (role instanceof DtlsServer)
```
bbaldino: i worry the way this is written is perhaps a bit misleading.  if i'm understanding it correctly, this code will return `actpass` if the role is null (meaning we haven't decided yet) or active/passive after it's already been set as client or server, right?  this makes it look like `role` could be an instance of some other class and that results in it being actpass.  i'm just thinking we could be a bit more explicit and do:

```
final DtlsRole role = dtlsStack.getRole();
final String setupRole;
if (role == null)
{
    // We've not chosen a role yet, so we can be either
    setupRole = "actpass";
}
else if (role instanceof DtlsClient)
{
    setupRole = "active";
}
else if (role instanceof DtlsServer)
{
    setupRole = "passive";
}
else
{
    // throw exception?
}
fingerprintPE.setSetup(setupRole);
```

though part of me also wonders if this may be better done as something like `dtlsStack.describeRole()` as well

mstyura:
@bbaldino I've fixed role description according to your suggestion. Regarding to introduction of `DtlsStack.describeRole` method I'm completely agree it's reasonable, but may we postpone it a bit? It will require PR into different repository, then `pom.xml` update in JVB repo... I think it will take some time and I'd like to have this change in JVB soon. I could do all of this of course, but I prefer to do it as an enhancement (refactoring) of current implementation as separate 2 PRs. Thank you very much for review!

PR #926, opened 2019-09-25
Fixed controlling value in `IceTransport` debug state.

mstyura:
Due to ice control role conflict resolution Agent's role might be different from what was used during initialization. It's better to poll current state when debug info object is filled.


mstyura:
Jenkins, test this please

PR #927, opened 2019-09-25
Ability to set ICE role during DtlsTransport creation.

mstyura:
This is the first step in restoring ability to control `JVB 2.0` ICE role during connection establishment.
`JVB 1.0` had ability to be either controlling/controlled agent based on `initiator` field in `IQ`.
Currently I'm suffering from [ICE role conflict issue](https://github.com/jitsi/ice4j/pull/187), but even when it is resolved there still be an issue with Firefox (I'm still investigating that Firefox bug).

So I'd like to have an explicit control over `JVB 2.0` ICE role, so I'll not rely on role conflict resolution implementation. This PR is first step to bringing back configurable ICE role of `JVB`.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -630,52 +630,57 @@ public class Conference
      */
     public AbstractEndpoint getEndpoint(String id)
     {
-        return getEndpoint(id, /* create */ false);
+        return endpoints.get(id);
     }
 
     /**
-     * Gets an <tt>Endpoint</tt> participating in this <tt>Conference</tt> which
-     * has a specific identifier/ID. If an <tt>Endpoint</tt> participating in
-     * this <tt>Conference</tt> with the specified <tt>id</tt> does not exist at
-     * the time the method is invoked, the method optionally initializes a new
-     * <tt>Endpoint</tt> instance with the specified <tt>id</tt> and adds it to
-     * the list of <tt>Endpoint</tt>s participating in this <tt>Conference</tt>.
-     *
-     * @param id the identifier/ID of the <tt>Endpoint</tt> which is to be
-     * returned
+     * Initializes a new <tt>Endpoint</tt> instance with the specified
+     * <tt>id</tt> and adds it to the list of <tt>Endpoint</tt>s participating
+     * in this <tt>Conference</tt>.
+     * @param id the identifier/ID of the <tt>Endpoint</tt> which will be
+     * created
      * @return an <tt>Endpoint</tt> participating in this <tt>Conference</tt>
-     * which has the specified <tt>id</tt> or <tt>null</tt> if there is no such
-     * <tt>Endpoint</tt> and <tt>create</tt> equals <tt>false</tt>
      */
-    private AbstractEndpoint getEndpoint(String id, boolean create)
-    {
-        AbstractEndpoint endpoint;
-        endpoint = endpoints.get(id);
-
-        if (endpoint == null && create)
+    @NotNull
+    public Endpoint createLocalEndpoint(String id) {
+        AbstractEndpoint existingEndpoint = endpoints.get(id);
+        if (existingEndpoint instanceof OctoEndpoint)
         {
-            synchronized (endpoints)
-            {
-                endpoint = new Endpoint(id, this, logger);
-                // The propertyChangeListener will weakly reference this
-                // Conference and will unregister itself from the endpoint
-                // sooner or later.
-                endpoint.addPropertyChangeListener(propertyChangeListener);
-                endpoints.put(id, endpoint);
-
-                updateEndpointsCache();
-            }
-
-            EventAdmin eventAdmin = getEventAdmin();
-            if (eventAdmin != null)
-            {
-                eventAdmin.sendEvent(
-                        EventFactory.endpointCreated(endpoint));
-            }
-
-            endpointsChanged();
+            // It is possible that an Endpoint was migrated from another bridge
+            // in the conference to this one, and the sources lists (which
+            // implicitly signal the Octo endpoints in the conference) haven't
+            // been updated yet. We'll force the Octo endpoint to expire and
+            // we'll continue with the creation of a new local Endpoint for the
+            // participant.
+            existingEndpoint.expire();
+        }
+        else if (existingEndpoint != null)
+        {
+            throw new IllegalArgumentException("Local endpoint with ID = "
+                + id + "already created");
         }
 
+        Endpoint endpoint;
+        synchronized (endpoints)
+        {
+            endpoint = new Endpoint(id, this, logger);
+            // The propertyChangeListener will weakly reference this
+            // Conference and will unregister itself from the endpoint
+            // sooner or later.
+            endpoint.addPropertyChangeListener(propertyChangeListener);
+            endpoints.put(id, endpoint);
+
+            updateEndpointsCache();
+        }
+
+        EventAdmin eventAdmin = getEventAdmin();
+        if (eventAdmin != null)
+        {
+            eventAdmin.sendEvent(
+                    EventFactory.endpointCreated(endpoint));
+        }
+
+        endpointsChanged();
         return endpoint;
     }
 
@@ -808,39 +813,24 @@ public class Conference
 
     /**
      * Gets an <tt>Endpoint</tt> participating in this <tt>Conference</tt> which
-     * has a specific identifier/ID. If an <tt>Endpoint</tt> participating in
-     * this <tt>Conference</tt> with the specified <tt>id</tt> does not exist at
-     * the time the method is invoked, the method initializes a new
-     * <tt>Endpoint</tt> instance with the specified <tt>id</tt> and adds it to
-     * the list of <tt>Endpoint</tt>s participating in this <tt>Conference</tt>.
+     * has a specific identifier/ID.
      *
      * @param id the identifier/ID of the <tt>Endpoint</tt> which is to be
      * returned
      * @return an <tt>Endpoint</tt> participating in this <tt>Conference</tt>
-     * which has the specified <tt>id</tt>
+     * or {@code null} if endpoint with specified ID does not exist in a
+     * conference
      */
-    public Endpoint getOrCreateLocalEndpoint(String id)
+    @Nullable
+    public Endpoint getLocalEndpoint(String id)
     {
-        AbstractEndpoint endpoint = getEndpoint(id, /* create */ true);
-        if (endpoint instanceof OctoEndpoint)
+        AbstractEndpoint endpoint = getEndpoint(id);
+        if (endpoint instanceof Endpoint)
         {
-            // It is possible that an Endpoint was migrated from another bridge
-            // in the conference to this one, and the sources lists (which
-            // implicitly signal the Octo endpoints in the conference) haven't
-            // been updated yet. We'll force the Octo endpoint to expire and
-            // we'll continue with the creation of a new local Endpoint for the
-            // participant.
-            endpoint.expire();
-            endpoint = getEndpoint(id, /* create */ true);
+            return (Endpoint) endpoint;
         }
 
-        if (!(endpoint instanceof Endpoint))
-        {
-            throw new IllegalStateException("Endpoint with id " + id +
-                    " should be an Endpoint instance");
-        }
-
-        return (Endpoint) endpoint;
+        return null;
     }
 
     /**
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -964,19 +964,19 @@ public class Conference
                 "conference = " + endpoint.getConference());
         }
 
-        final AbstractEndpoint overwrittenEndpoint;
+        final AbstractEndpoint replacedEndpoint;
         synchronized (endpoints)
         {
-            overwrittenEndpoint = endpoints.put(endpoint.getID(), endpoint);
+            replacedEndpoint = endpoints.put(endpoint.getID(), endpoint);
             updateEndpointsCache();
         }
 
         endpointsChanged();
 
-        if (overwrittenEndpoint != null)
+        if (replacedEndpoint != null)
         {
             logger.info("Endpoint with id " + endpoint.getID() + ": " +
-                overwrittenEndpoint + " has been overwritten by new " +
+                replacedEndpoint + " has been replaced by new " +
                 "endpoint with same id: " + endpoint);
         }
     }
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -956,6 +956,12 @@ public class Conference
      */
     public void addEndpoint(AbstractEndpoint endpoint)
     {
+        if (endpoint.getConference() != this)
+        {
+            throw new IllegalArgumentException("Endpoint belong to other " +
+                "conference = " + endpoint.getConference());
+        }
+
         final AbstractEndpoint overwrittenEndpoint;
         synchronized (endpoints)
         {
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -639,7 +639,9 @@ public class Conference
      * in this <tt>Conference</tt>.
      * @param id the identifier/ID of the <tt>Endpoint</tt> which will be
      * created
-     * @param iceControlling ICE control role of endpoint
+     * @param iceControlling {@code true} if the ICE agent of this endpoint's
+     * transport will initialized to serve as a controlling ICE agent;
+     * otherwise, {@code false}
      * @return an <tt>Endpoint</tt> participating in this <tt>Conference</tt>
      */
     public Endpoint createLocalEndpoint(String id, boolean iceControlling)
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -639,11 +639,13 @@ public class Conference
      * in this <tt>Conference</tt>.
      * @param id the identifier/ID of the <tt>Endpoint</tt> which will be
      * created
+     * @param iceControlling ICE control role of endpoint
      * @return an <tt>Endpoint</tt> participating in this <tt>Conference</tt>
      */
-    @NotNull
-    public Endpoint createLocalEndpoint(String id) {
-        AbstractEndpoint existingEndpoint = endpoints.get(id);
+    public Endpoint createLocalEndpoint(String id, boolean iceControlling)
+        throws IOException
+    {
+        final AbstractEndpoint existingEndpoint = endpoints.get(id);
         if (existingEndpoint instanceof OctoEndpoint)
         {
             // It is possible that an Endpoint was migrated from another bridge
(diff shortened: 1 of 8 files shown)


bgrozev:
I'm fine with this step, but can you include the rest as well so we can comment of them together?

mstyura:
@bgrozev sure, but before I proceed with implementation I'd like to briefly describe what I'm going to do, so you could tell me in advance if you are agree with proposed implementation:
1. I'm going to rename [`Endpoint.getTransportManager`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L988) to something like `Endpoint.getDtlsTransport`. `getTransportManager` looks like legacy name from `JVB 1.0` which hasn't been renamed during rewrite.
2. I'm going to move construction of `DtlsTransport` from `getTransportManager` (where there is not enough information to construct transport, I don't know whether it should be controlled or controlling) into [`setTransportInfo`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L1068)
3. I'm going to rename `setTransportInfo(IceUdpTransportPacketExtension info)` to something like `initializeDtlsTransport(IceUdpTransportPacketExtension info, boolean controlling)` and create `DtlsTransport` here.
4. I'm gong to [update `VideobridgeShim`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/shim/VideobridgeShim.java#L414) to pass `controlling` into `initializeDtlsTransport`.
5. Inside `Endpoint` there are 3 fields to handle state of `DtlsTransport` initialization: `transportManager`, `transportManagerException` and `transportManagerSyncRoot `. I think I have an idea how to use only one field, which should simplify things a bit.

Please let me know what do you think about these steps? Is there anything you are completely disagree, so I shouldn't even try to implement it?

bgrozev:
Sounds reasonable to me. How are you going to make the decision for the value of `controlling`?

mstyura:
@bgrozev [this commit](https://github.com/jitsi/jitsi-videobridge/pull/927/commits/82f01b62ba9565b6bedf246561e6442389a89940) adds bare minimal ad-hoc implementation over existing code without all the changes I've described in previous message.

The question is what to do with `IQ` which for example have `initiator = true` for video channel and `initiator = false` for video.
AFAIK this should be invalid, because `JVB 2.0` only support bundled media - so all channels should have same transport configuration.
The other question whether `controlling` should be computed from `initiator` (it was computed that way in `JVB 1.0`):
https://github.com/jitsi/jitsi-videobridge/blob/838e190c4dcf0ca5b94a4065780a59a82792a744/src/main/java/org/jitsi/videobridge/Videobridge.java#L938-L966

https://github.com/jitsi/jitsi-videobridge/blob/838e190c4dcf0ca5b94a4065780a59a82792a744/src/main/java/org/jitsi/videobridge/Conference.java#L1324-L1340

From `ICE` spec the controlling `ICE` agent is one who is generated offer, but I'm not sure if there is a special field for this in `IQ` other than `initiator`...

mstyura:
Jenkins test this please.

mstyura:
@bgrozev I've just finished my proposed solution to provide control over ICE role. I understand that changes might look big and scary, I'm really sorry for that. But there changes implement all the steps I've mentioned before. Could you please review them? Thank you very much, Boris!

bgrozev:
Thanks, @mstyura! I'll review later, but the tests fail because of jicofo's unit tests: http://pr-reports.jitsi.org/jitsi-videobridge-PR-test/1245/log.txt

mstyura:
According to lost there is also an evidence, that it's me broken some code. Will investigate tomorrow:
```
java.lang.IllegalStateException: DTLS transport is not yet initialized
	at org.jitsi.videobridge.Endpoint.getDtlsTransport(Endpoint.java:1009)
	at org.jitsi.videobridge.Endpoint.describe(Endpoint.java:1232)
	at org.jitsi.videobridge.shim.ConferenceShim.describeChannelBundles(ConferenceShim.java:126)
	at org.jitsi.videobridge.shim.VideobridgeShim.handleColibriConferenceIQ(VideobridgeShim.java:440)
	at org.jitsi.videobridge.Videobridge.handleColibriConferenceIQ(Videobridge.java:598)
	at org.jitsi.videobridge.FocusControlTest.noFocusControlTest(FocusControlTest.java:167)
```

mstyura:
Hello @bgrozev,
I've now fixed tests I've broken by yesterday's changes.
To do that I had to follow slightly different refactoring plan of [`getTransportManager`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L988) and [`setTransportInfo`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L1068) from what I've described [here](https://github.com/jitsi/jitsi-videobridge/pull/927#issuecomment-535001211).
With latest changes [`setTransportInfo`](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L1096) left untouched. 

From [`getTransportManager`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L988) automatic transport initialization were moved into dedicated method [initDtlsTransport](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L1060), so it necessary to initialize transport in advance with [`initDtlsTransport`](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L1060) before [`getDtlsTransport`](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L988) or [`setTransportInfo`](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L1096) can be called. I've fixed all users of [`getDtlsTransport`](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L988) which were relied on automatic transport initialization.

I know changes might look big, but I really think some things are now more explicit which is good and the complexity of implementation is worth of added ability to configure ICE role.

bbaldino:
Overall I do think it'd be nice to clean this up a bit (being more explicit about when the transport is initialized would be great, but as I think you've found it can be a bit tricky to nail down a deterministic flow there, which is why I think this ended up the way it did)--but it seems like we should be able to accomplish that with fewer changes.  Can we not just nail down the first possible time we  can definitively  initialize the transport and pass the controlling value in there and set the  member in Endpoint?  Code which needs to wait for it to be set can use the future (as before) and code that doesn't care can just use the  member directly (as before).

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -155,18 +149,8 @@
...
-     * The exception thrown from the attempt to initialize the transport manager,
-     * if any.
-     */
-    private IOException transportManagerException = null;
-
-    /**
-     * Synchronizes access to {@link #transportManager}.
-     */
-    private final Object transportManagerSyncRoot = new Object();
+    private final CompletableFuture<DtlsTransport> dtlsTransportFuture
```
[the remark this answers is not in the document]
mstyura: I'll try to explain my motivation behind the change and hope those 2 question will be implicitly answered.

Currently in [`master`](https://github.com/jitsi/jitsi-videobridge/commit/6afe06c99f2e4046bff409e6bb80631330b26a32) initialization of `DtlsTransport` is handled by 4 fields:
1. [`transportManager`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L158)
2. [`transportManagerException`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L164)
3. [`transportManagerSyncRoot`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L169)
4. [`onTransportManagerSet`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Endpoint.java#L129)

I've tried to understand motivation behind current implementation and deduced the following motivation:
1. `transportManager` field is necessary to hold reference to `DtlsTransport` instance once it successfully initialized;
2. `transportManagerException` field is necessary to remember the exception which happen during initialization of `DtlsTransport` is it happen;
3. `transportManagerSyncRoot` since there is basically two fields which handle the state of single logical thing - initialization of `DtlsTransport` - it necessary to synchronize access to both variables, such that only one of them is actually holding a value.
4. `onTransportManagerSet` - serves a kind of event that `DtlsTransport` is successfully initialized, basically it is a completion handler of successful `DtlsTransport` initialization.

With that understanding (which might be incorrect, because it was inferred from code, not from conversation with author of original code) I realized that this four fields is basically reinvent `CompletableFuture`.
`CompletableFuture` could either hold successfully computed value or exception happen during value computation, it may only be set once, it provides thread safety - so no external synchronization required and it may have completion handlers. All this functionality with single instance of `CompletableFuture`!
The only possible drawback of that is that whenever it necessary to read `DtlsTransport` it necessary to examine the state of `CompletableFuture`, but that's easily overcame with special getter, which in the proposed PR is [`tryGetDtlsTransport`](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L199).
On the other hand the explicit necessity to check `CompletableFuture` might be consider as a good thing, because it forces user of `DtlsTransport` to explicitly check the state of `DtlsTransport` initialization. For example, I think I've fixed [one potential NPE](https://github.com/mstyura/jitsi-videobridge/blob/802f7f4b6605a1a472c29916f811fb28990e0ab6/src/main/java/org/jitsi/videobridge/Endpoint.java#L739-L743) due to this.

I hope this explanation will give you more insight on my motivation behind this change and is answered the questions you've raised. 
Thanks a lot for review, Brian!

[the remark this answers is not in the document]
bbaldino: thanks for the great explanation, @mstyura, you've helped refresh my memory a bit and i think your assessment of replacing the current scheme with `CompletableFuture` makes sense.  looking at the old code again, i'm wondering if we can clean some things up here to improve this.  one thing i'm thinking when looking at this is why we can't just set up the transport in the endpoint constructor (and throw if it fails).  this would:
1) remove the ambiguity about when it is set
2) remove the need to 'remember' the thrown exception, as we'd (rightfully) prevent the endpoint from being created
3) remove the error checking in endpoint about retrieving it and it potentially being null

now, i _think_ i've had this thought before and there was a reason we couldn't, but perhaps in 2.0 that reason is now gone.  do you know, based on what you've seen, if that's not possible for some reason? i can look into it as well but would take some time before i'd be able.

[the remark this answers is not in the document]
mstyura: >  do you know, based on what you've seen, if that's not possible for some reason? i can look into it as well but would take some time before i'd be able.

I'll investigate this option tomorrow and let you know the result.

mstyura:
Jenkins test this please.

src/main/java/org/jitsi/videobridge/Endpoint.java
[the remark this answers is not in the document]
mstyura: > i'm wondering if we can clean some things up here to improve this. one thing i'm thinking when looking at this is why we can't just set up the transport in the endpoint constructor (and throw if it fails).

That would be ideal and today I did necessary refactorings to achieve the above stated goal.
It's occurred to be doable in `JVB 2.0`

With latest changes to this PR the following major things has been done:
1. [`DtlsTransport` is now initialized](https://github.com/mstyura/jitsi-videobridge/blob/d8a90777486b03e99109fda1c0a1494dc80aceec/src/main/java/org/jitsi/videobridge/Endpoint.java#L253) inside `Endpoint` constructor with initial `ICE` role specified in advance;
2. [`Conference.getOrCreateLocalEndpoint`](https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/src/main/java/org/jitsi/videobridge/Conference.java#L822) has been replaced with 2 dedicated methods [`getLocalEndpoint`](https://github.com/mstyura/jitsi-videobridge/blob/d8a90777486b03e99109fda1c0a1494dc80aceec/src/main/java/org/jitsi/videobridge/Conference.java#L826) and [`createLocalEndpoint`](https://github.com/mstyura/jitsi-videobridge/blob/d8a90777486b03e99109fda1c0a1494dc80aceec/src/main/java/org/jitsi/videobridge/Conference.java#L648).
3. During `ColibriConferenceIQ` handling all endpoints allocated and initialized [in the single place](https://github.com/mstyura/jitsi-videobridge/blob/d8a90777486b03e99109fda1c0a1494dc80aceec/src/main/java/org/jitsi/videobridge/shim/VideobridgeShim.java#L315);

@bbaldino could you please review latest changes? Thanks a lot in advance!

mstyura:
Jenkins test this please.

bbaldino:
🎉 🎉 🎉 🎉 !
At first glance this looks great...I think this is how the endpoint/transport creation should be (explicit and up-front).  I'll take another pass, but I assume things look ok in your testing?

bbaldino:
The tests are failing because of an issue related to the fact that Jicofo relies on the videobridge as a library for its unit tests:

```
[WARNING] /home/jenkins/jicofo/src/test/java/mock/media/MockVideoMediaStream.java:[64,22] getVisualComponent() in org.jitsi.service.neomedia.VideoMediaStream has been deprecated
[INFO] 19 warnings 
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/jenkins/jicofo/src/test/java/mock/jvb/MockVideobridge.java:[182,39] cannot find symbol
  symbol:   method getOrCreateLocalEndpoint(java.lang.String)
  location: variable conference of type org.jitsi.videobridge.Conference
[INFO] 1 error
```

bbaldino:
If you open a Jicofo PR with the branch name the same as this one and make the fix it'll pick it up for these tests. Then after this goes in we can  update the jvb version dep in jicofo

mstyura:
> I assume things look ok in your testing?

Yes, so far everything looks good! 

> If you open a Jicofo PR with the branch name the same as this one and make the fix it'll pick it up for these tests. Then after this goes in we can update the jvb version dep in jicofo

@bbaldino would you mind if I postpone this until Monday. In Belarus it's already late evening and I'm hardly be able to find time to do it today. Thanks for understanding!

bbaldino:
> @bbaldino would you mind if I postpone this until Monday. In Belarus it's already late evening and I'm hardly be able to find time to do it today. Thanks for understanding!

Sure, np.

mstyura:
Jenkins test this please.

mstyura:
> If you open a Jicofo PR with the branch name the same as this one and make the fix it'll pick it up for these tests. Then after this goes in we can update the jvb version dep in jicofo

I've tried this. Once I've managed to compile and run `jicofo` tests locally with [updated `JVB 2.0`](https://github.com/mstyura/jitsi-videobridge/commit/7db1bc00b11604f4939a19e552c5169adc8910bf) I've found NPE bug in my implementation, but it is fixed with latest commit. All `jicofo` tests are green now when run locally.

Unfortunately [`jicofo` PR](https://github.com/jitsi/jicofo/pull/391) did not help to fix [`jicoco` compilation issue](http://pr-reports.jitsi.org/jitsi-videobridge-PR-test/1261/build-logs/jicofo.txt).
I think that's probably because I'm working with forks of original repositories, obviously I don't have permission to have my branch in original repos. So build server does not see my fork to pick up changes from it.

damencho:
I think the problem is that you are using master branch and not a feature branch, I just ran the jicofo tests to see what happens will it detect and uses both PRs, but I'm pretty sure it will not because of the branch naming.

damencho:
Yeah, I just checked the PR scripts and it skips the PRs with branch master ...

mstyura:
@damencho yes it might be because I use `master` branch (unfortunately there is no known way to me to change branch after PR is created here in github). I've went another way, I adjusted `jicoco` PR such that changes are backward compatible with existing `JVB`. So https://github.com/jitsi/jicofo/pull/391 could be merged earlier than this PR. And then this PR's test could be re-run.

mstyura:
Jenkins test this please.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -630,50 +630,59 @@ public BundleContext getBundleContext()
      */
     public AbstractEndpoint getEndpoint(String id)
     {
-        return getEndpoint(id, /* create */ false);
+        if (id == null)
```
bgrozev: nit: How about `Objects.requireNonNull(id)`?

src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -112,19 +112,11 @@ private static void check(Conference conference)
...
-                throw new NullPointerException("Failed to create an endpoint.");
-            }
-
-            // Trigger the creation of the transport manager.
+            final Endpoint endpoint;
             try
             {
-                endpoint.getTransportManager();
+                endpoint = conference.createLocalEndpoint(
+                    generateEndpointID(), true);
```
bgrozev: I know we're not currently running ICE, but should we have different roles for the two endpoints?

src/main/java/org/jitsi/videobridge/Conference.java
bbaldino: and/or tag `id` with `NotNull`

src/main/java/org/jitsi/videobridge/health/Health.java
mstyura: I don't know what is correct, but you've just described behavior of `JVB 1.0` :) - so I've just added the same behavior to `JVB 2.0`
https://github.com/jitsi/jitsi-videobridge/blob/838e190c4dcf0ca5b94a4065780a59a82792a744/src/main/java/org/jitsi/videobridge/health/Health.java#L141-L143

src/main/java/org/jitsi/videobridge/Conference.java
mstyura: Done. Thank you very much for review!

mstyura:
Jenkins test this please.

bgrozev:
> Jenkins test this please.

You're whitelisted now, so Jenkins will automatically pick this up (I think it polls once every 5 minutes).

PR #928, opened 2019-09-25
Rest cleanup

bbaldino:
Not having dependency injection working was making the jersey code more complicated than it needed to be, so I finally got around to getting that figured out and working.  Once I got that setup, I also wanted to clean up the package structure so that the new REST resources will all contained within a single package.  These changes also required a new method for implementing whether or not some of the resources should be accessible based on config, so added some stuff for that too.

~This code is NOT done yet, it still needs some cleanup...but wanted to put it up so you could see some actual code after our discussion @bgrozev~


src/main/java/org/jitsi/videobridge/rest/root/Application.java
```diff
@@ -14,19 +14,25 @@
...
 
 import org.glassfish.jersey.server.*;
 import org.jitsi.utils.logging2.*;
-import org.jitsi.videobridge.util.*;
+import org.jitsi.videobridge.rest.binders.*;
+import org.jitsi.videobridge.rest.filters.*;
+import org.osgi.framework.*;
 
-public class ShutdownApp extends ResourceConfig
+public class Application extends ResourceConfig
```
bbaldino: NOTE that github got confused here.  `ShutdownApp` was removed entirely and `Application` is new

PR #929, opened 2019-09-27
New config

bbaldino:
This ports most of the `ConfigurationService` uses in JVB over to new config.

Some notes:

- For the most part I tried to just port things as they were, sometimes making notes of potential future improvements. There were one or two places where I  did refactor things a bit though.
- I didn't do constants for the config params.  They're only ever used in a single place, and, unlike the old code, the config *file* holds the documentation on each param (whereas before we had this on the constant for the config  key). Definitely open to other opinions on this one, but those were  my thoughts.
- The  `JvbConfig.getConfig().getConfig("...")` calls feel a little weird to me, I'm going to look at improving that.
- In a couple places where there were clusters of config values used I created a config class and parsed the config into that, I think there are a couple more places which could benefit from this so I may tweak those.

Some remaining TODOs before this is finished:

- [ ] Write some documentation on the new config scheme
- [x] Write a doc which includes a mapping of old config fields to new ones
- [ ] Some method of automatically translating old config fields to new ones
- [ ] More testing (unit tests and some more real calls)

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -58,3 +58,37 @@ Previously, we [read from a set of legacy videobridge properties](https://github
 
 ##### [4] Stats transports
 Previously, stats transports were set via a comma-separated string in the `org.jitsi.videobridge.STATISTICS_TRANSPORT` property.  The `pubsub` stats transport had extra settings which were set by the `org.jitsi.videobridge.PUBSUB_SERVICE` and `org.jitsi.videobridge.PUBSUB_NODE` properties.  The new config defines a `videobridge.stats.transports` property which is a list of stats transport config objects, correlating to the `org.jitsi.videobridge.stats.config.StatsTransportConfig` object.  There is a subclass `org.jitsi.videobridge.stats.config.PubSubStatsTransportConfig` for pubsub which holds the extra data relevant to the pubsub config.  The data in the config value is parsed directly into this config file instances and used in the code to instantiate the stats transports.
+
+## Legacy config
+In some places, JVB may use other Jitsi libraries which rely on the old `ConfigurationService` interface.  Other libraries, which still use the old `ConfigurationService` still rely on these libraries, so we don't want to migrate them to the new config.  For these, JVB defines its own `ConfigurationService` implementation: `LegacyConfigurationServiceShim` which provides the existing interface but reads from the new config.  Known properties that rely on this path are:
+
+##### AbstractJettyBundleActivator
+* `org.jitsi.videobridge.rest.private.jetty.sslContextFactory.keyStorePath`
+* `org.jitsi.videobridge.rest.private.jetty.port`
+* `org.jitsi.videobridge.rest`
+* `org.jitsi.videobridge.rest.jetty.sslContextFactory.keyStorePath`
+* `org.jitsi.videobridge.rest.jetty.port`
+
+##### PublicClearPortRedirectBundleAcivator
+* `org.jitsi.videobridge.rest.jetty.tls.port`
+
+##### BandwidthEstimatoImpl (This will go away when JMT is updated to use new config)
+* `org.jitsi.impl.neomedia.rtp.sendsidebandwidthestimation.BandwidthEstimatorImpl.START_BITRATE_BPS`
+* `org.jitsi_modified.impl.neomedia.rtp.sendsidebandwidthestimation.SendSideBandwidthEstimation.lossExperimentProbability`
+* `org.jitsi_modified.impl.neomedia.rtp.sendsidebandwidthestimation.SendSideBandwidthEstimation.timeoutExperimentProbability`
+*
+
+##### MucClientConfiguration (the usage here can be changed to use `loadFromMap` instead of `loadFromConfigurationService`)
+* All props with perfix `org.jitsi.videobridge.xmpp.user.`
+
+##### OctoRelayService (this will change to use new config)
+* `org.jitsi.videobridge.octo.BIND_ADDRESS`
+* `org.jitsi.videobridge.octo.PUBLIC_ADDRESS`
+* `org.jitsi.videobridge.octo.BIND_PORT`
+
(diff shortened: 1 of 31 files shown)


damencho:
Will this automatically migrate old config to the new config on upgrade?
Will these still work https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/resources/install/debian/postinst#L89?
There are few references of the sc.props file while doing the package. 
Also default installation of jitsi-meet configures jetty in the old config file ... Yeah we can take care of it when start depending on jvb2, but I'm just mentioning all I can think of when seeing the PR.
Ideally I would want to stop using jetty:) and default to nginx and I'm slightly testing some stuff in that direction.

damencho:
But the jetty thing means there will be hundreds of deployments that will upgrade and which will have it and we do not want to break it.

bbaldino:
Yeah, thanks for pointing these out!  

> Will this automatically migrate old config to the new config on upgrade?

This PR  does not yet migrate old to new; I was planning on doing that as the second part of the documentation task, I'll add it to the  TODO list.

> Will these still work
> 
> https://github.com/jitsi/jitsi-videobridge/blob/6afe06c99f2e4046bff409e6bb80631330b26a32/resources/install/debian/postinst#L89
> 
You're just talking about injecting the config values into System.properties?  We can definitely achieve that, but will need to take into account the 'translation' of property names.

> There are few references of the sc.props file while doing the package.

Ok, definitely want to make sure those are covered, though we may be able to phase them in, as the new config library does expose all system properties, just have to verify it's plumbed in correctly.

> Also default installation of jitsi-meet configures jetty in the old config file ... Yeah we can take care of it when start depending on jvb2, but I'm just mentioning all I can think of when seeing the PR.
> Ideally I would want to stop using jetty:) and default to nginx and I'm slightly testing some stuff in that direction.

I think we'll always need some amount of Jetty, right? At least the servlet containers?  But agree that I think it can be simplified and would  like this method as well.

Overall I still consider this a draft as there is lots to consider so wanted to get this up so we  can start talking about it, so thanks for the comments.

damencho:
The `JAVA_SYS_PROPS=\"-Dnet.java.sip.communicator` part is something people have in /etc/jitsi/videobridge/config file and is passed when starting the bridge. So we may need to handle at least those, that by default are there in that config file ... converting them in Main ... or where we handle those properties in the bridge, about changing config folder location and name and the log folder ... 

 -Dnet.java.sip.communicator.SC_HOME_DIR_LOCATION=/etc/jitsi
 -Dnet.java.sip.communicator.SC_HOME_DIR_NAME=videobridge
 -Dnet.java.sip.communicator.SC_LOG_DIR_LOCATION=/var/log/jitsi
 -Djava.util.logging.config.file=/etc/jitsi/videobridge/logging.properties

bbaldino:
Once I get the old->new mappings laid out, I think we could look at 'translating' in main or something.

bbaldino:
So I see a few options as far other libraries used by the bridge (actually I think Jicoco may be the only one?) which rely on `ConfigurationService`:

1. Change the libs to not rely on `ConfigurationService` and instead take in their config in a independent format (e.g. a plain map or config objects defined in the libs themselves which can be passed in by the application, or even System properties if we have to, as we can easily inject those 
 after parsing the new config)
1. Branch the libs into versions which use the new config library
1. Use multiple config files for JVB: the old one stays around with values used by the libs which still rely on it, JVB uses its new config file
1. Support both `ConfigurationService` and the new config library in these libs

Of those, I think options 2 and 3 are non-starters.  Options 1 seems like it would be a decent compromise but #4 may be worth looking into as well as it could give us the ease of integration the new config library provides while not breaking backwards-compatibility (It may be good to abstract the overall usage of config a bit such as not to make things to confusing if we go this route).   

In Jicoco, the users  of `ConfigurationService` are `AbstractJettyBundleActivator`, `ComponentBase`, and `MucClientConfiguration`.  I'll plan on playing with options 1 and 4 a bit this week.  Let me know if you have any thoughts.

JonathanLennox:
For option 4, wouldn't that mean that jicoco, and thus jvb, would still have a dependency on libjitsi?

jmt also still depends on `ConfigurationService`, in `BandwidthEstimatorImpl` and `SendSideBandwidthEstimation`.  I'm planning to remove these in the bandwidth estimation rewrite but that might be a bit further off than this project..

bbaldino:
> For option 4, wouldn't that mean that jicoco, and thus jvb, would still have a dependency on libjitsi?

This is true, but would at least be a step in the right direction.

> jmt also still depends on `ConfigurationService`, in `BandwidthEstimatorImpl` and `SendSideBandwidthEstimation`. I'm planning to remove these in the bandwidth estimation rewrite but that might be a bit further off than this project..

Yeah the plan for JMT will be to move to this new config as well, so those shouldn't be too bad.

bbaldino:
Ok, I think I've come up with something that will work well here and require minimal changes.  The bridge now has its own `ConfigurationService` implementation which it loads as a service (in place of loading the libjitsi one from before).  This service impl will rely on the new config underneath, and will allow us to define the 'legacy' config values in the new config file.  All we need to do is move the `ConfigurationService` interface to `jitsi-utils` (the impl will still stay in libjitsi) and then we can break that dependency.

JonathanLennox:
Do we want to put the non-libjitsi implementation of ConfigurationService in jitsi-utils also?

I'm concerned about things like unit test programs which may need the configuration to (at least) not fail.

bbaldino:
> Do we want to put the non-libjitsi implementation of ConfigurationService in jitsi-utils also?
> 
> I'm concerned about things like unit test programs which may need the configuration to (at least) not fail.

What scenario are you thinking about? A unit test where?

bbaldino:
superseded by https://github.com/jitsi/jitsi-videobridge/pull/934

PR #931, opened 2019-10-02
default the colibri rest resource to enabled

bbaldino:
this change allows a default value to be passed to the EnabledByConfig
annotation and, in the case of colibri resources, sets that default
value to true (enabled)


bbaldino:
The original code (from before the REST changes) is [here](https://github.com/jitsi/jitsi-videobridge/blob/jvb_1.0/src/main/java/org/jitsi/videobridge/rest/RESTBundleActivator.java#L107-L108).  Confirmed that colibri should be enabled by default and shutdown disabled.

PR #932, opened 2019-10-03
limits warnings about packet being dropped at the track projection level to VP8

gpolitis:


bbaldino:
remember to cherry-pick onto master as well (believe this code is the same there?)

gpolitis:
> remember to cherry-pick onto master as well (believe this code is the same there?)

I'll check 👍

PR #938, opened 2019-10-11
Debian: postinst - fix AUTHORIZED_SOURCE_REGEXP

iam-TJ:
Commit ac6c098277 introduces a bug in `resources/install/debian/postinst`
during installation that in some, rare, circumstances could cause extremely
hard to detect issues.

The problem here is that as a regular expression the domain-part separating dots
in JVB_HOSTNAME are interpreted as regular expression "match anything"
symbols so that all these would match:

JVB_HOSTNAME=meet.domain.tld

focus@auth1meet.domain.tld/.*
focus@auth.meet2domain.tld/.*
focus@auth3meet4domain.tld/.*
focus@auth5meet.domain6tld/.*

Signed-off-by: Tj <hacker@iam.tj>

Author:    Tj <hacker@iam.tj>

### DIFF
--- a/resources/install/debian/postinst
+++ b/resources/install/debian/postinst
@@ -97,7 +97,7 @@ case "$1" in
         sed -i 's/\$JVB_EXTRA_JVM_PARAMS//g' $CONFIG
 
         if ! grep -q "org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP" "$NEW_JITSI_CONFIG" ;then
-            echo "org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth.${JVB_HOSTNAME}/.*" >> $NEW_JITSI_CONFIG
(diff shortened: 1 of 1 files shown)


damencho:
So the current code (before your change) produces the following config property:
org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth.example.net/.*
And the following JVB log:
org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth.example.net/.*

After your changes the property is:
org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth\.example\.net/.*
and the log is:
INFO: [13] Videobridge.start#896: Authorized source regexp: focus@auth.example.net/.*

What, I see that your PR does not change anything in the current behavior. 
I think you want to have the following property:
org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth\\.example\\.net/.*

You also need to add additional tests to cover your cases: https://github.com/jitsi/jitsi-videobridge/blob/97aea491ebd0b41b04c6b520652997eabc14d6cf/src/test/java/org/jitsi/videobridge/FocusControlTest.java#L216

bgrozev:

PR #939, opened 2019-10-15
Config phase 1

bbaldino:
This PR includes the JVB property helpers as well as a single property moved over (the health check interval) to make things a bit more concrete.


CONFIG.md
```diff
@@ -0,0 +1,52 @@
...
+
+Moving forward, we'll be using lightbend's [config library]([https://github.com/lightbend/config](https://github.com/lightbend/config)).  Their GitHub page gives lots of details, but below are the changes most relevant to configuring the JVB.
+
+### Configuration sources
+During the transition, we'll continue to read from the original locations (`sip-communicator.properties` and program arguments) for all configuration properties, falling back to the new `reference.conf` file for defaults.  The `reference.conf` file holds defaults for all configuration properties, and shouldn't be modified in a deployment.  Instead, users can put a file named `application.conf` in the classpath: values can be put there to override ones in `reference.conf`.  The lightbend docs do a good job of describing [how configuration is loaded]([https://github.com/lightbend/config#standard-behavior](https://github.com/lightbend/config#standard-behavior)).  Here's a more concrete example of overriding a configuration value for JVB:
```
JonathanLennox: This should mention the `-Dconfig.file` option for the location of `application.conf`.

JonathanLennox: Doesn't `reference.conf` typically live inside a .jar file? If this file is intended to be user-facing rather than developer-facing it seems like mentioning it here at all might be confusing.

CONFIG.md
```diff
@@ -0,0 +1,52 @@
+# Configuration
+
+##### 10/14/19
+We've begun the migration to a new configuration framework.  The 'old' method (the `sip-communicator.properties` file and program arguments) will be supported throughout the transition and the bridge will be fully backwards-compatible with the old config.  No changes are necessary until we pull the plug on the old config completely, and we'll give ample warning before that time.
```
JonathanLennox: I'd be even more generous about how long we'll continue to support the old config file format -- I'm not sure this text should even be threatening to remove it.

That said, I'd suggest adding something to the effect of "new configuration options will only be supported in the new config file."

CONFIG.md
```diff
@@ -0,0 +1,52 @@
...
+    enabled-apis=[
+        {
+            // Here we'll customize
+            name="xmpp"
+            host="my.xmpp-host.com"
+            domain="my-xmpp-domain"
+            secret="my-xmpp-secret"
+            port=4242
+        }
+    ]
```
JonathanLennox: Maybe a comment that this is just an example and this config option isn't actually supported yet in the new config file?

src/main/resources/reference.conf
```diff
@@ -0,0 +1,6 @@
+videobridge {
+  health {
+    # The interval between health checks
+    interval=10 seconds
+  }
+}
```
JonathanLennox: Trailing newline!

src/test/java/org/jitsi/videobridge/health/HealthIntervalPropertyTest.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright @ 2018 - present 8x8, Inc.
```
JonathanLennox: This is a real quibble but brand new files (with no content copied from elsewhere) should probably say 2019 - present.

src/main/java/org/jitsi/videobridge/util/config/JvbConfig.java
```diff
@@ -0,0 +1,121 @@
...
+        commandLineArgs = commandLineArgsSupplier.get();
+
+        logger.info("Loaded complete config: " + config.withFallback(legacyConfig).root().render());
+        if (config.hasPath("videobridge"))
+        {
+            logger.info("Loaded JVB config: " + config.getConfig("videobridge").root().render());
+        }
+        else
+        {
+            logger.info("No new config found");
```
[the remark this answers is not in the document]
bbaldino: agree, fixed this

CONFIG.md
bbaldino: not sure what you mean by 'user facing', but i did want to mention this for developers since it's where the defaults live.  maybe i can make that clearer

CONFIG.md
bbaldino: i do like mentioning new config props will only be in the new file, will add that.

as far as the warning about deprecation...why not set the stage now?  i don't want to coddle people into thinking it'll be around forever, and i give no specific time (just mentioning that it'll be after the transition and 'ample' warning will be given)

CONFIG.md
bbaldino: yeah, this actually is a legit example of a later config, but to be honest i may end up changing this format by the time we get back to it.  i think i'll get rid of this chunk altogether

src/main/resources/reference.conf
bbaldino: 🙃

src/test/java/org/jitsi/videobridge/health/HealthIntervalPropertyTest.java
bbaldino: i haven't updated my template because i've been too lazy to verify that it can be done in a way that won't accidentally update old files as well.   i hope you'll forgive the transgression of leaving this as-is, as without the template i'll have to apply the new behavior manually and therefore inconsistently.

CONFIG.md
JonathanLennox: I mean, the primary audience for this file should be people deploying jvb and writing config files, and for them the exact mechanism by which defaults are defined isn't terribly relevant.

If you want to have a separate section for developers, that's fine, but better to keep them separate.

CONFIG.md
JonathanLennox: Okay, I guess that's fine. I just don't want people to panic thinking that they need to do this tomorrow, but I guess the "ample warning" wording is enough for that.

src/main/java/org/jitsi/videobridge/health/HealthIntervalProperty.java
```diff
@@ -0,0 +1,50 @@
...
+
+
+/**
+ * The property which configures the interval between health
+ * checks.
+ */
+public class HealthIntervalProperty extends AbstractConfigProperty<Integer>
+{
+    protected static final String legacyPropName = "org.jitsi.videobridge.health.INTERVAL";
+    protected static final String propName = "videobridge.health.interval";
```
JonathanLennox: In the old code the _PNAME constants were public. Do we want to make the new propName constants public too?

CONFIG.md
bbaldino: ok, that's seems reasonable. will move to a dev section

bbaldino: actually, thinking about this again i do think it's useful to mention this.  if you were deploying the old code and wanted to know whether or not you had to override a property, you'd probably look up its default value.  the default values used to live in the code, but now people will need to look elsewhere to find them, so i don't think it's useful, even for users just deploying, to be able to find those defaults.

src/main/java/org/jitsi/videobridge/health/HealthIntervalProperty.java
bbaldino: i don't think there's any use for that now, as previously we'd sometimes refer to a pname from multiple places and therefore it needed to have potentially public access...now the single source of truth is the property itself, so i think it's best to hide away the detail of the prop name.

CONFIG.md
JonathanLennox: The easiest way to find them is presumably to look at the logs at startup time, no? Rather than scraping the jar files.

bbaldino: hm, i was picturing someone going to the repo.  because of the json representation in the log it isn't the easiest to browse.  ok, i'll try and come up with a tweak here

bbaldino: hm, also, this would require running the bridge with your config cleared just to see all the default values.

bbaldino:
these will still fail due to a known npe from logcontext.  jonathan has a fix for it in his bwe refactor pr [here](https://github.com/jitsi/jitsi-videobridge/pull/936), so i'll wait until that goes in and rebase on top of it

bbaldino:
Ok, verified that the health interval property is set correctly when:

1) Old config is present and sets the interval value (uses that value from old config)
2) Old config is present and doesn't set the interval value (uses the default in reference.conf)
3) No old config is present (uses the default in reference.conf)
4) No old config is present, overrides file is passed (uses value in overrides file)

PR #945, opened 2019-10-23
Propagate selected/pinned events across all bridges.

gpolitis:
This work is a step towards interbridge last-n. By propagating
selected/pinned events across all bridges, every bridge in a particular
call is aware of the selected and pinned participants in every other
bridge, hence making it possible to determine whether a participant is
_globally_ off-stage or on-stage.

If the participant is _globally_ on-stage, her media needs to be
broadcasted across all bridges (in addition to the any participants that
are in the last-n set, of course). Otherwise (i.e. the participant is
off-stage), the bridge can notify the participant to suspend their SD/HD
layers.


JonathanLennox:
I'm not sure why pinning needs to be propagated to all bridges?  If I, a user connected to a bridge in us-east, pin you, a user on eu-central, why does the bridge at us-west need to know or care?

gpolitis:
> I'm not sure why pinning needs to be propagated to all bridges? If I, a user connected to a bridge in us-east, pin you, a user on eu-central, why does the bridge at us-west need to know or care?

We don't practice participant pinning nowadays, but in the past it was possible to chose multiple participants that you always wanted to view in the thumbnails. For the purposes of interbridge last-n, these participants need to be forwarded, in addition to the selected participants, in addition to the last-n participants.

JonathanLennox:
Right, sure, but if I pin participants X, Y, and Z, then X's bridge needs to know to forward X to my bridge, and Y's bridge needs to know to forward Y, etc.  But X's bridge doesn't need to know or care that I pinned Y or Z, and vice-versa.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -250,25 +212,21 @@ protected void onSelectedEndpointsChangedEvent(
...
-                newSelectedEndpoints.add((String)endpointId);
-            }
+            logger.warn(
+                "Unable to relay a JSON object, the conference is null or expired.");
+            return;
         }
 
-        endpoint.selectedEndpointsChanged(newSelectedEndpoints);
+        // Notify Cthulhu and its minions about selected/pinned events.
+        conference.sendMessage(jsonObject.toString(), Collections.EMPTY_LIST, true);
```
bbaldino: i'm confused by this: it's in `EndpointMessageTransport`, but only seems to handle the octo case?  shouldn't this be logic in `OctoEndpoindMessageTransport`? or, at least, some other logic to forward to the local endpoints?

gpolitis: Thanks Brian, you're right, my code is incomplete; apparently I forgot several super calls super.onXXXEndpointChangedEvent(). I'll update the PR.

I'm still going to add a short explanation here just for completeness and to make sure we agree on this approach and I haven't missed anything.

The code in `EndpointMessageTransport` executes when a message is received by a local endpoint, and so in that case we want to broadcast the message to all the participating bridges.

The code that actually does _stuff_ is moved to `AbstractEndpointMessageTransport` and its shared between local and remote endpoint.

There isn't anything special to do in `OctoEndpoindMessageTransport`.

This PR will in particular fix the `incrementSelectedCount` and `decrementSelectedCount` methods to track inter-bridge counts, so we can try enabling the off-stage participant layer suspension.

bbaldino: i just thought that the logic to transmit over octo 'datachannel' vs an endpoint datachannel was different, so was thinking there would still need to be some specific logic in `OctoEndpointMessageTransport`?

gpolitis: i had misunderstood how octo messaging works and missed the fact that `AbstractEndpointMessageTransport.endpoint` is null in the case of octo.. d5066f90f14bc399fa356d7bdb13ea0adf8b13c7 adds the octo specific logic.

I was thinking that for message propagation we could wrap the original message in an envelope and add the destination endpoint id as a property of the envelope, instead of adding it as a property to the original message (which is what I do now).

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -34,6 +34,12 @@
  */
 public abstract class AbstractEndpointMessageTransport
 {
+    /**
+     * The name of the JSON property that indicates the target Octo endpoint id
+     * of a propagated JSON message.
+     */
+    public final String PROP_TARGET_OCTO_ENDPOINT_ID = "targetOctoEndpointId";
```
bbaldino: if this is octo-specific, why not put it in `OctoEndpointMessageTransport`?

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -169,106 +169,49 @@ private void sendMessage(
...
-        JSONArray jsonArray = (JSONArray) o;
-        Set<String> newSelectedEndpoints = new HashSet<>();
-        for (Object endpointId : jsonArray)
-        {
-            if (endpointId != null && endpointId instanceof String)
-            {
-                newSelectedEndpoints.add((String)endpointId);
-            }
-        }
+        jsonObject.put(PROP_TARGET_OCTO_ENDPOINT_ID, endpoint.getID());
```
bbaldino: if this is propagating to octo, why is it in `EndpointMessageTransport` and not `OctoEndpointMessageTransport`?

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
gpolitis: It's actually common between `EndpointMessageTransport` and `OctoEndpointMessageTransport`: When the `EndpointMessageTransport` receives a message from an actual endpoint, it propagates it to all bridges where `OctoEndpointMessageTransport` intercept it.

When the message is propagated, `EndpointMessageTransport` puts the target endpoint id in the message and when the message is intercepted, `OctoEndpointMessageTransport` gets the target endpoint id from the message.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
gpolitis: same answer as above

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
bbaldino: got it, i didn't realize this was about propagating from the ingress to egress. thanks.

gpolitis:
jenkins, test this please

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -71,6 +87,70 @@
...
+        {
+            this.pinnedEndpoints = newPinnedEndpoints;
+
+            if (logger.isDebugEnabled())
+            {
+                logger.debug("Pinned "
+                    + Arrays.toString(pinnedEndpoints.toArray()));
+            }
+
+            // bitrateController.setPinnedEndpointIds(pinnedEndpoints);
```
bbaldino: should this be commented out? looks like it wasn't before...i'm assuming we don't have access to that here (in `AbstractEndpoint`), so maybe we'll still need some overridable method in the sub-classes so `Endpoint` can do this?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -71,6 +87,70 @@
...
+        {
+            this.selectedEndpoints = newSelectedEndpoints;
+
+            if (logger.isDebugEnabled())
+            {
+                logger.debug("Selected "
+                    + Arrays.toString(selectedEndpoints.toArray()));
+            }
+
+            // bitrateController.setSelectedEndpointIds(
```
bbaldino: same as above

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -38,6 +38,22 @@
 public abstract class AbstractEndpoint extends PropertyChangeNotifier
     implements EncodingsManager.EncodingsUpdateListener
 {
+    /**
+     * The name of the <tt>Endpoint</tt> property <tt>pinnedEndpoint</tt> which
+     * specifies the JID of the currently pinned <tt>Endpoint</tt> of this
```
bgrozev: s/JID/ID/

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -38,6 +38,22 @@
...
+     * <tt>Endpoint</tt>.
+     */
+    public static final String PINNED_ENDPOINTS_PROPERTY_NAME
+        = Endpoint.class.getName() + ".pinnedEndpoints";
+
+    /**
+     * The name of the <tt>Endpoint</tt> property <tt>selectedEndpoint</tt>
+     * which specifies the JID of the currently selected <tt>Endpoint</tt> of
```
bgrozev: s/JID/ID/

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: Thanks for catching this, Brian. I just pushed a fix.

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: Thanks Boris.

gpolitis:
@brianh5 could you please take another look? /cc @bgrozev @JonathanLennox

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -102,8 +102,11 @@
     /**
      * Sets the list of pinned endpoints for this endpoint.
      * @param newPinnedEndpoints the set of pinned endpoints.
+     * @return true if the underlying set of pinned endpoints has changed, false
+     * otherwise. The return value was introduced to enable overrides to
+     * act upon the underlying set changing.
      */
-    public void pinnedEndpointsChanged(Set<String> newPinnedEndpoints)
+    public boolean pinnedEndpointsChanged(Set<String> newPinnedEndpoints)
```
bbaldino: Isn't the purpose of the `PINNED_ENDPOINTS_PROPERTY_NAME` event already to notify when the list of pinned endpoints has changed?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -117,18 +120,23 @@ public void pinnedEndpointsChanged(Set<String> newPinnedEndpoints)
...
 
     /**
      * Sets the list of selected endpoints for this endpoint.
      * @param newSelectedEndpoints the set of selected endpoints.
+     * @return true if the underlying set of selected endpoints has changed,
+     * false otherwise. The return value was introduced to enable overrides to
+     * act upon the underlying set changing.
      */
-    public void selectedEndpointsChanged(Set<String> newSelectedEndpoints)
+    public boolean selectedEndpointsChanged(Set<String> newSelectedEndpoints)
```
bbaldino: Same comment as above

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: @brianh5 that thought crossed my mind and I can definitely leverage this functionality if you think it's cleaner, i.e. have the Endpoint instance register for events generated by its super. Is that what you had in mind? The only disadvantage that I see is that it's going to be slightly less efficient.

bbaldino: Actually, this way may take even longer to see the result than handling the event, since the event is fired/handled synchronously I think, isn't it?

gpolitis: You're right about that @brianh5, yeah. However I am noticing that there aren't that many places where we use these events. The pinned event is used nowhere.. so we could potentially axe it. And the selected event is only used to update the counters for the off-stage participant high-layer suspension functionality.

Do we have a new way of doing events? If so I can migrate the selected event to that new way without too much effort (and also remove the pinned event which is defunct).

bbaldino: We don't have any way of doing events, but I do think this method is nice--at least I wouldn't want us to rip it out without taking a deeper look there.

gpolitis:
@brianh5 could you please re-approve? i had to merge with master, and lost the approved status

gpolitis:
jenkins, test this please

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -72,6 +88,71 @@
...
+        {
+            this.pinnedEndpoints = newPinnedEndpoints;
+
+            if (logger.isDebugEnabled())
+            {
+                logger.debug("Pinned "
+                    + Arrays.toString(pinnedEndpoints.toArray()));
+            }
+
+            firePropertyChange(PINNED_ENDPOINTS_PROPERTY_NAME,
```
bgrozev: No changes expected, since this has been open for a long time and I never commented, but FYI: I've been trying to remove the usage of `PropertyChangeListener`, as it can be quite confusing (you have to do a string search to find usages, the value parameter is not typed, etc). I think an explicit method that Endpoint would override is a better option.

PR #946, opened 2019-10-23
Remove use of the Sigar library from Jitsi-videobridge

JonathanLennox:
This takes us a step towards not having non-jar's in the jvb distribution, which will simplify packaging.


JonathanLennox:
Tests on my Mac show that the "cpu_usage" stat in http://localhost:8080/colibri/stats updates properly after this change.  (This is the only consumer of Sigar APIs in JVB.)

JonathanLennox:
From git history it looks like @hristoterezov also worked on this, albeit more than five years ago.

bbaldino:
Also, I'm wondering at this point if we even need this at all? How is this different from pulling the CPU usage from the system (which we already do)?

damencho:
Yeah, I'm not sure I was doing something major with this. We were getting some complaints from security people about it, cause we were using an older version with some security vulnerabilities but in the end the way we use it was not exposing anything at risk.
I agree with Brian that we can rip this off entirely as there are more natural ways nowadays to get that stat.

JonathanLennox:
The stat is exposed in the stats REST API, and then in our datadog -- see https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1572268397397&to_ts=1572271997397&tile_size=m&exp_metric=jitsi.JVB.cpu_usage&exp_scope=environment%3Abeta-meet-jit-si&exp_group=host&exp_agg=max&exp_row_type=metric . So my inclination is to move to the native Java API to retrieve it rather than removing it entirely.

bbaldino:
yeah but from what i can tell this adds nothing over the system cpu stat already exposed in datadog?  @aaronkvanmeerten any thoughts?

bbaldino:
lgtm, but, i want to double check with @aaronkvanmeerten it's ok to remove the stats names completely (as opposed to a period where we hard-code them to 0 or something) in case any scripts will choke.

bbaldino:
aaron says we're all good on the scripts

PR #947, opened 2019-10-23
Optimize vp8 frame projection map

JonathanLennox:
This will fix the performance bottleneck from clearing old entries from the VP8 frame projection map.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
+++ b/src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
@@ -43,6 +43,13 @@ public class VP8AdaptiveTrackProjectionContext
     implements AdaptiveTrackProjectionContext
 {
     private final Logger logger;
+
+    /**
+     * The time (in clock rate samples) to wait before discarding a
+     * non-fully-projected frame (see {@link #cleanupFrameProjectionMap}).
+     */
+    private final long WAIT_SAMPLES = 45000;
+
     /**
      * A map of partially transmitted {@link VP8FrameProjection}s, i.e.
      * projections of VP8 frames for which we haven't transmitted all their
@@ -58,11 +65,14 @@ public class VP8AdaptiveTrackProjectionContext
      * Recovered packets of fully transmitted frames (this can happen for
      * example when the sending endpoint probes for bandwidth with duplicate
      * packets over the RTX stream) are dropped as they're not needed anymore.
-     *
-     * TODO fine tune the ConcurrentHashMap instance to improve performance.
      */
-    private final Map<Long, VP8FrameProjection>
-        vp8FrameProjectionMap = new ConcurrentHashMap<>();
+    private final SortedMap<Long, VP8FrameProjection>
+        vp8FrameProjectionMap = new ConcurrentSkipListMap<>(
+            /* This is only a valid Comparitor if timestamp diffs of all
+             * timestamps are within half the number space.  (This property is
+             * assured by #cleanupFrameProjectionMap.)
+             */
+            (t1, t2) -> (int)RtpUtils.getTimestampDiff(t1, t2));
 
     /**
      * A map that stores the maximum sequence number of frames that are not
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -179,6 +189,52 @@ public VP8AdaptiveTrackProjectionContext(
...
+     * more than #WAIT_SAMPLES older than the newest timestamp.
+     *
+     * Because the map is ordered, we can stop searching entries when we reach
+     * one that is new enough not to be removed, so this is amortized constant
+     * time per frame.
+     *
+     * Caller should be synchronized.
+     * */
+    private
+    void cleanupFrameProjectionMap(long newTimestamp)
```
bbaldino: nit: don't put the private on its own line

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -179,6 +189,52 @@ public VP8AdaptiveTrackProjectionContext(
...
+            return;
+        }
+
+        long threshold = RtpUtils.applyTimestampDelta(newTimestamp, -WAIT_SAMPLES);
+
+        Iterator<Long> it = vp8FrameProjectionMap.keySet().iterator();
+        while (it.hasNext())
+        {
+            Long key = it.next();
+            if (RtpUtils.isNewerTimestampThan(threshold, key))
```
bbaldino: i'd expect this to check if it's older than the threshold not newer?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -255,17 +311,26 @@ VP8FrameProjection createVP8FrameProjection(
             return null;
         }
 
+        if (lastVP8FrameProjection.getSSRC() != nextVP8FrameProjection.getSSRC())
+        {
+            /* We've changed the encoding we're forwarding, so previous frames
+             * in the map are no longer relevant.
+             */
+            vp8FrameProjectionMap.clear();
```
bbaldino: i can't remember, is the logic which handles the layer switching elsewhere? that is, we don't have to worry about 'finishing' a frame of a previous layer here, and therefore keeping its projection around, when we see a new ssrc?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
JonathanLennox: It's checking whether the threshold is newer than the key, rather than if the key is older than the threshold - so it's correct but confusing. I'll reverse it.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
JonathanLennox: The logic which handles layer switching is in `VP8QualityFilter`, which is why I have to do things in this somewhat indirect way.

We can only switch layers on a keyframe, so it's safe to abandon frames from before the switch. In some cases (e.g. dropping to a lower bandwidth layer due to bandwidth constraints) it's even helpful to do so.

bbaldino: ok, i thought there was some scenario which was like "we want to switch to this, but need to get the full keyframe" or something and there was a transition where we needed to keep both alive at once, but my memory is fuzzy so don't know for sure.

JonathanLennox: As far as I can tell, frames are only put in the projection map once they're actually ready to be routed, which implies we're actually switching.  But I'll double-check the flow.

bbaldino:
jenkins, test this please

PR #956, opened 2019-10-30
Add REST API for task pool stats.

JonathanLennox:


src/main/java/org/jitsi/videobridge/util/TaskPools.java
```diff
@@ -43,6 +44,37 @@
...
+        {
+            ThreadPoolExecutor ex = (ThreadPoolExecutor)es;
+            debugState.put("pool_size", ex.getPoolSize());
+            debugState.put("active_count", ex.getActiveCount());
+            debugState.put("completed_tasks", ex.getCompletedTaskCount());
+            debugState.put("core_pool_size", ex.getCorePoolSize());
+            debugState.put("maximum_pool_size", ex.getMaximumPoolSize());
+            debugState.put("largest_pool_size", ex.getLargestPoolSize());
+            debugState.put("queue_class", ex.getQueue().getClass().getSimpleName());
+            debugState.put("queue_size", ex.getQueue().size());
```
bbaldino: i'm wondering if we should use the more 'understandable' names for our stats here instead of matching the ones in the executor.

from our conversation, i think the ones i found confusing were just: 
`queue_size` -> `pending_job_count`
`active_count` -> `active_job_count`

PR #958, opened 2019-10-30
Health config

bbaldino:
This ports the remaining config properties used in `Health` over to new config


CONFIG.md
```diff
@@ -41,3 +41,5 @@ This list will be updated as properties are migrated:
 | Old property name | New property name | Notes |
 | -------- | ------- | ------- |
 | org.jitsi.videobridge.health.INTERVAL | videobridge.health.interval | The new config models this as a duration, rather than an amount of milliseconds |
+| org.jitsi.videobridge.health.TIMEOUT | videobridge.health.timeout | The new config models this as a duration, rather than an amount of milliseconds |
+| org.jitsi.videobridge.health.STICKY_FAILURES | videobridge.health.sticky-failures | |
```
JonathanLennox: Do we want to have something in this readme that says what the valid values are? I assume this is a boolean?

src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -272,23 +234,9 @@ private static String generateEndpointID()
...
-
-        timeout =
-            cfg == null ? TIMEOUT_DEFAULT
-                : cfg.getInt(TIMEOUT_PNAME, TIMEOUT_DEFAULT);
-
-        stickyFailures
-            = cfg == null ? STICKY_FAILURES_DEFAULT
-                : cfg.getBoolean(
-                    STICKY_FAILURES_PNAME, STICKY_FAILURES_DEFAULT);
+        super(videobridge, HealthConfig.HealthIntervalProperty.getValue(), true);
```
JonathanLennox: Shouldn't this use `HealthConfig.getInterval()`?

src/test/java/org/jitsi/videobridge/health/config/HealthTimeoutPropertyTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+        HealthConfig.HealthTimeoutProperty healthTimeoutProp = new HealthConfig.HealthTimeoutProperty();
+
+        Config changedConfig = ConfigFactory.parseString(HealthConfig.HealthTimeoutProperty.propName + "=90 seconds");
+        JvbConfig.configSupplier = () -> changedConfig;
+        JvbConfig.reloadConfig();
+
+        assertEquals(10000, (int)healthTimeoutProp.get());
+    }
+}
```
JonathanLennox: I feel like these property tests replicate a lot of the same test code.  Can this be abstracted or templatized in any way?

CONFIG.md
bbaldino: the intention here was just to document migration-related data (so just the mappings of old to new and any applicable notes), not a full documentation of all the properties

src/main/java/org/jitsi/videobridge/health/Health.java
bbaldino: yes, it should, nice catch.  maybe it'd be best to make these classes protected (so just the tests can access it to test directly)

src/test/java/org/jitsi/videobridge/health/config/HealthTimeoutPropertyTest.java
bbaldino: well, i added the helper for test setup, but everything that's left is pretty much what changes from test-to-test.  was there something that specifically stood out?

CONFIG.md
JonathanLennox: Ok.

I wonder if we _should_ have full documentation of config properties (so in particular we can distinguish between ones which are documented and which ones we consider internal) but that's a separate issue.

src/test/java/org/jitsi/videobridge/health/config/HealthTimeoutPropertyTest.java
JonathanLennox: Just that they felt fairly similar, I didn't read them in so much detail.

JonathanLennox: I don't know if it's possible to have a meta-test, but something like "here's the old property name, here's the new property name, here's an appropriate value if you're setting the old property, here's an appropriate value if you're setting the new property -- run through the full set of all the tests."

bbaldino: yeah that's an interesting idea...i'll take a shot at that

src/test/java/org/jitsi/testutils/ConfigPropertyTest.java
```diff
@@ -0,0 +1,123 @@
...
+package org.jitsi.testutils;
+
+import com.typesafe.config.*;
+import org.jitsi.utils.config.*;
+
+import java.util.function.*;
+
+import static org.junit.Assert.assertEquals;
+
+public class ConfigPropertyTest<T extends ConfigProperty, U>
```
JonathanLennox: Would `T extends ConfigProperty<U>` be valid Java syntax?  Would it be helpful?  (Are there any cases where it wouldn't be valid?)

src/test/java/org/jitsi/testutils/ConfigPropertyTest.java
```diff
@@ -0,0 +1,123 @@
...
+        ParamResult<U> newParamResult,
+        Supplier<T> propInstanceSupplier
+    )
+    {
+        // With the provided arguments, we test each of the following scenarios:
+        // Old config is present, but doesn't provide a value
+        // Old config isn't present
+        // Old config is present and provides a value
+        // NOTE: it's assumed that a new config file (in the form of the defaults)
+        // will always be present, as it is bundled with the JVB itself
```
JonathanLennox: It seems to me like it would be a good thing to test that the new config provides a value, to make sure that we don't forget to include it in the defaults file.

src/test/java/org/jitsi/testutils/ConfigPropertyTest.java
bbaldino: yeah, i think that would be better and does work. fixed.

src/test/java/org/jitsi/testutils/ConfigPropertyTest.java
bbaldino: well we're not using the actual config file here anyway, so those tests to me felt unnecessary.

CONFIG.md
bbaldino: yeah i agree, i have been trying to make the defaults file as self-documented as possible (with comments) in hopes we can avoid the need for a separate doc, so maybe a naming convention to separate internal from external?

src/test/java/org/jitsi/testutils/ConfigPropertyTest.java
JonathanLennox: Hm, maybe not as part of this, but a separate test that the param has a default value would probably be a good thing to add to the standard set of tests.

CONFIG.md
JonathanLennox: Sounds reasonable to me.  (Let's decide on a naming convention when we get to something we think should be internal.)

src/test/java/org/jitsi/testutils/ConfigPropertyTest.java
bbaldino: we'd have to load `reference.conf` manually, i think (as it wouldn't be on the classpath for the test by default, i think) and then we could probably do something like the [colibri resource test](https://github.com/jitsi/jitsi-videobridge/blob/master/src/test/java/org/jitsi/videobridge/rest/root/colibri/ColibriResourceTest.java) and find all the properties via reflection and then call `get` on them and make sure nothing throws.  i dunno if we'd get any false positives there, but could definitely try it out and see how it goes

src/test/java/org/jitsi/videobridge/health/config/HealthTimeoutPropertyTest.java
bbaldino: put up something for this...i think it turned out pretty good, should make testing a lot easier.

src/test/java/org/jitsi/testutils/ConfigPropertyTest.java
bbaldino: i'll take a shot at this in another PR

PR #959, opened 2019-10-31
`Endpoint.selectedCount` made final

mstyura:
The `selectedCount` fields is assigned only once and can be made final.


mstyura:
Decided to add this as part of different MR.

PR #960, opened 2019-10-31
Updateable media direction

mstyura:
The `JVB 1.0` [allowed to update media direction of `RTPChannel`](https://github.com/jitsi/jitsi-videobridge/blob/c29af045ca7081dce457ebd25a025248cfda3a67/src/main/java/org/jitsi/videobridge/RtpChannel.java#L1381-L1399) during it's lifetime either directly or [via colibri](https://github.com/jitsi/jitsi-videobridge/blob/c29af045ca7081dce457ebd25a025248cfda3a67/src/main/java/org/jitsi/videobridge/Videobridge.java#L953-L954) request.

This ability was used in a project I'm working on to implement hold functionality: leave connection between client and JVB, but do not forward media to a client.
I hope it was not intentional design decision to remove such functionality from `JVB 2.0`, so I've implemented it for `JVB 2.0`.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -1325,20 +1325,15 @@ public class Endpoint
         boolean acceptVideo = false;
         for (ChannelShim channelShim : channelShims)
         {
-            String direction = channelShim.getDirection();
-            if (direction != null)
+            if (channelShim.allowsReceivingMedia())
             {
-                if ("sendrecv".equalsIgnoreCase(direction) ||
-                    "recvonly".equalsIgnoreCase(direction))
+                if (MediaType.AUDIO == channelShim.getMediaType())
                 {
-                    if (MediaType.AUDIO == channelShim.getMediaType())
-                    {
-                        acceptAudio = true;
-                    }
-                    else if (MediaType.VIDEO == channelShim.getMediaType())
-                    {
-                        acceptVideo = true;
-                    }
+                    acceptAudio = true;
+                }
+                else if (MediaType.VIDEO == channelShim.getMediaType())
+                {
+                    acceptVideo = true;
                 }
             }
         }
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -1290,7 +1290,7 @@ public class Endpoint
     {
         if (channelShims.add(channelShim))
         {
-            refreshMediaDirection();
+            updateAcceptedMediaTypes();
         }
     }
 
@@ -1308,18 +1308,18 @@ public class Endpoint
             }
             else
             {
-                refreshMediaDirection();
+                updateAcceptedMediaTypes();
             }
         }
     }
 
     /**
-     * Refresh media direction based on direction of
-     * {@link ChannelShim}. An endpoint can accept
+     * Update accepted media types based on directions of
+     * {@link ChannelShim}s. An endpoint can accept
      * media, if its {@link ChannelShim} has either
      * 'sendrecv' or 'recvonly' media direction.
      */
-    public void refreshMediaDirection()
+    public void updateAcceptedMediaTypes()
     {
         boolean acceptAudio = false;
         boolean acceptVideo = false;
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -155,23 +155,24 @@ public class Endpoint
     private final Transceiver transceiver;
 
     /**
-     * The list of {@link ChannelShim}s associated with this endpoint. This
+     * The set of {@link ChannelShim}s associated with this endpoint. This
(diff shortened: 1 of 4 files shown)


mstyura:
Jenkins test this please.

mstyura:
Comment on implementation: I've decided to remove `synchronized` block around `channelShims`, because there was only two places where it was used under `synchronized`, but many other places accessed `channelShims` as is probably causing runtime exception, when modification is done during iteration.

mstyura:
Hello @bbaldino, could you please have a look at this PR? Thanks a lot in advance!

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1300,24 +1300,50 @@ public void addChannel(ChannelShim channelShim)
      */
     public void removeChannel(ChannelShim channelShim)
     {
-        synchronized (channelShims)
+        if (channelShims.remove(channelShim))
         {
-            switch (channelShim.getMediaType())
+            if (channelShims.isEmpty())
```
bbaldino: the `remove`/`isEmpty` check/accept video/audio before would've been atomic and now it's not.  are we sure that's not still needed?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1300,24 +1300,50 @@ public void addChannel(ChannelShim channelShim)
...
 
-            channelShims.remove(channelShim);
-            if (channelShims.isEmpty())
+    /**
+     * Refresh media direction based on direction of
+     * {@link ChannelShim}. An endpoint can accept
+     * media, if its {@link ChannelShim} has either
+     * 'sendrecv' or 'recvonly' media direction.
+     */
+    public void refreshMediaDirection()
```
bbaldino: `refreshMediaDirection` seems like a confusing name for this method since what's being modified is whether or not we accept audio/video (not the directions themselves).  `refreshAcceptedMediaTypes` maybe?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1300,24 +1300,50 @@ public void addChannel(ChannelShim channelShim)
...
+    {
+        boolean acceptAudio = false;
+        boolean acceptVideo = false;
+        for (ChannelShim channelShim : channelShims)
```
bbaldino: what if we add another helper to `ChannelShim`: `supportsReceivedMedia` that does the `sendrecv`/`recvonly` check, and then do:
```
this.acceptAudio = channelShims.stream().filter(ChannelShim::supportsReceivingMedia).anyMatch(c -> c.getMediaType() == MediaType.AUDIO)
this.acceptVideo = channelShims.stream().filter(ChannelShim::supportsReceivingMedia).anyMatch(c -> c.getMediaType() == MediaType.VIDEO)
```

src/main/java/org/jitsi/videobridge/Endpoint.java
mstyura: I could bring back `synchronized` section, if it purpose was to introduce atomic behavior between `remove` and `isEmpty`. But I'm actually can't figure out which problem it solved before? Currently ` if (channelShims.remove(channelShim))` is necessary to check if `channelShims` was modified during `remove` - it is optional check which could be omimted. The `if (channelShims.isEmpty())` necessary to check if there is no currently channels associated with endpoint.
As far as I understand endpoint become expired as soon as all channels removed.

src/main/java/org/jitsi/videobridge/Endpoint.java
mstyura: @bbaldino I've added dedicated method `allowsReceivingMedia` into `ChannelShim`, I've chosen the different name because endpoint might support media, but temporary don't want receiving it. I've also prefer `foreach` loop over `Stream API`, because it allows to compute `accept*` fields in single pass over shims collection.

src/main/java/org/jitsi/videobridge/Endpoint.java
mstyura: @bbaldino thanks for suggestion, I think `updateAcceptedMediaTypes` would be even better name.

src/main/java/org/jitsi/videobridge/Endpoint.java
bbaldino: the concern would be in the remove scenario if we ever failed to perform expiration when we should have.  i _think_ we're safe here without the sync block, though.  we are slightly more susceptible to the case where the remove gets run before an add (for the same channel) though, i think--the proper answer there would be to make sure we handle those calls from a single thread, as otherwise it's always possible to some degree.

src/main/java/org/jitsi/videobridge/Endpoint.java
bbaldino: yes, i think your naming is good for that reason.  as for saving the extra pass--i don't think it's a big deal as this should be an infrequent change and the list will be small so i think the more readable (in my opinion) stream version is nicer, but i don't feel strongly about it.

PR #961, opened 2019-10-31
Octo logging

bbaldino:
when i was trying to debug some octo issues i found myself wishing we had some extra logs to make things clear.  i suspect we may be removing a relay handler prematurely (something that could be coming from the shims which have changes in 2.0, so would explain why we're not seeing it in old bridge)

### DIFF
--- a/src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
+++ b/src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
@@ -19,7 +19,8 @@ import org.ice4j.socket.*;
 import org.jitsi.rtp.*;
 import org.jitsi.rtp.rtp.*;
 import org.jitsi.utils.*;
-import org.jitsi.utils.logging.*;
+import org.jitsi.utils.collections.*;
+import org.jitsi.utils.logging2.*;
 import org.jitsi.utils.stats.*;
 import org.jitsi.videobridge.util.*;
 import org.json.simple.*;
@@ -48,8 +49,7 @@ public class OctoRelay
      * The {@link Logger} used by the {@link OctoRelay} class and its
      * instances to print debug information.
      */
-    private static final Logger logger
-        = Logger.getLogger(OctoRelay.class);
+    private final Logger logger;
 
     /**
      * The receive buffer size to set of the socket.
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
```diff
@@ -19,7 +19,8 @@
 import org.jitsi.rtp.*;
 import org.jitsi.rtp.rtp.*;
 import org.jitsi.utils.*;
-import org.jitsi.utils.logging.*;
+import org.jitsi.utils.collections.*;
+import org.jitsi.utils.logging2.*;
```
JonathanLennox: Huh, we still had old logger in jvb in some places? I thought you'd switched everything over.  Are there any other places like this?

src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
```diff
@@ -312,8 +312,9 @@ public void addRtpExtension(RtpExtension rtpExtension)
      */
     public void expire()
     {
+        logger.info("Expiring");
```
JonathanLennox: This message is a bit terse.  Add what you're expiring?

src/main/java/org/jitsi/videobridge/octo/OctoRelay.java
bbaldino: yes, looks like there are.  i'll do a final sweep of this soon.

src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
bbaldino: it's terse just because the log context will already give the other info

JonathanLennox: I forget, does log context show up even if someone has the "wrong" log formatter configured?  I forget the order of operations.

bbaldino: no, that is done in the formatter, but we've already done this in some other places and, although i want to make sure things don't 'break' with old formatter i'm not really feeling like fully-accommodating it either, as we should just be migrating.  more motivation :)

JonathanLennox: Fair enough.

PR #962, opened 2019-10-31
Ep connection status config

bbaldino:


src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
```diff
@@ -54,48 +57,12 @@
     private final static String CFG_PNAME_BASE
         = "org.jitsi.videobridge.EndpointConnectionStatus";
```
JonathanLennox: Is `CFG_PNAME_BASE` still used after this change?

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
```diff
@@ -404,4 +360,64 @@ public void handleEvent(Event event)
             .filter(e -> e.getConference() == conference)
             .forEach(e -> sendEndpointConnectionStatus(e, false, endpoint));
     }
+
+    /**
+     * How long it can take an endpoint to send first data before it will
+     * be marked as inactive.
+     */
+    public static class FirstTransferTimeoutProperty extends AbstractConfigProperty<Long>
+    {
```
JonathanLennox: If these two properties are both scoped under `videobridge.ep-connection-status` in the config file, would it make sense to put them under the same master property object (like you did with `HealthConfig`)?

bbaldino: `EndpointConnectionStatus` is at the top level, so didn't wan to add a catch-all `config` package there, but do you mean just within the `EndpointConnecitonStatus` class itself?  like doing:

```
class EndpointConnectionStatus
{
  public static class Config
  {
    public static class FirstTransferTimeoutProperty
    {
    }
  }
}
```
?

Do you think it would add much? I could do it...don't feel strongly either way.

JonathanLennox: You don't like having an `EndpoingConnectionStatusConfig` class or something like that because it's at the top level?

bbaldino: `EndpointConnectionStatus` is in `org.jitsi.videobridge` and should probably be moved, as it's just a weird dumping ground for stuff, so i thought adding a `config` package there with config for all those classes would only add to the disarray, but i can do it.

JonathanLennox: Up to you, you have a better big-picture view of how everything's fitting together than I do.

bbaldino: ok, so i went to go add it, but i think it'll be misleading to have a `config` package at the top with all the other packages, and it didn't really feel 'connected' with the files at the top level, so it seemed confusing to me.  i think the long term fix here is to better organize the files and put the stuff at the top level in sub-packages

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
bbaldino: nope, removed

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
bbaldino: thought about this some more last night, and do think it'd be slightly better if i nested a `Config` static class here with the properties inside that...that way it's at least more consistent.  so will take a look at that today.

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
```diff
@@ -355,63 +355,76 @@ public void handleEvent(Event event)
...
+            }
+
+            public static Long getValue()
+            {
+                return singleton.get();
+            }
         }
 
-        public static Long getValue()
+        public static long getMaxInactivityLimt()
```
JonathanLennox: Typo in method name.

src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java
```diff
@@ -112,12 +112,12 @@ public void run()
             logger.error("Endpoint connection monitoring is already running");
         }
 
-        if (FirstTransferTimeoutProperty.getValue() <= MaxInactivityLimitProperty.getValue())
+        if (Config.getFirstTransferTimeout() <= Config.getMaxInactivityLimt())
         {
             throw new IllegalArgumentException(
                 String.format("FIRST_TRANSFER_TIMEOUT(%s) must be greater"
```
JonathanLennox: Should this exception message reference the new spelling of the property names?

bbaldino: yeah, i think i'll tweak it to reflect the property instead of the property name.  makes me think it'd also be cool if we could express this in the definitions themselves, maybe...something to think about for the future.

PR #964, opened 2019-11-04
Rework frame projection

JonathanLennox:
This reworks JVB's VP8 frame projection code, so it no longer needs to use a packet cache for piggybacking packets.

As a result there are also a number of additional packet loss pattern where previously we would have needed to request a keyframe, and now no longer do.

This depends on https://github.com/jitsi/jitsi-media-transform/pull/143 .


src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -53,131 +49,158 @@
...
+        byte[] buf = packet.getBuffer();
+        int payloadOffset = packet.getPayloadOffset(),
+            payloadLen = packet.getPayloadLength();
+
+        this.tl0PICIDX = packet.getTL0PICIDX();
+        this.isKeyframe = packet.isKeyframe();
+        this.pictureId = packet.getPictureId();
+        this.isReference =
+            /* TODO add this to Vp8Packet if we need it, otherwise remove it. */
+            DePacketizer.VP8PayloadDescriptor.isReference(buf, payloadOffset, payloadLen);
```
bbaldino: Guess we should add a helper on `Vp8Packet` for this.  Probably only `Vp8Packet` should use `DePacketizer` and expose nice methods for other places to use.

bbaldino: (not saying in this PR, just a thought)

JonathanLennox: Yeah, that's what I had in mind with that TODO.  If I even need `isReference`, which so far I haven't needed to use.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -191,64 +214,60 @@ long getTimestamp()
...
-    dependsOnSameTemporalBaseLayerFrame(@NotNull VP8Frame vp8Frame)
+    public VP8ProjectionRecord getProjectionRecord()
+    {
+        return projection;
+    }
+
+
+    public VP8FrameProjection getProjection()
+    {
+        if (projection == null || !(projection instanceof VP8FrameProjection))
```
bbaldino: i think `instanceof` handles null

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        return true;
+    }
+
+
+    /** Find a frame in the frame map, based on a packet. */
+    public synchronized VP8Frame findFrame(Vp8Packet packet)
+    {
+        int seq = packet.getSequenceNumber();
```
bbaldino: why does this use sequence number and not timestamp?

JonathanLennox: To make "previous frame" and "next frame" well-defined.  Technically, timestamps aren't required to be monotonic with respect to sequence numbers (though this is a feature that isn't often used for interactive video, because it adds encoding delay).

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
bbaldino: 🤦‍♂ sorry, managed to completely miss that comment.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
bbaldino: ah ok, i just thought it was the timestamp which separated/distinguished frames

JonathanLennox: Yeah, it's the timestamp that tells you whether two packets are the same frame, but it's the sequence numbers that tell you the decode order.

JonathanLennox: This function does two things -- it finds the frame to which a packet belongs, if it's already recorded; or if it's not it creates a new frame and finds the previous and next frames.

Probably should be documented/named better.

JonathanLennox: Following up on this - you'll see later in the same file there are now a bunch of functions to get the previous and next frames, either absolutely or with various condition tests.

I didn't see any reason to maintain two data structures (one mapping packets to frames, another maintaining frames in decode order) when I could do the job with one.

src/main/java/org/jitsi/videobridge/cc/GenericAdaptiveTrackProjectionContext.java
```diff
@@ -16,6 +16,7 @@
 package org.jitsi.videobridge.cc;
 
 import org.jetbrains.annotations.*;
+import org.jitsi.impl.neomedia.portaudio.*;
```
gpolitis: I'm wondering what could have caused the IDE to (presumably) automatically pull this package.

JonathanLennox: Probably an autocomplete typo.  Fixed.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -45,42 +42,16 @@
...
-
-    /**
-     * A map that stores the maximum sequence number of frames that are not
-     * (yet) accepted/projected. The map goes from ssrc -> timestamp -> highest
-     * sequence number.
-     */
-    private final Map<Long, LRUCache<Long, Integer>>
-        ssrcToFrameToMaxSequenceNumberMap = new HashMap<>();
+    private final Map<Long, VP8FrameMap>
+        vp8FrameMaps = new HashMap<>();
```
gpolitis: I'm still going through the code, so apologies for the potentially premature question; do we really need to remember frame projections from other ssrcs? isn't it sufficient to clean the map every time we switch simulcast streams (which refreshes the decoder as it can only happen on a keyframe, so we don't care if we leave undecodable frames at the receiver)

doc/svc.md
```diff
@@ -42,14 +42,97 @@ frame). This mode of operation has several implications.
...
+
+If there is a gap in receive sequence numbers between the top and a new frame,
+the SFU can determine the worst-case gap size of projected sequence numbers to
+leave such that any not-yet-received packets can be forwarded correctly.
+However, the SFU can't (in general) know the temporal layer of frames it has
+not yet received. If it turns out that the not-yet-received frames were in fact
+not ones that the SFU wanted to forward, this will result in sequence number
+gaps in the projected bitstream, which the receiver may interpret as packet
+loss.  (Care must be taken in bandwidth estimation algorithms not to incorrectly
+reduce bandwidth based on this "false" packet loss.)
```
gpolitis: For posterity, maybe we should point out here that when we're doing remb-based bwe, packet loss needs to be ignored (which is what we agreed upon) because we don't have a reliable way for computing loss?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -125,330 +96,393 @@ public VP8AdaptiveTrackProjectionContext(
...
 
-        // We have successfully projected the incoming frame and we've allocated
-        // a starting sequence number for it. Any previous frames can no longer
-        // grow.
-        vp8FrameProjectionMap.put(rtpPacket.getTimestamp(), nextVP8FrameProjection);
-        // The frame attached to the "last" projection is no longer the "last".
-        lastVP8FrameProjection = nextVP8FrameProjection;
+            if (accepted)
+            {
+                VP8FrameProjection projection = createProjection(frame, vp8Packet);
```
bbaldino: Is `createProjection` here doing more of a `getOrCreateProjection`?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -125,330 +96,393 @@ public VP8AdaptiveTrackProjectionContext(
...
-        if (frameToMaxSequenceNumberMap.containsKey(timestamp))
+        int projectedSeqGap = 1;
+
+        if (lastVP8FrameProjection.getVP8Frame() != null &&
+            !lastVP8FrameProjection.getVP8Frame().hasSeenEndOfFrame())
         {
-            int previousMaxSequenceNumber
-                = getMaxSequenceNumberOfFrame(ssrc, timestamp);
+            /* Leave a gap for the unfinished end of the previously routed frame. */
+            projectedSeqGap++;
```
bbaldino: Do we know here that we only need to leave a space for one?  Or could it be more and we don't know so we just leave a gap of 1?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -125,330 +96,393 @@ public VP8AdaptiveTrackProjectionContext(
...
-    public boolean needsKeyframe()
+    @NotNull
+    private VP8FrameProjection createInLayerProjection(@NotNull VP8Frame frame, @NotNull VP8Frame refFrame, @NotNull Vp8Packet initialPacket)
     {
-        if (vp8QualityFilter.needsKeyframe())
-        {
-            logger.debug(() -> "quality filter "
-                    + vp8QualityFilter.hashCode()
-                    + " says vp8 track needs keyframe");
+        long nowMs = System.currentTimeMillis();
```
bbaldino: Don't you have tests fo these? How do they work if they're using clock time?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -507,52 +536,59 @@ public PayloadType getPayloadType()
...
-        @NotNull VideoRtpPacket rtpPacket, PacketCache incomingPacketCache)
+    public void rewriteRtp(
+        @NotNull VideoRtpPacket rtpPacket)
         throws RewriteException
     {
         if (!(rtpPacket instanceof Vp8Packet))
         {
             logger.info("Got a non-VP8 packet.");
-            return null;
+            throw new RewriteException();
```
bbaldino: Seems like some context in these exceptions (of what exactly went wrong) would be useful.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -27,16 +27,11 @@
...
- * this class instances is final, the maxSequenceNumber and the
- * endingSequenceNumberIsKnown fields are not and it is therefore necessary
- * to synchronize reading/writing to those fields on this instance.
- *
- * Consider, for example, creating a new frame projection (that happens in the
- * translator thread and reads the maxSequenceNumber field) while packets are
- * being piggy-backed (that happens in the egress thread and could potentially
- * write the maxSequenceNumber field).
+ * this class instances is final, the sequence number ranges, haveStart/haveEnd,
+ * and isKeyframe.
```
bbaldino: I think this sentence is missing something?

doc/svc.md
JonathanLennox: I updated this to say that.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
JonathanLennox: If we've already received (and decided not to project) TL0s after a keyframe, we need to make sure that we don't subsequently project that keyframe, because the video will get stuck.

I wanted to keep the per-frame route-or-don't logic centralized in `VP8QualityFilter`, across all ssrcs, which means we need to maintain per-frame state for all the ssrcs.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -467,13 +467,10 @@ public void send(PacketInfo packetInfo)
         if (packet instanceof VideoRtpPacket)
         {
             //TODO(brian): we lose all information in PacketInfo here,
-            // unfortunately, because the BitrateController can return more
-            // than/less than what was passed in (and in different order) so
-            // we can't just reassign a transformed packet back into its
-            // proper PacketInfo. Need to change those classes to work with
+            // Need to change those classes to work with
```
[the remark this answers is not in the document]
JonathanLennox: I don't.  I guess there's nothing blocking us from doing that, now, though?  (Does any meta-data in `PacketInfo` need to be updated as we tranform the packet?)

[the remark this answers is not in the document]
bbaldino: No, not updated...but preserving what was originally in there let's us do things like track packet latency through the bridge.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
JonathanLennox: No, this is in the `isNewFrame` clause.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
JonathanLennox: It could be more than 1, but we close the old projection so that gap will never be filled in.  The gap is to make it clear that the old frame is incomplete.

Arguably this is unnecessary since the old code left frames incomplete without gaps, but I feel more comfortable making sure streams which are incomplete at the bitstream level are also incomplete at the RTP level.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
JonathanLennox: Yeah, this is slightly dodgy; it works by the fact that the minimum time difference it'll handle is 33 ms, which is more than the time the test should take to run.  There's a comment in the unit test that the test case will fail if you stop it in a debugger at the wrong moment.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -125,330 +96,393 @@ public VP8AdaptiveTrackProjectionContext(
...
-        if (lastVP8FrameProjection.getSSRC() != nextVP8FrameProjection.getSSRC())
+        VP8Frame frame = result.getFrame();
+
+        if (result.isNewFrame())
         {
-            /* We've changed the encoding we're forwarding, so previous frames
-             * in the map are no longer relevant.
+            if (vp8Packet.isKeyframe() && frameIsNewSsrc(frame))
+            {
+            /* If we're not currently projecting this SSRC, make sure we haven't
```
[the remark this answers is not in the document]
JonathanLennox: Re-worded it, any better?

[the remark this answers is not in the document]
bbaldino: Yes, much! Thanks.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
JonathanLennox: Added a message.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -125,330 +96,393 @@ public VP8AdaptiveTrackProjectionContext(
...
+     * Calculate the projected sequence number gap between two frames (of the same encoding),
+     * allowing collapsing for unaccepted frames.
+     */
+    private int seqGap(@NotNull VP8Frame frame1, @NotNull VP8Frame frame2)
+    {
+        int seqGap = RtpUtils.getSequenceNumberDelta(frame2.getEarliestKnownSequenceNumber(), frame1.getLatestKnownSequenceNumber());
+
+        if (!frame1.isAccepted() && !frame2.isAccepted() &&
+            frame2.getPictureId() ==
+                Vp8Utils.applyPictureIdDelta(frame1.getPictureId(), 1))
```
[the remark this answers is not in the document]
JonathanLennox: Done.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -125,330 +96,393 @@ public VP8AdaptiveTrackProjectionContext(
...
-
-        long threshold = RtpUtils.applyTimestampDelta(newTimestamp, -WAIT_SAMPLES);
-
-        Iterator<Long> it = vp8FrameProjectionMap.keySet().iterator();
-        while (it.hasNext())
+        else
         {
-            Long key = it.next();
-            if (RtpUtils.isOlderTimestampThan(key, threshold))
+            if (!frame1.isAccepted() && !frame1.hasSeenEndOfFrame() && seqGap > 1)
```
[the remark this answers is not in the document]
JonathanLennox: I added comments which hopefully should help.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
JonathanLennox: Fixed.

bbaldino:
> > I saw your other comment, but who are we really signaling this "to? at this point? The far-side decoder?
> 
> Yeah, that's the idea. Maybe it's silly?

Yeah I dunno.  Would the decoder not figure it out from the vp8 header information/payload descriptor?

bbaldino:
LGTM as far as the code is concerned.  I can't give as detailed of a review on the logic itself, so hopefully someone else can comment on that, but given the tests I'd be ok with merging if we want to start getting some usage on it.

src/test/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionTest.java
```diff
@@ -0,0 +1,858 @@
...
+            {
+                expectedTs = RtpUtils.applyTimestampDelta(expectedTs, 3000);
+            }
+        }
+    }
+
+    @Test
+    public void twoStreamsSwitchingTest() throws RewriteException
+    {
+        Vp8PacketGenerator generator1 = new Vp8PacketGenerator(3);
```
gpolitis: i'm wondering if it would be useful (and also if it's in scope for this PR) to have a test that feeds the code with real VP8 simulcast (stored in a pcap file, for instance)?

I seem to recall that the engine produces (and the pacer transmits) frames together, so you have frame A (packets x1,x2,x3) followed by frame B (packets y1,y2,y3) and so on and so forth, which is not captured by these (admittedly very much needed) test cases.

This could be useful for debugging purposes as well, i.e. capture ingress simulcast that we broke in the egress and play it back to see what went wrong; and it would enable us to easily write a test case to avoid the bug in the future.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
gpolitis: so you're talking about the scenario where we're forwarding stream_{A}, reject TL0_{n+1} of stream_{B} and then switch to stream_{B} upon receipt of keyframe_{n} of stream_{B}? Am I understanding correctly?

JonathanLennox: Right, exactly.  I have logic that checks for that case (line 297 of VP8AdaptiveTrackProjectionContext).

src/test/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionTest.java
JonathanLennox: I agree it seems like a useful tool.  Do you think it needs to be part of this PR, though?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -125,330 +96,411 @@ public VP8AdaptiveTrackProjectionContext(
...
+        return frameMap.findNextTl0(frame);
+    }
+
+    /**
+     * Calculate the projected sequence number gap between two frames (of the same encoding),
+     * allowing collapsing for unaccepted frames.
+     */
+    private int seqGap(@NotNull VP8Frame frame1, @NotNull VP8Frame frame2)
+    {
+        int seqGap = RtpUtils.getSequenceNumberDelta(frame2.getEarliestKnownSequenceNumber(), frame1.getLatestKnownSequenceNumber());
```
gpolitis: I haven't followed the call hierarchy (yet) so this is a more or less uneducated question but do we have to think about the case where frame1 is a really old frame (before switching away from this stream) and frame2 is the keyframe that brings us back to this stream and the gap is something randomly large like (1 << 15) - 1?

JonathanLennox: In normal cases the frame being referenced will be a recently-sent frame in this code path (it's only executed when projecting from existing frames) but I suppose there could be weird cases like a stream having a large outage.  I'll think about if there's a case that needs to be handled here.

For the normal cases, though, one of the reasons I'm tracking sequence numbers of the inactive streams is indeed to avoid sequence number confusion.

gpolitis:
Have we removed the decodability check? Suppose that we have the following sequence of pictures TL0_{1}, TL2_{2}, TL1_{3}, TL2_{4} and that we drop pictures 2 & 3 because the quality filter decides so. If the quality filter state changes and it accepts TL2_{4} are we going to accept it (even tho it's not decodable because TL1 are typically reference frames)?

JonathanLennox:
> Have we removed the decodability check? Suppose that we have the following sequence of pictures TL0_{1}, TL2_{2}, TL1_{3}, TL2_{4} and that we drop pictures 2 & 3 because the quality filter decides so. If the quality filter state changes and it accepts TL2_{4} are we going to accept it (even tho it's not decodable because TL1 are typically reference frames)?

Yes, we have.

Obviously this isn't doable in all general cases because of the out-of-order processing, but in normal cases (frames received in order without substantial loss) it should be achievable.

The worst case this would cause would be one frame of mis-predicted video (as TL2_{4} would use the wrong reference buffer) before the stream corrects itself.  How important do you think this is?

src/test/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionTest.java
gpolitis: No, it doesn't have to be, especially if we feel confident enough that the refactor doesn't introduce any significant regressions.

gpolitis:
> > Have we removed the decodability check? Suppose that we have the following sequence of pictures TL0_{1}, TL2_{2}, TL1_{3}, TL2_{4} and that we drop pictures 2 & 3 because the quality filter decides so. If the quality filter state changes and it accepts TL2_{4} are we going to accept it (even tho it's not decodable because TL1 are typically reference frames)?
> 
> Yes, we have.
> 
> Obviously this isn't doable in all general cases because of the out-of-order processing, but in normal cases (frames received in order without substantial loss) it should be achievable.
> 
> The worst case this would cause would be one frame of mis-predicted video (as TL2_{4} would use the wrong reference buffer) before the stream corrects itself. How important do you think this is?

I don't know how important it is as we don't have any data that show how often this happens, but I'd certainly feel more confortable if we kept it as we've struggled a lot with video corruption in the past. I suppose we can worry about it if we ever get any video corruption.

JonathanLennox:
> I don't know how important it is as we don't have any data that show how often this happens, but I'd certainly feel more confortable if we kept it as we've struggled a lot with video corruption in the past. I suppose we can worry about it if we ever get any video corruption.

I'll take a look at how hard it would be to do.  Thinking about it, it's probably not so hard.

JonathanLennox:
> > I don't know how important it is as we don't have any data that show how often this happens, but I'd certainly feel more confortable if we kept it as we've struggled a lot with video corruption in the past. I suppose we can worry about it if we ever get any video corruption.
> 
> I'll take a look at how hard it would be to do. Thinking about it, it's probably not so hard.

Not hard at all; I've added it.

gpolitis:
> > > I don't know how important it is as we don't have any data that show how often this happens, but I'd certainly feel more confortable if we kept it as we've struggled a lot with video corruption in the past. I suppose we can worry about it if we ever get any video corruption.
> > 
> > 
> > I'll take a look at how hard it would be to do. Thinking about it, it's probably not so hard.
> 
> Not hard at all; I've added it.

Thanks Jonathan, that's great. The only thing probably missing is a few svc rewriting examples in the svc document, like what we have here https://docs.google.com/document/d/1rdgvXhx63AR5GWkj9j0MURAvM5YlEzm5Wq93Kddlk3o/edit#heading=h.rf7xa7icdo4

lgtm modulo that

JonathanLennox:
> Thanks Jonathan, that's great. The only thing probably missing is a few svc rewriting examples in the svc document, like what we have here

I've added some, let me know if they seem complete?

gpolitis:
> > Thanks Jonathan, that's great. The only thing probably missing is a few svc rewriting examples in the svc document, like what we have here
> 
> I've added some, let me know if they seem complete?

It looks good, thanks Jonathan. The checks are failing on this non-functional commit which makes me think something is wrong with the tests, rather than the commit it self. I've logged into jenkins and requested a re-run, just in case.

PR #966, opened 2019-11-05
[ChannelShim::describe] Fill sourceGroup in channel IQ

Mikhalevich:
in JVB 2.0 in coparison with JVB 1.0 sourceGroup attribute is missing.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
+++ b/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
@@ -251,6 +251,10 @@ public class ChannelShim
 
                 iq.setSSRCs(ssrcs);
             }
+
+            if (sourceGroups != null) {
+                sourceGroups.forEach(iq::addSourceGroup);
+            }
         }
     }
 


bbaldino:
Not saying we shouldn't do this, but I actually can't find where jvb 1.0 set this field in this flow.  In the old `RtpChannel.java` [here](https://github.com/jitsi/jitsi-videobridge/blob/jvb_1.0/src/main/java/org/jitsi/videobridge/RtpChannel.java#L715) we set sources, but not source groups.  When I search for `addSourceGroup` in the old bridge, the only place I see it being done is in `JSONDeserializer` [here](https://github.com/jitsi/jitsi-videobridge/blob/jvb_1.0/src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java#L940), but we have that same code still in 2.0 [here](https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java#L901).  What's the call where you saw it in 1.0 but no 2.0?

Mikhalevich:
> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

done for individuals

Mikhalevich:
in JVB 1.0 to retrieve for example FID group we should write something like this:
```java
final MediaStreamTrackDesc[] mediaStreamTracks = rtpChannel.getStream().getMediaStreamTrackReceiver().getMediaStreamTracks();
        for (MediaStreamTrackDesc mediaStreamTrack : mediaStreamTracks) {
            for (RTPEncodingDesc rtpEncoding : mediaStreamTrack.getRTPEncodings()) {

                long primary = rtpEncoding.getPrimarySSRC();
                long secondary = rtpEncoding.getSecondarySsrc("rtx");
                String uniqueKey = Long.toString(primary) + Long.toString(secondary);

                if (sourceGroups.containsKey(uniqueKey)) {
                    continue;
                }

                final SourcePacketExtension secondarySsrc = new SourcePacketExtension();
                secondarySsrc.setSSRC(secondary);

                if (!secondarySsrc.hasSSRC()) {
                    continue;
                }

                final SourcePacketExtension primarySsrc = new SourcePacketExtension();
                primarySsrc.setSSRC(primary);

                final SourceGroupPacketExtension sourceGroup = new SourceGroupPacketExtension();
                sourceGroup.setSemantics("FID");
                sourceGroup.addSources(Arrays.asList(primarySsrc, secondarySsrc));

                sourceGroups.put(uniqueKey, sourceGroup);
            }
        }
```
in a new bridge we already have source groups attribute for fill up in describe method. I believe it's reasonable or I'm doing something wrong?

bbaldino:
Ok, am I understanding it correctly that you weren't previously getting this data from an external JVB API call, but with custom code within the JVB itself that you've added?  You're doing this with the code you pasted above in 1.0, and you were looking at how to achieve the same thing in 2.0?

Mikhalevich:
Yes, I can get this data in 1.0 by adding some code, but I don't know how to do the same thing in 2.0

Mikhalevich:
And it's looks strange for me that in describe method the ssrcs is filled and group is not :-)

bbaldino:
Ok, thanks for the info, I just wanted to make sure I understood the situation correctly.  I agree this does feel like it should be there, my only hesitation is that since it wasn't in 1.0 I worry something may be relying on it.  I'll trigger the tests (not sure why they didn't run) and we'll see if that turns anything up.

bbaldino:
jenkins, test this please

Mikhalevich:
Thx

PR #967, opened 2019-11-05
move the configuration of config source suppliers to another file

bbaldino:
i noticed this when working on another PR when i tried setting the
command-line args supplier to include the args given in main.  the
problem was that, since JvbConfig loaded the config in a static block,
any access on JvbConfig would trigger it--this includes setting a
supplier.  for the command-line args case, this means that when you
called JvbConfig.commandLineArgsSupplier = ..., the config would load
and log that no command line args were present (since the assignment was
not yet complete) which was misleading.  it also had the annoying
behavior of dumping any default-loaded config during tests when the
suppliers were overriden.

by separating that config out, we can manipulate the config source
suppliers without triggering a loading of the config, which makes more
sense.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/util/config/JvbConfig.java
+++ b/src/main/java/org/jitsi/videobridge/util/config/JvbConfig.java
@@ -22,8 +22,6 @@ import org.jitsi.utils.logging2.*;
 
 import java.util.*;
 
-import static org.jitsi.videobridge.util.config.ConfigSupplierSettings.*;
-
 /**
  * Helper class for retrieving configuration sources.  Currently, this supports reading config from
  * 1) Command line arguments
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/videobridge/util/config/JvbConfig.java
```diff
@@ -20,9 +20,9 @@
 import org.jetbrains.annotations.*;
 import org.jitsi.utils.logging2.*;
 
-import java.nio.file.*;
 import java.util.*;
-import java.util.function.*;
+
+import static org.jitsi.videobridge.util.config.ConfigSupplierSettings.*;
```
JonathanLennox: How messy would it be to not make this `import static`?  I found it sort of confusing how the `ConfigSupplierSettings` were affecting `JvbConfig` until I noticed it, so I'm wondering if it'd be clearer if `JvbConfig` explicitly mentioned `ConfigSupplierSettings` where it was needed.

src/test/java/org/jitsi/testutils/ConfigSetup.java
```diff
@@ -69,14 +69,14 @@ public void finishSetup()
         // Make sure config doesn't see any command line args from a previous setup
         if (commandLineArgs.isEmpty())
         {
-            JvbConfig.commandLineArgsSupplier =() -> new String[0];
+            ConfigSupplierSettings.commandLineArgsSupplier =() -> new String[0];
```
JonathanLennox: Space between `=` and `()`?  This isn't new but might as well fix it.

src/main/java/org/jitsi/videobridge/util/config/JvbConfig.java
bbaldino: not messy, i can change it

PR #968, opened 2019-11-05
Client connection config

bbaldino:


src/main/resources/reference.conf
```diff
@@ -17,4 +17,16 @@ videobridge {
     # be considered disconnected
     max-inactivity-limit=3 seconds
   }
+  # The APIs by which the JVB can be controlled
+  apis {
+    xmpp-client {
+      enabled=false
```
bbaldino: i added this separate `enabled` value because i thought it would be nice to be able to disable these without having to clear all the env configurations.

src/main/java/org/jitsi/videobridge/xmpp/config/Config.java
```diff
@@ -0,0 +1,117 @@
...
+    {
+        protected static class EnabledProperty extends AbstractConfigProperty<Boolean>
+        {
+            protected static final String propName = "videobridge.apis.xmpp-client.enabled";
+
+            private static EnabledProperty singleton = new EnabledProperty();
+
+            public EnabledProperty()
+            {
+                super(new JvbPropertyConfig<Boolean>()
```
bgrozev: I would much rather these properties use a more declarative syntax. One shouldn't have to follow code just to understand the characteristics of a simple boolean property.

src/test/resources/xmpp_client_configs.conf
```diff
@@ -0,0 +1,30 @@
...
+  videobridge {
+    apis {
+      xmpp-client {
+        configs {
+          shard-1 {
+            "DISABLE_CERTIFICATE_VERIFICATION" = true
+            "DOMAIN" = auth.brian.jitsi.net
+            "PASSWORD" = "password\\+"
+            "HOSTNAME" = brian.jitsi.net
+            "MUC" = "JvbBrewery@internal.auth.brian.jitsi.net"
```
bgrozev: I think `MUC` is obsolete in favour of `MUC_JIDS`. I don't see any code using `MUC`.

src/test/resources/xmpp_client_configs.conf
```diff
@@ -0,0 +1,30 @@
...
+            "USERNAME"=username
+          }
+          shard-2 {
+            "DISABLE_CERTIFICATE_VERIFICATION" = true
+            "DOMAIN" = auth.brian2.jitsi.net
+            "PASSWORD" = "password\\+"
+            "HOSTNAME" = brian2.jitsi.net
+            "MUC" = "JvbBrewery@internal.auth.brian2.jitsi.net"
+            "MUC_JIDS" = "JvbBrewery@internal.auth.brian2.jitsi.net"
+            "MUC_NICKNAME"=brian_local2
```
bgrozev: This doesn't have to be different from the one in shard-1. I'd use the same value to avoid implying that has to be different. I'd use something like `jvb-1` in both places to suggest it's an identifier of the bridge

src/main/java/org/jitsi/videobridge/xmpp/config/Config.java
bbaldino: Ok, I'm open to suggestions how to reimplement the config properties to achieve this.

src/test/resources/xmpp_client_configs.conf
bbaldino: For the purpose of a test, I think it's better to use distinct values...otherwise it's hard to ensure that both chunks were actually parsed.  I get that it doesn't necessarily mirror the real-life application, but this is more about testing the borders of the logic.

src/test/resources/xmpp_client_configs.conf
bbaldino: ok, i can strip that.

PR #970, opened 2019-11-06
Treat non-RTP traffic as endpoint's IO activity

mstyura:
In `JVB 1.0` endpoint connectivity monitoring [taken into account both `RTP` and `SCTP` packets](https://github.com/jitsi/jitsi-videobridge/blob/c29af045ca7081dce457ebd25a025248cfda3a67/src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java#L236-L262).
In `JVB 2.0` [only `RTP` packets](https://github.com/jitsi/jitsi-media-transform/blob/c1e931aa002f5f646a3032e61ae5f693d0a7dd83/src/main/kotlin/org/jitsi/nlj/Transceiver.kt#L129-L141) affect [`IO` activity](https://github.com/jitsi/jitsi-videobridge/blob/e3120b90798a30af78ae3ac1b05c47e977132576/src/main/java/org/jitsi/videobridge/Endpoint.java#L568-L573) during [connectivity check](https://github.com/jitsi/jitsi-videobridge/blob/e3120b90798a30af78ae3ac1b05c47e977132576/src/main/java/org/jitsi/videobridge/EndpointConnectionStatus.java#L184-L189), it causes frequent `EndpointConnectivityStatusChangeEvent` events via data channel toggling `active=true/false`, when client has temporary disable media (for example, during hold).

This PR changes IO activity counting to count any incoming/outgoing packet which passes `DtlsTransport` as an IO acivity. This is slightly different than `JVB 1.0` - it was explicitly captured `RTP` and `SCTP`, but hopefully it still valid.
As a followup `PR` packet IO activity field could be removed from Transceiver as long it is no longer necessary for activity computation.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -95,7 +95,6 @@ public class DtlsTransport extends IceTransport
     private final Node outgoingDtlsPipelineRoot;
     private final Node outgoingSrtpPipelineRoot;
     private boolean dtlsHandshakeComplete = false;
-
     /**
      * Measures the jitter introduced by the bridge itself (i.e. jitter calculated between
      * packets based on the time they were received by the bridge and the time they
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -502,8 +502,7 @@ public class DtlsTransport extends IceTransport
     protected void onIceConsentUpdated(long time)
     {
        super.onIceConsentUpdated(time);
-       packetIOActivity.setLastPacketReceivedTimestampMs(
-           System.currentTimeMillis());
+       packetIOActivity.setLastPacketReceivedTimestampMs(time);
     }
 
     /**
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -90,6 +90,12 @@ public class DtlsTransport extends IceTransport
     private final Node outgoingDtlsPipelineRoot;
     private final Node outgoingSrtpPipelineRoot;
     private boolean dtlsHandshakeComplete = false;
+
+    /**
+     * Packet IO activity of this transport.
+     */
+    private final PacketIOActivity packetIOActivity = new PacketIOActivity();
+
     /**
      * Measures the jitter introduced by the bridge itself (i.e. jitter calculated between
      * packets based on the time they were received by the bridge and the time they
@@ -137,6 +143,15 @@ public class DtlsTransport extends IceTransport
         incomingPipelineRoot = createIncomingPipeline();
         outgoingDtlsPipelineRoot = createOutgoingDtlsPipeline();
         outgoingSrtpPipelineRoot = createOutgoingSrtpPipeline();
+    }
+
+    /**
+     * Get packet IO activity of this transport.
+     * @return {@link PacketIOActivity} of this transport
+     */
+    public PacketIOActivity getPacketIOActivity()
+    {
+        return packetIOActivity;
     }
 
     /**
@@ -416,6 +431,7 @@ public class DtlsTransport extends IceTransport
                                 len);
                     PacketInfo pktInfo = new PacketInfo(pkt);
                     pktInfo.setReceivedTime(System.currentTimeMillis());
+                    packetIOActivity.setLastPacketReceivedTimestampMs(System.currentTimeMillis());
                     incomingPipelineRoot.processPacket(pktInfo);
 
                     p.setData(receiveBuf, 0, receiveBuf.length);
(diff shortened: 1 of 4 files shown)


JonathanLennox:
1.0 also had IO activity kept alive by ICE consent updates, I think, so that should be added as well.  I have a PR https://github.com/jitsi/jitsi-videobridge/pull/971 which should make it possible to add this.

mstyura:
@JonathanLennox that it might make sense to move `PacketIOActivity` into `IceTransport`?

JonathanLennox:
@mstyura Since the packets are actually received in `DtlsTransport` I think it makes sense to keep the logic about what counts as IO activity in one place.

mstyura:
Jenkins test this please

mstyura:
@JonathanLennox I've added IO activity update into `onIceConsentUpdated`, could you please check if I've done what you've meant? Thanks a lot!

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -479,6 +495,17 @@ protected void onIceFailed()
...
+     * {@inheritDoc}
+     */
+    @Override
+    protected void onIceConsentUpdated(long time)
+    {
+       super.onIceConsentUpdated(time);
+       packetIOActivity.setLastPacketReceivedTimestampMs(
+           System.currentTimeMillis());
+    }
+
```
JonathanLennox: The `time` parameter here is the `currentTimeMillis()` when the consent packet was received, so the additional call should be unnecessary.

mstyura: @JonathanLennox ok, I'll update. But the generic idea here was to not overwrite timestamp which is already newer and I assumed that System.currentTimeMillis will not go back in history, which is actually incorrect assumption, but used everywhere in a codebase... Just if-check before assignment is not enough due to field might be assigned from multiple threads. Ideally monotonic `System.nanoTime` should be used here and many other places in the project , which should guarantee non-decreasing and monotonic timestamps. Anyway changing now to just assigning `time`.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -568,7 +568,7 @@ public void addPayloadType(PayloadType payloadType)
     public long getLastActivity()
     {
         PacketIOActivity packetIOActivity
-                = this.transceiver.getPacketIOActivity();
+                = this.dtlsTransport.getPacketIOActivity();
```
bbaldino: i'm wondering if there's actually a use-case to distinguish last ICE activity from last media activity, in which case we should keep the `PacketIOActivity` instance in the `Transceiver` and just update it from here, rather than replacing that altogether.  if we _do_ go this route, we need an accompanying PR in JMT to remove `PacketIOActivity` from `Transceiver` so it's not accidentally used, I think.

I think I'm in favor of keeping this in `Transceiver` and just updating that instance from here, as then we can more easily expand the functionality here (separately tracking ICE and RTP activity, maybe even RTP/RTCP/audio/video activity, etc...we've got options for a lot more depth).

mstyura: > i'm wondering if there's actually a use-case to distinguish last ICE activity from last media activity, 

@bbaldino  There is one use-case I used. During hold (no media sent between client and jvb) endpoint is considered as disconnected, despite SCTP and ICE packets being sent/received.

> we need an accompanying PR in JMT to remove PacketIOActivity from Transceiver so it's not accidentally used, I think.

That's sounds reasonable.

> I think I'm in favor of keeping this in Transceiver and just updating that instance from here,

Have no objection to that. The only thing I wanted to fix actually is to properly handle case when no media is transmitted, but endpoint is actually connected to jvb. Today such endpoint is considered as having connectivity issues.

bbaldino: > Have no objection to that. The only thing I wanted to fix actually is to properly handle case when no media is transmitted, but endpoint is actually connected to jvb. Today such endpoint is considered as having connectivity issues.

Yep, agree.  Let me take a quick look at updating `PacketIOActivity` in JMT.  I'll put something up and let me know what you think.

bbaldino: What do you think of [this?](https://github.com/jitsi/jitsi-media-transform/pull/148)  You could then update `lastIceActivityTimestamp` here and query `latestOverallActivity` which would include ICE activity.

bbaldino: @mstyura https://github.com/jitsi/jitsi-media-transform/pull/148 just got merged...want to re-do this on top of that?

mstyura: @bbaldino thanks, I could redo but only tomorrow. If you have time and wish to do it yourself feel free to do so :)

mstyura: @bbaldino I've done change for ICE, but I think it also useful to have last SCTP send/receive timestamp. I could add it as well, but I prefer followup PR, because it will require PR to JMT.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -583,8 +583,8 @@ public void addPayloadType(PayloadType payloadType)
     public long getLastActivity()
     {
         PacketIOActivity packetIOActivity
-                = this.dtlsTransport.getPacketIOActivity();
-        return packetIOActivity.getLastOverallActivityTimestampMs();
+                = this.transceiver.getPacketIOActivity();
+        return packetIOActivity.getLatestOverallActivity().getEpochSecond();
```
bbaldino: this was millis before but you've got it as seconds here

mstyura: Nice catch, thanks Brian! Now fixed.

mstyura:
@bbaldino yes, I would prefer merging this (as it self contained) and adding SCTP handling as separate PR.

PR #972, opened 2019-11-06
Don't expose channel shims.

bgrozev:


mstyura:
Hello @bgrozev and @bbaldino!

I'd like to have better handling of scenario where JVB is used with mobile clients and mobile clients is get interrupted during video call, for example: minimized, put on hold by GSM call, screen is locked. 
In such case there is no sense in receiving incoming video from JVB, because it can't rendered (audio still can be played in case app is minimized, but not in case of GSM hold).

With JVB 1.0 I had an ability to stop JVB from sending media to client on the fly [by updating media direction of `RTPChannel`](https://github.com/jitsi/jitsi-videobridge/blob/c29af045ca7081dce457ebd25a025248cfda3a67/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java#L172-L178). It looked like:
```java
final List<RtpChannel> channels = endpoint.getChannels();
for (RtpChannel c : channels) {
     c.setDirection(/* desired direction here */);
}
```

It also was possible with JVB 2.0 when `ChannelShims` was exposed. Before this PR merged I was able to do:
```java
final Collection<ChannelShim> shims = endpoint.getChannelShims();
for (ChannelShim shim : shims) {
    shim.setDirection(/* desired direction here */);
}
```

Could you please consider bringing back endpoint's channel shims to public or suggest alternative way to on the fly updating media direction?
I've already tried to update direction via Colibri, but it had some undesired side effects I'd like to avoid: it touches too many things, not only direction when existing channel is updated.

Thank you very much in advance!

bbaldino:
We can bring this back, I'd just put a method like `setDirection` in `AbstractEndpoint`.  `OctoEndpoint` would ignore it, I think (maybe throw to make sure we didn't try this there?) and `Endpoint` could set the direction on the shims.

Also, might be nice to come up with better naming/vocabulary than `setDirection`/"direction"?  Nothing jumping to mind right now though.

bgrozev:
@bbaldino it would have to set different directions for audio and video.

The jitsi-meet client accomplishes the same by setting LastN=0 through the data channel.

PR #973, opened 2019-11-07
add an extra layer of tiimeout detection for endpoints

bbaldino:
before this, if an endpoint was allocated but never
started ice, it would never be expired as, if an ep has never had
activity, we waited for ice to fail...but you can't fail ice if you
never start ice...

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -102,6 +102,19 @@ public class Endpoint
     private SctpManager sctpManager;
 
     /**
+     * The time at which this endpoint was created (in millis since epoch)
+     */
+    //TODO: use clock/instant in this file
+    private final Long creationTimeMillis;
+
+    /**
+     * How long we'll give an endpoint to either successfully establish
+     * an ICE connection or fail before we expire it.
+     */
+    //TODO: make this configurable
+    private static final Duration EP_TIMEOUT = Duration.ofMinutes(2);
+
+    /**
      * TODO Brian
      */
     private final SctpHandler sctpHandler = new SctpHandler();
@@ -209,6 +222,7 @@ public class Endpoint
     {
         super(conference, id, parentLogger);
 
+        creationTimeMillis = System.currentTimeMillis();
         diagnosticContext = conference.newDiagnosticContext();
         transceiver
                 = new Transceiver(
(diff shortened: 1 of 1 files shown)


bbaldino:
no unit tests as we need to clean up `Endpoint` before we could reasonably do that (by using a `Clock`, mainly), but i was able to reproduce the issue locally and verify the fix.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -101,6 +101,19 @@
...
+     */
+    //TODO: use clock/instant in this file
+    private final Long creationTimeMillis;
+
+    /**
+     * How long we'll give an endpoint to either successfully establish
+     * an ICE connection or fail before we expire it.
+     */
+    //TODO: make this configurable
+    private static final Duration EP_TIMEOUT = Duration.ofMinutes(2);
```
bbaldino: this was somewhat arbitrary...seemed reasonable but i'd be fine with changing it.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -600,14 +614,20 @@ public boolean shouldExpire()
 
         long lastActivity
                 = packetIOActivity.getLastOverallActivityTimestampMs();
+        long now = System.currentTimeMillis();
         if (lastActivity <= 0)
         {
+            Duration timeSinceCreation = Duration.ofMillis(now - creationTimeMillis);
+            if (timeSinceCreation.compareTo(EP_TIMEOUT) > 0) {
+                logger.info("Endpoint's ICE connection has neither failed nor connected " +
+                    "after " + timeSinceCreation + ", expiring");
```
JonathanLennox: `Duration.toString()` output looks like `PT2M3.345S` for values in this range. Is this what we want?

bbaldino: i agree it's not the most legible, but i actually prefer it to reading a big number of milliseconds.  or is there a way to better format duration's output?

JonathanLennox:
Also note that the commit message says "tiimeout".

src/main/java/org/jitsi/videobridge/Endpoint.java
JonathanLennox: I agree it's reasonable.  (If anything it's probably over-generous.)

src/main/java/org/jitsi/videobridge/Endpoint.java
JonathanLennox: Yeah, it's better than a big number of ms.

Barring creating our own formatters for `Duration` (which we could do but is overkill for this) this should be good enough.

PR #975, opened 2019-11-08
update jmt version

bbaldino:
this gets the SR fix and the logging around 'missing packet detail'


JonathanLennox:
Maven version skew handling is always baffling, but should this update the jitsi-rtp version to make sure we get the SR cloning code?

bbaldino:
> Maven version skew handling is always baffling, but should this update the jitsi-rtp version to make sure we get the SR cloning code?

Gah, yes I think you're right.  Nice catch.

PR #978, opened 2019-11-12
Move to new PacketIOActivity fields based on Instant

bbaldino:
Fun fact: `Instant.MIN`, which we use for our definition of `NEVER` is unable to be converted to a number of milliseconds-since-epoch.  This is probably good for `NEVER`, but broke some things here, so had to migrate away from the usage of the old properties and use `Instant` everywhere.

I think `EndpointConnectionStatus` could really use a unit test for its logic, but it'll need to be refactored so we can swap in a fake executor service, etc. so left that for future work.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -31,9 +31,7 @@ import org.jitsi.rtp.rtp.*;
 import org.jitsi.utils.concurrent.*;
 import org.jitsi.utils.logging.*;
 import org.jitsi.utils.*;
-import org.jitsi.utils.logging2.*;
 import org.jitsi.utils.logging2.Logger;
-import org.jitsi.utils.logging2.LoggerImpl;
 import org.jitsi.videobridge.cc.*;
 import org.jitsi.videobridge.datachannel.*;
 import org.jitsi.videobridge.datachannel.protocol.*;
@@ -104,7 +102,7 @@ public class Endpoint
     /**
      * The time at which this endpoint was created (in millis since epoch)
      */
-    private final Instant creationTimeMillis;
+    private final Instant creationTime;
 
     /**
      * How long we'll give an endpoint to either successfully establish
@@ -229,7 +227,7 @@ public class Endpoint
 
         this.clock = clock;
 
-        creationTimeMillis = clock.instant();
+        creationTime = clock.instant();
         diagnosticContext = conference.newDiagnosticContext();
         transceiver = new Transceiver(
             id,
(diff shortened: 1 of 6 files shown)


bbaldino:
> LGTM.
> 
> Does this mean we can remove the deprecated flavors of the APIs from jmt?

Yeah, believe so.  Do you think I should do that change and update the JMT version in the pom as part of this PR?

JonathanLennox:
> > Does this mean we can remove the deprecated flavors of the APIs from jmt?
> 
> Yeah, believe so. Do you think I should do that change and update the JMT version in the pom as part of this PR?

Probably yes, to make sure we didn't miss something using the deprecated APIs.

bbaldino:
> > > Does this mean we can remove the deprecated flavors of the APIs from jmt?
> > 
> > 
> > Yeah, believe so. Do you think I should do that change and update the JMT version in the pom as part of this PR?
> 
> Probably yes, to make sure we didn't miss something using the deprecated APIs.

And indeed there was!  Will open the JMT PR, too.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -214,34 +219,37 @@
...
-                logger,
-                Clock.systemUTC());
+        transceiver = new Transceiver(
+            id,
+            TaskPools.CPU_POOL,
+            TaskPools.CPU_POOL,
+            TaskPools.SCHEDULED_POOL,
+            diagnosticContext,
+            logger,
+            Clock.systemUTC()
```
JonathanLennox: Presumably you should now pass `clock` to `Transceiver` rather than using `Clock.systemUTC()` explicitly?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -104,8 +104,7 @@
     /**
      * The time at which this endpoint was created (in millis since epoch)
      */
-    //TODO: use clock/instant in this file
-    private final Long creationTimeMillis;
+    private final Instant creationTimeMillis;
```
JonathanLennox: Shouldn't have `Millis` in the name if it's an `Instant`.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -622,21 +621,17 @@ public boolean shouldExpire()
...
+
+        Instant lastActivity = getLastActivity();
+        Instant now = clock.instant();
+        if (lastActivity == ClockUtils.NEVER)
         {
-            Duration timeSinceCreation = Duration.ofMillis(now - creationTimeMillis);
+            Duration timeSinceCreation = Duration.between(creationTimeMillis, now);
             if (timeSinceCreation.compareTo(EP_TIMEOUT) > 0) {
                 logger.info("Endpoint's ICE connection has neither failed nor connected " +
                     "after " + timeSinceCreation + ", expiring");
```
JonathanLennox: Now we're getting to the point where we might want a more human-friendly equivalent of `Duration.toString()`, but that can be for a separate patch.

bbaldino: sounds good

PR #984, opened 2019-11-19
New config impl

bbaldino:
This code rips out the old jvb config code and uses the new config lib in jitsi-utils-kotlin.


src/main/kotlin/org/jitsi/videobridge/config/JvbConfigk.kt
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jitsi.videobridge.config
+
+import org.jitsi.videobridge.util.config.JvbConfig
+
+class JvbConfigk {
```
bbaldino: in the final impl, this would replace `JvbConfig` (and be named that, no trailing `k`)

src/main/kotlin/org/jitsi/videobridge/health/config/HealthConfigk.kt
```diff
@@ -0,0 +1,69 @@
...
+package org.jitsi.videobridge.health.config
+
+import org.jitsi.utils.configk.dsl.multiProperty
+import org.jitsi.videobridge.config.JvbConfigk
+import org.jitsi.videobridge.config.legacyProperty
+import org.jitsi.videobridge.config.newProperty
+import org.jitsi.videobridge.config.simple
+import java.time.Duration
+
+class HealthConfigk {
```
bbaldino: in the real impl, this would replace `HealthConfig` and would be named that (no trailing `k`)

bgrozev:
Why do you have the old config overriding the new config? I'd expect the opposite.

src/main/kotlin/org/jitsi/videobridge/config/SimpleConfig.kt
```diff
@@ -0,0 +1,58 @@
...
+    legacyName: String,
+    newName: String,
+    readOnce: Boolean,
+    // Allow these to be passed in for testing
+    legacyConfig: ConfigSource = JvbConfig.legacyConfig,
+    newConfig: ConfigSource = JvbConfig.newConfig
+
+) : ConfigProperty<T> {
+    private val multiProp = MultiConfigPropertyBuilder(valueType).apply {
+        property {
```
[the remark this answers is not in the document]
bbaldino: I can't here because `multiProperty` requires a reified generic type to derive the value type, i.e.:

```
// reified T let's us do T::class in the method to get a KClass<T>
inline fun multiProperty<reified T : Any>(block: ...)
```

but we can't get that in a class (as it's only available to inline functions).  I'd have to create another `multiProperty` which could take in the valueType directly, like:
```
fun multiProperty<T : Any>(valueType: KCLass<T>, block: ...)
```
and call it as:
```
private val multiProp = multiProperty(valueType) {
...
}
```
but not sure it's worth it to avoid using the builder.

bbaldino:
> Why do you have the old config overriding the new config? I'd expect the opposite.

Since the new config is always present (due to `reference.conf` acting as defaults and shipping in the code), if we checked it first then we'd never see any overrides they may have had in their legacy config file (since we'd always find a value in `reference.conf` and then stop looking).  So we have to check for a value in the old config first.  Eventually we'll mark the old ones as deprecated (and print the warning) and then eventually we'd stop checking for the old file altogether.

PR #987, opened 2019-11-25
update jitsi-utils

bbaldino:
this change takes in the configuration service implementation moving to jitsi-utils (from libjitsi)


bbaldino:
jenkins, test this please

bbaldino:
the reformatting of the pom.xml file here is because i was testing `xmlstarlet` to change the libjitsi version for the build script, and it corrected some indentation, so i just left it.

PR #991, opened 2019-12-03
Move common config logic to Jicoco

bbaldino:
depends on https://github.com/jitsi/jicoco/pull/59


bgrozev:
LGTM once we have the jicoco and jmt versions

PR #992, opened 2019-12-04
Port bitrate controller config props

bbaldino:
This is built on top of the changes [here](https://github.com/jitsi/jitsi-videobridge/pull/991), so can ignore all but the last commit (but this will need a rebase before merging)


bgrozev:
LGTM

PR #993, opened 2019-12-04
Correct method for checking endpoint activity

Mikhalevich:
Changed the checking endpoint activity logic according to https://github.com/jitsi/jitsi-media-transform/pull/154

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-media-transform</artifactId>
-      <version>1.0-64-gb9dccad</version>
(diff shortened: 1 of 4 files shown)


Mikhalevich:
@bbaldino please review

Mikhalevich:
> Think this looks good. One question: was the failure to expire happening every time? or is there some scenario that triggers it? I only ask because I don't think we'd seen the no-expiration issue, but maybe it happens and just gets cleaned up when the last participant leaves?

Yes, it's happens every time when for example one participant loose connection to the bridge somehow (in my tests I just terminate a browser).
Yes when last active participant leaving from conference all dangling endpoints will be cleaned up.

Mikhalevich:
thx guys

PR #994, opened 2019-12-04
port bandwidth probing config

bbaldino:
I referred to the props differently here: did a static import instead of using the class.  It's nice that it shortens things a bit, and I think the import still makes it clear.  If we like this, then can fix the other ones at some point.


CONFIG.md
```diff
@@ -51,3 +51,4 @@ This list will be updated as properties are migrated:
 | org.jitsi.videobridge.ONSTAGE_PREFERRED_FRAME_RATE | videobridge.cc.onstage-preferred-framerate | |
 | org.jitsi.videobridge.ENABLE_ONSTAGE_VIDEO_SUSPEND | videobridge.cc.enable-onstage-video-suspend | |
 | org.jitsi.videobridge.TRUST_BWE | videobridge.cc.trust-bwe | |
+| org.jitsi.videobridge.PADDING_PERIOD_MS | videobridge.cc.padding-period | |
```
JonathanLennox: This should probably have the comment that this is a duration, like the other properties have.

bbaldino: yup, missed that.

PR #995, opened 2019-12-05
use new legacy config service shim, remove libjitsi dependency

bbaldino:
This change installs the legacy configuration service shim defined in jicoco  as the supplier for `ConfigurationService`, removing the need for libjitsi in JVB 💥  

Depends on https://github.com/jitsi/jicoco/pull/61

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <smack.version>4.2.4-47d17fc</smack.version>
     <jitsi.utils.version>1.0-24-g8f17c70</jitsi.utils.version>
-    <jicoco.version>1.1-SNAPSHOT</jicoco.version>
+    <jicoco.version>1.1-12-g2b135ca</jicoco.version>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/videobridge/osgi/ConfigurationActivator.java
```diff
@@ -37,19 +37,11 @@
...
-                    ConfigurationService.class.getName(),
-                    cfg,
-                    null);
-            logger.info("Registered the LibJitsi ConfigurationService in OSGi.");
-        }
-        else
-        {
-            logger.warn("Failed to register the Configuration service.");
-        }
+        bundleContext.registerService(
```
bgrozev: I wonder if this should now move to jicoco. Presumably jicofo will want to do the exact same thing.

bbaldino: At this rate I don't know when we'll even get to porting Jicofo, but I agree and can move it there if/when do the Jicofo work.

pom.xml
```diff
@@ -139,12 +139,6 @@
       <artifactId>jitsi-utils-kotlin</artifactId>
       <version>${jitsi.utils.version}</version>
     </dependency>
-    <dependency>
```
bgrozev: 💥 💥 💥

gpolitis: holy moly, nice!

PR #998, opened 2019-12-06
port websocket service config

bbaldino:


bbaldino:
Not ready yet, tweaking something.

bbaldino:
I played with a couple ways to implement "conditional" properties (props that should only be read based on the value of some other property, and for which no default may make sense, so we can't set one).

1) The one here in the PR:  "conditional" properties are set as lazy and are only read once accessed, meaning if the user only calls them when the other condition is true, they must be set; but otherwise they aren't and won't throw.

The advantage of this solution was simplicity.  We aren't forced to set some arbitrary default for a property where a default doesn't make sense, and we won't throw in the case where it doesn't need to be set and isn't (assuming the user doesn't try and access it)

2) A "conditionalProperty" helper which made things a bit more explicit.  An example implementation of the websocket `domain` config param using this method looked like:

```
private val domainProp = conditionalProperty(
    ::enabled,
    {
        simple<String>(
            readOnce = true,
            legacyName = "org.jitsi.videobridge.rest.COLIBRI_WS_DOMAIN",
            newName = "videobridge.websockets.domain"
        )
    },
    "Websocket domain only set if websockets are enabled"
)
```
Where the definition of `conditionalProperty` was:
```
/**
 * Define a property which only exists based on the result of the given
 * [predicate].  If [predicate] returns true, we define the prop via the
 * given [propDef], if not, we define a property which will throw on access
 * and use the given [notEnabledMessage]
 */
fun <T : Any >conditionalProperty(
    predicate: () -> Boolean,
    propDef: () -> ConfigProperty<T>,
    notEnabledMessage: String
): ConfigProperty<T> {
    return if (predicate()) {
        propDef()
    } else {
        object : ConfigProperty<T> {
            override val value: T
                get() = throw Exception("Property not enabled: $notEnabledMessage")
        }
    }
}
```

This achieves the same things, but makes the conditional nature a bit more explicit.  I'm not sure it's necessary, but I can change the PR to use this (or some other idea) if we think it'd be best.

src/main/resources/reference.conf
```diff
@@ -38,4 +38,11 @@ videobridge {
       }
     }
   }
+  websockets {
+    enabled=false
+    # Must be set when enabled = true
+    # server-id="default-id"
```
bgrozev: We could actually have a default value for this one

bbaldino: You're right: my first thought was that these are either there or not (depending on if enabled is true), but when enabled _is_ true, a default makes sense here but not for, say, domain.  I forgot about that situation.

bbaldino: What about the `tls` setting?

src/main/resources/reference.conf
```diff
@@ -38,4 +38,11 @@ videobridge {
...
+    # domain="some-domain"
+    # tls=true
```
bgrozev: This is an interesting case. The original code handles the case of a missing value gracefully by following the jetty configuration. This doesn't fit with the new idea of requiring values for all properties. We could:
1. Modify `useTls` to return `Boolean?`, catch the exception and return `null` if the property is not defined. 
2. Have a special value (e.g. the string `default`) in `reference.conf`.

I think 1 is cleaner, even though it requires catching an exception.

bbaldino: Ah yes, I forgot to take into account the passed value of `tls`.  I'll go with option 1 here.

bbaldino: It does feel like this is a legitimate use case for an 'optional' property.  I had originally done this by wiring 'optional' into the core code, but we could also do it with a layer on "top" here: by swallowing the exception and returning null.  Since `ConfigProperty` requires a non-nullable type on the generic, it'd be on top of the config code entirely.  I think I'll do the 'on top' version here, and we can see how much more this comes up and maybe put it back in the core config code.

bgrozev:
I like both approaches. I think your comment (`should only be accessed after verifying [enabled] is true`) is essential, so making it explicit would be nice. Hence, slight preference for 2.

bbaldino:
> I like both approaches. I think your comment (`should only be accessed after verifying [enabled] is true`) is essential, so making it explicit would be nice. Hence, slight preference for 2.

Cool, changed them over and tweaked the semantics around `useTls`

bgrozev:
LGTM. Are you going to move `conditionalProperty` to jicoco now or later?

bbaldino:
closed in favor of https://github.com/jitsi/jitsi-videobridge/pull/1004

PR #999, opened 2019-12-06
change rest api sys prop name

bbaldino:
The new config converts .properties values into a HOCON object, which is slightly lossy compared to a properties file. For example, in a properties like so:
```
a.b.c = true
a.b.c.d = 42
```
Is fine, but in a HOCON object there's an issue: `a.b.c` has to be one of the two values (a boolean or an object), so the value of `a.b.c` ends up as an object, making it impossible to test as a boolean.

This scenario was happening with `org.jitsi.videobridge.rest`.  Soon, we'll port the handling of the API properties to new config, but for now we'll change the value so there's no conflict.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -115,9 +115,16 @@ public class Videobridge
     /**
      * The (base) <tt>System</tt> and/or <tt>ConfigurationService</tt> property
      * of the REST-like HTTP/JSON API of Jitsi Videobridge.
+     *
+     * NOTE: Previously, the rest API name ("org.jitsi.videobridge.rest")
+     * conflicted with other values in the new config, since we have
+     * other properties scoped *under* "org.jitsi.videobridge.rest".   The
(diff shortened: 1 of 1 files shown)


damencho:
But does that mean that all that had enabled it now will have it disabled on upgrade? 
@bbaldino

bbaldino:
> But does that mean that all that had enabled it now will have it disabled on upgrade?
> @bbaldino

The command-line arg will continue to work as it did before.  The only thing that wouldn't work is if it was set only in the sip-communicator.properties file.  I talked with @bgrozev about this and he felt we hadn't ever really advertised the usage there so it seemed reasonable to make the change.

damencho:
Nope, it completely breaks the old deployments on upgrade.
2019-12-09 11:36:42.953 INFO: [11] AbstractJettyBundleActivator.start#613: Not starting the Jetty service for org.jitsi.videobridge.rest.PublicClearPortRedirectBundleActivator(port=8080)

damencho:
Nope, it was not broken in this PR as I installed and 2.1-37-g2c02cf5 and saw the same. I will create two internal issues with what I saw that is broken so we can fix it.

PR #1000, opened 2019-12-09
fix: Fixes setting the octo endpoints.

bgrozev:


bgrozev:
Fixes a regression introduced in #996

PR #1001, opened 2019-12-11
Uses new class-style config properties

bbaldino:
Adapts to jitsi/jicoco#63 and jitsi/jitsi-utils#32


src/main/kotlin/org/jitsi/videobridge/config/CommandLineArgsConfigSource.kt
```diff
@@ -25,4 +25,12 @@ class CommandLineArgsConfigSource(private val commandLineArgs: Array<String>) :
     override fun <T : Any> getterFor(valueType: KClass<T>): (String) -> T {
         TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
     }
+
+    override fun reload() {
+        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
```
bgrozev: Maybe remove the template string?

src/main/resources/application.conf
```diff
@@ -0,0 +1,20 @@
+jmt {
```
bgrozev: Does this replace the default values from `reference.conf` in JMT?

src/test/kotlin/org/jitsi/videobridge/JitsiConfigTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.jitsi.config.JitsiConfigFactory
+import org.jitsi.utils.config.ConfigSource
+import org.jitsi.videobridge.testutils.ConfigSourceWrapper
+
+/**
+ * This class is/should be a duplicate of the one in jicoco-kotlin.  We can't
+ * move it to jicoco-test-kotlin because it would cause a circular maven
+ * dependency
+ */
+abstract class JitsiConfigTest : ConfigTest() {
```
bgrozev: We already have this class in jicoco: https://github.com/jitsi/jicoco/pull/63/files#diff-c83b2ca7ca45151e7b86062f3434064f

src/main/resources/application.conf
bbaldino: Oops, this wasn't meant to be committed yet.

src/test/kotlin/org/jitsi/videobridge/JitsiConfigTest.kt
bbaldino: Yes, but since it's a test class I didn't want to define it there (and I couldn't put it in `jicoco-test-kotlin`).  I guess I can put it in the main jicoco-kotlin code, even though it's for testing.  Is that what you think would be best?

src/test/kotlin/org/jitsi/videobridge/health/config/HealthConfigTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.jitsi.config.DurationToLongMockConfigValueGenerator
+import org.jitsi.config.LongMockConfigValueGenerator
+import org.jitsi.config.runBasicTests
+import org.jitsi.videobridge.JitsiConfigTest
+
+class HealthConfigTest : JitsiConfigTest() {
+    init {
+        "health interval" {
+            runBasicTests(
+                legacyConfigName = "org.jitsi.videobridge.health.INTERVAL",
```
bgrozev: I don't understand the purpose of this test. The internals of the config code itself are already tested in jicoco, so essentially what the test does is compare the strings in `HealthIntervalProperty` to the string literals here, i.e. if forces us to keep two synchronized copies of the strings. Am I missing something?

src/test/kotlin/org/jitsi/videobridge/JitsiConfigTest.kt
bgrozev: If it doesn't bring in a lot of other classes, I think moving to jicoco-kotlin would be better than having two copies.

src/test/kotlin/org/jitsi/videobridge/health/config/HealthConfigTest.kt
bbaldino: Yeah, this is part of the reason why I originally thought we could get a way without tests for these properties.  You're not wrong: the health config properties don't make for the most compelling test.  But we do verify that the property class is properly wired in to both configs and is checking them in the correct priority order.  That being said, you'll notice I didn't do tests for the other ones here, largely because it didn't feel all that necessary (though we always can).  Things get a bit more interesting with the test for the websocket config (which I have in another branch and will put in a follow-up PR).

As far as the strings, I agree it's annoying to have to copy them.  I have some ideas on how to improve that, but it'd be when I bring back the config validation (which I've begun playing with).

src/test/kotlin/org/jitsi/videobridge/JitsiConfigTest.kt
bbaldino: Hmm, unfortunately it'd drag the kotlintest framework along with it.

bbaldino: Honestly the static singleton (`JitsiConfig`) really sucks.  It makes testing a pain and I'd really love to get rid of it (which would also likely eliminate the need for this sort of thing).

PR #1005, opened 2019-12-16
Unsubscribe endpoint from EncodingsManager on expiration.

JonathanLennox:


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -691,6 +691,7 @@ public void expire()
         }
         bandwidthProbing.enabled = false;
         recurringRunnableExecutor.deRegisterRecurringRunnable(bandwidthProbing);
+        getConference().encodingsManager.unsubscribe(this);
```
bgrozev: `AbstractEndpoint` implements `EncodingsManager.EncodingsUpdateListener`, and there is an implementation in `OctoEndpoint` as well. My intention was for Octo endpoint to also subscribe (though I need to verify the detils). I think we should move both `subscribe` and `unsubscribe` to `AbstractEndpoint`.

PR #1006, opened 2019-12-16
Fixes accidental object retention in lambda.

gpolitis:


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -732,13 +732,14 @@ private synchronized void update()
                 return adaptiveTrackProjection;
             }
 
+            final String endpointID = trackBitrateAllocation.endpointID;
```
JonathanLennox: Maybe comment this? Otherwise the rationale and the problem aren't obvious, at least to me.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -732,6 +732,13 @@ private synchronized void update()
                 return adaptiveTrackProjection;
             }
 
+            // XXX the lambda keeps a reference to the trackBitrateAllocation
```
bbaldino: What is the `XXX` convention?

PR #1009, opened 2019-12-19
Migrate to using mucs for jvb

damencho:
Not sure which one to merge first, I think tests that test the debian package deployment will fail till we have all 3 components jitsi-meet, jicofo and jvb changes merged.
Once this is merged I will cherry-pick it for jvb2.

https://github.com/jitsi/jitsi-meet/pull/4956
https://github.com/jitsi/jicofo/pull/413

### DIFF
--- a/lib/logging.properties
+++ b/lib/logging.properties
@@ -23,3 +23,5 @@ com.agafua.syslog.SyslogHandler.escapeNewlines = false
 
 # to disable double timestamps in syslog uncomment next line
 #net.java.sip.communicator.util.ScLogFormatter.disableTimestamp=true
+
+org.jitsi.xmpp.mucclient.MucClientManager.level=SEVERE
--- a/resources/install/debian/postinst
+++ b/resources/install/debian/postinst
@@ -56,7 +56,7 @@ case "$1" in
             echo "JVB_SECRET=$JVB_SECRET" >> $CONFIG
             echo >> $CONFIG
             echo '# extra options to pass to the JVB daemon' >> $CONFIG
-            echo "JVB_OPTS=\"\"" >> $CONFIG
+            echo "JVB_OPTS=\"--apis=rest\"" >> $CONFIG
             echo >> $CONFIG
 
         fi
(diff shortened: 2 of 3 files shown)


lib/logging.properties
```diff
@@ -23,3 +23,5 @@ com.agafua.syslog.SyslogHandler.escapeNewlines = false
 
 # to disable double timestamps in syslog uncomment next line
 #net.java.sip.communicator.util.ScLogFormatter.disableTimestamp=true
+
+org.jitsi.xmpp.mucclient.MucClientManager.level=SEVERE
```
bgrozev: Is this to silence "Setting a presence extension" logs?

resources/install/debian/postinst
```diff
@@ -96,8 +96,23 @@ case "$1" in
         # and this breaks startup script
         sed -i 's/\$JVB_EXTRA_JVM_PARAMS//g' $CONFIG
 
-        if ! grep -q "org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP" "$NEW_JITSI_CONFIG" ;then
-            echo "org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth.${JVB_HOSTNAME}/.*" >> $NEW_JITSI_CONFIG
+        if ! grep -q "#org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP" "$NEW_JITSI_CONFIG" ;then
```
bgrozev: Is the `if` necessary? The only case where it makes a difference is when the file contains both the commented-out and non-commented-out version.

Does this script execute when the debian package is updated?

resources/install/debian/postinst
```diff
@@ -96,8 +96,23 @@ case "$1" in
...
+
+        if ! grep -q "org.jitsi.videobridge.xmpp.user.shard.MUC_JIDS" "$NEW_JITSI_CONFIG" ;then
+            echo "org.jitsi.videobridge.ENABLE_STATISTICS=true" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.STATISTICS_TRANSPORT=muc" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.HOSTNAME=localhost" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.DOMAIN=auth.$JVB_HOSTNAME" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.USERNAME=jvb" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.PASSWORD=$JVB_SECRET" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.MUC_JIDS=JvbBrewery@internal.auth.$JVB_HOSTNAME" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.MUC_NICKNAME=$(uuidgen)" >> $NEW_JITSI_CONFIG
```
bgrozev: I don't think we need anything unique here, just "jvb" will do.

resources/install/debian/postinst
```diff
@@ -56,7 +56,7 @@ case "$1" in
             echo "JVB_SECRET=$JVB_SECRET" >> $CONFIG
             echo >> $CONFIG
             echo '# extra options to pass to the JVB daemon' >> $CONFIG
-            echo "JVB_OPTS=\"\"" >> $CONFIG
+            echo "JVB_OPTS=\"--apis=rest\"" >> $CONFIG
```
bgrozev: Does this change the default value of the rest interface?

lib/logging.properties
damencho: Exactly

resources/install/debian/postinst
damencho: Yes, this is executed on every update. The goal of this is to make current configs work(migrate them) and give people an option to anyway have something in there and we do not override it on update(having commented and not commented line). It is not documented but it is an option if someone complains about something like that.

resources/install/debian/postinst
damencho: What about adding a second jvb. You can do the standard installation on one machine(jitsi-meet) and on a second one just install jvb ...

resources/install/debian/postinst
damencho: Well, the intention is to turn off the component, as if we leave it, it fails health check. 
When I was initially adding the mucs support I added --no-component param for jigasi and jvb, but I think we merged only the jigasi part.

resources/install/debian/postinst
bgrozev: Okay..

resources/install/debian/postinst
bgrozev: I thought this required manual configuration anyway. I don't mind using a uuid

resources/install/debian/postinst
bgrozev: My question is about the rest interface, not the component. Does a default debian installation come with rest enabled or disabled? It looks like it used to be disabled, and now it's enabled.

damencho: Yeah, that's true ... I'm not sure how to disable the component another way. I will take a look.

bgrozev: I think just passing `--apis=` should do it

lib/logging.properties
bbaldino: Should we move that log to a lower level instead?

bgrozev: We could do it in addition. This PR targets `jvb_1.0` and can't use an updated jicofo (unless we maintain a branch, but that would be a lot of hassle to silence some logs)

resources/install/debian/postinst
damencho: > I think just passing `--apis=` should do it

@bgrozev that doesn't work

```
JVB 2020-01-07 17:46:15.489 SEVERE: [33] org.jitsi.meet.ComponentMain.log() host-unknown, host:localhost, port:5347
org.xmpp.component.ComponentException: host-unknown
	at org.jivesoftware.whack.ExternalComponent.connect(ExternalComponent.java:243)
```

damencho: Hum, but `--apis=,` works :) I will change it to that

bgrozev: Nice find :)

PR #1011, opened 2019-12-20
use maven shade plugin instead of maven assembly plugin

bbaldino:
Not sure this is ready yet, at this point I just want it to go through tests to see if it breaks anything there.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,6 +10,7 @@
   <properties>
     <assembly.skipAssembly>true</assembly.skipAssembly>
     <exec.mainClass>org.jitsi.videobridge.Main</exec.mainClass>
+
     <jetty.version>9.4.15.v20190215</jetty.version>
     <kotlin.version>1.3.21</kotlin.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


JonathanLennox:
With this change, it looks like the file that used to be called `jitsi-videobridge-2.1-SNAPSHOT.jar` is now called `original-jitsi-videobridge-2.1-SNAPSHOT.jar`, and the former `jitsi-videobridge-2.1-SNAPSHOT-jar-with-dependencies.jar` is now `jitsi-videobridge-2.1-SNAPSHOT.jar`.  This doesn't seem like what we want?

This now no longer generates platform-specific zip files, but more than that, there's now no longer any zip file at all.  The zip files were what used to contain `jvb.sh`, `lib/logging.properties`, and `lib/videobridge.rc`.

Since we no longer generate platform-specific zip packages, we'd need to also update at least:
* The ant script (`build.xml`) - probably this should be replaced entirely?
* `make-run-test-projects.sh` in the PR testing script

bgrozev:
The zip packages are also distributed here, though 2.0 versions are missing and I don't think anyone's complained:
https://download.jitsi.org/jitsi-videobridge/linux/

@damencho are these also used in the process of generating `.deb`s?

damencho:
Nope, those are not used for debian packages, but I know there are many people using those.

bbaldino:
I think for things like the zips and debian packages we should just do them separately (after the maven build).  The maven build/pom is complicated enough, and those things seem like separate, subsequent steps anyway, no?  Is there any reason they need to be part of the maven build itself?

bbaldino:
replaced by https://github.com/jitsi/jitsi-videobridge/pull/1045

PR #1015, opened 2020-01-02
Improve performance of the new frame projection code.

JonathanLennox:


bbaldino:
Seems reasonable to me.  Why not pass the loggers in anymore?

JonathanLennox:
> Seems reasonable to me. Why not pass the loggers in anymore?

They're not used, and constructing a logger per-frame is expensive.

If we need a logger in `VP8FrameProjection` for something going forward, passing in a parent logger should be fine, but I don't think we want to construct a child logger per frame.

bbaldino:
> > Seems reasonable to me. Why not pass the loggers in anymore?
> 
> They're not used, and constructing a logger per-frame is expensive.
> 
> If we need a logger in `VP8FrameProjection` for something going forward, passing in a parent logger should be fine, but I don't think we want to construct a child logger per frame.

Ah definitely, didn't realize we were creating one per frame.  Sgtm.

PR #1018, opened 2020-01-03
Add ktlint checks to verify; fix existing ktlint issues

bbaldino:


bbaldino:
> Do you want an .editorconfig with `disabled_rules=import-ordering`?

I forgot about this.  I'll add that.

PR #1022, opened 2020-01-07
Prevent null pointer exception.

JonathanLennox:
When AdaptiveTrackProjection.getSource() (which uses a weak ref) returns null.


gpolitis:
Thinking about this a bit more, NPE spamming at this point indicates something's gone really wrong (potentially with signaling?) and the bridge may drop packets that it should forward. I'm worried we're masking some underlying issue by preventing the NPE.

PR #1023, opened 2020-01-07
Handle and ignore an obscure corner case in the frame projection unit tests.

JonathanLennox:


src/test/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionTest.java
```diff
@@ -124,121 +124,137 @@ private void runInOrderTest(Vp8PacketGenerator generator, int targetIndex)
...
+        for (int i = 0; i < reorderSize; i++)
+        {
+            buffer.add(generator.nextPacket());
+        }
 
-            for (int i = 0; i < reorderSize; i++)
-            {
-                buffer.add(generator.nextPacket());
-            }
+        Random random = new Random(seed);
```
bbaldino: kotlintest actually has some support for this (random but including some pre-defined constants), maybe that'd work well here? https://github.com/kotlintest/kotlintest/blob/master/doc/reference.md#custom-generators

JonathanLennox: Well, this is a Java test case...

bbaldino: Oh, oops.

JonathanLennox:
BTW because indentation changed this looks like a lot more changed than really did; I suggest reviewing the diff with the "hide whitespace changes" checkbox set.

PR #1024, opened 2020-01-08
Migrate to using mucs for jvb (#1009)

damencho:
* Migrates to use mucs by default.

* Disables just component and leaves defaults.

### DIFF
--- a/resources/install/debian/postinst
+++ b/resources/install/debian/postinst
@@ -111,8 +111,13 @@ case "$1" in
             echo "org.jitsi.videobridge.xmpp.user.shard.MUC_NICKNAME=$(uuidgen)" >> $NEW_JITSI_CONFIG
         fi
 
-        if ! grep -q -e "--apis=," "$CONFIG" ;then
-            sed -i 's/JVB_OPTS=\"\"/JVB_OPTS=\"--apis=,\"/g' $CONFIG
+        APIS_VALUE=$(grep -oPe '--apis=\K(\w*,*\w*)' $CONFIG)
+        if [[ $APIS_VALUE == *"xmpp"* ]]; then
+            APIS_NEW_VALUE=${APIS_VALUE/xmpp/}
+            if [ -z "$APIS_NEW_VALUE" ]; then
+                APIS_NEW_VALUE=","
+            fi
+            sed -i "s/--apis=$APIS_VALUE/--apis=$APIS_NEW_VALUE/g" $CONFIG
         fi
 
         # we don't want to start the daemon as root
--- a/resources/install/debian/postinst
+++ b/resources/install/debian/postinst
@@ -111,7 +111,7 @@ case "$1" in
             echo "org.jitsi.videobridge.xmpp.user.shard.MUC_NICKNAME=$(uuidgen)" >> $NEW_JITSI_CONFIG
         fi
 
-        if ! grep -q "--apis=," "$CONFIG" ;then
+        if ! grep -q -e "--apis=," "$CONFIG" ;then
             sed -i 's/JVB_OPTS=\"\"/JVB_OPTS=\"--apis=,\"/g' $CONFIG
         fi
 
--- a/resources/install/debian/postinst
+++ b/resources/install/debian/postinst
@@ -56,7 +56,7 @@ case "$1" in
             echo "JVB_SECRET=$JVB_SECRET" >> $CONFIG
             echo >> $CONFIG
             echo '# extra options to pass to the JVB daemon' >> $CONFIG
-            echo "JVB_OPTS=\"\"" >> $CONFIG
+            echo "JVB_OPTS=\"--apis=,\"" >> $CONFIG
             echo >> $CONFIG
 
         fi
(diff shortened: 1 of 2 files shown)


resources/install/debian/postinst
```diff
@@ -116,7 +116,7 @@ case "$1" in
             echo "org.jitsi.videobridge.xmpp.user.shard.MUC_NICKNAME=$(uuidgen)" >> $NEW_JITSI_CONFIG
         fi
 
-        if ! grep -q "--apis=," "$CONFIG" ;then
+        if ! grep -q "\--apis=," "$CONFIG" ;then
```
JonathanLennox: `grep -q -e "--apis=,"` would also work -- -e means "interpret the next arg as the pattern, even if it starts with a dash."

lib/logging.properties
```diff
@@ -37,3 +37,5 @@ timeseries.level=OFF
 timeseries.org.jitsi.videobridge.cc.vp8.level=ALL
 timeseries.useParentHandlers = false
 timeseries.handlers = java.util.logging.FileHandler
+
+org.jitsi.xmpp.mucclient.MucClientManager.level=SEVERE
```
bgrozev: Let me fix the logging instead please (PR coming)

resources/install/debian/postinst
```diff
@@ -96,8 +96,28 @@ case "$1" in
         # and this breaks startup script
         sed -i 's/\$JVB_EXTRA_JVM_PARAMS//g' $CONFIG
 
-        if ! grep -q "org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP" "$NEW_JITSI_CONFIG" ;then
-            echo "org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP=focus@auth.${JVB_HOSTNAME}/.*" >> $NEW_JITSI_CONFIG
+        if ! grep -q "#org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP" "$NEW_JITSI_CONFIG" ;then
+            sed -i 's/org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP/#org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP/g' $NEW_JITSI_CONFIG
+        fi
+
+        # disables tcp harvester to make sure jvb will not take port 443
```
bgrozev: With the new config code switching a default looks like this :)
https://github.com/jitsi/jitsi-videobridge/pull/1030/commits/8f4e05a9d917319dcae988df08d2989fc78eebb2

lib/logging.properties
damencho: I had removed this change.

resources/install/debian/postinst
damencho: I had dropped this commit from this PR.

resources/install/debian/postinst
```diff
@@ -96,8 +96,23 @@ case "$1" in
...
+            echo "org.jitsi.videobridge.STATISTICS_TRANSPORT=muc" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.HOSTNAME=localhost" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.DOMAIN=auth.$JVB_HOSTNAME" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.USERNAME=jvb" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.PASSWORD=$JVB_SECRET" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.MUC_JIDS=JvbBrewery@internal.auth.$JVB_HOSTNAME" >> $NEW_JITSI_CONFIG
+            echo "org.jitsi.videobridge.xmpp.user.shard.MUC_NICKNAME=$(uuidgen)" >> $NEW_JITSI_CONFIG
+        fi
+
+        if ! grep -q -e "--apis=," "$CONFIG" ;then
```
bgrozev: Does this disable REST if it has been enabled manually?

bgrozev: And also remove other `JVB_OPTS` that have been added

damencho: oh, maybe you are right, that can break stuff ... I just need to disable the component if enabled ... I need to re-think it ... We already merged it for jvb1.0 ... probably needs an update and there 
So it can be any of those then:
```
--apis=xmpp
--apis=rest
--apis=xmpp,rest
--apis=rest,xmpp
```
And it must become `--apis=,` or `--apis=rest`.

damencho: @bgrozev WDYT about this:
```
APIS_VALUE=`grep -oPe '--apis=\K(\w*,*\w*)' $CONFIG`
if [[ $APIS_VALUE == *"xmpp"* ]]; then
    APIS_NEW_VALUE=${APIS_VALUE/xmpp/}
    if [ -z "$APIS_NEW_VALUE" ]; then
        APIS_NEW_VALUE=","
    fi
    sed -i 's/--apis=$APIS_VALUE/--apis=$APIS_NEW_VALUE/g' $CONFIG
fi
```
That produces as a result:
```
--apis=,
--apis=rest
--apis=,rest
--apis=rest,
```

PR #1025, opened 2020-01-08
Disables tcp harvester by default.

damencho:


resources/install/debian/postinst
```diff
@@ -100,6 +100,11 @@ case "$1" in
             sed -i 's/org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP/#org.jitsi.videobridge.AUTHORIZED_SOURCE_REGEXP/g' $NEW_JITSI_CONFIG
         fi
 
+        # disables tcp harvester to make sure jvb will not take port 443
```
bgrozev: I think we should migrate the properties to the new config and switch the default there instead of adding more to the legacy config. I volunteer to do it.

damencho: This is for jvb1 :)

bgrozev: D'oh

PR #1030, opened 2020-01-09
Migrates the ice-related configuration

bgrozev:


JonathanLennox:
I notice that the TcpTest in the PR tests is now failing -- which looks like it's deliberate.  Do we need to update the PR tester for this?

bgrozev:
Yes, I've opened a PR to exlude this test

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1002,27 +1002,6 @@ private void startIce4j(
...
-                HarvesterConfiguration.STUN_MAPPING_HARVESTER_ADDRESSES})
-            {
-                String propertyValue = cfg.getString(propertyName);
-
-                if (propertyValue != null)
-                {
-                    String newPropertyName
-                        = newPrefix
-                                + propertyName.substring(oldPrefix.length());
-                    System.setProperty(newPropertyName, propertyValue);
```
bbaldino: So we don't need to inject these as system properties anymore?  I thought we still did.

src/main/java/org/jitsi/videobridge/ice/Harvesters.java
```diff
@@ -0,0 +1,154 @@
...
+            {
+                singlePortHarvesters = null;
+                classLogger.info("No single-port harvesters created.");
+            }
+
+            healthy = singlePortHarvesters != null;
+
+            if (Config.tcpEnabled())
+            {
+                for (int port : Config.tcpPortsToTry())
```
bbaldino: 🎉

src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
```diff
@@ -0,0 +1,185 @@
...
+                    fromConfig(JitsiConfig.newConfig)
+                }
+            )
+            private val tcpFallbackPortProperty = TcpFallbackPortProperty()
+
+            /**
+             * The list of ports to try. Fallback can be disabled by setting the fallback port to <=0.
+             */
+            @JvmStatic
+            fun tcpPortsToTry() = if (tcpFallbackPortProperty.value > 0)
```
bbaldino: Would you mind adding braces or using a `when` here?

src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
```diff
@@ -0,0 +1,185 @@
...
+                legacyName = "org.jitsi.videobridge.TCP_HARVESTER_MAPPED_PORT",
+                newName = "videobridge.ice.tcp.mapped-port"
+            )
+            private val tcpMappedPortProperty = TcpMappedPortProperty()
+
+            /**
+             * Returns the additional port to advertise, or [null] if none is configured.
+             */
+            @JvmStatic
+            fun tcpMappedPort(): Int? = try {
```
bbaldino: Any reason that for `tcpMappedPort` we don't set it at all if we don't want one, but for fallback we use a negative value?  It seems like either scheme would work for both?  Is it that it doesn't make sense to provide a reasonable default for `tcpMappedPort`?

src/main/resources/reference.conf
```diff
@@ -46,4 +46,26 @@ videobridge {
     # Must be set when enabled = true
     # domain="some-domain"
   }
+  ice {
+    tcp {
+      # Whether ICE/TCP is enabled.
+      enabled = false
+      port = 443
```
bbaldino: Comment for this field?

src/main/resources/reference.conf
```diff
@@ -46,4 +46,26 @@ videobridge {
...
+      # Port to fallback to in case binding to 'port' fails. Set to -1 to disable fallback.
+      fallback-port = 4443
```
bbaldino: I'm curious: is there a reason you didn't end up using a list in the new config for the ports?

src/main/resources/reference.conf
```diff
@@ -46,4 +46,26 @@ videobridge {
...
+      # An optional additional port to advertise.
+      # mapped-port = 8443
+      # Whether to use "ssltcp" or plain "tcp".
+      ssltcp = true
+    }
+    # The port for ICE/UDP.
+    port = 10000
```
bbaldino: Should we name the config value something more descriptive to be more self-documenting?  Or nest it within a `UDP` block?

src/main/resources/reference.conf
```diff
@@ -46,4 +46,26 @@ videobridge {
...
+    # An optional prefix to include in STUN username fragments generated by the bridge.
+    # ufrag-prefix = "jvb-123:"
+
+    # Which candidate pairs to keep alive. One of "selected_and_tcp", "selected_only", or "all_succeeded".
```
bbaldino: There's an enum for this, right? Should we refer to it here?

damencho:
Will this allow us to set the google's stun servers by default to discover public/private addresses?

bgrozev:
> Will this allow us to set the google's stun servers by default to discover public/private addresses?

Not yet, for this we need to migrate ice4j itself

src/main/java/org/jitsi/videobridge/Videobridge.java
bgrozev: The code which reads `org.ice4j` properties from our file and injects them as system properties still exists. The removed code reads `org.videobridge` properties and injects them as `org.ice4j`. The properties have been marked deprecated for a long time, and I verified we don't use them anywhere.

src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
bgrozev: Yes, the mapped port is disabled by default, while the fallback port is 4443 by default. Do you prefer setting `mappedPort = -1` in `reference.conf`?

bbaldino: No, I was thinking at some point we should add a `isPresent` or `exists` field to `ConfigProperty` that would do the try/catch work and then stuff like this could use it.  But this is fine.

src/main/resources/reference.conf
bgrozev: Yes, because a list suggests that we're binding on multiple ports.

bbaldino: That's really just a matter of how it's named/documented, no?

src/main/resources/reference.conf
bgrozev: If you are a user reading the config file the raw values are more useful than the class name.

bbaldino: Sure, it's just also more likely to get out of date.  Not opposed to keeping the strings here, but referring to the enum name would be nice.

src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
```diff
@@ -0,0 +1,187 @@
...
+            fun iceSslTcp() = sslTcpProperty.value
+
+            /**
+             * The property that configures the ICE UDP port.
+             */
+            class PortProperty : LegacyFallbackConfigProperty<Int>(
+                Int::class,
+                readOnce = true,
+                legacyName = "org.jitsi.videobridge.SINGLE_PORT_HARVESTER_PORT",
+                newName = "videobridge.ice.port"
```
bbaldino: Think this needs to change as well to match the new location `videobridge.ice.udp.port`

src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
bgrozev: Yes, this would be nice. We could also avoid try/catch and implement it as checks on the `ConfigResult`.

src/main/resources/reference.conf
bgrozev: Exactly. I think `port` and `fallback-port` serve as a good description of what they can do. It's immediately obvious how to change the port. With a list one has to read the documentation to figure out what it means and how to change the port.

I don't think we'll even need more than one fallback port, so what would be the reasons for exposing a list in the config file?

bbaldino: Even for 2 ports, I think a list is more ergonomic as we just iterate through until one is successful.  It also avoids the problem of having to use `-1` if no fallback is desired and, if someone wanted to fallback to more ports they could do so easily.

bgrozev: I don't think we should consider the case of more than two ports, there's no real use for it.

Do you know if the config library provides a way to read either an int or a list of ints? I'd be nice to just have `port = 443`, but allow it to be overriden by `port = [443, 4443]`

bbaldino: I'd say the utility of multiple ports degrades with each one, but I don't think there's a cliff from useful -> not useful between 2 and 3, say.

As for the Int or list of Ints, I think we could do that but we'd have to do it on our own.  Basically we could have a fallback which first tried it as one type and then another, but we'd need to have a common 'final' type which would have to be a list of Ints.

bgrozev:
Merging, because it is only failing the TCP test, which was not properly disabled (it is now).

PR #1031, opened 2020-01-09
attempts to fix the npe spam issue that breaks the media flow by removing the weakref

gpolitis:


bgrozev:
We should definitely remove the weak reference!

I wonder if looping over all endpoints and their tracks has a performance impact. If I understand correctly this code executes every 15ms, which is quite often. We could keep a regular reference to the source track in `AdaptiveTrackProjection`. A `MediaStreamTrackDesc` doesn't reference anything external, so this would be safe.

gpolitis:
> I wonder if looping over all endpoints and their tracks has a performance impact. If I understand correctly this code executes every 15ms, which is quite often. 

yes, I agree that we should optimize this; the simplest thing we could do is have it run less often but we risk breaking probing or cause issues by stressing the link (with a stale padding estimate that is too high).

Do you have any ideas that you'd like me to incorporate in this PR or can we leave this as a future step? (Note that what I'm changing here has the same space/time complexity as the previous code: the previous version had a loop over the track projections whereas now we have a loop over the endpoints, but the number of track projections that we have is equal to the number of endpoints).

gpolitis:
> We could keep a regular reference to the source track in AdaptiveTrackProjection.

I don't feel super confortable about having egress objects strongly referencing ingress objects and vice versa as it could lead to issues in the future, but today there's no risk as you point out so we could do it this way.

JonathanLennox:
@gpolitis : can you do a YourKit profile of a bridge with your patch, to see if the loop over all the endpoints has a significant performance cost? I don't want to spend effort optimizing it or breaking architectural assumptions before we know that it's actually a problem.

JonathanLennox:
I did a YourKit profile on this code, and the performance impact of this seems negligible -- as in, it's so small it's hard to even find, and click on, the column represented by this new code.

So I don't think we need to worry about the performance of George's current PR.

JonathanLennox:
Screen Shot 2020-02-11 at 1 57 49 PM

PR #1032, opened 2020-01-09
Fix JSON of VP8 adaptive track projection context debug output.

JonathanLennox:


src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -646,13 +646,14 @@ public synchronized JSONObject getDebugState()
                 "class",
                 VP8AdaptiveTrackProjectionContext.class.getSimpleName());
 
-        JSONObject[] mapSizes = new JSONObject[vp8FrameMaps.size()];
+        JSONArray mapSizes = new JSONArray();
         int i = 0;
         for (long ssrc: vp8FrameMaps.keySet())
```
bbaldino: Since we end up retrieving the value, we can just iterate of the entries of the map (instead of the keySet)

PR #1037, opened 2020-01-15
port octo region to new config

bbaldino:


src/main/resources/reference.conf
```diff
@@ -38,6 +38,12 @@ videobridge {
       }
     }
   }
+  octo {
+    # A string denoting the 'region' of this JVB.  This region
+    # will be used by Jicofo in the selection of a bridge for
+    # a client by comparing it to the client's region.
+    #region
```
bgrozev: Add a value to the commented example?
`#region = "region1"`?

bbaldino: Sure

src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -95,12 +90,7 @@ public VideobridgeStatistics()
         BundleContext bundleContext
             = StatsManagerBundleActivator.getBundleContext();
 
-        ConfigurationService cfg
-            = ServiceUtils2.getService(bundleContext, ConfigurationService.class);
-        if (cfg != null)
-        {
-            region = cfg.getString(REGION_PNAME, region);
-        }
+        region = Config.region();
```
bgrozev: I suggest accessing it as `OctoConfig.Config.region()` since this the config is not specific to `VideobridgeStatistics`

bbaldino: Good idea, will change that.

bgrozev: Also, the initialization can be moved to the field declaration, it doesn't need to be in the `static` block anymore

PR #1038, opened 2020-01-15
Octo relay config

bbaldino:


src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -38,6 +38,60 @@ class OctoConfig {
...
+                Int::class,
+                "org.jitsi.videobridge.octo.BIND_PORT",
+                "videobridge.octo.bind-port",
+                readOnce = true
+            )
+
+            private val bindPortProp = BindPortProperty()
+
+            @JvmStatic
+            fun bindPort(): Int {
```
bgrozev: We should make the port required and use 4096 as the default

src/main/resources/reference.conf
```diff
@@ -43,6 +43,13 @@ videobridge {
     # will be used by Jicofo in the selection of a bridge for
     # a client by comparing it to the client's region.
     #region="us-west-1"
+    # The address on which the Octo relay should bind
+    #bind-address=198.51.100.1
+    # The port to which the Octo relay should bind
+    #bind-port=4096
+    # The address which controls the public address which
+    # will be part of the Octo relayId
+    #public-address=198.51.100.1
```
bgrozev: If we set a default of `${videobridge.octo.bind-address}`, is it going to work as expected when `bind-address` is overriden in `application.conf`, but `public-address` isn't?

src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -38,6 +38,60 @@ class OctoConfig {
...
+                String::class,
+                "org.jitsi.videobridge.octo.BIND_ADDRESS",
+                "videobridge.octo.bind-address",
+                readOnce = true
+            )
+
+            private val bindAddressProp = BindAddressProperty()
+
+            @JvmStatic
+            fun bindAddress(): String? {
```
bgrozev: Since we're updating the config, can we add a `videobridge.octo.enabled` flag, and make the bind address required?

src/main/resources/reference.conf
bbaldino: The properties are completely independent, so they'd be overridden separately.  What would be expected?

bbaldino:
This has gotten complicated enough that I want to add some tests, but let me know what you think of the code.

src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -16,12 +16,56 @@
...
+                // file.
+                legacyConfigAttributes {
+                    name("org.jitsi.videobridge.octo")
+                    readOnce()
+                    retrievedAs<ConfigObject>() convertedBy {
+                        val cfg = it.toConfig()
+                        if (cfg.hasPath("BIND_ADDRESS") && cfg.hasPath("BIND_PORT")) {
+                            val bindAddress = cfg.getString("BIND_ADDRESS")
+                            val bindPort = cfg.getInt("BIND_PORT")
+                            bindAddress != null && (bindPort.isUnpriviligedPort())
```
bgrozev: This looks so much like the bug in `MediaStreamTracks`.  Maybe change to `return if` to make it clear?

bbaldino: Return isn't allowed here unfortunately

src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -16,12 +16,56 @@
...
+
+            private val enabledProp = EnabledProperty()
+
+            @JvmStatic
+            fun enabled() = enabledProp.value
+
+            class RegionProperty : ConditionalProperty<String>(
+                ::enabled,
+                Region(),
+                "Octo region is only parsed when Octo is enabled"
```
bgrozev: What is the value of `ConditionalProperty` if the getter function will swallow the exception?

src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -38,6 +82,86 @@ class OctoConfig {
                     null
                 }
             }
+
+            class BindAddressProperty : ConditionalProperty<String>(
+                ::enabled,
+                BindAddress(),
+                "Octo bind address is only parsed when Octo is enabled"
```
bgrozev: I don't think the code follows this, it retrieves the address unconditionally.

src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
bbaldino: What do you mean by 'getter function' here, `Region()`?  We try and access the value of that property, which can throw on its own (this doesn't use the `region` getter, but an instance of the property class).

PR #1039, opened 2020-01-15
port expire thread config

bbaldino:


src/main/resources/reference.conf
```diff
@@ -1,4 +1,10 @@
 videobridge {
+  entity-expiration {
+    # If an entity has no activity after this timeout, it is expired
+    timeout=1 minute
+    # The interval at which the videobridge will check for expired entities
+    check-interval=${videobridge.inactivity-timeout}
```
bgrozev: I think this should be `videobridge.entity-expiration.timeout`

PR #1040, opened 2020-01-16
Remove native code building/assembly

bbaldino:
Similar to https://github.com/jitsi/jitsi-videobridge/pull/1011, this should be considered a bit of a 'test': I went in and ripped things out but not sure how best to validate nothing has broken, so starting here with a test run.

As far as I know, all the native code the bridge still relies on has been moved out into other projects (jitsi-sctp, jitsi-srtp) so I don't think we need the code to build it here anymore 🎉 .  This step could actually be a step towards https://github.com/jitsi/jitsi-videobridge/pull/1011.


bbaldino:
Jenkins, test this please.

pom.xml
```diff
@@ -270,99 +270,30 @@
...
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-assembly-plugin</artifactId>
+          <executions>
+            <execution>
+              <id>assembly-native-libraries</id>
```
JonathanLennox: I don't think this id name makes sense for this execution anymore.

pom.xml
```diff
@@ -270,99 +270,30 @@
...
-              <descriptor>src/assembly/windows-x86-bin-archive.xml</descriptor>
-              <finalName>${project.artifactId}-windows-x86-${project.version}</finalName>
-              <skipAssembly>${assembly.skipAssembly}</skipAssembly>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-assembly-plugin</artifactId>
```
JonathanLennox: You don't have a version specified for maven-assembly-plugin -- I realize this isn't a new bug, but you're working in this space now.

As a result, at least in my install, it pulls version 2.2-beta-5, which as far as I can tell from Google is from 2009 (!).

The latest version of it seems to be 3.2.0.

This also fixes an issue of it not being thread-safe, which would let us add -T options (parallelize build) to maven.

pom.xml
```diff
@@ -270,99 +270,30 @@
...
+              <phase>package</phase>
+              <goals>
+                <goal>single</goal>
+              </goals>
+              <configuration>
+                <appendAssemblyId>true</appendAssemblyId>
+                <descriptors>
+                  <descriptor>src/assembly/archive.xml</descriptor>
+                </descriptors>
+                <skipAssembly>${assembly.skipAssembly}</skipAssembly>
```
JonathanLennox: I think skipAssembly was only necessary before because building zip files for so many platforms was so slow.  It now seems pretty speedy (at least after the first time when it downloads a bunch of maven packages), so I think we can remove this step and just always build the zip.

resources/install/jvb.sh
```diff
@@ -32,4 +31,4 @@ fi
 
 if [ -z "$VIDEOBRIDGE_MAX_MEMORY" ]; then VIDEOBRIDGE_MAX_MEMORY=3072m; fi
 
-LD_LIBRARY_PATH=$libs exec java -Xmx$VIDEOBRIDGE_MAX_MEMORY $VIDEOBRIDGE_DEBUG_OPTIONS -XX:+UseConcMarkSweepGC -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp -Djava.library.path=$libs $LOGGING_CONFIG_PARAM $JAVA_SYS_PROPS -cp $cp $mainClass $@
+exec java -Xmx$VIDEOBRIDGE_MAX_MEMORY $VIDEOBRIDGE_DEBUG_OPTIONS -XX:+UseConcMarkSweepGC -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp -Djava.library.path=$LOGGING_CONFIG_PARAM $JAVA_SYS_PROPS -cp $cp $mainClass $@
```
bgrozev: `-Djava.library.path=$LOGGING_CONFIG_PARAM` doesn't see right

bbaldino: Good catch

bbaldino: Fixed

pom.xml
bbaldino: I just tested and do see they're about the same (if the downloads aren't done).  With the downloads it goes from ~50 seconds to ~90 seconds on my machine.  It seems harmless enough to keep, no?  Might still be nice to avoid it sometimes?

JonathanLennox: Once all the downloads have been run, on my (slow) lonely instance, plain `mvn clean install` is 01:09, whereas `mvn clean install -Dassembly.skipAssembly` is 01:23.

At the very least I'd say flip the default?

bbaldino: Sure, done.

bbaldino:
> I see
> 
> ```
> [INFO] --- maven-assembly-plugin:3.2.0:single (assemble-archive) @ jitsi-videobridge ---
> [INFO] Reading assembly descriptor: src/assembly/archive.xml
> [WARNING] The assembly descriptor contains a *nix-specific root-relative reference (starting with slash). This is not portable and might fail on Windows: /
> [WARNING] The assembly descriptor contains a *nix-specific root-relative reference (starting with slash). This is not portable and might fail on Windows: /
> [WARNING] The assembly descriptor contains a *nix-specific root-relative reference (starting with slash). This is not portable and might fail on Windows: /
> [INFO] Building zip: /home/jlennox/Git/jitsi-videobridge/target/jitsi-videobridge-2.1-SNAPSHOT-archive.zip
> ```
> 
> I suspect this is a new warning generated by the newer version of the assembly plug-in, but is it worth investigating? (None of the generated archives seem to be root-relative as far as I can tell.)

Interesting, I didn't notice that.  It looks like we can use `<outputDirectory>${file.separator}</outputDirectory>`.  I'll do that.

bbaldino:
I think build.xml will need some changes to for the deb package.

JonathanLennox:
The deb package should also change to `Architecture: all`, I think, since this package will no longer contain anything arch-specific.

PR #1042, opened 2020-01-20
Remove ant

damencho:


bbaldino:
lgtm, I'd just say let's put a 'build_deb_package.sh` script or whatever in the resources dir that does all the work, so we can just call it from the command line or jenkins and pass in what's needed.

PR #1045, opened 2020-01-21
Combine reference conf correctly when building a fat jar

bbaldino:
This adds a new method for building a fat jar which correctly concatenates `reference.conf` files if there are multiple (which is the case since JVB and JMT have them).

Although building the fat jar only took ~3 seconds longer on my machine, I made it off by default.  It can be built by passing `-P buildFatJar` to maven.

`maven-assembly-plugin` had some support for this, but it didn't support concatenating the files for the final jar, so we no longer use it for building the fat jar, and use `maven-shade-plugin` instead which is recommended in the `maven-assembly-docs`:
```
Note that jar-with-dependencies provides only basic support for uber-jars. For more control, use the Maven Shade Plugin.
```
[here](https://maven.apache.org/plugins/maven-assembly-plugin/descriptor-refs.html#jar-with-dependencies)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,33 @@
     <smack.version>4.2.4-47d17fc</smack.version>
     <jitsi.utils.version>1.0-33-g2ed4090</jitsi.utils.version>
     <jicoco.version>1.1-17-g0f9a752</jicoco.version>
+    <maven-shade-plugin.version>3.2.1</maven-shade-plugin.version>
   </properties>
+
+  <profiles>
+    <profile>
+      <!-- Building a fat jar is disabled by default. To build it
+           pass '-P buildFarJar
+      -->
+      <id>buildFatJar</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -16,7 +16,33 @@
     <smack.version>4.2.4-47d17fc</smack.version>
     <jitsi.utils.version>1.0-33-g2ed4090</jitsi.utils.version>
     <jicoco.version>1.1-17-g0f9a752</jicoco.version>
+    <maven-shade-plugin.version>3.2.1</maven-shade-plugin.version>
   </properties>
+
+  <profiles>
+    <profile>
+      <!-- Building a fat jar is disabled by default. To build it
+           pass '-P buildFarJar
```
JonathanLennox: Typo

bbaldino: fixed

PR #1046, opened 2020-01-21
Optimize VP8FrameMap to use a linear map indexed by picture ID.

JonathanLennox:


src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
```diff
@@ -148,51 +82,31 @@ private FrameInsertionResult doFrameInsert(@NotNull VP8Frame frame, Vp8Packet pa
...
 
-        int seq = packet.getSequenceNumber();
-        if (!cleanupFrameMap(seq))
-            return null;
-
         frame = new VP8Frame(packet);
 
-        vp8FrameMap.put(timestamp, frame);
-        vp8OrderedFrameMap.put(seq, frame);
+        if (!frameHistory.insert(pictureId, frame))
```
bbaldino: braces please

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
```diff
@@ -271,4 +170,127 @@ public boolean isNewFrame()
...
+
+    private static class FrameHistory extends ArrayTracker<VP8Frame>
+    {
+        FrameHistory(int size) {
+            super(size);
+        }
+
+        PictureIdIndexTracker indexTracker = new PictureIdIndexTracker();
+
+        /* TODO: equivalent of rfc3711 tracker */
```
bbaldino: Is this TODO still relevant?

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
```diff
@@ -0,0 +1,213 @@
...
+    protected val syncRoot = Any()
+    /**
+     * The index in [cache] where the item with the highest index is stored.
+     */
+    private var head = -1
+
+    protected val lastIndex: Int
+        get() = if (head == -1) -1 else cache[head].index
+
+    protected var firstIndex: Int = -1
```
bbaldino: Could you add some docs to `firstIndex`, and `lastIndex`?

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
```diff
@@ -0,0 +1,195 @@
...
+import java.lang.Integer.max
+import org.jitsi.nlj.util.floorMod
+import java.lang.Integer.min
+
+/**
+ * Implements a fixed-sized tracker based on a pre-filled array. The main use-case is the frame projection tracker.
+ *
+ * @author Jonathan Lennox, based on [org.jitsi.nlj.util.ArrayCache] by Boris Grozev
+ */
+open class ArrayTracker<T>(
```
bgrozev: What are the differences between `ArrayTracker` and `ArrayCache`? Would it be hard to extract the common logic to avoid duplication?

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
```diff
@@ -0,0 +1,213 @@
...
+     * Called when an item in the cache is replaced/discarded.
+     */
+    protected open fun discardItem(item: T) {}
+
+    /**
+     * Gets an item from the cache with a given index. Returns 'null' if there is no item with this index in the cache.
+     * The item is wrapped in a [Container] to allow access to the time it was added to the cache, and we provide a
+     * copy.
+     */
+    fun getContainer(index: Int): Container? {
```
bbaldino: Can you document what `index` represents here?

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Implements a fixed-sized tracker based on a pre-filled array. The main use-case is the frame projection tracker.
+ *
+ * @author Jonathan Lennox, based on [org.jitsi.nlj.util.ArrayCache] by Boris Grozev
+ */
+open class ArrayTracker<T>(
+    val size: Int
+) {
+    private val cache: Array<Container> = Array(size) { Container() }
+    protected val syncRoot = Any()
```
bbaldino: This doesn't appear to be used anywhere? Do we need synchronization in this class?

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    /**
+     * Called when an item in the cache is replaced/discarded.
+     */
+    protected open fun discardItem(item: T) {}
+
+    /**
+     * Gets an item from the cache with a given index. Returns 'null' if there is no item with this index in the cache.
+     * The item is wrapped in a [Container] to allow access to the time it was added to the cache, and we provide a
+     * copy.
```
bbaldino: I see a couple references to making/returning a copy, but I don't think we're making any copies in here.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
```diff
@@ -271,4 +172,127 @@ public boolean isNewFrame()
...
+                else {
+                    v = roc;
+                }
+
+                if (updateRoc && isNewerThan(seqNum, highestSeqNumReceived))
+                {
+                    highestSeqNumReceived = seqNum;
+                }
+
+                return 0x8000 * v + seqNum;
```
bgrozev: This will not work as expected if the sender uses only 7-bit PictureIDs. I don't think any implementations actually do this, but it is explicitly allowed in RFC7741: `Having sent a 7-bit PictureID with all bits set to 1, the sender may either wrap the PictureID to 0 or extend to 15 bits and continue incrementing.`

Can you think of a way to detect it if it happens, so we can log?

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
JonathanLennox: No, you're right, this is an artifact of forking off `ArrayCache`.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
JonathanLennox: Nope!

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
JonathanLennox: The main difference is that `ArrayTracker` clears out all entries older than `lastIndex - size`, whereas `ArrayCache` doesn't.  (The other differences are all reconcilable, I think.)

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
JonathanLennox: I have some vague recollection that Chrome actually insists on 15-bit PictureIDs (i.e. even if one packet uses 7-bit, it expects the packet after 0x7f to be a 15-bit 0x80).  But now I can't find where this is documented.

Probably we can just treat 7-bit packets as though they didn't include a picture ID at all, since they're optional.  (Which leads to the point that we don't handle that case, currently.)

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
JonathanLennox: Ok.

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
JonathanLennox: Fixed.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
JonathanLennox: Possibly streams without picture IDs should just be handled by the generic tracker?  I'm not sure how much we care about this case, because all the browsers we support will generate them.

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
```diff
@@ -35,9 +35,15 @@ open class ArrayTracker<T>(
      */
     private var head = -1
 
+    /**
+     * The numerically highest index value seen, or -1.
```
bbaldino: Hm, I guess what I had in mind was something more about what an index represents.  If this is as detailed as it can get (because the class is generic) then I can see why you left them off :)

src/main/kotlin/org/jitsi/videobridge/util/ArrayTracker.kt
JonathanLennox: This could probably be added as a conditional on `ArrayCache`, or made possible for a subclass -- do you think it's worth it?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
bgrozev: Okay, if wrapping at 0x7f breaks Chrome it won't be that hard to discover the probem.

> Probably we can just treat 7-bit packets as though they didn't include a picture ID at all

This seems more likely to cause problems in practice (when senders wrap at 0x7fff but encode small values as 7-bit)

src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameMap.java
```diff
@@ -199,32 +205,91 @@ public VP8Frame get(int pictureId)
...
+            {
+                return;
+            }
+            if (lastIndex - oldlastIndex >= getSize())
+            {
+                flush();
+            }
+            for (int i = oldlastIndex + 1; i < lastIndex; i++)
+            {
+                discard(i);
```
bgrozev: I wonder if these actually need to be discarded. Can we get away with just masking them in `get()` (i.e. return `null` if we're asked for an old entry even though one exists).

JonathanLennox: I guess that would work.  Is your goal to avoid the linear-time array clearing?

bgrozev: Looks like the only reason we explicitly discard is to accurately know the number of cached items, which is only used in the debug output?

JonathanLennox: Yes, and I suppose to reduce the lifetime of Vp8Frame objects once they're not reachable.  But thinking about it the time efficiency is probably more important than the memory efficiency here.

bgrozev:
lgtm

PR #1048, opened 2020-01-22
have adaptivetrackprojection refer to the same map of payload types used

bbaldino:
by the bitratecontroller to avoid any out-of-sync issues.

There existed a race between creating an `AdaptiveTrackProjection` and being notified of a new payload type that resulted in the `AdaptiveTrackProjection` not knowing of a payload type and logging `AdaptiveTrackProjection.getContext#278: No payload type object signaled for payload type...`.  This change removes the chance for the two payload type maps (one in `BitrateController`, one in `AdaptiveTrackProjection`) getting out of sync by using only a single map.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
+++ b/src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
@@ -113,26 +113,29 @@ public class AdaptiveTrackProjection
      */
     private int targetIndex = RTPEncodingDesc.SUSPENDED_INDEX;
 
-    //TODO(brian): we need this to know which frameprojectioncontext to make
-    // based on the payload type of a packet.  is there a better way?
-    private final Map<Byte, PayloadType> payloadTypes = new HashMap<>();
+    private final Map<Byte, PayloadType> payloadTypes;
 
     /**
      * Ctor.
      *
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -113,26 +113,29 @@
      */
     private int targetIndex = RTPEncodingDesc.SUSPENDED_INDEX;
 
-    //TODO(brian): we need this to know which frameprojectioncontext to make
-    // based on the payload type of a packet.  is there a better way?
-    private final Map<Byte, PayloadType> payloadTypes = new HashMap<>();
+    private final Map<Byte, PayloadType> payloadTypes;
```
JonathanLennox: Do we want to declare this as type `ConcurrentMap<Byte, PayloadType>` (and similarly in the constructor and in BitrateController) to ensure that the API enforces the thread-safety we need?

bbaldino: I thought it was best to leave that responsibility to the caller and leave this as the more basic type.  At the end of the day this class just expects to be able to read from this map to find things, and there are a couple ways a caller might implement that safely so the flexibility seemed desirable.

PR #1049, opened 2020-01-22
Reduce log level of 'unknown ssrc' log in BitrateController

bbaldino:
This log isn't completely unexpected, and can happen for a brief period when an endpoint first joins (before it's updated about all other endpoints in the conference), so reduce the log level and track the number of occurrences in a stat.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -601,6 +601,7 @@ public class Endpoint
             if (logger.isDebugEnabled() && getConference().includeInStatistics())
             {
                 logger.debug(transceiver.getNodeStats().prettyPrint(0));
+                logger.debug(bitrateController.getDebugState().toJSONString());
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -601,6 +601,7 @@ public void expire()
             if (logger.isDebugEnabled() && getConference().includeInStatistics())
             {
                 logger.debug(transceiver.getNodeStats().prettyPrint(0));
+                logger.debug(bitrateController.getDebugState().toJSONString());
```
JonathanLennox: Why isn't this method just using `Endpoint.getDebugState()` here anyway?

bbaldino: One unfortunate thing is that JSONObject doesn't support pretty-printing, so unfortunately the output is much less friendly to read in the logs.

PR #1050, opened 2020-01-23
Adapts to JMT (renamed fields).

bgrozev:


JonathanLennox:
Need the jmt version update in the pom

PR #1053, opened 2020-01-23
Fixes upgrade when no `--apis` param is available.

damencho:


resources/install/debian/postinst
```diff
@@ -116,14 +116,18 @@ case "$1" in
...
         if [[ $APIS_VALUE == *"xmpp"* ]]; then
             APIS_NEW_VALUE=${APIS_VALUE/xmpp/}
             if [ -z "$APIS_NEW_VALUE" ]; then
                 APIS_NEW_VALUE=","
             fi
             sed -i "s/--apis=$APIS_VALUE/--apis=$APIS_NEW_VALUE/g" $CONFIG
         fi
+        if [ -z "$APIS_VALUE" ]; then
+            # no apis setting, component is turned on by default, so let's disable it
+            sed -i "s/JVB_OPTS=\"/JVB_OPTS=\"--apis=, /g" $CONFIG
```
bgrozev: 1. This removes previous options. How about
`sed -E -i 's/JVB_OPTS="(.*)"/JVB_OPTS="\1 --apis=,"/'`
2. Do we care about the case where there is no `JVB_OPTS` defined in the file at all?

damencho: > 1. This removes previous options. How about

What do you mean? The logic is if we do not find and apis we add one. So we replace`JVB_OPTS="` with `JVB_OPTS="--apis=, `

> Do we care about the case where there is no JVB_OPTS defined in the file at all?

Nope, by default we were adding an empty one... maybe we can add in that case ... I donn't know, I'm wondering. WDYT?

bgrozev: If the file contains `JVB_OPTS=--xxx`, line 119 will not find a line matching `--apis`, and line 129 will replace `JVB_OPTS=--xxx` with `JVB_OPTS=--apis=,`

Ignoring the case of `JVB_OPTS` completely missing should be fine

damencho: But line 119 searches for `--apis` not `JVB_OPTS=--apis`. 
So I still think it will work correctly.
If someone has somewhere in the file a commented line containing `--apis` it may not work as expected based on whats in the JVB_OPTS value, but we cannot handle everything.

bgrozev: > But line 119 searches for `--apis` not `JVB_OPTS=--apis`.

Precisely my point. If they have `JVB_OPTS`, but no `--apis`, this will replace their existing `JVB_OPTS` line without preserving the old values.

damencho: This is desired, I explicitly tested this and added that if. 
So if you don't have `--apis` this means the component is running by default, so we add `--apis=,` in `JVB_OPTS` to disable the component.
And we do not replace the whole line, just the start of it and we preserve whatever was there in `JVB_OPTS` after the `JVB_OPTS="` part.

bgrozev: Aha! Sorry, I misread that part.

PR #1059, opened 2020-01-23
Endpoint message transport tweaks

bbaldino:
I happened to be looking at the endpoint message transport classes and noticed they had a bunch of warnings and reach into all sorts of places (`Endpoint`, `Conference`, `Videobridge`) so wanted to start cleaning some of those up.  The commits here are pretty independent, so looking one-by-one may be easiest.


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -230,7 +230,11 @@ protected void consume(@NotNull PacketInfo packetInfo)
             });
         bitrateController = new BitrateController(this, diagnosticContext, logger);
 
-        messageTransport = new EndpointMessageTransport(this, logger);
+        messageTransport = new EndpointMessageTransport(
+            this,
+            () -> getConference().getVideobridge().getStatistics(),
```
bgrozev: You can also just pass the reference

bbaldino: How do you use method references with chained calls like that?

PR #1061, opened 2020-01-24
Adds a query param to the main debug call in the colibri http api.

gpolitis:
This avoids having to do 2 API calls during testing, first get the conference id and then get the full description of the call.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -1098,7 +1098,7 @@ public class Videobridge
      * to include. If not specified, all of the conference's endpoints will be
      * included.
      */
-    public OrderedJsonObject getDebugState(String conferenceId, String endpointId)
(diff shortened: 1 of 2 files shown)


bbaldino:
Would it be easy to add a test to verify the default behavior still works?

gpolitis:
> Would it be easy to add a test to verify the default behavior still works?

@brianh5 do you see this as parsing the json and making sure it doesn't have the "full" stats or something smarter?

bbaldino:
> > Would it be easy to add a test to verify the default behavior still works?
> 
> @brianh5 do you see this as parsing the json and making sure it doesn't have the "full" stats or something smarter?

Yeah, I don't think that would be too bad?  If it turns out to be problematic we can skip it.  (btw that git username is defunct now, I'm just `bbaldino`).

gpolitis:
@bbaldino ah, so, it turns out it's not so practical because I have to setup the mock JVB to return something when its `getDebugState` method is called which is meaningless because then we're not checking the result of an actual computation.

PR #1064, opened 2020-01-28
AdaptiveTrackProjection log: null value check for srcEpId

Mikhalevich:
possible exception in LogContext  when use ImmutableMap.copyOf(context), if source.getOwner() is null


Mikhalevich:
@bbaldino please review

bbaldino:
Is the scenario where getOwner is null here legit? Or does it only happen when something goes wrong?

bgrozev:
I think jicofo always includes the `owner` field in colibri, but the only place we actually need it and use it is Octo. We could always set it for `MediaStreamTrackDesc` objects.

bgrozev:
@Mikhalevich thanks for the contribution!

PR #1065, opened 2020-01-28
ByteBufferPool bookkeeping enhancements

JonathanLennox:


src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
```diff
@@ -113,6 +127,19 @@ private static long threadId()
         return Thread.currentThread().getStackTrace();
     }
 
+    /**
+     * Gets a stack trace as a multi-line string.
+     */
+    private static String stackTraceAsString(StackTraceElement[] stack)
```
bbaldino: While you're doing this, can you write a kotlin one which is inlined so we don't get these methods as part of the stack?  I'm fine with just doing it in JVB.

JonathanLennox: I can use `UtilKt.getStackTrace()` but Java won't inline Kotlin, so they still show up on the stack.  Probably still worth it, though.

src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
```diff
@@ -198,10 +228,23 @@ else if (len < 2000)
...
+            {
+                logger.info("Thread " + threadId()
+                    + " returned a previously-returned buffer at\n"
+                    + getStackTraceAsString() +
+                    "previously returned at\n" +
+                    stackTraceAsString(returnedBookkeeping.get(arrayId).deallocTrace));
+            }
+            else
             {
                 logger.info("Thread " + threadId()
```
bbaldino: As I'm doing debugging, I'm finding I wish we had the size of the buffer as well here.

src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
bbaldino: Ah, bummer.

src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
```diff
@@ -198,14 +208,27 @@ else if (len < 2000)
...
+                returnedBookkeeping.put(arrayId, new ReturnedBufferBookkeepingInfo(bookkeeping.get(arrayId), UtilKt.getStackTrace()));
+                bookkeeping.remove(arrayId);
+            }
+            else if (returnedBookkeeping.containsKey(arrayId))
+            {
+                logger.info("Thread " + threadId()
+                    + " returned a previously-returned buffer at\n"
+                    + UtilKt.getStackTrace() +
+                    "previously returned at\n" +
+                    returnedBookkeeping.get(arrayId).deallocTrace);
```
bbaldino: I've seen a race happen here where this causes an NPE...so we should check and save the entry in one step.

PR #1066, opened 2020-01-29
Address some buffer issues

bbaldino:
Marking this as a draft for now, as I plan on adding some comments to each fix about things we might want to change before merging.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/DtlsTransport.java
+++ b/src/main/java/org/jitsi/videobridge/DtlsTransport.java
@@ -87,6 +87,7 @@ public class DtlsTransport extends IceTransport
     private final Endpoint endpoint;
 
     private final SocketSenderNode packetSender = new SocketSenderNode();
+    private final RtpSenderNode rtpPacketSender = new RtpSenderNode(packetSender);
     private final Node incomingPipelineRoot;
     private final Node outgoingDtlsPipelineRoot;
     private final Node outgoingSrtpPipelineRoot;
@@ -353,7 +354,7 @@ public class DtlsTransport extends IceTransport
     private Node createOutgoingSrtpPipeline()
     {
         PipelineBuilder builder = new PipelineBuilder();
-        builder.node(packetSender);
+        builder.node(rtpPacketSender);
         return builder.build();
     }
 
@@ -560,20 +561,55 @@ public class DtlsTransport extends IceTransport
         return debugState;
     }
 
+
     /**
-     * A terminating {@link Node}, which sends the packets through a
-     * datagram socket.
+     * A {@link ConsumerNode} which sends a packet out over a socket
      */
     private class SocketSenderNode extends ConsumerNode
     {
         public DatagramSocket socket = null;
 
-        /**
-         * Initializes a new {@link SocketSenderNode}.
-         */
         SocketSenderNode()
         {
             super("Socket sender");
+        }
+
+        @Override
+        protected void consume(@NotNull PacketInfo packetInfo)
+        {
+            if (socket != null)
+            {
+                try
+                {
+                    socket.send(
+                        new DatagramPacket(
+                            packetInfo.getPacket().getBuffer(),
+                            packetInfo.getPacket().getOffset(),
+                            packetInfo.getPacket().getLength()));
+                }
+                catch (IOException e)
+                {
+                    logger.error("Error sending packet: " + e.toString());
+                    throw new RuntimeException(e);
+                }
+            }
+
+        }
+    }
+
+    /**
+     * A terminating {@link Node}, to send out RTP packets whose buffer
+     * came from the pool.  It tracks certain RTP-related statistics
+     * (jitter, etc.)
+     */
+    private class RtpSenderNode extends ConsumerNode
+    {
+        private final SocketSenderNode socketSenderNode;
+
+        RtpSenderNode(SocketSenderNode socketSenderNode)
+        {
+            super("RTP Socket sender");
+            this.socketSenderNode = socketSenderNode;
         }
 
         /**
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -300,12 +301,15 @@ private Node createIncomingPipeline()
         dtlsPath.setPredicate(DTLS_PREDICATE);
         PipelineBuilder dtlsPipelineBuilder = new PipelineBuilder();
         dtlsPipelineBuilder.node(dtlsReceiver);
-        dtlsPipelineBuilder.simpleNode(
-                "sctp app packet handler",
-                packetInfo -> {
-                    endpoint.dtlsAppPacketReceived(packetInfo);
-                    return null;
-        });
+        ConsumerNode sctpHandler = new ConsumerNode("sctp app packet handler")
```
bbaldino: These two fixes (this one and the one for SRTP below) I think are correct.  Previously, these were implemented as `TransformerNode`s which returned null, so it was returning the buffer to the pool, but functionally they are `ConsumerNode`s.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -557,48 +564,74 @@ public JSONObject getDebugState()
         return debugState;
     }
 
+
     /**
-     * A terminating {@link Node}, which sends the packets through a
-     * datagram socket.
+     * A {@link ConsumerNode} which sends a packet out over a socket
      */
     private class SocketSenderNode extends ConsumerNode
```
bbaldino: This class, the way it's implemented here, is incorrect, I think, and should be tweaked.  This is implemented as a `ConsumerNode` but it does *not* take ownership of the buffer, as it doesn't free it.  The issue was that we wanted to return SRTP packets to the pool but not those coming from the DTLS pipeline.  In fact, I think this change may be defunct now because I started copying outgoing DTLS packets into pooled buffers, so this may currently be a leak (which just means we can go back to using a single sender for both DTLS and SRTP, which would be better anyway).  I'll look into this.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -694,8 +694,15 @@ public void createSctpConnection()
...
+                    // The buffer we get here came from usrsctp, and therefore
+                    // shouldn't be returned to the pool.  We can either
+                    // never return any packets from this path, or copy here
+                    // (actually, ideally, closer to SCTP, since this is
+                    // usrsctp specific) into a pooled buffer.
+                    byte[] newBuf = ByteBufferPool.getBuffer(length);
+                    System.arraycopy(data, offset, newBuf, 0, length);
                     PacketInfo packet
-                        = new PacketInfo(new UnparsedPacket(data, offset, length));
+                        = new PacketInfo(new UnparsedPacket(newBuf, 0, length));
```
bbaldino: Packets coming from SCTP are in buffers allocated by usrsctp (as far as I can tell), so we either have to have a way to know not to return them, or, we put them in a pooled buffer.  Here I'm doing the latter, which seemed like a good place to start: I hope the traffic won't be too significant, but if we find it's a problem we can improve it later.  Before merging, though, I would like to put the work of copying into a pooled buffer "closer" to the SCTP code, if possible.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1012,6 +1019,7 @@ protected void consume(PacketInfo packetInfo)
                 else
                 {
                     sctpManager.handleIncomingSctp(packetInfo);
+                    ByteBufferPool.returnBuffer(packetInfo.getPacket().buffer);
```
bbaldino: `usrsctp` copies all buffers into a buffer of its own (synchronously down this call chain), so we have to return here.  This is another places where it'd be nice if we could get the return "closer" to the SCTP level, as it's usrsctp-specific.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
bbaldino: Ok, confirmed that this is no longer necessary (since we now copy outgoing SCTP buffers into pooled buffers).

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -591,7 +597,7 @@ protected void consume(@NotNull PacketInfo packetInfo)
                                 packetInfo.getPacket().getBuffer(),
                                 packetInfo.getPacket().getOffset(),
                                 packetInfo.getPacket().getLength()));
-                    ByteBufferPool.returnBuffer(packetInfo.getPacket().getBuffer());
+                    packetDiscarded(packetInfo);
```
bbaldino: Hm, I thought `packetDiscarded` would be more appropriate here but I've just realized it'll be misleading, so we _should_ return it to the pool directly.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -52,10 +52,10 @@
 import java.nio.*;
 import java.time.*;
 import java.util.*;
-import java.util.stream.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
 import java.util.function.*;
+import java.util.stream.*;
```
JonathanLennox: Are these changes just IntelliJ having fun?

bbaldino: Yeah, I ran optimize imports because I thought I had removed something from here and wanted to get rid of it, but guess it did some ordering tweaks instead.

src/main/java/org/jitsi/videobridge/sctp/SctpManager.java
```diff
@@ -16,10 +16,14 @@
...
 package org.jitsi.videobridge.sctp;
 
+import org.jetbrains.annotations.*;
 import org.jitsi.nlj.*;
 import org.jitsi.utils.logging2.*;
+import org.jitsi.videobridge.util.*;
 import org.jitsi_modified.sctp4j.*;
 
+import java.nio.*;
+
```
JonathanLennox: This isn't used (as far as I can tell).  Bad autocomplete?

bbaldino: From some old code I think.  Removed.

PR #1072, opened 2020-01-31
Dtls transport layer

bbaldino:
Rather than having `DtlsTransport` extend `IceTransport`, it now wraps `IceTransport` instead.

When looking at endpoint queues for Octo I found myself wanting to make the Octo endpoints more like normal endpoints but with a shared transport, so I started poking around at that and felt this would be a good step in the right direction.

To me this better reflects the 'layers' of what is actually going on, and the plan would be to add some more fundamental 'Transport' interfaces as next steps and get to a place where `OctoRelay` can implement a form of transport and then just be shared amongst the OctoEndpoints in a way that looks the same as local endpoints having their own transports.

I think even `DtlsTransport` could be given an `IceTransport` (in the form of the generic `Transport` interface) and then become even further decoupled from it, which would make it feasible to add some tests for it, as well as do more fully functional 'test harnesses' (like we have in JMT) all the way down to this layer as well (since it would be easy to plug in a dummy transport).

I had originally started with something more drastic and was porting `IceTransport` and `DtlsTransport` to kotlin, but the inheritance actually made that a little difficult (as the classes are in different packages and kotlin treats the visibility keywords slightly differently), so this would be a good step towards that as well.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -1350,13 +1350,13 @@ public class Conference
         /**
          * Whether at least one endpoint in this conference failed ICE.
          */
-        boolean hasIceFailedEndpoint = false;
+        public boolean hasIceFailedEndpoint = false;
 
         /**
          * Whether at least one endpoint in this conference completed ICE
          * successfully.
          */
-        boolean hasIceSucceededEndpoint = false;
+        public boolean hasIceSucceededEndpoint = false;
 
         /**
          * Gets a snapshot of this object's state as JSON.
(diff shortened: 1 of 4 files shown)


JonathanLennox:
Would this allow us to do an ICE restart without a DTLS teardown/restart?

bbaldino:
> Would this allow us to do an ICE restart without a DTLS teardown/restart?

The changes here don't enable that yet, but I do think it's a step in the right direction as the ICE layer becomes more independent from the DTLS layer.  

I've been continuing to play out this refactor and ended up where the Endpoint is aware both layers (ICE and DTLS) exist, but DTLS, for example, just knows it has some 'transport' beneath it it sends packets via and receives packets from.  I think we'd also end up with a somewhat more 'direct' route for SRTP packets, as technically they can go directly from ICE to the Endpoint/Transceiver.

bbaldino:
Took another shot at this here: https://github.com/jitsi/jitsi-videobridge/pull/1104

PR #1076, opened 2020-02-04
Spotbugs

bgrozev:
I fixed or supressed the obvious warnings, but there are 9 warning left that are valid and require some thought. That's why I've not added it to the `verify` phase in maven yet.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -15,7 +15,9 @@
  */
 package org.jitsi.videobridge;
 
+import edu.umd.cs.findbugs.annotations.*;
 import org.jetbrains.annotations.*;
+import org.jetbrains.annotations.Nullable;
 import org.jitsi.eventadmin.*;
 import org.jitsi.nlj.*;
 import org.jitsi.rtp.*;
@@ -38,6 +40,8 @@ import org.osgi.framework.*;
 
 import java.beans.*;
 import java.io.*;
+import java.lang.*;
+import java.lang.SuppressWarnings;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
@@ -91,6 +95,10 @@ public class Conference
      * The indicator which determines whether {@link #expire()} has been called
      * on this <tt>Conference</tt>.
      */
+    @SuppressFBWarnings(
+            value = "IS2_INCONSISTENT_SYNC",
+            justification = "The value is deemed safe to read without " +
+                "synchronization.")
     private boolean expired = false;
 
     /**
@@ -123,6 +131,10 @@ public class Conference
      * <tt>Conference</tt>. In the time interval between the last activity and
      * now, this <tt>Conference</tt> is considered inactive.
      */
+    @SuppressFBWarnings(
+            value = "IS2_INCONSISTENT_SYNC",
+            justification = "The value is deemed safe to read without " +
+                    "synchronization.")
     private long lastActivityTime;
 
     /**
@@ -885,9 +897,6 @@ public class Conference
      */
     public boolean isExpired()
     {
-        // Conference starts with expired equal to false and the only assignment
-        // to expired is to set it to true so there is no need to synchronize
-        // the reading of expired.
         return expired;
     }
 
@@ -1319,7 +1328,7 @@ public class Conference
     /**
      * Holds conference statistics.
      */
-    public class Statistics
+    public static class Statistics
     {
         /**
          * The total number of bytes received in RTP packets in channels in this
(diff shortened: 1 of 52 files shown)


src/main/java/org/jitsi/videobridge/rest/ColibriWebSocketServlet.java
```diff
@@ -26,6 +27,7 @@
 /**
  * @author Boris Grozev
  */
+@SuppressFBWarnings("SE_BAD_FIELD")
```
bbaldino: Reading the description of this warning, I wonder if it's wonder we should disable overall?

src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -88,7 +88,7 @@ private static void check(Conference conference)
                 throw new RuntimeException(ioe);
             }
 
-            endpoints.add(endpoint);
+            //endpoints.add(endpoint);
```
bbaldino: What's this?

src/main/java/org/jitsi/videobridge/cc/vp8/VP8QualityFilter.java
```diff
@@ -382,6 +384,11 @@ private static int getSpatialLayerId(int index)
      * Gets a JSON representation of the parts of this object's state that
      * are deemed useful for debugging.
      */
+    @SuppressWarnings("unchecked")
+    @SuppressFBWarnings(
+            value = "IS2_INCONSISTENT_SYNC",
+            justification = "We intentionally avoid synchronizing while reading" +
```
bbaldino: Reading through this class, I find the synchronization we're doing a bit confusing (I probably haven't parsed it fully) but I wonder if it's worth a look.

src/main/java/org/jitsi/videobridge/cc/AdaptiveTrackProjection.java
```diff
@@ -96,6 +99,10 @@
      * payload type (could be VP9, could be H264, could be VP8) so it has to be
      * created on packet arrival.
      */
+    @SuppressFBWarnings(
```
bbaldino: Instead of this, I wonder if we should just get rid of the `synchronized` on `getContext` (the comments there even state it isn't necessary now).

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -885,9 +897,6 @@ public final Videobridge getVideobridge()
      */
     public boolean isExpired()
     {
-        // Conference starts with expired equal to false and the only assignment
-        // to expired is to set it to true so there is no need to synchronize
-        // the reading of expired.
         return expired;
```
JonathanLennox: I'd keep this explanation, which is better than just "is deemed safe" imo.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -484,12 +478,14 @@ private VP8FrameProjection createInLayerProjection(@NotNull VP8Frame frame,
         }
         else
         {
-            do {
+            do
+            {
                 f2 = prevFrame(f1);
```
JonathanLennox: This will probably conflict when you rebase on master, so you might want to do that sooner rather than later.

src/main/java/org/jitsi/videobridge/health/Health.java
bgrozev: There's a block of code below which uses `endpoints`, but it's commented out.

bbaldino: So nothing is using `endpoints`?

bbaldino: Ah, ok...I expanded the file and see now.

src/main/java/org/jitsi/videobridge/cc/vp8/VP8QualityFilter.java
bgrozev: I don't know whether the syncronization in the rest of the class makes sense, but for `getDebugState` my intention is as described: read the int fields without aquiring any locks to reduce the risk of a deadlock (since this runs in a completely different context)

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -145,7 +145,7 @@ static MediaType readMediaType(byte[] buf, int off, int len)
         case OCTO_MEDIA_TYPE_DATA:
             return MediaType.DATA;
         default:
-            return null;
+            throw new IllegalArgumentException("Invalid media type value" + mediaType);
```
JonathanLennox: Missing space here.

PR #1077, opened 2020-02-04
Have matchesFrameConsistently describe what doesn't match (if it doesn't).

JonathanLennox:


src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -339,15 +339,60 @@ boolean matchesFrame(@NotNull Vp8Packet pkt)
...
      *
      * @param pkt the RTP packet to check whether its parameters match this frame.
-     * @return true if the specified RTP packet is consistent with this frame, false
-     * otherwise.
+     * @return null if the specified RTP packet is consistent with this frame; a string
+     *  describing the problem otherwise
      */
-    boolean matchesFrameConsistently(@NotNull Vp8Packet pkt)
+    @Nullable
+    String matchesFrameConsistently(@NotNull Vp8Packet pkt)
```
bbaldino: We don't expect this to happen a lot, right? Why not throw in the error case instead of returning `null` on success, which feels a little odd.

JonathanLennox: I don't want this to actually throw and block frame processing - this is just a sanity check.  Do you mean for the immediate calling function to catch and log?

bbaldino: Yeah

PR #1079, opened 2020-02-04
add extra information when logging unknown active speaker

bbaldino:


src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -193,13 +193,21 @@ private void activeSpeakerChanged(long ssrc)
             AbstractEndpoint endpoint
                 = conference.findEndpointByReceiveSSRC(ssrc);
 
+            if (endpoint == null)
+            {
+                logger.warn("Unable to find endpoint corresponding to active" +
+                    "speaker SSRC " + ssrc);
```
JonathanLennox: Missing space at the line break.

bbaldino: Fixed

PR #1081, opened 2020-02-06
Optimize off-by-one in Vp8FrameMap.findAfter; plus minor tweaks.

JonathanLennox:
`getIndex` of `getLastIndex() - getSize()` always returns `null`, so we can safely lower-bound `findAfter`'s search at `getLastIndex() - getSize() + 1`.

(The probably has pretty negligible performance impact, but it's nice for correctness.)

Also clean up a log message.


JonathanLennox:
Closed in favor of https://github.com/jitsi/jitsi-videobridge/pull/1084 which incorporates it.

PR #1083, opened 2020-02-07
Reduce the verbosity of 'No available transport channel' logs

bbaldino:
1. Move the log to debug
1. Add a stat to track the number of messages dropped
1. Log an error if no transport is setup within 30 seconds of DTLS finishing.


src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
```diff
@@ -147,6 +147,13 @@ protected void onLastNChangedEvent(
         logUnexpectedMessage(jsonObject.toJSONString());
     }
 
+    @Override
+    public boolean isConnected()
+    {
+        //TODO(before merge): is this correct?
+        return true;
```
bgrozev: Yes, we always have an octo transport

PR #1086, opened 2020-02-10
Added method to update endpoint media direction.

mstyura:
This PR is attempt to allow scenario described [here](https://github.com/jitsi/jitsi-videobridge/pull/972#issuecomment-569266366).

An implementation is my interpretation of [@bbaldino suggestion](https://github.com/jitsi/jitsi-videobridge/pull/972#issuecomment-569309700)

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -1324,6 +1324,45 @@ public class Endpoint
     }
 
     /**
+     * Update media direction of {@link ChannelShim}s associated
+     * with this Endpoint.
+     *
+     * When media direction is set to 'sendrecv' JVB will
+     * accept incoming media from endpoint and forward it to
+     * other endpoints in a conference. Other endpoint's media
+     * will also be forwarded to current endpoint.
+     * When media direction is set to 'sendonly' JVB will
+     * NOT accept incoming media from this endpoint (not yet implemented), but
+     * media from other endpoints will be forwarded to this endpoint.
+     * When media direction is set to 'recvonly' JVB will
+     * accept incoming media from this endpoint, but will not forward
+     * other endpoint's media to this endpoint.
+     * When media direction is set to 'inactive' JVB will
+     * neither accept incoming media nor forward media from other endpoints.
+     *
+     * @param type media type.
+     * @param direction desired media direction:
+     *                       'sendrecv', 'sendonly', 'recvonly', 'inactive'
+     */
+    public void updateMediaDirection(MediaType type, String direction) {
+        switch (direction) {
+            case "sendrecv":
+            case "sendonly":
+            case "recvonly":
+            case "inactive": {
+                for (ChannelShim channelShim : channelShims) {
+                    if (Objects.equals(channelShim.getMediaType(), type)) {
+                        channelShim.setDirection(direction);
+                    }
+                }
+                break;
+            }
+            default:
+                throw new IllegalArgumentException("Media direction unknown: " + direction);
+        }
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


bbaldino:
Is the goal still to be able to have an endpoint stop receiving media?  If so you can set last-n to 0 like boris mentioned in that other bug.

mstyura:
@bbaldino as I currently understand `Endpoint.setLastN(0)` will only stop sending video to endpoint. I want to have separate control over audio and video to properly handle scenarios described below:
1. I want to forward only audio to endpoint when client application is minimized by user on mobile. `setLastN(0)` fits this scenario.
2. Putting call on hold on mobile due to incoming GSM call. Currently there is no way to stop forwarding both audio and video on JVB 2.0, but there was such option in JVB 1.0 [by updating media direction of RTPChannel](https://github.com/jitsi/jitsi-videobridge/blob/c29af045ca7081dce457ebd25a025248cfda3a67/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java#L172-L178)

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1323,6 +1323,45 @@ public void removeChannel(ChannelShim channelShim)
...
+     *                       'sendrecv', 'sendonly', 'recvonly', 'inactive'
+     */
+    public void updateMediaDirection(MediaType type, String direction) {
+        switch (direction) {
+            case "sendrecv":
+            case "sendonly":
+            case "recvonly":
+            case "inactive": {
+                for (ChannelShim channelShim : channelShims) {
+                    if (Objects.equals(channelShim.getMediaType(), type)) {
```
bbaldino: I think a normal `==` works fine here?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1323,6 +1323,45 @@ public void removeChannel(ChannelShim channelShim)
...
+     * accept incoming media from this endpoint, but will not forward
+     * other endpoint's media to this endpoint.
+     * When media direction is set to 'inactive' JVB will
+     * neither accept incoming media nor forward media from other endpoints.
+     *
+     * @param type media type.
+     * @param direction desired media direction:
+     *                       'sendrecv', 'sendonly', 'recvonly', 'inactive'
+     */
+    public void updateMediaDirection(MediaType type, String direction) {
```
bbaldino: What do you see the interactions with this code looking like? (As in, who would call it?)

mstyura: @bbaldino I use jvb as a library for an application which basically implement signaling for jvb (I think it does similar things as jicofo does). So I just use public API of Videobridge to configure desired behavior. I think in my signaling code I'll use  `updateMediaDirection` in following way:
1. To enable hold
```
localEndpoint.updateMediaDirection(MediaType.VIDEO, "inactive");
localEndpoint.updateMediaDirection(MediaType.AUDIO, "inactive");
```
2. To disable video when client is minimized:
```
localEndpoint.updateMediaDirection(MediaType.VIDEO, "inactive");
```

bbaldino: Gotcha.  And, with the changes to IO activity to take ICE into account, there's no issues with expiration when a caller is put on hold?

mstyura: Yes, exactly! Endpoint is still treated as connected because there is ICE activity during hold and hence endpoint is not expired.
Fix for IO activity was previoulsy done in https://github.com/jitsi/jitsi-videobridge/pull/970

mstyura:
jenkins test this please

PR #1088, opened 2020-02-11
Keep track of the processing time for XMPP requests.

bgrozev:
Depends on jitsi/jitsi-utils#45 and jitsi/jitsi-media-transform#217

### DIFF
--- a/src/main/java/org/jitsi/videobridge/xmpp/XmppCommon.java
+++ b/src/main/java/org/jitsi/videobridge/xmpp/XmppCommon.java
@@ -15,6 +15,7 @@
  */
 package org.jitsi.videobridge.xmpp;
 
+import org.jitsi.nlj.stats.*;
 import org.jitsi.osgi.*;
 import org.jitsi.utils.logging2.*;
 import org.jitsi.utils.version.*;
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/videobridge/xmpp/XmppCommon.java
```diff
@@ -214,6 +239,16 @@ else if (request instanceof ShutdownIQ)
...
             response = null;
         }
+
+        if (delayStats != null)
+        {
+            long delay = System.currentTimeMillis() - start;
+            if (delay > 100)
+            {
+                logger.warn("Took " + delay + "ms to handle IQ: " + request.toXML());
+            }
```
JonathanLennox: Missing space before `"ms`?

bbaldino:
Should we expose these through datadog as well?

bbaldino:
(Sorry, closed on accident)

src/main/java/org/jitsi/videobridge/xmpp/XmppCommon.java
bgrozev: No, that was intentional. I've *always* formatted it this way! Looks like I've always been wrong :)

bgrozev:
> Should we expose these through datadog as well?

I think this wouldd be part of the infra scripts. I'll ping you about it

PR #1089, opened 2020-02-12
wip: initial stress-level computation

gpolitis:


gpolitis:
We're switching to a bitrate-based stress definition.

PR #1090, opened 2020-02-12
Filter out the destination endpoint when looping over tracks.

bgrozev:


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -431,6 +431,7 @@ StatusSnapshot getStatusSnapshot()
         long nowMs = System.currentTimeMillis();
         for (MediaStreamTrackDesc sourceTrack : destinationEndpoint
             .getConference().getEndpoints().stream()
+            .filter(e -> !destinationEndpoint.equals(e))
```
bbaldino: Wondering if we should just compare ep IDs here.

bbaldino: I don't know why...I think the object equality should work fine, but I worry it may be more brittle.  Probably being paranoid: if we define some custom equals for endpoint in the future, then certainly we should ensure an endpoint equals itself, so probably nothing to worry about there.  I guess the only potential would be a duplicate endpoint instance, and that also seems unlikely.

JonathanLennox: This will compare objects -- are there any cases where we'll have two different `AbstractEndpoint` objects with the same epID?

bbaldino: We shouldn't ever have that

JonathanLennox: The worst case with a duplicate endpoint ID is that we get the log message we're getting now, right?  It sounds harmless.

PR #1091, opened 2020-02-12
Add more detail to VP8Frame "inconsistent frame" log, and fix spacing.

JonathanLennox:


src/main/java/org/jitsi/videobridge/cc/vp8/VP8Frame.java
```diff
@@ -351,17 +351,23 @@ void validateConsistent(@NotNull Vp8Packet pkt)
...
-        StringBuilder s = new StringBuilder().append("Packet ")
-            .append(pkt.getSequenceNumber())
-            .append(" is not consistent with frame with timestamp ")
-            .append(timestamp)
-            .append(":");
+        StringBuilder s = new StringBuilder().append("Packet")
+            .append(" ssrc ").append(pkt.getSsrc())
+            .append(", seq ").append(pkt.getSequenceNumber())
+            .append(", picture id ").append(pkt.getPictureId())
+            .append(", timestamp ").append(pkt.getTimestamp())
```
bgrozev: Minor (I don't mind leaving it as it is): All these except PictureID are included in `RtpPacket.toString()`. Should we extend `Vp8Packet.toString()` to append the PictureID and just print the packet?

JonathanLennox: Potentially -- the former is certainly a good idea, independent of the latter -- but it can be a followup PR.

PR #1092, opened 2020-02-13
Fix pruning the list of inactive endpoints

bgrozev:


bgrozev:
As discussed with @paweldomas offline, this was intentional in order to continue broadcasting the "inactive" state to other endpoint (in case the endpoint is still in the MUC). The side effects are that we keep a reference to the endpoint, and we log a debug message.

PR #1097, opened 2020-02-17
Ignore remote candidates that need DNS resolution.

bgrozev:


bgrozev:
Works as expected when GUM access is denied:
Screenshot 2020-02-17 at 19 27 14

I don't knonw why this test failed though:
https://ci.jitsi.org/view/pr-testing/job/jitsi-videobridge-PR-test/1776/testReport/org.jitsi.meet.test/ActiveSpeakerTest/

src/main/java/org/jitsi/videobridge/ice/IceTransport.java
```diff
@@ -453,6 +464,14 @@ private int addRemoteCandidates(
             if ((relAddr = candidate.getRelAddr()) != null
                     && (relPort = candidate.getRelPort()) != -1)
             {
+                boolean relAddrNeedsResolution
+                        = !IPAddressUtil.isIPv4LiteralAddress(relAddr)
+                            && !IPAddressUtil.isIPv6LiteralAddress(relAddr);
+                if (relAddrNeedsResolution && !Config.resolveRemoteCandidates())
+                {
+                    logger.debug(() -> "Ignoring remote candidate with non-literal related address: " + address);
+                    continue;
```
JonathanLennox: I'm not sure it makes sense to completely ignore a candidate with a non-literal related address, as opposed to just passing null for that parameter?  I'd be surprised if we use remote candidates' related addresses at all.

(That said, I'd also be surprised to have a candidate with an IP literal for the candidate, and a hostname for the related, so this probably doesn't matter so much.)

bgrozev:
> Is it worth logging that we're ignoring the related address?

I don't think it is (I don't even think reading the related address is necessary).

PR #1102, opened 2020-02-21
Do not describe expired channels or channels of expired endpoint.

mstyura:
In some scenarios when endpoint is expired, the corresponding channel shims are not removed. Generation of IQ with expired channels via:
```java
final ColibriConferenceIQ iq = new ColibriConferenceIQ();
final ConferenceShim confShim = conference.getShim();
confShim.describeDeep(iq);
```

The fix for channel shims outliving owning endpoint will be done separately.


mstyura:
Not relevant because https://github.com/jitsi/jitsi-videobridge/pull/1103 merged long time ago.

PR #1104, opened 2020-02-21
Ice transport refactor

bbaldino:
We should hold off on this until after going stable, but I've been thinking about ICE restarts on the bridge and some other cleanup so have been tinkering around with some refactoring in the transports and came up with this as a first step.

I have a branch where I implemented this layering scheme all the way up; this code represents what I thought would be a good first step so as not to change things all at once.

This change includes a refactoring of `IceTransport` that makes it act as an independent layer.  `DtlsTransport` is independent of `IceTransport` (it no longer extends it).  `IceTransport` has an API for handling incoming data and sending data out, as well as a set of events it fires.  `DtlsTransport` has been tweaked to add a `dataReceived` method as well as specifying a `sender`.  `Endpoint` is responsible for wiring the data coming **in** from `IceTransport` into `DtlsTransport` and the data coming **out** of `DtlsTransport` into `IceTransport`.

The next step after this would be to refactor `DtlsTransport` and tweak `Endpoint` a bit to have SRTP packets skip `DtlsTransport` completely and go directly into the `Transceiver` (there's no reason for those packets to go through a "DTLS transport").  This will also make it easy to avoid adding the RTP before/after padding to DTLS packets.

After that SCTP and Datachannel would be tweaked to fit into the 'layer' model as well (which will be nice, as they're quite awkward to force into the 'pipeline' scheme now: the stacks typically function by saying "I'll let you know when I have data"; this is awkward for a synchronous pipeline but fits perfectly in this 'layer' model).  

I tried to illustrate how I see the before and after here:

Before:
image

After:
image


src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -223,30 +205,50 @@ else if (!StringUtils.isNullOrEmpty(setup))
...
+        catch (Throwable e)
+        {
+            logger.error("Error during DTLS negotiation: " + e.toString() +
+                    ", closing this transport manager");
+        }
     }
 
     /**
      * Returns {@code true} if this {@link DtlsTransport} is connected. It is
      * considered connected if the underlying ICE connection has been
```
JonathanLennox, line 230: Is the comment about the "underlying ICE connection" still correct/relevant here?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -988,8 +1026,32 @@ public void sendLastNEndpointsChangeEvent(
...
+            }
+        });
+        dtlsTransport.setRemoteFingerprints(remoteFingerprints);
+        if (!fingerprintExtensions.isEmpty())
+        {
+            String setup = fingerprintExtensions.get(0).getSetup();
+            dtlsTransport.setSetupAttribute(setup);
+        }
+
+        iceTransport.startConnectivityEstablishment(transportInfo);    }
```
JonathanLennox, line 1054: Format

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,443 @@
...
+     */
+    private val iceConnected = AtomicBoolean(false)
+    /**
+     * Whether or not this [IceTransport] has failed to connect.
+     */
+    private val iceFailed = AtomicBoolean(false)
+
+    fun hasFailed(): Boolean = iceFailed.get()
+
+    fun isConnected(): Boolean = iceConnected.get()
```
JonathanLennox, line 76: Having `hasFailed` and `isConnected` as separate booleans seems odd -- e.g. there's no way they can both be true at the same time, can there?  Should this be some sort of enum instead?

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,443 @@
...
+                0 -> {
+                    // XXX Effectively, the check above but realizing that all
+                    // candidates were ignored:
+                    // iceAgentStateIsRunning && candidates.isEmpty().
+                }
+                else -> iceComponent.updateRemoteCandidates()
+            }
+        } else if (remoteCandidateCount != 0) {
+            // Once again, because the ICE Agent does not support adding
+            // candidates after the connectivity establishment has been started
```
JonathanLennox, line 154: ??? Really?  It seems like we do that a lot.

src/main/java/org/jitsi/videobridge/Endpoint.java
bbaldino: Fixed

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: They can't both be true, but they can both be false :)

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: This is a copy-paste of some existing logic (including the comment).  But yeah I'm not sure if this still applies.

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
JonathanLennox: Yeah, that's why I was thinking some kind of enum state might work better.

Possibly the same states as `IceProcessingState`, though I don't know if this needs the `COMPLETED`/`TERMINATED` distinction.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
bbaldino: Nope, fixed

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: So something like `RUNNING`, `FAILED`, `SUCCEEDED` ?

JonathanLennox: Yeah.  (Is there ever a case where this will be in a `WAITING` state, and if so does this API care?)

JonathanLennox: ICE restarts are tricky, because usually you want the `RUNNING` -> `SUCCEEDED` transition to gate things like starting the DTLS handshake, and you don't have to do that for ICE restarts.

bbaldino: I'm good with the enum change, but I was trying to change as little other code as possible as part of this PR (and the ensuing PRs) because they include such fundamental code I didn't want to make the PR bigger/touch anything I didn't have to.

bbaldino: > ICE restarts are tricky, because usually you want the `RUNNING` -> `SUCCEEDED` transition to gate things like starting the DTLS handshake, and you don't have to do that for ICE restarts.

Yeah, and I think if we _did_ need to trigger something, we can use the events from `IceTransport` (since we'll fire a new `connected` event in that case)

JonathanLennox: Ok, yeah, if this is just the Kotlin-ization of the existing code it's fine. (Because of the Kotlin translation it's hard to tell how much is new in `IceTransport`.)

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,444 @@
+package org.jitsi.videobridge.transport.ice
```
bgrozev, line 1: License header

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,444 @@
...
+                receivedTime = clock.instant()
+            } catch (e: SocketClosedException) {
+                logger.info("Socket closed, stopping reader")
+                break
+            } catch (e: IOException) {
+                logger.warn("Stopping reader", e)
+                break
+            }
+            packetStats.numPacketsReceived++
+            incomingDataHandler?.dataReceived(receiveBuf, 0, packet.length, receivedTime) ?: run {
```
bgrozev, line 192: Is it possible that `receive` set a non-zero offset?

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,444 @@
...
+        logger.info("No longer running, stopped reading packets")
+    }
+
+    /**
+     * Send data out via this transport
+     *
+     * TODO: should we ensure that we're running before sending?
+     */
+    fun send(data: ByteArray, off: Int, length: Int) {
+        iceComponent.socket.send(DatagramPacket(data, off, length))
```
bgrozev, line 206: This is weird -- how is the `IOException` from `send` handled? The ice4j component socket can throw if it doesn't have an underllying socket to send through.

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,444 @@
...
+        packetStats.numPacketsSent++
+    }
+
+    /**
+     * NOTE(brian): this method only exists because we want to use a Consumer<T> in
+     * SocketSenderNode in DtlsTransport.  When that gets ported to Kotlin it will be easier to
+     * make the T a function type (which can take multiple arguments), but until then we provide this.
+     */
+    fun send(packet: DatagramPacket) {
+        send(packet.data, packet.offset, packet.length)
```
bgrozev, line 216: Why does `send(data: ByteArray, off: Int, length: Int)` exist? We take a `DatagramPacket`, break it down to its components, then create a new `DatagramPacket`.

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,444 @@
...
+    private fun iceStateChanged(ev: PropertyChangeEvent) {
+        val oldState = ev.oldValue as IceProcessingState
+        val newState = ev.newValue as IceProcessingState
+        val transition = IceProcessingStateTransition(oldState, newState)
+
+        logger.info("ICE state changed old=$oldState new=$newState")
+
+        when {
+            transition.completed() -> {
+                if (iceConnected.compareAndSet(false, true)) {
```
bgrozev, line 310: This is cool

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,444 @@
...
+/**
+ * Models a transition from one ICE state to another and provides convenience
+ * functions to test the transition.
+ */
+private data class IceProcessingStateTransition(
+    val oldState: IceProcessingState,
+    val newState: IceProcessingState
+) {
+    // We should be using newState.isEstablished() here, but we see
+    // transitions from RUNNING to COMPLETED, which should not happen and
```
bgrozev, line 389: I believe `COMPLETED` should say `TERMINATED` (I know the original has the same mistake).

bgrozev: If you fix this, maybe add a note that this happens when the Agent is freed prior to being started

bgrozev: Also, the comment should apply to `failed` below. Looking at the git history it was me who messed it up originally...

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: I dunno, but agree it would be best to use `packet.offset` (I think I forgot it had one).

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: Ah, it's not (Kotlin doesn't have checked exceptions so won't complain).  I'll add a try/catch here.

EDIT: Actually, we used to catch the `IOException` but then throw a `RuntimeException` in that case anyway (I guess propagating up to the queue error hander).  So I'll look into what would be best here now.

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: It's an annoying side effect (which I tried to describe in the comment) that will go away in the next phase (when `DtlsTransport` is redone).  `DtlsTransport` has function to set its "sender" (`setSender`) as we use a `Consumer` there.  `Consumer` only supports a single argument, so I had `DtlsTransport` take a `Consumer<DatagramPacket>`.  The other option would be to have `DtlsTransport` define a `OutgoingDataHandler` (and this is exactly what I did in the next phase of the refactor), but here in order to minimize the change I just added the shim in `IceTransport`, expecting that the next phase will follow this shortly (code is pretty much done there), so we won't be living with it for long.

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: Sure, will tweak the comment.

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -0,0 +1,444 @@
...
+        val remoteCandidates = transportPacketExtension.getChildExtensionsOfType(CandidatePacketExtension::class.java)
+        if (iceAgentStateIsRunning && remoteCandidates.isEmpty()) {
+            logger.cdebug { "Ignoring transport extensions with no candidates, " +
+                    "the Agent is already running." }
+            return
+        }
+
+        val remoteCandidateCount = addRemoteCandidates(remoteCandidates, iceAgentStateIsRunning)
+        if (iceAgentStateIsRunning) {
+            when (remoteCandidateCount) {
```
bbaldino, line 160: @bgrozev do you think it's worth preserving the 'empty' case here?  I.e. the commented:
```
// XXX Effectively, the check above but realizing that all
// candidates were ignored:
// iceAgentStateIsRunning && candidates.isEmpty().
```
Or should I just get rid of it and turn this into a:
```
if (remoteCandidateCount != 0) {
    iceComponent.updateRemoteCandidates()
}
```
?

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bbaldino: Ok, the context of this call (with regards to the pool/queue) is the same, so I added a try/catch to match the old code.

src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bgrozev: I don't think we need to preserve this case

PR #1105, opened 2020-02-24
makes queue sizes configurable

gpolitis:


src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -21,12 +21,25 @@ import org.jitsi.config.LegacyFallbackConfigProperty
...
+import org.jitsi.utils.config.SimpleProperty
 import org.jitsi.utils.config.exception.ConfigValueParsingException
 
 class OctoConfig {
     class Config {
         companion object {
 
+            class QueueSizeProperty : SimpleProperty<Int>(
+                    newConfigAttributes {
+                        name("videobridge.octo.q-size")
```
JonathanLennox: I'd spell out "queue" in the parameter names.

src/main/resources/reference.conf
```diff
@@ -117,4 +120,10 @@ videobridge {
     # such candidates will be ignored.
     resolve-remote-candidates = false
   }
+
+  dtls {
+    # packets from the egress end-up in this queue, right before encryption. Its
```
JonathanLennox: This isn't right - the DtlsTransport doesn't actually do the encryption of the SRTP/SRTCP.

src/main/java/org/jitsi/videobridge/DtlsTransport.java
```diff
@@ -131,7 +131,7 @@ public DtlsTransport(Endpoint endpoint, boolean controlling, Logger parentLogger
                         getClass().getSimpleName() + "-outgoing-packet-queue",
                         TaskPools.IO_POOL,
                         this::handleOutgoingPacket,
-                        1024);
+                        DtlsTransportConfig.Config.queueSize());
```
bbaldino: nit: I think in other places we've imported `DtlsTransportConfig.Config` so we can just refer to the config values via `Config.queueSize`

src/main/java/org/jitsi/videobridge/octo/OctoTransceiver.java
```diff
@@ -92,7 +93,7 @@
                 "octo-tranceiver-incoming-packet-queue",
                 TaskPools.CPU_POOL,
                 this::processPacket,
-                1024);
+                OctoConfig.Config.queueSize());
```
bbaldino: nit: same as above

src/main/resources/reference.conf
```diff
@@ -117,4 +120,10 @@ videobridge {
     # such candidates will be ignored.
     resolve-remote-candidates = false
   }
+
+  dtls {
```
bbaldino: I'm wondering if, at least for the non-octo queue, we should scope this under something else rather than `dtls`, as the fact that it happens to live in the DTLS transport is implementation specific.  It's not really a "DTLS" queue, but really the main "ingress" queue.  Also, what about the sender-side queue size?

bbaldino: Sorry, I've noticed now that this is actually the `egress` side, but I still have the same thought about the scoping of the parameter.

src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
bgrozev: Perhaps also add "receive" to the name. We currently don't have a send queue, but we might add one in the future.

src/main/resources/reference.conf
gpolitis: yeah, that sounds like a good idea and I was wondering about this myself. Any suggestions?

src/main/resources/reference.conf
gpolitis: right, so this is right after encryption then if i'm reading the code correctly the pipeline is as simple as writing to the socket?

JonathanLennox: Yes, I believe that's right.  (It nonetheless takes up 15% of bridge CPU time...)

src/main/resources/reference.conf
bbaldino: maybe:
```
transport {
  send {
    queue-size
  }
}
```
?

src/main/resources/reference.conf
gpolitis: I've updated the description, please let me know if you feel it needs any other changes.

src/main/resources/reference.conf
gpolitis: done

src/main/java/org/jitsi/videobridge/DtlsTransportConfig.kt
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.videobridge
+
+import org.jitsi.config.newConfigAttributes
+import org.jitsi.utils.config.SimpleProperty
+
+class DtlsTransportConfig {
```
JonathanLennox: If the attribute is now named `videobridge.transport.send.queue-size`, should the config class be named to match?

src/main/resources/reference.conf
gpolitis: @bbaldino shall the element be named `transport` or `dtls-transport` ?

bbaldino: I think just transport

src/main/java/org/jitsi/videobridge/DtlsTransportConfig.kt
gpolitis: done

gpolitis:
@bbaldino @JonathanLennox do you guys feel confortable with these defaults, or you would prefer some more conservative values? They're worst-case scenarios over-estimations and they seem to be working quite well in my tests.

JonathanLennox:
What I'd like to do is figure out some way we can get reasonable instrumentation of queue sizes, so we can see average sizes and distributions...it wouldn't be too hard to add this to the Colibri debug output, though it's less clear how to get it in a non per-endpoint form.

gpolitis:
> What I'd like to do is figure out some way we can get reasonable instrumentation of queue sizes, so we can see average sizes and distributions

Yeah, I agree that'd be useful. Shall this be implemented as part of this PR (which has been reduced to what the title says, i.e. making the queue sizes configurable, until there's more clarity on what we want).

JonathanLennox:
> Yeah, I agree that'd be useful. Shall this be implemented as part of this PR (which has been reduced to what the title says, i.e. making the queue sizes configurable, until there's more clarity on what we want).

IMO now that this isn't changing the defaults, the instrumentation doesn't need to be part of this PR, but I'll let others weigh in too.

gpolitis:
@bgrozev could you please re-stamp, I've updated jmt and lost your review.

PR #1106, opened 2020-02-24
Fail to project (and log packet contents) for VP8 packets without picture IDs.

JonathanLennox:
The frame map indexes by picture ID.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
+++ b/src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
@@ -676,6 +676,13 @@ public class VP8AdaptiveTrackProjectionContext
         }
         Vp8Packet vp8Packet = packetInfo.packetAs();
 
+        if (vp8Packet.getPictureId() == -1)
+        {
+            /* Should have been rejected in accept(). */
+            logger.info("VP8 packet does not have picture ID, cannot track in frame map.");
+            throw new RewriteException("VP8 packet without picture ID in VP8 track projeciton");
+        }
+
         VP8Frame vp8Frame = lookupVP8Frame(vp8Packet);
         if (vp8Frame == null)
         {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveTrackProjectionContext.java
```diff
@@ -676,6 +676,13 @@ public void rewriteRtp(
         }
         Vp8Packet vp8Packet = packetInfo.packetAs();
 
+        if (vp8Packet.getPictureId() == -1)
```
bgrozev: Currently `setPictureId` fails when the packet has a the 7 bit version of the field. I'm pretty sure we've seen such packets in the past. Should we augment this to check for [`hasExtendedPictureId`](https://github.com/jitsi/jitsi-media-transform/blob/master/src/main/java/org/jitsi_modified/impl/neomedia/codec/video/vp8/DePacketizer.java#L291)?

I opened jitsi/jitsi-media-transform#226 to help us see if this condition actually happens.

JonathanLennox: I don't know if we want 7-bit picture IDs to actually cause projection to fail (as missing picture IDs will), but we could add a log for it.

bgrozev: I'm afraid 7-bit picture IDs already cause things to go wrong. The projection logic works fine, but when it comes to actually re-write the field the call to [`setPictureId`](https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/java/org/jitsi/videobridge/cc/vp8/VP8FrameProjection.java#L172) silently fails leaving the original value in the buffer. Am I missing something?

JonathanLennox: Yeah, that happens.  The question is whether the receiver will still be able to decode something -- which I think it should if there's not otherwise packet loss? -- so it's better to route the packets even without the picture ID rewriting.

JonathanLennox:
I removed the log message for VP8 packets missing picture IDs; I'd rather log them as in https://github.com/jitsi/jitsi-media-transform/pull/227, so it only occurs once per packet.

bgrozev:
Closing in favor of #1107

PR #1108, opened 2020-02-26
'Removes the obsolete code for serving static content and proxying.'

bgrozev:


damencho:
I'm not sure about that ... As in jitsi-meet I think we print a warning when we see this configuration... but we leave it ... It is true that new deployments will not use it, but we do not remove old config and migrate it to nginx

damencho:
https://github.com/jitsi/jitsi-meet/blob/94c48ec8387647606f98805d536bd5011032c8f8/debian/jitsi-meet-web-config.postinst#L116

bgrozev:
Okay, let's keep it for now. Do you think we can merge this a few months from now?

Do you know whether the "alias" properties are used anywhere? If so, they are broken. This is what made me remove the code.

damencho:
Well, alias we use for config.js, do you know when we broke it?
We can work on adding some debian version shinanigans to disable upgrades of currently installed stable versions ... or at least after some version ... so we do not break people's deployment, but just don't update them ... Not sure it is the best approach ...

damencho:
For sure alias is working in the current stable version.

bgrozev:
> For sure alias is working in the current stable version.

I guess it works for config.js, but fails with an exception if the path contains special characters. We can live with that for now.

PR #1112, opened 2020-03-02
Adds more statistics.

bgrozev:


bgrozev:
This depends on jitsi/jitsi-media-transform#232 and jitsi/jitsi-xmpp-extensions#22.

It adds the following stats:
`octo_conferences` -- number of conferences with octo enabled.
`inactive_conferences` -- number of conferences in which no endpoints are sending audio or video (includes p2p conferences)
`p2p_conferences` -- number of inactive conferences with two endpoints, which are probably using peer-to-peer (there are false positives with e.g. firefox)
`receive_only_endpoints` -- number of endpoints, which are in an active conference, but are themselves not sending audio or video.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1333,5 +1333,11 @@ public Version getVersion()
          */
         public AtomicInteger numEndpointsNoMessageTransportAfterDelay =
             new AtomicInteger();
+
+        /**
+         * The total number of times the dominant speaker in any conference
+         * changed.
+         */
+        public AtomicInteger totalDominantSpeakerChanges = new AtomicInteger();
```
JonathanLennox: Would this be better as a `LongAdder` rather than an `AtomicLong`?

src/main/java/org/jitsi/videobridge/octo/OctoEndpoint.java
```diff
@@ -153,4 +153,18 @@ void setReceiveSsrcs(Map<MediaType, Set<Long>> ssrcsByMediaType)
             ssrcs.forEach(ssrc -> streamInformationStore.addReceiveSsrc(ssrc, mediaType));
         });
     }
+
+    @Override
+    public boolean isSendingAudio()
+    {
+        // TODO implement detection
```
JonathanLennox: Do you consider this TODO as needing to be resolved before this PR gets merged?

bgrozev: No, I would like to get this merged, so we can start gathering data while we have octo turnedd off. Implementing this on top of Brian's transport refactor (and some more octo endpoint refactoring) would be cleaner, which is why I left it out for now.

src/main/java/org/jitsi/videobridge/Videobridge.java
bgrozev: Because of contention? Dominant speaker changes don't happen very often.

JonathanLennox: Not a big deal either way, I guess, I just thought `LongAdder` was supposed to be more efficient when you don't need test-and-set semantics.

PR #1113, opened 2020-03-05
Expose forwaded endpoints change event from BitrateController.

mstyura:
Currently `BitrateController` controls the set of endpoints whose video is forwarded to target endpoint.
It does it automatically from external user point of view. Whenever set of forwarded endpoints is changed by `BitrateController` it automatically sends `LastNEndpointsChangeEvent` message over data channel to target endpoint.

I'd like to have an extensibility point at that point and have custom handling when set of forwarded endpoints changed. 
There could be several options here I can think of:
1. Propose PR which will allow using custom `EndpointMessageTransport`&`EndpointMessageBuilder`, which could serve as an extensibility point to implement custom handling of all events and messages produced by JVB;
2. Expose an additional event from `BitrateController` which will not affect already sent message over data channel, but will provide an extensibility point for additional custom handling of forwarded endpoints change event.

This PR implements option 2. 
I've considered option 1, but is seems it will require much more changes and has no clear benefit for Jitsi team/product, so that's why I'm proposing option 2.

I hope it's not much harmful for JVB to have such an event.
Please let me know what do you think.

PS. In my case custom logic I'd like to have when forwarded endpoints changed includes 1) modification of some internal state I maintain 2) send a custom message to endpoint 3) record extra metrics.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -55,6 +55,7 @@ import org.jitsi_modified.impl.neomedia.rtp.*;
 import org.jitsi_modified.sctp4j.*;
 import org.json.simple.*;
 
+import javax.annotation.*;
 import java.beans.*;
 import java.io.*;
 import java.nio.*;
@@ -79,7 +80,8 @@ import static org.jitsi.videobridge.EndpointMessageBuilder.*;
 public class Endpoint
     extends AbstractEndpoint implements PotentialPacketHandler,
         PropertyChangeListener,
-        EncodingsManager.EncodingsUpdateListener
+        EncodingsManager.EncodingsUpdateListener,
+        BitrateController.BitrateControllerListener
 {
     /**
      * Track how long it takes for all RTP and RTCP packets to make their way through the bridge.
@@ -172,6 +174,7 @@ public class Endpoint
     /**
      * The bitrate controller.
      */
+    @Nonnull
     private final BitrateController bitrateController;
 
     /**
@@ -299,6 +302,7 @@ public class Endpoint
                 }
             });
         bitrateController = new BitrateController(this, diagnosticContext, logger);
+        bitrateController.subscribe(this);
 
         outgoingSrtpPacketQueue = new PacketInfoQueue(
             getClass().getSimpleName() + "-outgoing-packet-queue",
@@ -1096,6 +1100,20 @@ public class Endpoint
     }
 
     /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onForwardedEndpointsChanged(
+        Collection<String> forwardedEndpoints,
+        Collection<String> endpointsEnteringLastN,
+        Collection<String> conferenceEndpoints) {
+        sendLastNEndpointsChangeEvent(
+            forwardedEndpoints,
+            endpointsEnteringLastN,
+            conferenceEndpoints);
+    }
+
+    /**
      * Sends a message to this {@link Endpoint} in order to notify it that the
      * list/set of {@code lastN} has changed.
      *
@@ -1105,7 +1123,7 @@ public class Endpoint
      * @param conferenceEndpoints the collection of all endpoints in the
      * conference.
      */
-    public void sendLastNEndpointsChangeEvent(
+    private void sendLastNEndpointsChangeEvent(
         Collection<String> forwardedEndpoints,
         Collection<String> endpointsEnteringLastN,
         Collection<String> conferenceEndpoints)
@@ -1597,6 +1615,15 @@ public class Endpoint
     public Transceiver getTransceiver()
     {
         return transceiver;
+    }
+
+    /**
+     * @return this {@link Endpoint}'s bitrate controller
+     */
+    @Nonnull
+    public BitrateController getBitrateController()
+    {
+        return bitrateController;
     }
 
     /**
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -189,6 +189,11 @@ public class BitrateController
     private Set<String> pinnedEndpointIds = Collections.emptySet();
 
     /**
+     * Set of event listeners of current {@code BitrateController}
+     */
+    private final Set<BitrateControllerListener> listeners = ConcurrentHashMap.newKeySet();
+
+    /**
      * The last-n value for the endpoint to which this {@link BitrateController}
      * belongs
      */
@@ -722,7 +727,7 @@ public class BitrateController
         {
             // TODO(george) bring back sending this message on message transport
             //  connect
-            destinationEndpoint.sendLastNEndpointsChangeEvent(
+            notifyForwardedEndpointsChanged(
                 newForwardedEndpointIds,
                 endpointsEnteringLastNIds,
                 conferenceEndpointIds);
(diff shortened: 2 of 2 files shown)


mstyura:
Hello @bbaldino!
Could you please take a look? 
Thanks a lot in advance.

bbaldino:
We recently found an issue related to last-n in JVB 2 and I think we're going to be changing logic in this area somewhat soon, so my thought would be to hold off on merging this but to enable this scenario when we tweak that code (as I think it makes sense).  Seem alright?

mstyura:
Sound good to me, thanks @bbaldino! Could you please also mention me in corresponding PR with last-n fix when it will appear? Thanks a lot!

bbaldino:
> Sound good to me, thanks @bbaldino! Could you please also mention me in corresponding PR with last-n fix when it will appear? Thanks a lot!

Will do.  And feel free to keep this open/ping us if you don't see anything in the next couple weeks.

MMX777:
> We recently found an issue related to last-n in JVB 2 and I think we're going to be changing logic in this area somewhat soon, so my thought would be to hold off on merging this but to enable this scenario when we tweak that code (as I think it makes sense).  Seem alright?

@bbaldino, is the first JVB also affected?

bbaldino:
> > We recently found an issue related to last-n in JVB 2 and I think we're going to be changing logic in this area somewhat soon, so my thought would be to hold off on merging this but to enable this scenario when we tweak that code (as I think it makes sense).  Seem alright?
> 
> @bbaldino, is the first JVB also affected?

We don't know, as this code has changed between JVB 1 and 2.  The issue in 2 is that we didn't call `BitrateController#update` when the last-n ordering changed.  [This PR](https://github.com/jitsi/jitsi-videobridge/pull/1114) fixes the issue in JVB 2, but we decided to hold off for a better cleanup of that code as it's a bit confusing to follow now.

mstyura:
Hello @bbaldino 
It there any changes this PR is being merged soon? Or maybe there are plans to expose endpoint's last-N changed event in a different way? I have custom logic when Last-N changes and would like to know when it happen, but a message automatically sent over message transport is accessible only by clients.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -172,6 +174,7 @@ public static OrderedJsonObject getPacketDelayStats()
     /**
      * The bitrate controller.
      */
+    @Nonnull
```
bbaldino, line 177: I think we typically use `NotNull`

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -299,6 +302,7 @@ public void trace(@NotNull Function0<Unit> f)
                 }
             });
         bitrateController = new BitrateController(this, diagnosticContext, logger);
+        bitrateController.subscribe(this);
```
bbaldino, line 305: Can you create an anonymous class to add as the handler?  That way we don't make the API of Endpoint too broad (nothing outside cares or should leverage `Endpoint` as a `BitrateController.BitrateControllerListener`)

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1095,6 +1099,20 @@ public void decrementSelectedCount()
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onForwardedEndpointsChanged(
```
bbaldino, line 1106: I think it's unwritten, but we try to use `onXXX` names for "subscription" calls, and just something like `forwardedEndpointsChanged` for when the event occurs.

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1599,6 +1617,15 @@ public Transceiver getTransceiver()
         return transceiver;
     }
 
+    /**
+     * @return this {@link Endpoint}'s bitrate controller
+     */
+    @Nonnull
+    public BitrateController getBitrateController()
```
bbaldino, line 1624: I don't see anything using this?  I'd rather not expose this and keep it encapsulated.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1182,6 +1187,54 @@ public boolean transformRtp(@NotNull PacketInfo packetInfo)
...
+        for (BitrateControllerListener listener : listeners) {
+            listener.onForwardedEndpointsChanged(
+                forwardedEndpoints, endpointsEnteringLastN, conferenceEndpoints);
+        }
+    }
+    /**
+     * Interface for listeners of events generated
+     * by {@code BitrateController}
+     */
+    public interface BitrateControllerListener {
```
bbaldino, line 1219: Can you rename this to `EventHandler`? Then it gets referred to as `BitrateController.EventHandler`

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1182,6 +1187,54 @@ public boolean transformRtp(@NotNull PacketInfo packetInfo)
         }
     }
 
+    /**
+     * Subscribe to {@link BitrateController}'s events.
+     * @param listener event listener
+     */
+    public void subscribe(BitrateControllerListener listener) {
```
bbaldino, line 1194: `subscribe` -> `addEventHandler`

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1182,6 +1187,54 @@ public boolean transformRtp(@NotNull PacketInfo packetInfo)
...
+        listeners.add(listener);
+    }
+
+    /**
+     * Unsubscribe from {@link BitrateController}'s events.
+     * @param listener event listener
+     */
+    public void unsubscribe(BitrateControllerListener listener) {
```
bbaldino, line 1202: `unsubscribe` -> `removeEventHandler`

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1182,6 +1187,54 @@ public boolean transformRtp(@NotNull PacketInfo packetInfo)
...
+         * Notify listener that the collection of forwarded endpoints by
+         * current {@link BitrateController} has changed.
+         *
+         * @param forwardedEndpoints the collection of forwarded endpoints.
+         * @param endpointsEnteringLastN the <tt>Endpoint</tt>s which are entering
+         * the list of <tt>Endpoint</tt>s defined by <tt>lastN</tt>
+         * @param conferenceEndpoints the collection of all endpoints in the
+         * conference.
+         */
+        default void onForwardedEndpointsChanged(
```
bbaldino, line 1231: I don't think we need the default implementation.

src/main/java/org/jitsi/videobridge/Endpoint.java
mstyura: @bbaldino the reason why I exposed it, it's because in my code which uses `JVB` as a library I only have an access to endpoint instance and few other public classes and I wanted having access to `Endpoint` object to understand when the set of forwarded endpoints is changed. 
So without access to bitrate controller I can't listen it's events.

So if you think `BitrateController` must stay private to endpoints, would you mind if I add an forwarded endpoint changed event on endpoint level? I mean adding `EventHandler` for endpoint and putting `forwardedEndpointChanged` here.
I think now that I would prefer `EventHandler` approach, because I've just recently realized I also want an event when endpoint is selected by some other endpoint. Currently, endpoint just directly send message to a message transport, but it would be really nice to have an event here.
And once I written all of this I realized that with `EventHandler` on endpoint level I could reduce diff significantly... so I'll will create another PR to show an idea.

bbaldino: Yeah, I think that sounds good.

mstyura:
Deprecated in favor of https://github.com/jitsi/jitsi-videobridge/pull/1221

PR #1114, opened 2020-03-05
Update the BitrateController any time the active speaker changes

bbaldino:
`BitrateController` needs to be told any time the order of active speakers changes (and have the new, correct order) for last-n to work correctly.  This issue did not affect non-last-n calls, as a `dominantSpeakerChanged` message would be sent to the client which would respond with a `SelectedEndpointMessage`.  Since selected endpoints are always given priority, that client would always be prioritized.  With last-n enabled, though, the `BitrateController` must have an accurate, up-to-date ordering of active endpoints so that the proper ones make it into last-n.

Note: in debugging this, we found the flow around updating the interested parties when active-speaker-list changes happen (which can be driven by a few things) feels pretty messy and would be worth revisiting; so this fix feels more like a hack on top of hacks.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -447,10 +447,7 @@ public class Conference
             getVideobridge().getStatistics().totalDominantSpeakerChanges.increment();
         }
 
-        List<String> sortedEndpointIds = speechActivity.getEndpointIds();
-        endpointsCache.stream()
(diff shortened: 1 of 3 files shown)


bbaldino:
Will hold off on merging this for now.  Last-n is a "big conferences" blocker, but we don't need it before then.  We can either end up going with this simple fix or getting a larger refactor in around active-speaker-change propagation in the bridge.

PR #1115, opened 2020-03-09
Add a queue to send out Octo packets on the I/O pool.

JonathanLennox:


src/main/resources/reference.conf
```diff
@@ -69,6 +69,10 @@ videobridge {
     # set its size (in terms of millis) to be 16ms.
     # With interbridge last-n the load would be much smaller.
     recv-queue-size=1024
+    # The size of the outgoing octo queue.  This is a tentacle level queue and
+    # it's where we temporarily store packets to be sent to remote bridges.
+    # TODO: tune this.
+    send-queue-size=1024
```
gpolitis: A quick n' dirty calculation of the queue depth would be nice (but not mandatory).

JonathanLennox:
So testing on jitsi-net, what I see is that there are dropped packets on the Octo send queue when the conference shuts down -- possibly this is an artifact of how the queue is shut down, so I need to investigate that.

A 20 sender/100 participant call (all senders on one bridge) consistently gets Octo receive queue drops in one region only (not the region with the senders).  I'm not sure why the asymmetry.

src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
```diff
@@ -353,6 +409,13 @@ public JSONObject getDebugState()
         debugState.put("relay", relay.getDebugState());
         debugState.put("targets", targets.toString());
 
+        /* TODO: do we really want to dump debug state for all the queues? */
```
bbaldino: Good question.  Should we just sum up the stats?

JonathanLennox: If I can get my queue stats patch in a usable state, we can get global state for the octo send queue.

Absent that, I don't think there's a whole lot of value for the non-statistics-enabled PacketQueue debug state per queue.

PR #1116, opened 2020-03-11
Calculate loss_rate_upload based on received TCC reports.

JonathanLennox:


src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -391,7 +390,7 @@ private void generate0()
             // TODO verify
             unlockedSetStat(LOSS_RATE_DOWNLOAD, lossRateDownload);
             // TODO verify
-            unlockedSetStat(LOSS_RATE_UPLOAD, lossRateUpload);
+            unlockedSetStat(LOSS_RATE_UPLOAD, lossRateReported);
```
bbaldino: I wonder if we should call this out specifically as loss reported by TCC

JonathanLennox: I'm in favor of it in theory, but I'm also trying to avoid having to change the existing stats names here.  (Both to avoid changing the xmpp extensions, and to avoid changing our existing DataDog reporting.)

JonathanLennox:
Close in favor of https://github.com/jitsi/jitsi-videobridge/pull/1455.

PR #1117, opened 2020-03-12
Track how many endpoints connect to the bridge using a relayed ICE candidate.

JonathanLennox:


pom.xml
```diff
@@ -112,7 +112,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-media-transform</artifactId>
-      <version>1.0-135-g687d806</version>
+      <version>1.0-SNAPSHOT</version>
```
bgrozev: This doesn't need a JMT update, right?

JonathanLennox: No, you're right - I was task switching between too many different PRs when I wrote this.

PR #1125, opened 2020-03-17
Start jitsi-videobridge after nginx

vosdev:
`apt install jitsi-meet` results in an nginx that is unable to start because jitsi-videobridge has claimed port 443.

```
11396#11396: bind() to 0.0.0.0:443 failed (98: Address already in use)
11396#11396: bind() to 0.0.0.0:443 failed (98: Address already in use)
11396#11396: bind() to 0.0.0.0:443 failed (98: Address already in use)
11396#11396: bind() to 0.0.0.0:443 failed (98: Address already in use)
11396#11396: bind() to 0.0.0.0:443 failed (98: Address already in use)
11396#11396: still could not bind()
```

Making nginx start before jitsi-videobridge has solved this issue for me and can be done with the patch in this PR.

But why does the package `jitsi-meet` install two services that both want to bind on 443? (And why is jitsi-videobridge fine with not being able to bind on 443 because nginx has now claimed it?)

### DIFF
--- a/debian/jitsi-videobridge2.service
+++ b/debian/jitsi-videobridge2.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Jitsi Videobridge
-After=network.target
+After=network.target nginx.service
 
 [Service]
 # allow bind to 80 and 443


damencho:
What will happen if nginx is not installed?

rlaager:
If nginx is not installed, then this is a no-op. An After= only sets an order; it is not a dependency.

If Apache is also supported, then it should likely be listed too (apache2.service).

damencho:
Jvb actually should not bind to port 443 by default. Are you sure you don't have such setting in your config?

vosdev:
@damencho by default it does, https://github.com/jitsi/jitsi-videobridge/blob/master/doc/tcp.md#configuration

Perhaps a better option would be to change this default when you install the `jitsi-meet` debian package?

damencho:
We need to update that doc. Default now is to enable it, otherwise, it is disabled by default. 
https://github.com/jitsi/jitsi-videobridge/commit/753d3de37cff6187e7f148970984b89030f187c1#diff-5d225913d631e97e8291eeb263af444cR40

PR #1126, opened 2020-03-17
deps: updates jicoco to bring in MUC XMPP port option

paweldomas:
Updates jicoco to 1.1-20-ge4ade96.


bbaldino:
Are no other changes in the JVB itself needed?

paweldomas:
Nope, it only reads one extra MUC config option: https://github.com/jitsi/jicoco/commit/e4ade960604368b0342cf1e629a518079ca3cff5

paweldomas:
MUC config option are based on a prefix, so this change will read one more automatically

bbaldino:
Ok, that's part of what I wanted to check since JVB does config differently now.  But I guess you tested with JVB and saw it works?

paweldomas:
yes

PR #1128, opened 2020-03-18
Return XMPP errors for error conditions in handleIQRequest.

JonathanLennox:


JonathanLennox:
This adds indentation, so review hiding whitespace changes.

Question: do we want to move the assignment of `delayStats` so that exceptions are counted toward them?

PR #1140, opened 2020-03-24
debian: change name of init daemon

saghul:


saghul:
Nah, I thought I had caught the upgrade bug but nope.

PR #1152, opened 2020-03-30
Add optional sentry error tracking

sapkra:
Because Jitsi is using `java.util.logging` it is quite easy to implement:
https://docs.sentry.io/clients/java/integrations/#javautillogging

To configure the Sentry client the SENTRY_DSN environment variable has to be set. But there are also some other possibilities.

The same changes for the other jitsi repositories can be found here:
- https://github.com/jitsi/jigasi/pull/247
- https://github.com/jitsi/jicofo/pull/449

### DIFF
--- a/lib/logging.properties
+++ b/lib/logging.properties
@@ -1,5 +1,6 @@
 handlers= java.util.logging.ConsoleHandler
 #handlers= java.util.logging.ConsoleHandler, com.agafua.syslog.SyslogHandler
+#handlers= java.util.logging.ConsoleHandler, io.sentry.jul.SentryHandler
 
 java.util.logging.ConsoleHandler.level = ALL
 java.util.logging.ConsoleHandler.formatter = org.jitsi.utils.logging2.JitsiLogFormatter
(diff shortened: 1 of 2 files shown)


bbaldino:
Does sentry handle all logs? Or only errors?  Can't tell from looking at their site.

sapkra:
@bbaldino  It can handle all logs but in my opinion it only makes sense for warnings and errors which is also sentry's default. But of cause you can change the log level in the following line:

https://github.com/jitsi/jitsi-videobridge/blob/97262e9d5e4c3d2f120e40ee5cc157085938168f/lib/logging.properties#L26

bbaldino:
> @bbaldino It can handle all logs but in my opinion it only makes sense for warnings and errors which is also sentry's default. But of cause you can change the log level in the following line:
> 
> https://github.com/jitsi/jitsi-videobridge/blob/97262e9d5e4c3d2f120e40ee5cc157085938168f/lib/logging.properties#L26

Cool, thanks.  Do you have another centralized logging service you're using for JVB?  Or only the warnings/errors (handled by sentry)?

sapkra:
I'm just using sentry to track my warnings and errors of my entire deployment (also for my own servers and custom client). All the other logs I can see in OpenShift. The jitsi server part is still missing to debug most problems a bit better.

PR #1157, opened 2020-04-01
Remove system-level statistics from README

itrich:
According to https://community.jitsi.org/t/proposed-videobridge-stats-removal/20431 the system-level statistics cpu_usage, total_memory and used_memory have been removed. It took me a while to find this news post and stop wondering why these three values aren't reported by my instance. In order to prevent confusion, these three values should also be removed from the docs.

### DIFF
--- a/doc/rest-colibri.md
+++ b/doc/rest-colibri.md
@@ -124,8 +124,6 @@ The respective response could look like:
 		200 OK with a JSON object which represents the statistics report.
 <pre>
 {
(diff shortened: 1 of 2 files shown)


bgrozev:
Thanks for the fix! We'll need your CLA before we can merge it.

itrich:
> Thanks for the fix! We'll need your CLA before we can merge it.

Done!

PR #1161, opened 2020-04-03
jvb ENABLE_STATISTICS defval is true

pedro-nonfree:
seen in two servers with default configuration

```
root@masked:/etc/jitsi/videobridge# grep -ir org.jitsi.videobridge.ENABLE_STATISTICS *
sip-communicator.properties:org.jitsi.videobridge.ENABLE_STATISTICS=true
```

### DIFF
--- a/doc/statistics.md
+++ b/doc/statistics.md
@@ -120,7 +120,7 @@ The statistics can also be published periodically via XMPP (which allows jicofo 
 The statistics reporting functionality can be configured with the following properties:
 
  * **org.jitsi.videobridge.ENABLE_STATISTICS** - boolean property.
-The default value is `false`
+The default value is `true`
  * **org.jitsi.videobridge.STATISTICS_TRANSPORT** - string property.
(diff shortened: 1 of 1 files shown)


bgrozev:
The default value is `false`, that's why it's necessary to override it in the config file:
https://github.com/jitsi/jitsi-videobridge/blob/master/src/main/resources/reference.conf#L49

pedro-nonfree:
Aha, ok, got it. But that is confusing. When you are going to enable stats, you see *default value is false*, but when you look at your instance config is set to true. Then:

- it is false by default in `src/main/resources/reference.conf`
- it is true by default in `/etc/jitsi/meet/*-config.js`

but I don't know how to solve the situation (and if that applies to other variables). Now I see that some information is missing

bgrozev:
I understand it can be confusing, especially with the migration to the new config style. The documentation in this repository documents the default values that jitsi-videobridge will use unless they are overriden in a config file. The jitsi-meet debian package overrides some of the defaults.

Issue #1163, opened 2020-04-03
Support for endpoints which uses different payload types for same codec.


mstyura:
Hi,
I'd like to discuss the support of heterogenous WebRTC-endpoints in JVB 2.0.

Please note, I'll describe the scenario in a context of my use-case: I use JVB with my own signaling on top of it (not using jicofo/jitsi meet).

The flow of joining the conference in my case is the following:

1. Client create peer connection. Adds send-only `RTCRtpTransceiver` for outbound audio and video. 
2. Client create offer and send it to custom signaling to join the conference.
 Typical offer would look like this ()
```
m=audio 9 UDP/TLS/RTP/SAVPF 111
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
a=fmtp:111 minptime=10;useinbandfec=1
a=mid:0
a=sendrecv
a=msid:c3ce0974-437c-4792-95e7-7bac8fcadb36 34517d54-6f21-4527-a017-12eac50a96ef
a=ssrc:2285997888 cname:bfGDkSeJ5uT0ADKP
m=video 9 UDP/TLS/RTP/SAVPF 96 97
a=rtpmap:96 VP8/90000
a=rtcp-fb:96 goog-remb
a=rtcp-fb:96 transport-cc
a=rtcp-fb:96 ccm fir
a=rtcp-fb:96 nack
a=rtcp-fb:96 nack pli
a=rtpmap:97 rtx/90000
a=fmtp:97 apt=96
a=mid:1
a=sendrecv
a=msid:c3ce0974-437c-4792-95e7-7bac8fcadb36 7730f97d-31a1-4937-9233-06312c3f002f
a=ssrc-group:FID 2546541619 25938643
a=ssrc:2546541619 cname:bfGDkSeJ5uT0ADKP
a=ssrc:25938643 cname:bfGDkSeJ5uT0ADKP
m=application 9 UDP/DTLS/SCTP webrtc-datachannel
a=mid:2
```

3. In response to join request client receives answer:
```
m=audio 9 UDP/TLS/RTP/SAVPF 111
a=recvonly
a=mid:0
m=video 9 UDP/TLS/RTP/SAVPF 96
a=recvonly
a=mid:1
m=application 9 UDP/DTLS/SCTP webrtc-datachannel
a=mid:2
```
[...]

PR #1164, opened 2020-04-03
Use payload types map of endpoint to which track belongs

mstyura:
It seems that this change fixes an issue (or not an issue, I'm not sure) described here #1163.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -771,6 +771,27 @@ public class BitrateController
             // creating local final variables and pass that to the lambda function
             // in order to avoid that.
             final String endpointID = trackBitrateAllocation.endpointID;
+
+            final Endpoint endpointOwningTrack = destinationEndpoint.getConference().getLocalEndpoint(endpointID);
+            if (endpointOwningTrack == null)
+            {
+                return null;
+            }
+
+            // Populate payload types from endpoint which owns the track.
+            // In JVB scenarios (Jitsi Meet) where all payload type assignments are guaranteed
+            // to be the same across all endpoints it would be the same set of payload types
+            // as of current endpoint payload types.
+            // In JVB scenarios where payload types might be different across different endpoints
+            // the payload types would be populated from endpoint which originates track, i.e.
+            // sending endpoint, rather than payload types of receiving endpoint. This allows
+            // having endpoints with Unified Plan SDP semantics in the same conference and not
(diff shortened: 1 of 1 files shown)


mstyura:
Jenkins test this please.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -771,14 +771,21 @@ private synchronized void update()
             // creating local final variables and pass that to the lambda function
             // in order to avoid that.
             final String endpointID = trackBitrateAllocation.endpointID;
+
+            final Endpoint endpointOwningTrack = destinationEndpoint.getConference().getLocalEndpoint(endpointID);
```
bbaldino, line 775: We'll have to discuss if we want to fix this or not (as you've seen, quite a few things assuming the same payload types), but if we do do this, we should expose the payload types a different way, I think.  The first thing that comes to mind is having the `Transceiver` expose a read-only version of its `StreamInformationStore` (which we could then expose at the `Endpoint` level)

bbaldino: See https://github.com/jitsi/jitsi-media-transform/blob/master/src/main/kotlin/org/jitsi/nlj/Transceiver.kt#L75

mstyura:
@bbaldino sorry for bothering you, did you have a chance to discuss this change in a team? So far I don't see any problem with this change: running native Android/iOS WebRTC in same conference with Chromium's WebRTC and Firefox's WebRTC. As long as for jitsi meed use-case where all endpoints have same payload types assignment this change is actual no-op and maybe this PR can be safely merged? Thank you very much!

mstyura:
Seems like the scenario I've tried to fix is not valid according to [BUNDLE spec](https://tools.ietf.org/html/draft-ietf-mmusic-sdp-bundle-negotiation-54#section-9.1.1):
```
9.1.1.  Payload Type (PT) Value Reuse

   Multiple bundled "m=" sections might describe RTP based media.  As
   all RTP based media associated with a BUNDLE group belong to the same
   RTP session, in order for a given payload type value to be used
   inside more than one bundled "m=" section, all codecs associated with
   the payload type number MUST share an identical codec configuration.
   This means that the codecs MUST share the same media type, encoding
   name, clock rate and any parameter that can affect the codec
   configuration and packetization.
   [I-D.ietf-mmusic-sdp-mux-attributes] lists SDP attributes, whose
   attribute values are required to be identical for all codecs that use
   the same payload type value.
```
So all endpoints in same conference should have same payload type assignment.

Issue #731, opened 2018-12-12
jvb service not passing --host flag | "Something went wrong" loop


Ark74:
## Description
---
On the latest update from debian unstable repository, I'm having some issues with the jvb looping the conference as described here: https://community.jitsi.org/t/stuck-in-something-went-wrong-loop-once-second-member-connected/16091/12?u=ark74
## Current behavior
---
Whenever a user set a conference and is joined by a second one, this breaks and loops back.
Seems that this is only present on the unstable branch, since on the stable one other users can still join.
## Expected Behavior
---
Being able to join a conference without issues.
## Possible Solution
---
Seems like with the use of the systemd service for jvb, on older installations don't pass the --host flag even when it's set on /etc/jitsi/videobridge/config

Manually edit: /lib/systemd/system/jitsi-videobridge.service
And set localhost manually :laughing: 
**It works!**

## Steps to reproduce
---
1. Install a fresh unstable branch 
2. Try to join the conference two persons or more, it will crash.
3. Edit /etc/jitsi/videobridge/config to set JVB_HOST as localhost and restart the system.
4. `systemctl status jitsi-videobridge.service `

Will show that the JVB_HOST as localhost flag is not being passed.

`"...zrtp4j-light-3.2.0-jitsi-1-20150723.002345-1.jar org.jitsi.videobridge.Main --host= --domain=jitsi.si..."
`
# Environment details
---
*buntu 16.04 // Others find this issue at the community forum.

damencho:
Does it work if you add something else, other than localhost ... 
A few months ago I've added the systemd unit file https://github.com/jitsi/jitsi-videobridge/blob/master/resources/install/debian/jitsi-videobridge.service#L19
And it should be working just fine.

Ark74:
No, as for the date tested, nothing passed, 127.0.0.1, localhost, they passed on stable.

I see today there are new updates, I'll update and report back.

Ark74:
Just tested today with a new instance.
The JVB_HOST in /etc/jitsi/videobridge/config is not passing to actual executed service.

The service file reads on the ExecStart,

`ExecStart=/bin/bash -c "exec /usr/share/jitsi-videobridge/jvb.sh --host=${JVB_HOST:-localhost} --domain=${JVB_HOSTNAME} --port=${JVB_PORT} --secret=${JVB_SECRET} ${JVB_OPTS} ...`

And systemctl reports,

```
● jitsi-videobridge.service - Jitsi Videobridge
[...]

PR #1167, opened 2020-04-06
Fix typo in systemd unit

licaon-kter:
...else component won't register to the XMPP server (at least on ejabberd)

### DIFF
--- a/debian/jitsi-videobridge2.service
+++ b/debian/jitsi-videobridge2.service
@@ -17,7 +17,7 @@ TasksMax=65000
 # allow more open files for this process
 LimitNPROC=65000
 LimitNOFILE=65000
-ExecStart=/bin/bash -c "exec /usr/share/jitsi-videobridge/jvb.sh --host=$${JVB_HOST:-localhost} --domain=${JVB_HOSTNAME} --port=${JVB_PORT} --secret=${JVB_SECRET} ${JVB_OPTS} < /dev/null >> ${LOGFILE} 2>&1"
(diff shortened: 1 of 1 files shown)


damencho:
@licaon-kter This was a fix contributed by the community and was tested on several different OS versions.
https://github.com/jitsi/jitsi-videobridge/issues/731

What is your OS and version you are seeing this problem?

licaon-kter:
Debian Buster arm64, DEBs generated with your scripts.

damencho:
We do not do debs.

licaon-kter:
Who is "we"?

What is this for: https://jitsi.org/downloads/ubuntu-debian-installations-instructions/ ?

Why is this here: https://github.com/jitsi/jitsi-videobridge/blob/master/resources/build_deb_package.sh ?

Are we lost in translation?

Anyway from [here](https://www.freedesktop.org/software/systemd/man/systemd.service.html): `To pass a literal dollar sign, use "$$". Variables whose value is not known at expansion time are treated as empty strings. Note that the first argument (i.e. the program to execute) may not be a variable.`

So `$${JVB_HOST:-localhost}` would be what exactly?

damencho:
> Who is "we"?

The jitsi team.



> So `$${JVB_HOST:-localhost}` would be what exactly?

If there is no JVB_HOST set, it will use localhost as value.

damencho:
> Why is this here: https://github.com/jitsi/jitsi-videobridge/blob/master/resources/build_deb_package.sh ?

This is how we build a binary package. We do not do debian source packages.

licaon-kter:
Yes, I've retested with empty, localhost and actual host and they all work now.

Strange that it didn't when I setup it, maybe there was some other change that I did at the same time.

Sorry for the noise.

> This is how we build a binary package. We do not do debian source packages.

I didn't say that...

PR #1168, opened 2020-04-06
$$ is the process ID

aboettger-tuhh:


rubenk:
This is a duplicate of #1167

damencho:
More on the subject.
https://github.com/jitsi/jitsi-videobridge/issues/731

https://unix.stackexchange.com/questions/216780/why-does-bash-parameter-expansion-not-work-inside-systemd-service-files

PR #1175, opened 2020-04-07
Add config option for ICE nomination strategy.

JonathanLennox:
Depends on https://github.com/jitsi/ice4j/pull/197.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,7 +135,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
-      <version>3.0-SNAPSHOT</version>
+      <version>3.0-10-g982e782</version>
     </dependency>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -135,7 +135,7 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>ice4j</artifactId>
-      <version>3.0-1-ge854e4e</version>
+      <version>3.0-10-g982e782</version>
```
bbaldino, line 138: Any other changes in this version bump we should be worried about?

JonathanLennox: Reviewing them I think they're all desirable or harmless.

JonathanLennox:
I think this will need https://github.com/jitsi/jicofo/pull/460 to be merged first or we'll start getting PR testing failures.

PR #1178, opened 2020-04-08
Bump ice4j version.

JonathanLennox:


JonathanLennox:
These ice4j changes don't change its API, so I think this time we don't need jicofo to change.

PR #1184, opened 2020-04-10
Fix various buffer pool leaks

bbaldino:
Depends on https://github.com/jitsi/jitsi-utils/pull/53 and https://github.com/jitsi/jitsi-media-transform/pull/251.


bbaldino:
> This isn't complete (we still have leaks on Octo queues shutdown) but it's monotonically better that what we had before.

Well the PR didn't say it fixed _all_ buffer pool leaks :)

PR #1186, opened 2020-04-10
Allow VP8 layers without temporal scalability

JonathanLennox:
This works around a FF 75 bug https://bugzilla.mozilla.org/show_bug.cgi?id=1628851 .

Also improve some logging.


JonathanLennox:
Note: depends on https://github.com/jitsi/jitsi-media-transform/pull/252 .

PR #1188, opened 2020-04-10
Move health checks code to jicoco.

bgrozev:


src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -311,26 +230,11 @@ protected void doRun()
...
-                    + timeSinceLastResult + "ms ago.");
-        }
-
-        if (lastResult != null)
-        {
-            throw new Exception(lastResult);
-        }
-
-        // We've had a recent result, and it is successful (no exception).
+        doCheck(videobridge);
```
bbaldino, line 237: I'm too lazy to expand the whole file in the diff...does `doCheck` get called from multiple places? Could we just put the logic here?

bgrozev:
This was tested as part of jitsi/jicofo#467

Issue #1189, opened 2020-04-10
IceTransport.kt: Symbol is declared in module 'java.base' which does not export package 'sun.net.util'


licaon-kter:
Debian Buster, arm64

Since 498d17b2355ceb235597328392a400f74b7fd216 I can't build anymore:
```
[ERROR] Failed to execute goal org.jetbrains.kotlin:kotlin-maven-plugin:1.3.61:compile (compile) on project jitsi-videobridge: Compilation failure: Compilation failure: 
[ERROR] /home/pi/jitsi/source/jitsi-videobridge/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt:[38,21] Symbol is declared in module 'java.base' which does not export package 'sun.net.util'
[ERROR] /home/pi/jitsi/source/jitsi-videobridge/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt:[423,10] Symbol is declared in module 'java.base' which does not export package 'sun.net.util'
[ERROR] /home/pi/jitsi/source/jitsi-videobridge/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt:[423,24] Symbol is declared in module 'java.base' which does not export package 'sun.net.util'
[ERROR] /home/pi/jitsi/source/jitsi-videobridge/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt:[423,53] Symbol is declared in module 'java.base' which does not export package 'sun.net.util'
[ERROR] /home/pi/jitsi/source/jitsi-videobridge/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt:[423,67] Symbol is declared in module 'java.base' which does not export package 'sun.net.util'
```
...what am I missing?

bbaldino:
I think this is a java version thing. Are you using java 8?

licaon-kter:
Nope, 11 is Buster [default](https://packages.debian.org/buster/default-jdk)

gramakri:
I get similar errors with OpenJDK 8 (ubuntu).

```
[ERROR] /app/code/jvb-build/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt: (38, 21) Symbol is declared in module 'java.base' which does not export package 'sun.net.util'
[...]

PR #1195, opened 2020-04-14
Use Guava InetAddresses to determine if something is a literal IP address.

JonathanLennox:
Rather than sun.net.util.IPAddressUtil.  Fixes Java 11 build.


JonathanLennox:
This should fix https://github.com/jitsi/jitsi-videobridge/issues/1189.

PR #1200, opened 2020-04-15
have PotentialPacketHandler interface extend PacketHandler

bbaldino:


bbaldino:
After thinking this over some more, I think it's probably better to keep these two interfaces distinct.  We want to reinforce that a `PotentialPacketHandler` must have its `wants` called before its `send`, which isn't the case for a `PacketHandler` (whose `processPacket` should just be called directly).  We should think of a way in the future to tighten up the `PotentialPacketHandler` behavior, if we can.

PR #1202, opened 2020-04-16
Octo transport refactor phase 1

bbaldino:
This phase takes the current `OctoRelay` class and splits the logic into two new classes:

1) `UdpTransport` which is a generic, raw UDP transport classes that owns a UDP socket and can receive from and send on that socket
2) `OctoTransport` which handles Octo packet parsing and can send and receive media and 'messages'


src/main/java/org/jitsi/videobridge/octo/OctoRelayService.java
```diff
@@ -68,17 +85,30 @@ public void start(BundleContext bundleContext)
...
         {
-            logger.error("Failed to initialize Octo relay with address "
-                             + address + ":" + port + ". ", e);
+            logger.error("Failed to initialize Octo UDP transport with " +
+                "address " + address + ":" + port + ".", e);
+            return;
         }
+
+        octoTransport = new OctoTransport(publicAddress + ":" + port, logger);
+        logger.info("Created OctoTransport");
```
bgrozev, line 99: Move that log to `OctoTransport` so  it can print it's ID?

src/main/java/org/jitsi/videobridge/octo/OctoRelayService.java
```diff
@@ -87,9 +117,29 @@ public void start(BundleContext bundleContext)
...
         }
+        if (octoTransport != null)
+        {
+            octoTransport.stop();
+        }
+    }
+
+    public OctoRelayServiceStats getStats()
+    {
+        return new OctoRelayServiceStats(
```
bgrozev, line 132: Can we use JSON or a map? Using that many parameters is error prone.

bbaldino: I agree, and considered porting `OctoRelayService` to kotlin just to get named parameters here :) Neither JSON nor a map work great, because then the caller has to look them up by string which I wasn't crazy about.  I can change this to a builder-style so it's nice from java.

src/main/kotlin/org/jitsi/videobridge/transport/octo/OctoTransport.kt
```diff
@@ -0,0 +1,269 @@
...
+        }
+    }
+
+    fun stop() {
+    }
+
+    fun dataReceived(buf: ByteArray, off: Int, len: Int, receivedTime: Instant) {
+        val (conferenceId, mediaType, sourceEndpointId) = try {
+            Triple(
+                OctoPacket.readConferenceId(buf, off, len),
```
bgrozev, line 112: Is this efficient? Since it happens for each octo packt received, even something like allocating a `Triple` could have an effect (probably small, but still).

src/main/kotlin/org/jitsi/videobridge/transport/octo/OctoTransport.kt
```diff
@@ -0,0 +1,269 @@
...
+                OctoPacket.readMediaType(buf, off, len),
+                OctoPacket.readEndpointId(buf, off, len)
+            )
+        } catch (iae: IllegalArgumentException) {
+            logger.warn("Invalid Octo packet, len=$len", iae)
+            stats.invalidPacketReceived()
+            return
+        }
+        val handler = incomingPacketHandlers[conferenceId] ?: run {
+            logger.warn("No conference found for Octo packet with conference id $conferenceId")
```
bgrozev, line 122: Where did the log suppresion logic go?

src/main/kotlin/org/jitsi/videobridge/transport/octo/OctoTransport.kt
```diff
@@ -0,0 +1,269 @@
...
+        return String(
+            buf,
+            off + OCTO_HEADER_LENGTH,
+            len - OCTO_HEADER_LENGTH
+        ).also {
+            logger.cdebug { "Received a message in an Octo data packet: $it" }
+        }
+    }
+
+    private class Stats {
```
bgrozev, line 225: There's a single `OctoTransport` for the whole bridge, right? Doesn't it need to be thread safe?

src/main/kotlin/org/jitsi/videobridge/transport/udp/UdpTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+     */
+    fun startReadingData() {
+        val buf = ByteArray(1500)
+        val packet = DatagramPacket(buf, 0, 1500)
+        while (running.get()) {
+            // TODO: needed?
+            packet.setData(buf, 0, 1500)
+            try {
+                socket.receive(packet)
+                clock.instant().also { now ->
```
bgrozev, line 85: Is ther a reason this block is inside `try`?

Personal preference: I find this easier to read
```
val now = clock.instant()
stats.packetReceived(...)
...
```

src/main/kotlin/org/jitsi/videobridge/transport/udp/UdpTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+    /**
+     * Get the stats for this transport
+     */
+    fun getStatsJson(): OrderedJsonObject = OrderedJsonObject().apply {
+        put("bind_address", bindAddress)
+        put("bind_port", bindPort)
+        putAll(stats.toJson())
+    }
+
+    class Stats {
```
bgrozev, line 142: Same note about thread safety

src/main/kotlin/org/jitsi/videobridge/transport/udp/UdpTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+                sendPacketRate.update(1, timeMs)
+                sendBitRate.update(numBytes, timeMs)
+            }
+        }
+
+        fun incomingPacketDropped() {
+            incomingPacketsDropped++
+        }
+
+        fun outgoingPacketsDropped() {
```
bgrozev, line 181: Drop the `s` in `Packets`?

src/main/java/org/jitsi/videobridge/octo/OctoRelayService.java
bgrozev: We could have `VideobridgStatistics` export all contents of the map. Just use the existing constants so we don't break dashboards. I did something like this in jicofo recently and I think it's much cleaner.

src/main/java/org/jitsi/videobridge/octo/OctoRelayService.java
bbaldino: Good idea, fixed.

src/main/kotlin/org/jitsi/videobridge/transport/udp/UdpTransport.kt
bbaldino: Fixed

src/main/kotlin/org/jitsi/videobridge/transport/octo/OctoTransport.kt
bbaldino: Oops, totally blew that away. Fixed.

src/main/kotlin/org/jitsi/videobridge/transport/udp/UdpTransport.kt
bbaldino: Fixed

src/main/kotlin/org/jitsi/videobridge/transport/octo/OctoTransport.kt
bbaldino: True...I think the receive-side stats should technically be fine (`UdpTransport` will always call them from the same thread) but maybe I'll tweak all of them for consistency (and since technically it could be fed by something else).

src/main/java/org/jitsi/videobridge/octo/OctoRelayService.java
bbaldino: I tried the map approach, but it doesn't give a good way to fill default values if the relay service is null, which I think is necessary.

src/main/kotlin/org/jitsi/videobridge/transport/octo/OctoTransport.kt
bbaldino: I changed it...it feels painfully un-kotlin, and I'm tempted to say the `Triple` allocation won't be a big deal, but I'm not sure I'll get around to profiling it one way vs another.

src/main/kotlin/org/jitsi/videobridge/octo/OctoRelayServiceStats.kt
```diff
@@ -0,0 +1,70 @@
...
+        fun bytesSent(bytesSent: Long) = apply { this.bytesSent = bytesSent }
+        fun packetsReceived(packetsReceived: Long) = apply { this.packetsReceived = packetsReceived }
+        fun packetsSent(packetsSent: Long) = apply { this.packetsSent = packetsSent }
+        fun packetsDropped(packetsDropped: Long) = apply { this.packetsDropped = packetsDropped }
+        fun receiveBitrate(receiveBitrate: Long) = apply { this.receiveBitrate = receiveBitrate }
+        fun receivePacketRate(receivePacketRate: Long) = apply { this.receivePacketRate = receivePacketRate }
+        fun sendBitrate(sendBitrate: Long) = apply { this.sendBitrate = sendBitrate }
+        fun sendPacketRate(sendPacketRate: Long) = apply { this.sendPacketRate = sendPacketRate }
+        fun relayId(relayId: String) = apply { this.relayId = relayId }
+        fun build(): OctoRelayServiceStats = OctoRelayServiceStats(
```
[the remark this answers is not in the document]
bbaldino: Adding any stats to this flow requires knowing to add it in multiple places already (`VideobridgeStatistics`).  I looked at doing a map-based approach as boris suggested, which would be nice, but it doesn't work because `VideobridgeStatistics` relies on filling in fields with default values if they're null.  

What did you have in mind?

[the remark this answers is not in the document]
JonathanLennox: I'm not sure, it'd just be nice if missing a stat someplace was a compile-time error rather than a runtime error.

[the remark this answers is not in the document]
bbaldino: I had that originally, but boris brought up the issue of it being easy to mismatch the arguments (since java doesn't have named arguments), so I went with what I thought was the closest Java equivalent which is a builder.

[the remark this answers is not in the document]
bbaldino: I said above, but I thought about re-writing `OctoRelayService` in kotlin just to get the named args and then we can avoid this.  I could take a look at that?

PR #1207, opened 2020-04-17
Update dependencies to latest versions

123Haynes:
This PR updates the dependencies to their latest versions.

The exception is `org.igniterealtime.smack:smack-extensions` because of https://issues.igniterealtime.org/browse/SMACK-769

This requires a change in jicoco first.
I've successfully tested the changes against java 8 and java 11 😃


bgrozev:

PR #1215, opened 2020-04-21
Don't send in IceTransport if it's not running

bbaldino:
Sending here increased the likelihood that we'd send on a socket that was closed, since we free the `IceAgent` in `IceTransport#stop`.

### DIFF
--- a/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
+++ b/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
@@ -215,16 +215,16 @@ class IceTransport @JvmOverloads constructor(
 
     /**
      * Send data out via this transport
-     *
-     * TODO: should we ensure that we're running before sending?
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -215,28 +215,19 @@ class IceTransport @JvmOverloads constructor(
...
-            logger.error("Error sending packet", e)
-            throw RuntimeException()
+        if (running.get()) {
+            try {
+                iceComponent.socket.send(DatagramPacket(data, off, length))
+                packetStats.numPacketsSent++
+            } catch (e: IOException) {
+                logger.error("Error sending packet", e)
+                throw RuntimeException()
+            }
```
JonathanLennox, line 227: You just ignore the send entirely if you're not running - do you want to accumulate a stat or something?

PR #1221, opened 2020-04-23
Expose forwarded events changed and selection changed from Endpoint.

mstyura:
This PR deprecates https://github.com/jitsi/jitsi-videobridge/pull/1113.

I'm using `JVB 2.0` as a library to add custom logic on top of an existing functionality. Custom logic is injected in {{JVB}} as a custom bundle. So in implementing additional functionality I'm limited to public API of {{JVB}}.
Currently I very much need to know when endpoint's set of forwarded endpoints has changed, and need to know when endpoint is selected/unselected by other endpoints. There is no public way to know when this happen, because event notification is sent silently (to my custom logic) over message transport.

It would be really nice to have these events exposed from `Endpoint`, so I (and maybe others) is able to intercept an events and inject custom handler.

Thanks you very much for review and considering this functionality!


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1123,6 +1130,8 @@ public void sendLastNEndpointsChangeEvent(
             forwardedEndpoints, endpointsEnteringLastN, conferenceEndpoints);
 
         sendMessage(msg);
+
+        notifyForwardedEndpointsChanged(forwardedEndpoints, endpointsEnteringLastN, conferenceEndpoints);
```
bbaldino, line 1134: Hm, wondering if this feels a bit weird since this call already came from _outside_ the endpoint.  If anything, it seems like we shouldn't forward this here and instead have BitrateController be the source of truth (like in your other PR)

mstyura: @bbaldino ah, very good point! Then I'll bring back event from bitrate controller and move message sending and forwarding event into event hander of bitrate controller.

mstyura: @bbaldino I think I've fixed an issues you've raising during previous review. Could you please give another chance for this PR and have a look at it? Thanks a lot!

PR #1228, opened 2020-04-26
Fixed broken link to statistics.md

allo-:


allo-:
I do not like to sign a CLA, but you can use my patch under Apache-2.0.

I think this should be okay for such a trivial patch and Apache-2.0 permits commercial use, so I do not see a need for a CLA.

By the way, your CLA page tries to access my browser location. I think this is a bit intrusive, maybe you can configure your document viewer not to do this.

allo-:
cloes due to inactivity

PR #1231, opened 2020-04-28
Move octo pipelines to individual OctoEndpoint instances

bbaldino:
This is another step in making Octo/remote endpoints act more like local endpoints.  This change moves from a single `OctoTransceiver` instance per conference to one per `OctoEndpoint`, and the `OctoTentacle` demuxes incoming Octo packets to their appropriate `OctoEndpoint` for processing (note that a conference-level `OctoTransceiver` does still exist to process incoming Octo packets which originated from a remote bridge and have no source endpoint ID set).

This means that we no longer have a single incoming Octo packet queue per conference, but instead individual queues per remote endpoint, just like local endpoints.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
+++ b/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
@@ -224,7 +224,7 @@ public class ConfOctoTransport extends PropertyChangeNotifier
         }
         else
         {
-            // Packets without an endpoint ID originalted from within the bridge
+            // Packets without an endpoint ID originated from within the bridge
             // itself and, in practice, are things like keyframe requests.  We
             // send them out directly (without queueing).
             doSend(packet);
@@ -383,7 +383,7 @@ public class ConfOctoTransport extends PropertyChangeNotifier
     public void expire()
     {
         logger.info("Expiring");
-        setRelays(new LinkedList<>());
+        setRelays(Collections.emptyList());
         octoEndpoints.setEndpoints(Collections.emptySet());
         outgoingPacketQueues.values().forEach(PacketInfoQueue::close);
         outgoingPacketQueues.clear();
--- a/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
+++ b/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
@@ -115,6 +115,11 @@ public class ConfOctoTransport extends PropertyChangeNotifier
     private final Clock clock;
 
     /**
+     * Whether or not this {@link ConfOctoTransport} is currently active
+     */
+    private final AtomicBoolean running = new AtomicBoolean(true);
+
+    /**
      * Initializes a new {@link ConfOctoTransport} instance.
      * @param conference the conference.
      */
@@ -169,7 +174,10 @@ public class ConfOctoTransport extends PropertyChangeNotifier
      */
     public void addPayloadType(PayloadType payloadType)
     {
-        octoEndpoints.addPayloadType(payloadType);
+        if (running.get())
+        {
+            octoEndpoints.addPayloadType(payloadType);
+        }
     }
 
     /**
@@ -179,9 +187,13 @@ public class ConfOctoTransport extends PropertyChangeNotifier
      */
     public void setRelays(Collection<String> relays)
     {
+        if (!running.get())
+        {
+            return;
+        }
         Objects.requireNonNull(
             bridgeOctoTransport,
-                "Octo requested but not configured");
+            "Octo requested but not configured");
 
         Set<SocketAddress> socketAddresses = new HashSet<>();
         for (String relay : relays)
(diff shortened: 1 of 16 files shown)


bbaldino:
Here's a diagram of the flow after these changes:

HB6fpNsErrmd (2)

src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
```diff
@@ -311,31 +398,116 @@ public void expire()
...
+        bridgeOctoTransport.sendString(
             message,
             targets,
             conference.getGid()
         );
     }
 
+    public void addHandler(String epId, IncomingOctoEpPacketHandler handler)
+    {
+        logger.info("Adding handler for ep ID " + epId);
```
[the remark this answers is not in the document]
bbaldino: I was thinking this isn't too bad, since it's just one per endpoint and, if for some reason we thought packets weren't flowing the first questions we would ask would be if the handler was put in so it seemed valuable.

src/main/java/org/jitsi/videobridge/octo/OctoTentacle.java
```diff
@@ -302,7 +388,8 @@ public void expire()
         logger.info("Expiring");
         setRelays(new LinkedList<>());
```
JonathanLennox, line 389: If you make this `Collections.emptyList()` you'll get an immutable list, which ensures no relays are added to an expired tentacle.

bbaldino: We always reassign a new value though (as opposed to adding and removing)--though it should use `Collections.emptyList` anyway.

JonathanLennox:
Now that octo receive queues are per remote endpoint, we should probably change `jvb.octo.recv-queue-size` to be scaled like the standard endpoint queue.

src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
```diff
@@ -0,0 +1,561 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean wants(PacketInfo packetInfo)
+    {
+        // Cthulhu devours everything (as long as it's not coming from
+        // itself, and we have targets).
+        return running.get() &&
```
bgrozev, line 219: This is probably premature optimization, but...

Most rejected packets will come from the `instanceof` check. Do we want to put it in front to avoid some calls to `AtomicBoolean.get()` (coming from multiple threads, iirc)?

bbaldino: I can do that...if it makes you feel any better `AtomicBoolean` calls didn't come up at all in a 5 minute, 20 person conference (split across 2 bridges, sampled from one).  But I did CPU sampling, not tracing.  I can run a trace real quick and get more data.

src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
```diff
@@ -0,0 +1,561 @@
...
+        }
+        stats.packetReceived(packetInfo.getPacket().length, clock.instant());
+        IncomingOctoEpPacketHandler handler = incomingPacketHandlers.get(packetInfo.getEndpointId());
+        if (handler != null)
+        {
+            handler.handleIncomingPacket(packetInfo);
+        }
+        else
+        {
+            stats.incomingPacketDropped();
```
bgrozev, line 286: Return the buffer to the pool?

bbaldino: Good catch!

src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
bbaldino: I tracked this and it looks like it gets inlined entirely, so I think we're good.

src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
```diff
@@ -0,0 +1,562 @@
...
+            OrderedJsonObject debugState = new OrderedJsonObject();
+            debugState.put("packets_received", packetsReceived);
+            debugState.put("receive_packet_rate_pps", receivePacketRate.getRate());
+            debugState.put("bytes_received", bytesReceived);
+            debugState.put("receive_bitrate_bps", receiveBitRate.getRate());
+            debugState.put("incoming_packets_dropped", incomingPacketsDropped);
+
+            debugState.put("packets_sent", packetsSent.sum());
+            debugState.put("send_packet_rate_pps", sendPacketRate.getRate());
+            debugState.put("bytes_sent", bytesSent);
```
bgrozev, line 552: `bytesSent.sum()`

src/main/resources/reference.conf
```diff
@@ -65,9 +65,9 @@ videobridge {
     # The size of the incoming octo queue. This is a conference level queue and
     # it's where we temporarily store packets from remote bridges.
     #
-    # With 4096 we observe no drops in a 100-participant call with 
-    # 20 senders across 4 bridges (c5.xlarge)
-    recv-queue-size=4096
+    # This queue is per-remote-endpoint, so it matches what we use for local
```
bgrozev, line 68: Remote the "this is a conference level queue" text above?

PR #1241, opened 2020-05-03
Don't expose secret in process listing

### LINKED ISSUE #1237: Don't expose password in process listing
## Description
The service will use the authentication password as a command line option (`--secret=`).
This will allow **any** user on this system to obtain the password from the process table (`ps ...`)

---

## Current behavior
The service will use the authentication password as a command line option (`--user_password=`).
This will allow **any** user on this system to obtain the password from the process table (`ps ...`)

---

## Possible Solution
I haven't touched Java for years, but if I'm recalling correctly the command line arguments  are not immutable in Java (they aren't in C/Go/Rust).

Unlike "classic" UNIX systems the process arguments on Linux are not read only. Changing the relevant argument to some obfuscated value (after processing them) will be reflected in the process output of the operating system.

This can be used to prevent access to the password/shared secrets to anyone running `ps`.

obfusk:
Fixes  #1237.


bgrozev:
Thanks for the contribution. We're not ignoring it, we'll address this soon

obfusk:
Fixed by #1266.

PR #1248, opened 2020-05-06
doc(pom): add java version

alexseedkou:
add java version in pom.xml to avoid any build error caused by java version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,6 +8,7 @@
   <name>jitsi-videobridge</name>
   <url>https://github.com/jitsi/jitsi-videobridge</url>
   <properties>
+    <java.version>1.8</java.version>
     <assembly.skipAssembly>false</assembly.skipAssembly>
     <exec.mainClass>org.jitsi.videobridge.Main</exec.mainClass>
     <jetty.version>9.4.15.v20190215</jetty.version>


alexseedkou:
> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

I have done that.

bgrozev:
Thanks for the contribution (confirmed the CLA).

Why is this necessary? The maven compiler plugin is already configured with source and  target of 1.8: https://github.com/jitsi/jitsi-videobridge/blob/master/pom.xml#L483

alexseedkou:
> Thanks for the contribution (confirmed the CLA).
> 
> Why is this necessary? The maven compiler plugin is already configured with source and target of 1.8: https://github.com/jitsi/jitsi-videobridge/blob/master/pom.xml#L483

Hi,

Thanks for your feedback, I faced  a build failure before the change.

I was trying to produce the bug today and it seems that failure is related to the network which results in a failure to download a dependency for build. Therefore I think we are good here.

PR #1249, opened 2020-05-06
Allocate bitrate evenly among all selected endpoints.

JonathanLennox:
This should cause fewer "ninjas" in tile view, at the expense of tile view sources sometimes being shown in lower resolutions or frame rates.


JonathanLennox:
Closed in favor of https://github.com/jitsi/jitsi-videobridge/pull/1250 .

PR #1250, opened 2020-05-07
ref: Unifies pinned, selected and receiver video constraints.

### LINKED ISSUE #147: Fine-Grained Control of Forwarded Simulcast Layers

gpolitis:
This is a WIP effort to unify the pinned, selected and receiver video
constraints messages to the bridge, all while providing full backwards
compatibility with existing client side code.

Phase 2 (implying this WIP is phase 1) is to propagate endpoint
constraints across bridges to enable off-stage video suspension. To
minimize and simplify messaging we consider two types of receivers for a
given stream of a given endpoint, as per @bbaldino's suggestion: a)
endpoint receivers and b) bridge receivers.

The endpoint receivers are straightforward. An endpoint X on bridge B1
receiving 720p from endpoint Y on the same bridge will count as 1
receiver (of 720p from endpoint Y).

The bridge receivers can be thought of this way: Any number of endpoints
on bridge B2 receiving 720p from endpoint Y on bridge B1 will again
count as 1 receiver.

Fixes #147.
Also addresses https://github.com/jitsi/jitsi-videobridge/pull/1249.


src/main/java/org/jitsi/videobridge/Constraints.java
```diff
@@ -0,0 +1,26 @@
+package org.jitsi.videobridge;
+
+public class Constraints
```
bbaldino, line 3: Do we need a separate class for this?  Could endpoints just have 2 instances of `EndpointConstraints`? One for global and one for their own?

bbaldino: We could even have a single instance of the 'global constraints' per conference and give the endpoints a read-only form of it?  That would mean instead of `Constraints` and `EndpointConstraints`, we have `EndpointConstraints` and `ReadOnlyEndpointConstraints`.  An `Endpoint` would have an instance of `EndpointConstraints` for its own, specific constraints and be given an instance of `ReadOnlyEndpointConstraints` which was the shared, global endpoint constraints for that conference?

gpolitis: > Do we need a separate class for this? Could endpoints just have 2 instances of EndpointConstraints? One for global and one for their own?

You're right, this feels redundant.

> We could even have a single instance of the 'global constraints' per conference and give the endpoints a read-only form of it?

'Global constraints' are per receiver tho, or am I reading this wrong?

bbaldino: Maybe I misunderstood "Global constraints", I assumed they were constraints set on, say, a conference level (as opposed to an Endpoint level) that were applied to all Endpoints?

gpolitis: @bbaldino ah, I see now where's the confusion. Admittedly I chose a very misleading name.

Imagine that this `bitrateController.globalConstraints` effectively replaces the `bitrateController.maxRxFrameHeightPx ` constraint. So, it's not conference level, it's endpoint (receiver) level but for all the endpoints (senders) that the receiver receives.

We should probably rename this to something else..

Maybe a little context here would help, so for the time being I'm trying to refactor what's already there, and then add the new stuff we had been discussing.

mstyura:
Hello @gpolitis!
I'm glad to see this PR intended to unify and enrich control over video sent/received by endpoints. May I ask you to consider enriching control over video even more? I'm currently working on a project which does user `JVB` as `SFU`, but with own signaling on top of it (not using `jicofo/meet`). I have clients which could send several video tracks, for example, screen and camera. What I've found very useful, but currently impossible, is to request receive video quality **per video track**. Current `JVB` `API` only allows binary control over video quality: for selected endpoint high quality video will be forwarded to destination, for not selected - low quality video will be forwarded. It believe it would be very useful to implement ability to control receive video quality not per whole endpoint, but per video track which is sent by endpoint. This will allow scenarios, where for example only 1 of N video tracks forwarded to destination endpoint or when screen share forwarded in high quality, while camera video in preview quality. I believe such flexibility of video quality control will allow great new scenarios bot for future `Jitsi Meet` and side project, which uses `JVB` as `SFU`. Thanks a lot in advance for considering this option!

gpolitis:
> What I've found very useful, but currently impossible, is to request receive video quality per video track.

We have considered [this](https://github.com/jitsi/jitsi-videobridge/pull/1250/files#diff-166b674840c19251cc8750754e36b96dR22) already but unfortunately it is out of reach right now, especially given that this work is an intermediate step towards something else.

The bitrate controller is already track aware (it distributes bandwidth to tracks, not endpoints), but neither the bridge nor the client keeps track of media stream track identifiers (mids) and it would be too time consuming to add that on top of the other changes that we're planning.

We will eventually get there.

mstyura:
@gpolitis thanks for your response!
> The bitrate controller is already track aware (it distributes bandwidth to tracks, not endpoints), but neither the bridge nor the client keeps track of media stream track identifiers (mids) and it would be too time consuming to add that on top of the other changes that we're planning

Would it be possible to consider the following solution, which is very good for me and hopefully acceptable to `Jitsi`. 
As you've said `BitrateController` is already track aware, so it's API could be enriched to expose methods to explicitly request desired quality per  video  track.
`JVB` will not use this `API` until some moment in a future.
The presence of such `API` will allow other projects which uses `JVB` to get an access to `BitrateController` and ask for a desired quality of track "manually".

For example, in my project I already keep track of all endpoint tracks :) and their state, if I could get a reference to `BitrateController` instance and explicitly ask for desired track quality I would be very happy with that solution. An I hopefully it would be fine to `JVB` as well, because it will offload necessity to keep track of media tracks to a application which uses `JVB`.

So if me or someone else can provide a PR which will enrich public `API` of `BitrateController`, would it be possible to get it merged, even so `JVB` itself will not use it, because it does not keep track of video tracks?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -927,27 +1004,47 @@ private synchronized void update()
...
+        Map<String, EndpointConstraints> endpointConstraintsMapCopy = endpointConstraintsMap;
+        EndpointConstraints globalConstraintsCopy = globalConstraints;
+
+        List<EndpointMultiRank> endpointMultiRankList = IntStream
+            .range(0, conferenceEndpoints.size() - 1)
+            .mapToObj(i -> {
+                AbstractEndpoint endpoint = conferenceEndpoints.get(i);
+                EndpointConstraints endpointConstraints = globalConstraintsCopy
+                    .of(endpoint.getID())
+                    .unless(endpointConstraintsMapCopy.get(endpoint.getID()));
```
bbaldino, line 1030: So we translate the 'global constraints' into instances of Endpoint constraints for every endpoint (that doesn't already have a specific constraint set)?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +267,53 @@ private static boolean changeIsLargerThanThreshold(
             ||  deltaBwe < -1 * previousBwe * Config.bweChangeThresholdPct() / 100;
     }
 
+    static class EndpointMultiRank
+    {
+        final int speakerRank;
+        final EndpointConstraints endpointConstraints;
+        final AbstractEndpoint endpoint;
```
bbaldino, line 274: Can you document which endpoint this refers to? Is it the one for whom the constraints apply (the sender)? Or the one applying the constraint (the receiver)?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +267,53 @@ private static boolean changeIsLargerThanThreshold(
             ||  deltaBwe < -1 * previousBwe * Config.bweChangeThresholdPct() / 100;
     }
 
+    static class EndpointMultiRank
+    {
+        final int speakerRank;
```
bbaldino, line 272: doc please

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +267,53 @@ private static boolean changeIsLargerThanThreshold(
...
+            // loops over the endpoints multiple times, improving their target
+            // bitrate at ever step, until no further improvement is possible.
+            //
+            // In this multi-rank implementation, endpoints that have 720p ideal
+            // height will be given bandwidth before endpoints that have 180p
+            // ideal height. If two endpoints have the same ideal height, then
+            // we look at their speech rank (whoever spoke last has is ranked
+            // higher).
+
+            int idealHeightDiff = o1.endpointConstraints.getIdealHeight() - o2.endpointConstraints.getIdealHeight();
```
bbaldino, line 305: Making sure I understand it right: this ranking is specifically for bandwidth allocation, which is why we prioritize first by ideal height, and then by speaker rank?

src/main/java/org/jitsi/videobridge/Constraints.java
bbaldino: Ah, ok, I think I see now.  I had the scopes wrong.  The name isn't too bad, actually, maybe we just document it somewhere.  Sorry for being slow on this, I left some more comments.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: Yes, which means that we allocate bandwidth first for "selected" endpoints (720p ideal resolution), starting with the most recent active speaker in that group, then 180p again starting with the most recent active speaker in that group and then everything else.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -386,6 +424,43 @@ public JSONObject getDebugState()
...
+            this.endpointConstraintsMap = newEndpointConstraintsMap;
+        }
+
+        if (needsUpdate)
+        {
+            update();
+        }
+    }
+
+    public void setGlobalConstraints(EndpointConstraints newGlobalConstraints)
```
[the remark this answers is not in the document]
gpolitis: Yes!!

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -386,6 +424,43 @@ public JSONObject getDebugState()
         return debugState;
     }
 
+    public void setEndpointConstraints(Set<EndpointConstraints> newEndpointConstraints)
+    {
+        boolean needsUpdate = false;
+        Map<String, EndpointConstraints> oldEndpointConstraintsMap = this.endpointConstraintsMap;
+        if (newEndpointConstraints.size() != oldEndpointConstraintsMap.size())
```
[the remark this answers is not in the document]
gpolitis: 🤦

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: Yes, that was the thinking. But we no longer need to translate now that we have `VideoConstraints`.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -390,9 +392,17 @@ protected void onSelectedEndpointsChangedEvent(
...
      * @param newPinnedEndpoints the new pinned endpoints
      */
-    protected abstract void onPinnedEndpointsChangedEvent(
+    protected void onPinnedEndpointsChangedEvent(
         JSONObject jsonObject,
-        Set<String> newPinnedEndpoints);
+        Set<String> newPinnedEndpoints)
+    {
+        // Note that this captures the set.
+        pinnedEndpointConstraintsStream = newPinnedEndpoints
```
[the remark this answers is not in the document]
gpolitis: Are you talking about removing the memory of what's pinned and what's selected? We need that, consider, for example, the case where a pinned endpoint comes on-stage and then off-stage again. It's still a pinned endpoint (the video constraints class has good info on what's pinned and what's selected in case it's confusing).

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -386,6 +424,43 @@ public JSONObject getDebugState()
...
+        {
+            Map<String, EndpointConstraints> newEndpointConstraintsMap
+                = new HashMap<>(newEndpointConstraints.size());
+
+            for (EndpointConstraints endpointConstraints : newEndpointConstraints)
+            {
+                String endpointId = endpointConstraints.getEndpointId();
+                newEndpointConstraintsMap.put(endpointId, endpointConstraints);
```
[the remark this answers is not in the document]
gpolitis: Things got refactored, this doesn't apply anymore.

gpolitis:
> For example, in my project I already keep track of all endpoint tracks :) and their state, if I could get a reference to BitrateController instance and explicitly ask for desired track quality I would be very happy with that solution.

@mstyura even if your client is msid aware, the bridge is not (the bandwidth allocation algorithm is track aware, but this is only half of the story). Maybe the work that @JonathanLennox's track refactoring will make what you're suggesting possible?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1346,10 +1342,14 @@ private TrackBitrateAllocation(
...
             {
-                if (maxFrameHeight >= 0 && encoding.getHeight() > maxFrameHeight)
+                int idealHeight = videoConstraints.getIdealHeight();
+                if (idealHeight >= 0 && encoding.getHeight() > idealHeight)
                 {
                     continue;
                 }
+
+                boolean selected = videoConstraints
+                    .equals(VideoConstraints.SELECTED_ENDPOINT_CONSTRAINT);
```
bbaldino, line 1352: I thought the `idealHeight` in the constraint obsoletes the idea of explicitly setting something as 'selected'?

gpolitis: The `VideoConstraints.SELECTED_ENDPOINT_CONSTRAINT` is meant to be used when translating the selected endpoint messages into video constraints messages for backwards compatibility purposes so that we're not forced to change anything in the client.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -386,6 +449,46 @@ public JSONObject getDebugState()
         return debugState;
     }
 
+    public void setVideoConstraints(Map<String, VideoConstraints> newVideoConstraintsMap)
+    {
+        Map<String, VideoConstraints>
+            videoConstraintsMapCopy = this.videoConstraintsMap;
```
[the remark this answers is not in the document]
gpolitis: The purpose isn't to create a deep copy, it's to prevent the reference from changing while we're accessing it (we're doing `VideoConstraints oldVideoConstraints = videoConstraintsMapCopy.get(endpointId);` in this method.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -386,6 +449,46 @@ public JSONObject getDebugState()
...
+                videoConstraintsEntry : newVideoConstraintsMap.entrySet())
+            {
+                String endpointId = videoConstraintsEntry.getKey();
+                VideoConstraints newVideoConstraints = videoConstraintsEntry.getValue();
+                VideoConstraints oldVideoConstraints = videoConstraintsMapCopy.get(endpointId);
+                if (!newVideoConstraints.equals(oldVideoConstraints))
+                {
+                    needsUpdate = true;
+                    break;
+                }
```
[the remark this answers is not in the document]
gpolitis: Yes!

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -46,8 +46,20 @@
...
     protected final @NotNull Logger logger;
-    private Map<String, VideoConstraints> pinnedEndpointConstraintsMap;
-    private Map<String, VideoConstraints> selectedEndpointConstraintsMap;
+
+    /**
+     * The last pinned endpoints set signaled by the client, converted into
+     * a map of endpoint id -> video constraints. We need this for backwards
+     * compatibility (see more bellow in {@link #onVideoConstraintsChangedEvent()}
+     */
+    private Map<String, VideoConstraints> pinnedVideoConstraintsMap;
```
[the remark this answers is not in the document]
gpolitis: Do we want to provide backwards compatibility for pinned endpoints or do we decide to axe it?

The state is part of the compatibility layer that handles cases like this: a pinned endpoint comes on-stage and then off-stage again. It's still a pinned endpoint.

With the new video constraints this would client side work: The client would send 180p video constraint to "pin" an endpoint A, then a 720p video constraint when endpoint A becomes selected (i.e. comes on-stage) and then a 180p video constraint again when endpoint A goes off-stage (but the user still would like to see endpoint A in the thumbnails).

We don't have these mechanics in the client right now (nor people who rely on selected and pinned messages) but we can chose to not provide backwards compatibility for pinned endpoints (as we've dropped them a while ago).

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: I misunderstood you, the track allocation should operate on constraints only. Is that your point?

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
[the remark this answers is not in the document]
bbaldino: Thanks for explaining the scenario.  It still seems like the message transport is a weird place to keep this state, could we have this compatibility logic in Endpoint?

Also, do we have a plan for how we'd implement this behavior in the new scheme?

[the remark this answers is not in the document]
bbaldino: > We don't have these mechanics in the client right now (nor people who rely on selected and pinned messages) but we can chose to not provide backwards compatibility for pinned endpoints (as we've dropped them from Jitsi Meet a while ago, not sure when).

Oh, really?  If so then I think we can avoid the backwards compatibility in the bridge for pinned and we'll just do some new scheme when we bring it back.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +267,78 @@ private static boolean changeIsLargerThanThreshold(
...
+    /**
+     * A helper class that is used to determine the bandwidth allocation
+     * rank/priority of an endpoint that is based on its speaker rank and its
+     * video constraints. See {@link EndpointMultiRanker} for more information
+     * on how the ranking works.
+     */
+    static class EndpointMultiRank
+    {
+        /**
+         * The speaker rank of the {@link #endpoint}.
```
bbaldino, line 279: Can you document the order of the ranking (is 0 high or low?)

src/main/kotlin/org/jitsi/videobridge/octo/OctoEndpoint.kt
```diff
@@ -72,6 +73,14 @@ class OctoEndpoint(
         // single OctoEndpoints instance.
     }
 
+    override fun setVideoConstraints(newVideoConstraints: MutableMap<String, VideoConstraints>?) {
+        TODO("Not yet implemented")
```
bbaldino, line 77: If we want this to be a no-op, we should remove the `TODO` as it'll throw an exception if called. (and below)

src/main/java/org/jitsi/videobridge/VideoConstraintsCompatibility.java
```diff
@@ -92,10 +92,41 @@
...
+            // https://github.com/jitsi/jitsi-meet/commit/29bc18df01c82cefbbc7b78f5aef7b97c2dee0e4
+            // https://github.com/jitsi/jitsi-meet/commit/e63cd8c81bceb9763e4d57be5f2262c6347afc23
+            //
+            // This means that the condition selectedEndpoints.size() > 1 is
+            // equivalent to tile-view.
+
+            // In tile view we set the ideal height but not the preferred height
+            // nor the preferred frame-rate because we want even even
+            // distribution of bandwidth among all the tiles to avoid ninjas.
+            selectedEndpointConstraints
```
bbaldino, line 120: Maybe not now, but I think in the future we should consider a 'receiver quality policy' to handle this: basically we'd have a policy to prefer quality or number of participants viewed.

Preferring quality would mean that we go through the list of endpoints and pick the maximum quality that can fit for each one until we run out of 'room'

Preferring number of participants viewed would iterate over all the endpoints and get them in their lowest quality, then iterate again to try and improve quality for each one until we ran out of room.

'Stage view' would be the first policy (maximize quality of active speaker), 'tile view' would be the second (maximize the number of participants seen).

gpolitis: Yes, the receiver quality policy idea makes sense.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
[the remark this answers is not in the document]
gpolitis: I'll mark this conversation as resolved because we achieved this with the compatibility layer.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: I'll mark this conversation as resolved because I think this has now been addressed.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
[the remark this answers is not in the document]
gpolitis: We've removed the complexity from the transport and encapsulated all the backwards compatibility logic in a separate component, hopefully making things a bit clearer. I'll mark this conversation as resolved now.

src/main/java/org/jitsi/videobridge/VideoConstraintsCompatibility.java
JonathanLennox: Doesn't the "preferred" constraint per source handle this?  If you want to prefer number of participants, set the preferred quality per-source to something low.

src/main/kotlin/org/jitsi/videobridge/octo/OctoEndpoint.kt
gpolitis: I don't think octo video constraint propagation will go through this so I'll make it a NO-OP.

gpolitis:
I'm thinking to add the colibri messages in this PR before we merge it, to make it actually possible for the client to make use of the new message. What do you guys think?

src/main/java/org/jitsi/videobridge/VideoConstraintsCompatibility.java
gpolitis: Τhanks for pointing that out @JonathanLennox, that's exactly the intention

bbaldino: > Doesn't the "preferred" constraint per source handle this? If you want to prefer number of participants, set the preferred quality per-source to something low.

I'm not sure I see how it does. How does a receiver know what to set the preferred resolution to?  What if it sets 180, but it actually could have received 360 for all (or some) of the senders?

bbaldino:
> I'm thinking to add the colibri messages in this PR before we merge it, to make it actually possible for the client to make use of the new message. What do you guys think?

I'd be fine with doing those separately, even if it meant this code was somewhat "dead" until they were in, just in the interest of separating the PRs.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -364,9 +432,7 @@ public JSONObject getDebugState()
         debugState.put("forwardedEndpoints", forwardedEndpointIds.toString());
         debugState.put("trustBwe", Config.trustBwe());
         debugState.put("lastBwe", lastBwe);
-        debugState.put("maxRxFrameHeightPx", maxRxFrameHeightPx);
-        debugState.put("selectedEndpointIds", selectedEndpointIds.toString());
-        debugState.put("pinnedEndpointIds", pinnedEndpointIds.toString());
+        debugState.put("videoConstraints", videoConstraintsMap);
```
bbaldino, line 435: I don't think this will format correctly when printed, will it?

src/main/kotlin/org/jitsi/videobridge/cc/config/BitrateControllerConfig.kt
```diff
@@ -114,6 +114,24 @@ class BitrateControllerConfig {
 
             @JvmStatic
             fun trustBwe() = trustBweProp.value
+
+            /**
+             * The property for the max resolution to allocate for the onstage
+             * participant.
+             */
+            class OnstageMaxHeightPixelsProperty : LegacyFallbackConfigProperty<Int>(
```
bbaldino, line 122: Did this exist previously?  If not, just do a `SimpleProperty` which reads only from the new config

src/main/java/org/jitsi/videobridge/VideoConstraintsCompatibility.java
JonathanLennox: "Preferred" is not "ideal" - it's the minimum to give this source before moving on to the next source.  If every source has reached its "preferred", then the server starts upping resolutions (round-robin, starting with the first) up to "ideal".

Perhaps this is badly named - better names suggested?

bbaldino: If that's true then I do think we should come up with a better name.  Maybe `minimum`, as that seems to be what it is.  

But I still think the receiver is not in a good position to set this constraint.  Consider the 'main stage' case: a receiver would set `ideal` and `minimum` to 720p, I assume? What if they don't have the bandwidth to fit that stream?  If the JVB strictly obeys, then it would forward nothing.  It seems like letting the receiver set a policy as to how the JVB would fallback would make it easier for the receiver to express what it wants without code having to 'guess' (either on the receiver side or the JVB's side), then it can express an 'ideal' but let the JVB fit things in based on the available bandwidth.

JonathanLennox: It's not 'minimum' either - the bridge will give less than this value, but if so, won't allocate any bandwidth to any later sources.

To give an example, right now, in main stage the bridge implicitly sets this value to 360p/30 for the selected participant, and 180p/30 for all others.  So it allocates the selected source 360p/30, then allocates bandwidth for the filmstrip sources, and then if there's still available bandwidth ups the main stage to 720p/30.

bbaldino: I'd say this other field feels pretty awkward to describe then.  If there's some kind of over-arching 'receiver policy', then a client need only give its ideal height for each sender (or 'pseudo sender', e.g. 'current active speaker') and the policy for how the bridge should degrade if it can't fit that ideal.

1. Stage view: receiver sets ideal to 720p for active speaker, 180p for everyone else.  Policy is "prefer quality": bridge would sort by ideal resolution for senders and fulfill each one completely before moving to the next.

2. Tile view: receiver sets ideal to 720p for all speakers (say there are only 4, so enough screen to fit them in 720p), policy is "prefer number of video streams": bridge iterates through all streams and fits in one layer at a time, going back over them and upgrading the layers while there's still room.

There are some slight twists there with how we decide what order to go over the streams in, but I think that exists in any implementation.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: oh, it will https://github.com/jitsi/jitsi-videobridge/pull/1250/commits/df9fef8495802b61b6a98590e8974db8bb0cef48

bbaldino: But this is a `Map<String, VideoConstraints>` ?

src/main/kotlin/org/jitsi/videobridge/cc/config/BitrateControllerConfig.kt
gpolitis: Done that and I've also renamed the config property to `ideal` height to match the video constraint field that it pertains to. Unfortunately the new name creates a small incoherence with the max thumbnail height config property name.

To be consistent and correct, we should probably rename the thumbnail property name, but I'm worried this may break existing systems..

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: It works (I've tested it) as you'd expect because the root JSONObject knows how to convert the map values to JSON by calling their `toString` method.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +269,80 @@ private static boolean changeIsLargerThanThreshold(
...
+     * we look at their speech rank (whoever spoke last has is ranked
+     * higher).
+     */
+    static class EndpointMultiRanker
+        implements Comparator<EndpointMultiRank>
+    {
+        @Override
+        public int compare(EndpointMultiRank o1, EndpointMultiRank o2)
+        {
+            int idealHeightDiff = o1.videoConstraints.getPreferredHeight() - o2.videoConstraints.getPreferredHeight();
```
bgrozev, line 334: The doc and variable name say "ideal", but it uses "preferred". Is this intentional?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +269,80 @@ private static boolean changeIsLargerThanThreshold(
...
+            if (idealHeightDiff != 0)
+            {
+                return idealHeightDiff;
+            }
+            else
+            {
+                return o1.speakerRank - o2.speakerRank;
```
bgrozev, line 341: Is this backwards? According to the docs, speaker rank 0 is the highest

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -927,27 +993,42 @@ private synchronized void update()
...
+                VideoConstraints videoConstraints = videoConstraintsMapCopy
+                    .getOrDefault(endpoint.getID(), defaultVideoConstraints);
+
+                int rank = conferenceEndpoints.indexOf(endpoint);
+                return new EndpointMultiRank(rank, videoConstraints, endpoint);
+            })
+            .sorted(new EndpointMultiRanker())
+            .collect(Collectors.toList());
+
+        for (EndpointMultiRank endpointMultiRank : endpointMultiRankList)
```
bgrozev, line 1021: You don't use the rank other than to access the endpoint. I think you can you just `.map(r -> r.endpoint)` after sorting the stream above and then work with the endpoints.

src/main/java/org/jitsi/videobridge/VideoConstraints.java
```diff
@@ -0,0 +1,122 @@
...
+        VideoConstraints that = (VideoConstraints) o;
+        return idealHeight == that.idealHeight
+            && preferredHeight == that.preferredHeight
+            && preferredFps == that.preferredFps;
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(idealHeight);
```
bgrozev, line 110: What is the reason for not including `preferredHeight` and `preferredFps` here? Looking at `BitrateController#setVideoConstraints`, I think it might fail to call `update` if the only thing that changed is the preferred height or FPS. Or is this on purpose?

gpolitis: > Or is this on purpose?

No!!

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: Thanks Boris!

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: Thanks Boris!

gpolitis:
It would be nice to have this merged to test it on beta. I also have a 2nd PR coming that builds upon this one. @bbaldino @bgrozev @JonathanLennox could you please take a final look?

bgrozev:
lgtm

src/main/java/org/jitsi/videobridge/VideoConstraints.java
```diff
@@ -0,0 +1,122 @@
...
+     */
+    private final double preferredFps;
+
+
+    /**
+     * Ctor.
+     *
+     * @param idealHeight The ideal height of the constrained endpoint.
+     * @param preferredFps
+     * @param preferredHeight
```
JonathanLennox, line 53: Params are out of order in javadoc

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +269,86 @@ private static boolean changeIsLargerThanThreshold(
...
+    }
+
+    /**
+     * An endpoint that has higher priority/rank will be allocated
+     * bandwidth prior to other endpoints with lower priority/rank
+     * (see the allocate method bellow)
+     *
+     * Once the endpoints are ranked, the bandwidth allocation algorithm
+     * loops over the endpoints multiple times, improving their target
+     * bitrate at ever step, until no further improvement is possible.
```
JonathanLennox, line 320: Typo: "every step"

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -275,6 +269,86 @@ private static boolean changeIsLargerThanThreshold(
...
+     *
+     * In this multi-rank implementation, endpoints that have a preferred height
+     * set (on-stage endpoints in Jitsi Meet) will be given bandwidth first.
+     * Then we prioritized endpoints that have higher ideal height (this rule is
```
JonathanLennox, line 324: Wrong tense: should be "Then we prioritize"

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -927,27 +999,42 @@ private synchronized void update()
...
-        // be visible).
-        for (Iterator<AbstractEndpoint> it = conferenceEndpoints.iterator();
-             it.hasNext() && endpointPriority < adjustedLastN;)
+        // Before endpoint constraints we had had the notion of selected and
+        // pinned endpoints. "selected" endpoints were endpoints that had
+        // 720p ideal height and prioritized over everything else and were
+        // allocated bandwidth first (up to their preferred resolution).
+        // "Pinned" endpoints where those thee had 180p ideal height and were
+        // ranked lower than selected endpoints but before the rest of the pack.
+        // Finally, any remaining endpoints were ranked after pinned endpoints.
```
JonathanLennox, line 1011: I don't think we need a comment describing what the code used to do -- that's what git history is for.  Better to describe what it's doing now.

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1346,37 +1315,42 @@ private TrackBitrateAllocation(
...
+                // The guard below should lead to no video for that endpoint.
+                //
+                // It works today because of how we interpret signaling from the
+                // endpoints: the bridge assumes that the first encoding is 180p
+                // (and that the second is 360p and the third 720p).
+                //
+                // When we change how tracks are signaled or how we discover the
+                // resolution of each individual encoding, this guard will have
+                // to be tweaked.
+                if (idealHeight >= 0 && encoding.getHeight() > idealHeight)
```
JonathanLennox, line 1333: How hard would it be to fix this now?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1346,37 +1315,42 @@ private TrackBitrateAllocation(
...
-                        ratedPreferredIdx = ratesList.size() - 1;
-                    }
-                }
-                else if (encoding.getHeight() <= Config.thumbnailMaxHeightPx())
+                // For the "selected" participant we favor frame rate over
+                // resolution. Basically what we want for the on-stage
+                // participant is 180p7.5fps, 180p15fps, 180p30fps, 360p30fps
+                // and 720p30fps. For the thumbnails, we consider all temporal
+                // layers of the low resolution stream.
+
```
JonathanLennox, line 1343: Does this generalize?  Can it be described in terms of the generalized video constraints, with this common case as an example?

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1385,6 +1359,11 @@ else if (encoding.getHeight() <= Config.thumbnailMaxHeightPx())
                     ratesList.add(
                         new RateSnapshot(encodingBitrateBps, encoding));
                 }
+
+                if (encoding.getHeight() <= videoConstraints.getPreferredHeight())
```
JonathanLennox, line 1363: This could probably use a comment, explaining that we jump immediately to the preferred encoding.

bbaldino:
lgtm

src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: I think we could fix this by having at least 1 rated encoding.

gpolitis:
@JonathanLennox I think I've addressed all your comments. Please let me know if you there's anything else you would like me to change/address/clarify.

gpolitis:
Thanks @JonathanLennox, I was going to forget this..

PR #1252, opened 2020-05-07
InterruptedExceptions should never be ignored in the code.

kamaci:


bbaldino:
Can you educate me a bit on the importance of this?

kamaci:
> Can you educate me a bit on the importance of this?

When InterruptedException is thrown it clears the interrupted state of the Thread. So, the fact that the thread was interrupted will be lost if you don't handle the exception properly.

You can find more detail here: https://www.ibm.com/developerworks/java/library/j-jtp05236/index.html?ca=drs-#2.1

Also can check conversations of my another PR related to this: https://github.com/apache/druid/pull/6943

kamaci:
Is there anything to be done at my PR?

PR #1253, opened 2020-05-08
Drop IOException from Endpoint constructor signature.

mstyura:
As long as `Endpoint` no longer generate `IOException` it could be removed, what do you think?


mstyura:
@bbaldino could you please have a look? Thanks a lot in advance!

PR #1255, opened 2020-05-11
Migrating ice/Harvesters and ice/TransportUtils to kotlin

jespino:
This is a small PR to migrate things to kotlin, I'm not an expert on Jitsi, and
I'm not sure what is the right code styling, I followed the `ktlint` linter
default rules.

I used a `object` (singleton) to represent the
`all-methods-all-attributes-static` classes that I have migrated.

for `TransportUtils` probably makes sense to migrate it to a simple function,
instead of a singleton with a single method.


bbaldino:
Thanks, will try and find some to take a look at this.

jespino:
@bbaldino thanks, I'm working in parallel to add some unit tests to that part in an independent PR. I'm struggling now to change the IceConfig.Config data on the fly to test different branches in the code, but I think I'll figure it out.

bbaldino:
> @bbaldino thanks, I'm working in parallel to add some unit tests to that part in an independent PR. 

That'd be great, thanks.

> I'm struggling now to change the IceConfig.Config data on the fly to test different branches in the code, but I think I'll figure it out.

Yes this is, unfortunately, a very annoying situation and it's my least-favorite part of the new config.  There are some examples of how to do this.  I'd look at [WebsocketServiceConfigTest.kt](https://github.com/jitsi/jitsi-videobridge/blob/master/src/test/kotlin/org/jitsi/videobridge/websocket/config/WebsocketServiceConfigTest.kt).

jespino:
Great, thanks for the pointer, I'll try to add to follow that approach and see if that works for me.

jespino:

PR #1258, opened 2020-05-13
Ice TransportUtils and Harvesters tests

jespino:
This tests are not working, this is the last approach that I have used. I used
things like `System.setProperty` to set new and legacy properties and then
`JitsiConfig.reload()`, it didn't work, I tried other approaches using mocks,
but nothing has worked so far, any advice?


jespino:
@bbaldino can you give me some pointer from here? I have tried multiple things. Trying to reset the properties, executing JitsiConfig.reload(), but all this weird mess of config handling is beating. Any idea of how to solve this?

jespino:

PR #1259, opened 2020-05-14
Migrate rest stats to kotlin

jespino:
I added this migration because there were already unit tests for this, so is
more easy to migrate without breaking anything.


jespino:

PR #1261, opened 2020-05-16
WIP: Use usrsctp only from dedicated single thread.

mstyura:
This PR moves calling usrsctp API into dedicated single thread to reduce chances of multi-threading issues.
When `usrsctp` is itself updated to the version which has support of disabling internal `SCTP Timer` thread then `usrsctp` will not be used concurrently at all.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/sctp/SctpManager.java
+++ b/src/main/java/org/jitsi/videobridge/sctp/SctpManager.java
@@ -22,6 +22,8 @@ import org.jitsi.utils.logging2.*;
 import org.jitsi.videobridge.util.*;
 import org.jitsi_modified.sctp4j.*;
 
+import java.util.concurrent.*;
+
 /**
  * Manages the SCTP connection and handles incoming and outgoing SCTP packets.
  *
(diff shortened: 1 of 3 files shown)


bbaldino:
Did you run into an issue with this?

src/main/java/org/jitsi/videobridge/sctp/SctpManager.java
```diff
@@ -144,7 +164,13 @@ public int send(byte[] data, int offset, int length)
         {
             byte[] newBuf = ByteBufferPool.getBuffer(length);
             System.arraycopy(data, offset, newBuf, 0, length);
-            return innerSctpDataSender.send(newBuf, 0, length);
+
+            // Leave SCTP thread ASAP
+            TaskPools.IO_POOL.submit(() -> 
+            {
+                innerSctpDataSender.send(newBuf, 0, length);
```
bertm, line 171: Delegating to just any IO thread breaks outgoing SCTP packet order when the machine is under load.

bgrozev: The underlying transport (udp) doesn't guarantee order either

bertm: You're right. Thanks for pointing out that we're not dealing with user message here, but rather with the underlying protocol. I missed that.

mstyura:
@bbaldino ah, sorry, I've missed your comment. What I was trying to fix is to move `usrsctp` usage into a single thread to avoid multi threading bugs in `usrsctp`.

bgrozev:
usrsctp removed

PR #1264, opened 2020-05-18
debian: Ignores commented lines searching for apis. Fixes #1254.

### LINKED ISSUE #1254: debian postinst gets multiline value for $APIS_VALUE
This line (120):
`APIS_VALUE=$(grep -oPe '--apis=\K(\w*,*\w*)' $CONFIG || true)`
needs a next line to avoid installation failure (I've tested with upgrading):
`APIS_VALUE="$(echo $APIS_VALUE | sed -e 's/ //g')"`

damencho:


obfusk:
I just proposed #1273, which also skips comments in the sed commands.

PR #1266, opened 2020-05-19
Remove the XMPP component mode.'

bgrozev:


src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -192,7 +161,7 @@ protected void performCheck()
             throw new Exception("Failed to bind single-port");
         }
 
-        checkXmppConnection(videobridge);
+        // TODO: check if ClientConnectionImpl is configured and connected.
```
bbaldino, line 164: Do we need to address this before merging?

src/main/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfig.kt
```diff
@@ -147,22 +143,6 @@ class StatsManagerBundleActivatorConfig {
              * Because of that, we define these here to handle the above use cases, but
              * implement the property that's used differently (see above)
              */
-            @Suppress("unused")
```
bbaldino, line 150: Now that there is only one remaining property here, we can use a 'normal' property (instead of these three unused ones and the 'meta' property above).  We can save that for another PR though.

src/main/java/org/jitsi/videobridge/health/Health.java
bgrozev: No, the behavior doesn't change. We were only checking the component connection anyway.

PR #1268, opened 2020-05-22
JVB http (ws) api v1

bbaldino:
This is what looks like will be v1 of the JVB http (websocket, in this case) API.  It is implemented as 3 artifacts (independent of the JVB artifact) living in the JVB repo:

1. `jvb-api-common`: common code used by both `jvb-api-client` and` jvb-api-server`.
1. `jvb-api-client`: the api "client" code.  client here refers to the entity controlling the jvb.
1. `jvb-api-server`: the api "server" code.  this is what jvb itself will integrate with to handle requests.

v1 of this API defines 2 methods:
1. `sendIqAndGetReply(IQ): IQ` which synchronously sends an IQ and returns the reply
1. `sendAndForget(IQ)`: which sends an IQ but does not wait for the response

Basically, v1 does nothing but change the transport over which existing COLIBRI messages are sent.  This implementation uses websockets in order to preserve the ordering of the requests (which COLIBRI requires).

This PR contains only the API code itself (jvb changes to integrate it are not included here) because I thought that would be cleaner.  Don't bother looking commit-by-commit, as it was a g and winding path to this point.


jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/SynchronousWebSocketClient.kt
```diff
@@ -0,0 +1,130 @@
...
+            client.ws(host = host, port = port, path = path) {
+                try {
+                    for (msg in msgsToSend) {
+                        send(msg)
+                        val resp = incoming.receive()
+                        msgsReceived.send(resp)
+                    }
+                } finally {
+                    logger.info("Websocket closing")
+                    close()
```
JonathanLennox, line 116: What does this do to any pending coroutines?

jvb-api/jvb-api-client/pom.xml
```diff
@@ -0,0 +1,179 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.jitsi</groupId>
+        <artifactId>jvb-api-parent</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jvb-api-client</artifactId>
```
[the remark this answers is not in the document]
bbaldino: For now this will update every time JVB does, the other thing would be nice to have but I don't think we should block on it.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/client/v1/JvbApi.kt
```diff
@@ -0,0 +1,75 @@
...
+    fun sendAndForget(iq: IQ) {
+        wsClient.sendIqAndForget(iq)
+    }
+}
+
+/**
+ * Send [IQ]s and get their replies via a [SynchronousWebSocketClient].
+ */
+private fun SynchronousWebSocketClient.sendIqAndGetReply(iq: IQ): IQ {
+    val resp = sendAndGetReply(SmackXmlSerDes.serialize(iq))
```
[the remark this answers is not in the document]
bbaldino: This doesn't timeout, as, from what I can tell, the XMPP version didn't.  Do you mean if the connection closes?  That would be a good thing to add a test for, I'll look at that.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/SynchronousWebSocketClient.kt
bbaldino: This should clean things up correctly, but I think it'd be good to have a test to verify that...I'll add one.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/client/v1/JvbApi.kt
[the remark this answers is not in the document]
bbaldino: So it's a bit awkward to add a unit tests for this since there's a blocking call involved (there will be races between the blocking thread and the one calling cancel).  But, I did write one for testing and cleaned this up a bit by throwing an exception if things go south while waiting for the response.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/SynchronousWebSocketClient.kt
bbaldino: I think the close here was actually redundant, I added some better handling here.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/client/v1/JvbApi.kt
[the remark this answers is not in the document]
bgrozev: The XMPP implementation does timeout and return `null` when it does:
https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/impl/protocol/xmpp/XmppProtocolProvider.java#L644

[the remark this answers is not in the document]
bbaldino: Thanks for the pointer, I'll add a timeout.

bbaldino:
@bgrozev caught that adding of the timeout made it such that the simple request/response correlation scheme would no longer work, so I made some changes to fix that:

1. There's now a simple `WebSocketClient` which can send messages and invokes a handler when a message is received
2. There's now a `SynchronousXmppWsClient` which wraps a `WebSocketClient` and implements request/response correlation based on stanza ID.

jvb-api/jvb-api-common/src/test/kotlin/org/jitsi/videobridge/api/types/SupportedApiVersionsTest.kt
```diff
@@ -0,0 +1,60 @@
...
+            }
+            context("where there is no intersection") {
+                val s2 = SupportedApiVersions()
+                should("return null") {
+                    s1.maxSupported(s2) shouldBe null
+                }
+            }
+        }
+    }
+}
```
JonathanLennox, line 60: Are these files getting ktlint'd? I'd have thought this would give "no newline at end of file".

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
```diff
@@ -0,0 +1,89 @@
...
+        iq.stanzaId = "$id"
+        val response = CompletableFuture<Stanza>()
+        responseHandlers["$id"] = response
+        wsClient.sendString(SmackXmlSerDes.serialize(iq))
+
+        return try {
+            response.get(requestTimeout.toMillis(), TimeUnit.MILLISECONDS)
+        } catch (t: Throwable) {
+            responseHandlers.remove("$id")
+            // TODO: handle the specific exception types?
```
JonathanLennox, line 81: Do you want any kind of logging here?

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
```diff
@@ -0,0 +1,89 @@
...
+            null
+        }
+    }
+
+    fun sendIqAndForget(iq: IQ) = wsClient.sendString(SmackXmlSerDes.serialize(iq))
+
+    fun run() = wsClient.run()
+}
```
JonathanLennox, line 89: You don't seem to have a `stop()` at this level.  Should there be one?  I imagine it should stop the websocket client, and cancel all the pending futures?

jvb-api/jvb-api-common/src/test/kotlin/org/jitsi/videobridge/api/types/SupportedApiVersionsTest.kt
bbaldino: I think not, thanks for the reminder.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
bbaldino: I think this comment made more sense in the old style, where it could've been a timeout or a coroutine job cancellation, but I don't know that we need logging: this can only return null in the case of timeout.  I did consider not catching the exception here and letting it bubble up, but I dunno if that's really all that valuable?

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
bbaldino: This probably does make sense at this level for completeness, but I'm not sure we'd end up calling it from `JvbApi`, as that seems like a leaky abstraction.  I think instead we'd just rely on it being destroyed when the user was done and that cleaning everything up.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
```diff
@@ -35,15 +35,41 @@ import java.util.concurrent.atomic.AtomicInteger
  * 'sending-and-forgetting': sending an IQ but not waiting for its response.
  */
 class XmppWebSocketClient(
+    /**
+     * The underlying [HttpClient] to use.  It must supports websockets.
```
JonathanLennox, line 39: "Must support"

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
```diff
@@ -121,4 +121,9 @@ class XmppWebSocketClient(
      * Connect the websocket client
      */
     fun run() = wsClient.run()
+
+    /**
+     * Disconnect the websocket client
+     */
+    fun stop() = wsClient.stop()
```
JonathanLennox, line 128: Does this need to cancel the futures, also?  Otherwise won't `sendIqAndGetReply` block forever?

bbaldino: There are timeouts there, so they won't block forever.  Also, I'm pretty sure I thought about this exact scenario last week and had a reason why I thought this would be sufficient, but it escapes me now, so I could go either way.  We could cancel the futures here and also set a sentinel to make sure any new requests returned immediately.

bbaldino: I looked at doing this...I'd have to add some locking to make sure no new futures were added to the map when we were shutting down in addition to the sentinel (to make sure everything got cancelled), so I thought maybe it's not worth it...but let me know if you have thoughts.

jvb-api/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.jitsi</groupId>
+    <artifactId>jvb-api-parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <name>jvb-api-parent</name>
```
JonathanLennox, line 25: I take it this pom is entirely separate from the jvb pom, as far as maven is concerned?  What will we need to change for the PR testing scripts, the CI, and the like?

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
JonathanLennox: A timeout on `stop` is not great but is probably okay, if you've tested that it works.

jvb-api/pom.xml
bbaldino: Arg, we don't have a build script in the jvb repo for jenkins to use.  Maybe I'll do a PR for that first.

jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/XmppWebSocketClient.kt
JonathanLennox: The fancier stop seems like it'd be a good idea, but maybe not essential if the timeout works...I don't imagine our normal JVB/Jicofo usecases will ever want to stop the JVB API code?

bbaldino: I guess we'd stop the client when we ended a session with the bridge (because of how Jicofo is set up, we'll have a jvb api client per bridge per conference), but I don't think the timeout behavior feels like a big deal there.  I did add a test for stop and verified pending calls timeout.

JonathanLennox: As long as a race condition doesn't cause jicofo in general to go into a timeout-block that blocks other transactions, I think this is fine.

bbaldino: `stop` itself blocks on cancelling the job the websocket client is in, and anything waiting for a response is blocked on a future we created, so there shouldn't be any deadlock scenario there.

jvb-api/jvb-api-server/src/main/kotlin/org/jitsi/videobridge/api/server/v1/Colibri.kt
```diff
@@ -0,0 +1,81 @@
...
+ */
+private val WebSocketSession.incomingStanzas: ReceiveChannel<Stanza>
+    get() = produce {
+        for (msg in incoming) {
+            when (msg) {
+                is Frame.Text -> {
+                    val stanza = SmackXmlSerDes.deserialize(msg.readText())
+                    send(stanza)
+                }
+                else -> throw IllegalStateException("Expected only text frames")
```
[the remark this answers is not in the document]
bbaldino: It will...this would basically be a programming bug so the idea was to fail fast so it could be caught and fixed quickly.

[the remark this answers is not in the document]
JonathanLennox: Well, this could be a programming error but it could also be an attack (someone connects to your ws port and sends a binary data frame).  If it just kills that ws connection that's ideal, but if it causes anything else to go wrong that's a problem.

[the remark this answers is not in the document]
bbaldino: It won't affect the other websocket sessions.  I wrote a test to verify this, but I'm not sure it makes sense to actually commit.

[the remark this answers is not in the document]
bbaldino: I did tweak this line a bit which made the test nicer--I think it won't matter in a real session but pushed it anyway.

bbaldino:
Had to rebase to pick up the build script changes.

jvb-api/jvb-api-client/src/test/kotlin/org/jitsi/videobridge/api/util/WebSocketClientTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlin.random.Random
+import kotlin.time.ExperimentalTime
+import kotlin.time.seconds
+
+@ExperimentalTime
+class WebSocketClientTest : ShouldSpec() {
+    override fun isolationMode(): IsolationMode? = IsolationMode.InstancePerLeaf
+
+    private val wsPort = Random.nextInt(1024, 65535).also {
+        println("Server running on port $it")
```
JonathanLennox, line 43: I'm worried we'll get spurious test failures if this picks a port that number happens to be in use.

It looks like ktor doesn't support the "bind with port 0 and then read back the port it bound on" idiom (https://github.com/ktorio/ktor/issues/909) so maybe catch the "Address already in Use" exception and re-try binding?

PR #1275, opened 2020-05-26
remove residual use of options obsoleted by #1266

### LINKED ISSUE #1237: Don't expose password in process listing

obfusk:
Closes #1237 (thanks to #1266).

- [x] CLA signed :)

### DIFF
--- a/resources/mvn-exec.sh
+++ b/resources/mvn-exec.sh
@@ -8,4 +8,4 @@ fi
 
 . $CONFIG
 
-exec mvn compile exec:exec -Dexec.executable=java -Dexec.args="-cp %classpath ${JAVA_SYS_PROPS} org.jitsi.videobridge.Main --opts=${JVB_OPTS}"
+exec mvn compile exec:exec -Dexec.executable=java -Dexec.args="-cp %classpath ${JAVA_SYS_PROPS} org.jitsi.videobridge.Main ${JVB_OPTS}"
--- a/resources/mvn-exec.sh
+++ b/resources/mvn-exec.sh
@@ -8,4 +8,4 @@ fi
 
 . $CONFIG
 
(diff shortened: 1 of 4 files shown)


resources/mvn-exec.sh
```diff
@@ -8,4 +8,4 @@ fi
 
 . $CONFIG
 
-exec mvn compile exec:exec -Dexec.executable=java -Dexec.args="-cp %classpath ${JAVA_SYS_PROPS} org.jitsi.videobridge.Main --domain=\"${JVB_HOSTNAME}\" --host=\"${JVB_HOST}\" --port=\"${JVB_PORT}\" --secret=\"${JVB_SECRET}\" --opts=${JVB_OPTS}"
+exec mvn compile exec:exec -Dexec.executable=java -Dexec.args="-cp %classpath ${JAVA_SYS_PROPS} org.jitsi.videobridge.Main --opts=${JVB_OPTS}"
```
obfusk, line 11: Should `--opts=${JVB_OPTS}` just be `${JVB_OPTS}` or am I missing something? I don't see a `--opts` defined anywhere.

bgrozev: Yes, I think so, good catch

obfusk: I added a commit to fix that.

bgrozev:
Thank you for the contribution!

PR #1279, opened 2020-05-28
debian: Remove obsolete configuration (fixes #1277).

### LINKED ISSUE #1277: debian/postinst still generates config files w/ JVB_HOST etc.
#1266 removed `--host`, `--secret` etc. but the `postinst` still generates config files with those options.

bgrozev:


debian/postinst
```diff
@@ -31,30 +31,12 @@ case "$1" in
 
         # generate config on new install or when we are reconfiguring
         # and all install is different than current one
-        if [ "$OLDCONFIG" = "false" ] || [ "$JVB_HOSTNAME" != "$JVB_HOSTNAME_IN" ]; then
```
damencho, line 34: This check is being dropped which was taking care of dpkg-reconfigure we are checking whether the hostname has changed. With the new check we will not reconfigure jvb I think, it will only work on new installs.
There was also the logic where we only create the secret on clean install ... I'm not sure whether this should be dropped or not ... as if you are changing the hostname you need to create the account in prosody with the new hostname ... so the new password for that is fine, you cannot reuse the old account anyway...

bgrozev: I'm confused :)

Do we use the JVB_HOSTNAME value stored in /etc/jitsi/videobridge/config to detect when the hostname changes?

damencho: Yes.

We get what is stored in config: https://github.com/jitsi/jitsi-videobridge/blob/0d6fb601878cbc735faa7261b9cc06195c842e41/debian/postinst#L19

We ask the user for the new value: https://github.com/jitsi/jitsi-videobridge/blob/0d6fb601878cbc735faa7261b9cc06195c842e41/debian/postinst#L30

And we compare it: `"$JVB_HOSTNAME" != "$JVB_HOSTNAME_IN"`

PR #1281, opened 2020-05-28
Fix warnings and lgtm.com alerts.

bgrozev:


src/main/java/org/jitsi/videobridge/stats/BundleContextHolder2.java
```diff
@@ -92,21 +94,22 @@ void start(BundleContext bundleContext)
...
                 oldValue = this.bundleContext;
                 this.bundleContext = null;
-                newValue = this.bundleContext;
             }
         }
-        if (oldValue != newValue)
-            bundleContextChanged(oldValue, newValue);
+        if (oldValue != null)
+        {
+            bundleContextChanged(oldValue, null);
```
bbaldino, line 112: I think this should be `bundleContextChanged(oldValue, this.bundleContext);` ? Also, if `this.bundleContext` _was_ null and it was changed to something non-null, we wouldn't fire the event.

bgrozev: I wasn't sure what the original intention was, but I don't think I changed the semantics. `newValue` was always `null`, and I just replaced it with an explicit `null. 

If `this.bundleContext` was null, it wouldn't be replaced because of the `==` check on line 102/104. I think the idea is to only act for the context we're started in (if any), and then set it to `null` (since this is `stop()`).

bbaldino: > I wasn't sure what the original intention was, but I don't think I changed the semantics. newValue was always null, and I just replaced it with an explicit `null.

Oh, wow, now I see it.  The old code here is mind-boggling...

> If this.bundleContext was null, it wouldn't be replaced because of the == check on line 102/104. I think the idea is to only act for the context we're started in (if any), and then set it to null (since this is stop()).

Ok, I think you're right.

PR #1288, opened 2020-06-01
Some small cleanup - mainly removing Localpart from Conference.java

bbaldino:
Pass a`String` instead of a `Localpart` for the conference name.  This is something I came across when doing JVB API investigation, since we won't have a `Localpart` in the future.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -430,7 +430,14 @@ public class Conference
         iq.setID(getID());
         try
         {
-            iq.setName(Localpart.from(getName()));
+            if (conferenceName == null)
+            {
+                iq.setName(null);
+            }
+            else
+            {
+                iq.setName(Localpart.from(conferenceName));
+            }
         } catch (XmppStringprepException e)
         {
             logger.error("Error converting conference name to a localpart ", e);
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -1322,7 +1322,7 @@ public class Conference
     {
         JSONObject debugState = new JSONObject();
         debugState.put("id", id);
-        debugState.put("name", conferenceName == null ? null : conferenceName);
+        debugState.put("name", conferenceName);
 
         if (full)
         {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -427,7 +428,21 @@ else if (logger.isDebugEnabled())
...
-        iq.setName(getName());
+        try
+        {
+            if (conferenceName == null)
+            {
+                iq.setName(null);
+            }
+            else
+            {
+                iq.setName(Localpart.from(conferenceName));
```
JonathanLennox, line 439: Conference names currently always come from localparts, so in practice this shouldn't contain any forbidden characters (throwing XmppStringprepException), right?

bbaldino: Yeah, the idea is that if they need to be put back into a Localpart, they must have come from a Localpart.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1307,7 +1322,7 @@ public JSONObject getDebugState(boolean full, String endpointId)
     {
         JSONObject debugState = new JSONObject();
         debugState.put("id", id);
-        debugState.put("name", name == null ? null : name.toString());
+        debugState.put("name", conferenceName == null ? null : conferenceName);
```
bgrozev, line 1325: That's equivalent to just `conferenceName`. Did you want the string "null"?

bbaldino: Ha, oops.  I just wanted to recreate the old logic, which looks like would be just using `conferenceName`

bbaldino: Fixed, thanks.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -219,7 +219,7 @@ public Videobridge()
      */
     public @NotNull Conference createConference(Jid focus, Localpart name, String gid)
     {
-        return this.createConference(focus, name, /* enableLogging */ true, gid);
+        return this.createConference(focus, name == null ? null : name.toString(), /* enableLogging */ true, gid);
```
bgrozev, line 222: Nit:  I think we can push this back one more level (`VideobridgeShim`) and simplify to `Objects.toString(name)`

PR #1294, opened 2020-06-04
make the version strings used consistent across all poms

bbaldino:
I think this must have not gotten caught due to some cached versions (from before the pom hierarchy change) as part of https://github.com/jitsi/jitsi-videobridge/pull/1268


bbaldino:
Merging because the failures here are flaky tests, I'll be working on those next.

PR #1295, opened 2020-06-04
don't use mvn dependency:resolve to cache deps

bbaldino:
mvn dependency:resolve expects all packages to be in a remote repo, but
with the jvb api we have the packages locally.  since this call was only
made to make sure everything was downloaded, use mvn package instead
which doesn't expect everything to be remote.


bbaldino:
I think there must be something later which is re-calling `mvn package` (which would include the effect of the `mvn versions:set` call, but I still moved `mvn versions:set` to _after_ the new `mvn package` so we get the version update there, just in case.

bbaldino:
The flaky gh test is one I'm in the process of fixing, so ignoring that here.

PR #1297, opened 2020-06-04
refactor the xmpp ws client test to not rely on a local server

bbaldino:
Since `WebSocketClient` is doing all the real work, we can change `XmppWebSocketClient` to allow taking in a `WebSocketClient` such that it can be mocked and make this test a lot simpler and deterministic.


jvb-api/jvb-api-client/src/main/kotlin/org/jitsi/videobridge/api/util/WebSocketClient.kt
```diff
@@ -47,8 +47,8 @@ class WebSocketClient(
      * The path of the remote websocket URL
      */
     private val path: String,
-    private val incomingMessageHandler: (Frame) -> Unit = {},
     parentLogger: Logger,
+    var incomingMessageHandler: (Frame) -> Unit = {},
```
JonathanLennox, line 51: Having this be a `var` seems a little odd to me, semantically - is the idea that this can only be changed "early" in the initialization process?  Is there any way to constrain this?

bbaldino: Yeah, this isn't ideal.  The reason is that `XmppWebSocketClient` needs to set this handler to one of its own methods, so if we want to be able to pass in an already-created `WebSocketClient`, this needs to be able to be changed after `WebSocketClient`s creation.  I considered passing a `WebSocketClient` "creator" function instead (i.e. `(HttpClient, String, Int, String, Logger, (Frame) -> Unit, CoroutineDispatcher) -> WebSocketClient` but the argument list there was long enough that that also felt awkward.

The only way I can think to constrain the value is to have another field, defined like:

```
val actualIncomingMessageHandler: (Frame) -> Unit by lazy { incomingMessageHandler }
```
Which would mean: the message handler could be changed up until the first time it was invoked, but after that no changes would be reflected.  This feels worse, though, as at least the current scheme is consistent with the idea that the handler can be changed and those changes would be reflected.

JonathanLennox: Yeah, there's nothing wrong conceptually with changing the message handler, so long as when it takes effect is atomic and well-defined.  If you change it from within the handling of a message, is the next message handled with the new one?

bbaldino: I don't know what you mean by 'atomic and well-defined' beyond the variable assignment, but to me this seems fine as a "user-beware" situation.

JonathanLennox: I mostly mean does the client not check the value of the handler for the next message until after the previous message has finished processing - so a handler can reset itself.  If that's the case it's probably fine.

bbaldino: well `WebSocketClient` has no sense of requests and replies, it just sends messages out and notifies the handler when there is a message is received--that message may not even be a response to something, just a message from the other side.  When a message is received, we invoke whatever the value of `incomingMessageHandler` is at that time

JonathanLennox: Right - my question is whether the message processing is single-threaded, or if it might start processing the next message (picking up the old `incomingMessageHandler`) before the previous message finished being handled.

I'm thinking of a model where the contents of one message changes the way subsequent messages will be processed, so the handler for the first message installs a new message handler.

PR #1300, opened 2020-06-05
Rest api jvb integration

bbaldino:
This implements v1 of the new JVB API


jvb_pom.xml
```diff
@@ -114,6 +114,11 @@
       <artifactId>sctp</artifactId>
       <version>1.0-20190319.172750-1</version>
     </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>jvb-api-server</artifactId>
+      <version>2.1-227-g3ce5bf97</version>
```
JonathanLennox, line 120: Should this be ${project.version}?

src/main/kotlin/org/jitsi/videobridge/signaling/api/SignalingApiConfig.kt
```diff
@@ -0,0 +1,68 @@
...
+        fun bindPort(): Int = bindPortProperty.value
+
+        class BindAddressProperty : SimpleProperty<String>(
+            newConfigAttributes {
+                readOnce()
+                name("videobridge.apis.jvb-api.bind-address")
+            }
+        )
+
+        private val bindAddressProperty = BindAddressProperty()
```
JonathanLennox, line 53: Do you have any desire to validate this as being a valid IP address, or do you just want to let whatever exception gets thrown at initialization handle it?

jvb_pom.xml
bbaldino: Yup, thanks

src/main/kotlin/org/jitsi/videobridge/signaling/api/SignalingApiConfig.kt
bbaldino: I think elsewhere we just let it fail further down which seems fine to me.

PR #1301, opened 2020-06-05
Add a stat for endpoints with ICE but no DTLS. Cleanup.

bgrozev:


src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -416,37 +416,17 @@ public void consentUpdated(@NotNull Instant time) {
 
     private void setupDtlsTransport()
     {
-        dtlsTransport.incomingDataHandler = new DtlsTransport.IncomingDataHandler()
+        dtlsTransport.incomingDataHandler = this::dtlsAppPacketReceived;
```
bbaldino, line 419: I didn't use method references here because I think it's misleading of intellij to suggest using them for interfaces that happen to have only a single method (I think it's clearer to make it obvious this is implementing an interface, not just a single method)--but I'm not necessarily against changing it, that's just my preference/thought on it.

JonathanLennox:
I'd have put the cleanup in a separate PR from the new stat, but I guess it's okay.

PR #1305, opened 2020-06-10
fix: The on-stage participants becomes suspended when last-n equals 1.

gpolitis:
@paweldomas discovered this issue where the prioritization order that the EndpointMultiRanker implements is inversed, leading to the wrong endpoints being allocated bandwidth first. The comments in the patch provide further information.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -332,19 +332,27 @@ public class BitrateController
         @Override
         public int compare(EndpointMultiRank o1, EndpointMultiRank o2)
         {
-            int preferredHeightDiff = o1.videoConstraints.getPreferredHeight() - o2.videoConstraints.getPreferredHeight();
+            // We want "o1 has higher preferred height than o2" to imply "o1 is
+            // smaller than o2" as this is equivalent to "o1 needs to be
+            // prioritized first".
+            int preferredHeightDiff = o2.videoConstraints.getPreferredHeight() - o1.videoConstraints.getPreferredHeight();
             if (preferredHeightDiff != 0)
             {
                 return preferredHeightDiff;
             }
             else
             {
-                int idealHeightDiff = o1.videoConstraints.getIdealHeight() - o2.videoConstraints.getIdealHeight();
+                // We want "o1 has higher ideal height than o2" to imply "o1 is
+                // smaller than o2" as this is equivalent to "o1 needs to be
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -332,19 +332,27 @@ private static boolean changeIsLargerThanThreshold(
         @Override
         public int compare(EndpointMultiRank o1, EndpointMultiRank o2)
         {
-            int preferredHeightDiff = o1.videoConstraints.getPreferredHeight() - o2.videoConstraints.getPreferredHeight();
+            // We want "o1 has higher preferred height than o2" to imply "o1 is
```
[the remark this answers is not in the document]
gpolitis: @bbaldino could you please take another look?

src/test/java/org/jitsi/videobridge/cc/EndpointMultiRankerTest.java
```diff
@@ -0,0 +1,88 @@
...
+    final VideoConstraints stageVideoConstraints
+        = new VideoConstraints(/* idealHeight */ 720);
+    final VideoConstraints thumbnailVideoConstraints
+        = new VideoConstraints(/* idealHeight */ 180);
+
+    @Test
+    public void followActiveSpeaker()
+    {
+        EndpointMultiRank activeSpeakerMultiRank
+            = makeNextSpeakerWithVideoConstraints(stageVideoConstraints);
```
bbaldino, line 38: nit: I think the naming of these variables (`activeSpeakerMultiRank`, `speaker2MultiRank`, `speaker3MultiRank`) are maybe a bit confusing, because (as the next test shows) this code prioritizes based on their ideal height, not whether or not they're the active speaker (which this code isn't even aware of).

gpolitis: active speaker rank is the tie breaker if both the ideal and preferred height is equal and the order in which the endpoints are added to the list determines their speaker rank. I should probably add this note in the code

gpolitis: @bbaldino I've improved the comment that explains the assertions. Let me know if it's better now

bbaldino: 🤦 I didn't realize speaker rank was part of that structure, sorry!  Thanks.

PR #1307, opened 2020-06-12
wip: Octo propagation of video constraints and off-stage participant …

gpolitis:
…layer suspension.

### DIFF
--- a/jvb_pom.xml
+++ b/jvb_pom.xml
@@ -64,12 +64,6 @@
       <artifactId>dom4j</artifactId>
       <version>1.6.1</version>
     </dependency>
-    <!-- temp dependency -->
-    <dependency>
-      <groupId>com.google.code.gson</groupId>
-      <artifactId>gson</artifactId>
-      <version>2.8.6</version>
-    </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
--- a/jvb_pom.xml
+++ b/jvb_pom.xml
@@ -64,6 +64,12 @@
       <artifactId>dom4j</artifactId>
       <version>1.6.1</version>
     </dependency>
+    <!-- temp dependency -->
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.8.6</version>
+    </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -355,7 +355,6 @@ public abstract class AbstractEndpoint
             .max(Comparator.comparingInt(VideoConstraints::getIdealHeight))
             .orElse(DEFAULT_MAX_RECEIVER_CONSTRAINTS);
 
-        // for intra-bridge traffic we only care about the ideal height.
         if (!newReceiverMaxVideoConstraints.equals(oldReceiverMaxVideoConstraints))
         {
             maxReceiverVideoConstraints = newReceiverMaxVideoConstraints;
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -87,8 +87,9 @@ public abstract class AbstractEndpoint
         DEFAULT_MAX_RECEIVER_CONSTRAINTS = THUMBNAIL_VIDEO_CONSTRAINTS;
 
     /**
-     * The max video constraints that the bridge should receive from this
-     * endpoint.
+     * The maximum set of constraints applied by all receivers of this endpoint
+     * in the conference. The client needs to send _at least_ this to satisfy
+     * all receivers.
      */
     private VideoConstraints maxReceiverVideoConstraints
         = new VideoConstraints(BitrateControllerConfig.Config.onstageIdealHeightPx());
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -84,7 +84,7 @@ public abstract class AbstractEndpoint
      * The default video constraints to assume when nothing is signaled.
      */
     private final VideoConstraints
-        defaultReceiverVideoConstraints = THUMBNAIL_VIDEO_CONSTRAINTS;
+        DEFAULT_MAX_RECEIVER_CONSTRAINTS = THUMBNAIL_VIDEO_CONSTRAINTS;
 
     /**
      * The max video constraints that the bridge should receive from this
@@ -352,7 +352,7 @@ public abstract class AbstractEndpoint
             .values()
             .stream()
             .max(Comparator.comparingInt(VideoConstraints::getIdealHeight))
-            .orElse(defaultReceiverVideoConstraints);
+            .orElse(DEFAULT_MAX_RECEIVER_CONSTRAINTS);
 
         // for intra-bridge traffic we only care about the ideal height.
         if (!newReceiverMaxVideoConstraints.equals(oldReceiverMaxVideoConstraints))
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -315,6 +315,8 @@ public abstract class AbstractEndpoint
     public JSONObject getDebugState()
     {
         JSONObject debugState = new JSONObject();
+        debugState.put("receiveVideoConstraints", videoConstraintsMap);
+        debugState.put("maxReceiveVideoConstraints", maxReceiverVideoConstraints);
         debugState.put("displayName", displayName);
         debugState.put("expired", expired);
         debugState.put("statsId", statsId);
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -315,8 +315,8 @@ public abstract class AbstractEndpoint
     public JSONObject getDebugState()
     {
         JSONObject debugState = new JSONObject();
-        debugState.put("receiveVideoConstraints", videoConstraintsMap);
-        debugState.put("maxReceiveVideoConstraints", maxReceiverVideoConstraints);
+        debugState.put("receiverVideoConstraints", receiverVideoConstraintsMap);
+        debugState.put("maxReceiverVideoConstraints", maxReceiverVideoConstraints);
         debugState.put("displayName", displayName);
         debugState.put("expired", expired);
         debugState.put("statsId", statsId);
@@ -404,7 +404,7 @@ public abstract class AbstractEndpoint
     /**
      * The map of receiver endpoint id -> video constraints.
      */
-    private final Map<String, VideoConstraints> videoConstraintsMap = new ConcurrentHashMap<>();
+    private final Map<String, VideoConstraints> receiverVideoConstraintsMap = new ConcurrentHashMap<>();
 
     /**
      * Notifies this instance that the specified endpoint wants to receive
@@ -417,11 +417,11 @@ public abstract class AbstractEndpoint
      */
     public void addReceiver(String endpointId, VideoConstraints newVideoConstraints)
     {
-        VideoConstraints oldVideoConstraints = videoConstraintsMap.put(endpointId, newVideoConstraints);
+        VideoConstraints oldVideoConstraints = receiverVideoConstraintsMap.put(endpointId, newVideoConstraints);
         if (oldVideoConstraints == null
             || !oldVideoConstraints.equals(newVideoConstraints))
         {
-            setReceiverVideoConstraints(ImmutableMap.copyOf(videoConstraintsMap));
+            setReceiverVideoConstraints(ImmutableMap.copyOf(receiverVideoConstraintsMap));
         }
     }
 
@@ -434,9 +434,9 @@ public abstract class AbstractEndpoint
      */
     public void removeReceiver(String endpointId)
     {
-        if (videoConstraintsMap.remove(endpointId) != null)
+        if (receiverVideoConstraintsMap.remove(endpointId) != null)
         {
-            setReceiverVideoConstraints(ImmutableMap.copyOf(videoConstraintsMap));
+            setReceiverVideoConstraints(ImmutableMap.copyOf(receiverVideoConstraintsMap));
         }
     }
 }
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -350,8 +350,7 @@ public abstract class AbstractEndpoint
             .orElse(defaultVideoConstraints);
 
         // for intra-bridge traffic we only care about the ideal height.
-        if (newReceiverMaxVideoConstraints.getIdealHeight()
-            != oldReceiverMaxVideoConstraints.getIdealHeight())
+        if (!newReceiverMaxVideoConstraints.equals(oldReceiverMaxVideoConstraints))
         {
             maxReceiverVideoConstraints = newReceiverMaxVideoConstraints;
             onMaxReceiverVideoConstraintsChanged(newReceiverMaxVideoConstraints);
--- a/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -359,7 +359,7 @@ public abstract class AbstractEndpoint
         if (!newReceiverMaxVideoConstraints.equals(oldReceiverMaxVideoConstraints))
         {
             maxReceiverVideoConstraints = newReceiverMaxVideoConstraints;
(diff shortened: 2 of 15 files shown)


src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -74,6 +77,22 @@
      */
     private boolean expired = false;
 
+    /**
+     * The instance responsible for bookkeeping of who (which receiver) is
```
bgrozev, line 81: When reading the java doc, I would like to find out how this relates to this AbstractEndpoint

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -325,6 +325,46 @@ public JSONObject getDebugState()
...
+            = this.maxReceiverVideoConstraints;
+
+        VideoConstraints defaultVideoConstraints
+            = new VideoConstraints(BitrateControllerConfig.Config.thumbnailMaxHeightPx());
+
+        VideoConstraints newReceiverMaxVideoConstraints = newVideoConstraints
+            .values()
+            .stream()
+            .max(Comparator.comparingInt(VideoConstraints::getIdealHeight))
+            .orElse(defaultVideoConstraints);
```
bgrozev, line 350: It looks like we only care about `idealHeight`, but the event is fired any time the `VideoConstrant` instances don't match. Maybe fire the event if `newReceivedMaxVideoConstaints.getIdealHeight() != oldReceiverMaxVideoConstraints.getIdealHeight()`?

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -417,9 +421,12 @@ protected void onSelectedEndpointsChangedEvent(
...
     }
 
-    protected void onVideoConstraintsChangedEvent(Map<String, VideoConstraints> newVideoConstraints)
+    protected void onVideoConstraintsChangedEvent(Map<String, VideoConstraints> videoConstraintMap)
     {
-        endpoint.setVideoConstraints(newVideoConstraints);
+        // Don't "pollute" the video constraints map with constraints for this
+        // endpoint.
+        videoConstraintMap.remove(endpoint.getID());
+        endpoint.setSenderVideoConstraints(ImmutableMap.copyOf(videoConstraintMap));
```
bgrozev, line 429: We modify the map on the line above, so this code must own the map. If that's the case, why is the copy necessary?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -357,5 +397,69 @@ public JSONObject getDebugState()
...
+
+    /**
+     * The class that is responsible for bookkeeping of who (which receiver) is
+     * viewing what (video constraints). Internally it maintains a map of
+     * receiver endpoint id -> video constraints. When the map changes, it sets
+     * the new map by calling {@link #setReceiverVideoConstraints(ImmutableMap)}.
+     *
+     * The main benefit of having this class (as opposed to spilling the
+     * implementation in AbstractEndpoint) is to keep the endpoint video
+     * constraints API symmetric (i.e. provide setter for send/recv video
```
bgrozev, line 421: I am confused. Why would having `addReceiver` and `removeReceiver` in `AbstractEndpoint` make the api non-symmetric? Half the "API" is private anyway. I think having that logic in a separate class is nice, but having to use `getReceiverVideoConstraintsBroker()` is unnecessary detail. I was first going to suggest to add `addReceiver` and `removeReceiver` to `AbstractEndpoint`, calling the broker internally, but since the amount of code is not large I think it's best put it all in `AbstractEndpoint`

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -638,9 +633,56 @@ public void addRtpExtension(RtpExtension rtpExtension)
...
+            oldVideoConstraints = bitrateController.getVideoConstraints();
+
         bitrateController.setVideoConstraints(newVideoConstraints);
+
+        Set<String> endpointsThatNoLongerCare
+            = new HashSet<>(oldVideoConstraints.keySet());
+        endpointsThatNoLongerCare.removeAll(newVideoConstraints.keySet());
+
+        // Endpoints that no longer care what they receive.
+        for (String id : endpointsThatNoLongerCare)
```
bgrozev, line 648: I am probably misunderstanding something, but if these are endpoints that no longer have constraints, that means we may want to send them high quality layers, right? Why are we doing `removeReceiver`?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1530,7 +1535,6 @@ public JSONObject getDebugState()
     {
         JSONObject debugState = super.getDebugState();
 
-        debugState.put("selectedCount", selectedCount.get());
```
bgrozev, line 1533: Add all the constraints to the debug state of `AbstractEndpoint` instead?

src/main/kotlin/org/jitsi/videobridge/octo/OctoEndpoint.kt
```diff
@@ -95,6 +97,17 @@ class OctoEndpoint(
         return transceiver.mediaSources
     }
 
+    override fun onMaxReceiverVideoConstraintsChanged(maxVideoConstraints: VideoConstraints?) {
+        // TODO send to the specific bridge that the sender connects
+        val octoEndpointId = "[bridgeId]$id"
```
bgrozev, line 102: The semantics are that this bridge wants to receive `maxVideoConstraints` for `localEndpoint`, right? Then this ID will be the same on all remote bridges, and they will replace each others' constraints.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
gpolitis: > We modify the map on the line above, so this code must own the map.

That is correct.

> If that's the case, why is the copy necessary?

Not necessary strictly speaking, I just wanted to express in the API that the map is immutable and that requires a new object. We could go with normal maps if you'd prefer that.

src/main/java/org/jitsi/videobridge/Endpoint.java
gpolitis: > I am probably misunderstanding something, but if these are endpoints that no longer have constraints, that means we may want to send them high quality layers, right?

Currently endpoints that haven't signalled any video constraints will get 180p which is the default resolution that we've used since forever.

src/main/kotlin/org/jitsi/videobridge/octo/OctoEndpoint.kt
gpolitis: > The semantics are that this bridge wants to receive maxVideoConstraints for localEndpoint, right?

Yes!

> Then this ID will be the same on all remote bridges, and they will replace each others' constraints.

This is just a placeholder of the actual `bridgeId` of the local bridge, i just needed to find where to get the local bridge id.

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: Yes, and I think this emphasizes how awkward the `preferredHeight` and `preferredFps` are in the VideoConstraints, which takes us back to @bbaldino suggestion to have a separate object for bandwidth allocation preferences/policy.

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -354,8 +386,57 @@ public JSONObject getDebugState()
...
+    /**
+     * Notifies this instance that the max video constraints that the bridge
+     * needs to receive from this endpoint has changed. Each implementation
+     * handles this notification differently.
+     *
+     * @param maxVideoConstraints the max video constraints that the bridge
+     *                            needs to receive from this endpoint
+     */
+    protected abstract void
+    onMaxReceiverVideoConstraintsChanged(VideoConstraints maxVideoConstraints);
```
bbaldino, line 402: nit: for these types of methods I think we don't use the `on` prefix, so just `maxReceiverVideoConstraintsChanged` (a method with an `on` would be used when adding a handler for an event, and the event occurring itself is without the `on`).

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -318,13 +315,48 @@ public void describe(ColibriConferenceIQ.ChannelBundle channelBundle)
...
+        VideoConstraints defaultVideoConstraints
+            = new VideoConstraints(BitrateControllerConfig.Config.thumbnailMaxHeightPx());
+
+        VideoConstraints newReceiverMaxVideoConstraints = newVideoConstraints
+            .values()
+            .stream()
+            .max(Comparator.comparingInt(VideoConstraints::getIdealHeight))
+            .orElse(defaultVideoConstraints);
+
+        // for intra-bridge traffic we only care about the ideal height.
```
bbaldino, line 351: I don't get why this commented refers to 'intra-bridge traffic', because `onMaxReceiverVideoConstraintsChanged` doesn't say anything about being specific to only intra-bridge traffic cases?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -354,8 +386,57 @@ public JSONObject getDebugState()
...
+     * @param newVideoConstraints the video constraints that the receiver
+     *                            wishes to receive.
+     */
+    public void addReceiver(String endpointId, VideoConstraints newVideoConstraints)
+    {
+        VideoConstraints oldVideoConstraints = receiverVideoConstraintsMap.put(endpointId, newVideoConstraints);
+        if (oldVideoConstraints == null
+            || !oldVideoConstraints.equals(newVideoConstraints))
+        {
+            setReceiverVideoConstraints(ImmutableMap.copyOf(receiverVideoConstraintsMap));
```
bbaldino, line 424: What is the difference between adding these constraints to `receiverVideoConstraintsMap` and calling `setReceiverVideoConstraints` ?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -638,9 +633,54 @@ public void addRtpExtension(RtpExtension rtpExtension)
...
+
+    @Override
+    protected void onMaxReceiverVideoConstraintsChanged(VideoConstraints maxVideoConstraints)
+    {
+        // Note that it's up to the client to respect these constraints.
+        String senderVideoConstraintsMessage = createSenderVideoConstraintsMessage(maxVideoConstraints);
+
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("Is now selected, sending message: " + senderVideoConstraintsMessage);
```
bbaldino, line 680: This log message seems inaccurate, it's not necessarily selected, right?  Just constraints changed?

jvb_pom.xml
```diff
@@ -64,6 +64,12 @@
       <artifactId>dom4j</artifactId>
       <version>1.6.1</version>
     </dependency>
+    <!-- temp dependency -->
```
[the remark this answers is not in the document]
gpolitis: @bbaldino I'm using this to deserialize the video constraints messages from other bridges. What we typically do for data channel messages is write our own serializers/deserializers, so I was planning on doing that (write a custom deserializer), and only use Guava for my testing (this is still a draft).

If we don't mind adding an extra dependency, I would be inclined to keep it and potentially scrap our custom deserializers, but I'm not sure if we want to do that?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: @bbaldino @bgrozev I've taken the preferred fps/resolution out of the video constraints class and put it in a video allocation policy here ef9e0270.

Three motivations behind this change:

1.  It doesn't feel right to signal the preferred fps/resolution to the sender for layer suspension
2. @bbaldino has been talking about an allocation policy for a while now and I agree with that, I just didn't know where/how to fit it
3. @bgrozev's comment here really nailed it: in the future when we extend the video constraints class with an ideal frame rate we need to go come back here to fix a bug.

Let me know what you think

src/main/kotlin/org/jitsi/videobridge/octo/OctoEndpoint.kt
gpolitis: bb530216fd8d5ed4df642befe0c6fceb98e9ccfd addresses this

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: cool, I renamed a few more methods.

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: `setReceiverVideoConstraints` (which I renamed to `receiverVideoConstraintsChanged`) is called when the receiver video constraints change (which happens when a receiver is added or removed to/from the map) and it computes the new max receiver video constraints. If that's different from the old (current) max receiver video constraints, then it calls the `maxReceiverVideoConstraints` method, effectively yielding to the implementation class.

src/main/java/org/jitsi/videobridge/Endpoint.java
bgrozev: Oh, I didn't know that.

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -354,8 +391,58 @@ public JSONObject getDebugState()
...
+     * @param newVideoConstraints the video constraints that the receiver
+     *                            wishes to receive.
+     */
+    public void addReceiver(String endpointId, VideoConstraints newVideoConstraints)
+    {
+        VideoConstraints oldVideoConstraints = receiverVideoConstraintsMap.put(endpointId, newVideoConstraints);
+        if (oldVideoConstraints == null
+            || !oldVideoConstraints.equals(newVideoConstraints))
+        {
+            receiverVideoConstraintsChanged(ImmutableMap.copyOf(receiverVideoConstraintsMap));
```
bgrozev, line 430: I think in this particular case we should avoid the new map, because we're just calling a private method and it doesn't actually need a map. How about just passing `receiverVideoConstraintsMap.values()` (or `Collection.unmodifiableList(receiverVideoConstraintsMap.values())`)?

jvb_pom.xml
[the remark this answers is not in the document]
bbaldino: I'd look at using jackson for this...it's what jersey uses for the rest code and I've used it elsewhere and had good experiences.

src/main/java/org/jitsi/videobridge/cc/VideoAllocationPolicy.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+package org.jitsi.videobridge.cc;
+
+import org.jitsi.videobridge.*;
+import org.json.simple.*;
+
+import java.util.*;
+
+public class VideoAllocationPolicy
+    extends VideoConstraints
```
bbaldino, line 24: I'm confused by this extending `VideoConstraints`, as I think the two things are very different.  The way I see it is:

1) Receivers create `VideoConstraints` instances that express how they would, ideally, view other senders.
2) Senders combine all the `VideoConstraints` that have been set for them to determine a lowest common denominator and determine if they can suspend certain layers (e.g. if the highest expressed constraint is 360p, it can stop sending 720p).
3. Totally separately, the `BitrateController` for an endpoint, when deciding what to forward, uses a `VideoAllocationPolicy` to determine how to allocate the bandwidth available across the streams.  For example, say a receiver was in tile view and had expressed they'd, ideally, see every sender in 720p, but their bandwidth is insufficient to handle this.  A `PreferMoreVideo` `VideoAllocationPolicy` would tradeoff quality to forward more streams (i.e. forward the 360p layer of all senders).  A `PreferQuality` `VideoAllocationPolicy` would trade off more streams to send better quality (i.e. maximize the forwarded quality of each sender one by one and, for some senders, not end up forwarding any video at all).

src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
```diff
@@ -206,6 +206,11 @@ public void setRelays(Collection<String> relays)
         setTargets(socketAddresses);
     }
 
+    public String getRelayId()
```
bbaldino, line 209: Looks like this isn't used anywhere?  (Maybe this is part of what you're planning on pulling out)

bgrozev: It's used in kotlin as `relayId` (I had the same thought)

src/main/java/org/jitsi/videobridge/cc/VideoAllocationPolicy.java
gpolitis: @bbaldino would a reasonable compromise at this point be to separate the `VideoAllocationPolicy` from the `VideoConstraints` but leave the fields as they are now (i.e. `VideoAllocationPolicy { preferredFps, preferredHeight }`

gpolitis: Actually I may be able to tweak the bitrate allocator to suit your suggestion. I'll work on this tomorrow.

gpolitis:
I'm closing this one as it's started growing too big. I'll open a smaller PR that implements single-bridge layer suspension and then we can think about video allocation policies and octo propagation of video constraints.

PR #1309, opened 2020-06-12
Fix DataChannel SCTP packets arriving out-of-order

bertm:
When opening a DataChannel SCTP connection on a heavily loaded machine, the logs often indicate:
> SEVERE: Could not find data channel for sid <sid>

This indicates that data packets arrived before the OpenChannelMessage was processed.

The cause of this issue is that incoming packet processing was scheduled on independent IO threads.
As such, they would be handled in any particular order on a loaded machine.
While JVB could recover from this condition for future packets, initial packets would get lost.
I have observed examples where this impacted initial Last-N selection, for which the command is sent from Jitsi Meet upon connection opened.

This change ensures that incoming SCTP packets are processed in the exact order that they arrived in.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Endpoint.java
+++ b/src/main/java/org/jitsi/videobridge/Endpoint.java
@@ -878,6 +878,7 @@ public class Endpoint
                 logger.info("SCTP connection is disconnected.");
             }
         };
+        SequentialExecutor receiveExecutor = new SequentialExecutor(TaskPools.IO_POOL);
         socket.dataCallback = (data, sid, ssn, tsn, ppid, context, flags) -> {
             // We assume all data coming over SCTP will be datachannel data
             DataChannelPacket dcp
@@ -886,8 +887,7 @@ public class Endpoint
             // holding a lock inside the SctpSocket which can cause a deadlock
             // if two endpoints are trying to send datachannel messages to one
             // another (with stats broadcasting it can happen often)
-            TaskPools.IO_POOL.execute(
-                    () -> dataChannelHandler.consume(new PacketInfo(dcp)));
+            receiveExecutor.execute(() -> dataChannelHandler.consume(new PacketInfo(dcp)));
         };
         socket.listen();
         sctpSocket = Optional.of(socket);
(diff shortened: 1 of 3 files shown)


bertm:
Done that.

bbaldino:
How about just putting the incoming datachannel packets into a blocking queue which a single IO thread reads from?

bertm:
That could work, however at the expense of maintaining yet another thread
per Endpoint. I'm not yet convinced that's worth it.

Ideally the message handling would be done on the shared CPU pool (while
sending any follow-up packets through an IO queue to prevent blocking the
CPU pool) because it primarily revolves around packet parsing and internal
routing. But that requires some time to analyze all flows for potential
blockers.

As soon as all DataChannel message handling is free of blocking operations,
switching to the CPU pool would constitute a single-line change after this
proposed change.

Op za 13 jun. 2020 05:00 schreef bbaldino <notifications@github.com>:

bertm:
Or do you mean a single IO thread, shared across all Endpoints?
That might not be too bad.

Obviously that solution would not scale, but that's unlikely to be a
problem given the relaitively small number of messages that the
DataChannels carry.

Op za 13 jun. 2020 05:00 schreef bbaldino <notifications@github.com>:

mstyura:
@bbaldino @bertm maybe it make sence to use `PacketsQueue` to handle sequential receiving of incoming sctp messages? It seems to solve exactly this issue.

mstyura:
I've provided alternative way to fix an issue based on `PacketInfoQueue`: https://github.com/jitsi/jitsi-videobridge/pull/1318

bbaldino:
> @bbaldino @bertm maybe it make sence to use `PacketsQueue` to handle sequential receiving of incoming sctp messages? It seems to solve exactly this issue.

Good idea, looking at that PR that seems like a good way to do it.  I'll take a closer look there soon.  @bertm any thoughts?

bertm:
I believe that the solution proposed by @mstyura is equally valid.
Given that `Endpoint` has now moved to Kotlin (with the same issue still present), neither of our PRs are in a mergeable state.

I'll give it a try to rebase and rework @mstyura's approach into Kotlin. Closing this PR.

PR #1311, opened 2020-06-15
Octo header and GID update

bgrozev:
This is a draft, because we need to have jicofo `>= b848a9fa19` running on meet.jit.si before merging.

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -104,10 +104,10 @@ public class Conference
 
     /**
      * The "global" id of this conference, set by the controller (e.g. jicofo)
-     * as opposed to the bridge. This defaults to {@code null} unless it is
-     * specified.
+     * as opposed to the bridge. This is a 32-bit unsigned integer, or {@code -1}
+     * if it is not set.
      */
-    private final String gid;
+    private final long gid;
 
     /**
      * The world readable name of this instance if any.
@@ -188,17 +188,18 @@ public class Conference
                       String id,
                       String conferenceName,
                       boolean enableLogging,
-                      String gid)
+                      long gid)
     {
+        if (gid < -1 || gid > 0xffff_ffffl)
+        {
+            throw new IllegalArgumentException("Invalid GID:" + gid);
+        }
         this.videobridge = Objects.requireNonNull(videobridge, "videobridge");
         Level minLevel = enableLogging ? Level.ALL : Level.WARNING;
         Map<String, String> context = JMap.ofEntries(
-            entry("confId", id)
+            entry("confId", id),
+            entry("gid", String.valueOf(gid))
         );
-        if (gid != null)
-        {
-            context.put("gid", gid);
-        }
         if (conferenceName != null)
         {
             context.put("conf_name", conferenceName);
@@ -969,10 +970,10 @@ public class Conference
     }
 
     /**
-     * @return the global ID of the conference, or {@code null} if none has been
+     * @return the global ID of the conference, or {@code -1} if none has been
      * set.
      */
-    public String getGid()
+    public long getGid()
     {
         return gid;
     }
@@ -1070,6 +1071,11 @@ public class Conference
      */
     public ConfOctoTransport getTentacle()
     {
+        if (gid < 0)
+        {
+            throw new IllegalStateException(
+                    "Can not enable Octo without a valid GID.");
+        }
         if (tentacle == null)
         {
             tentacle = new ConfOctoTransport(this);
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -61,6 +61,11 @@ public class Conference
         AbstractEndpointMessageTransport.EndpointMessageTransportEventHandler
 {
     /**
+     * The constant denoting that {@link #gid} is not specified.
+     */
+    public static final long GID_NOT_SET = -1;
+
+    /**
      * The endpoints participating in this {@link Conference}. Although it's a
      * {@link ConcurrentHashMap}, writing to it must be protected by
      * synchronizing on the map itself, because it must be kept in sync with
@@ -98,14 +103,28 @@ public class Conference
     private boolean expired = false;
 
     /**
-     * The (unique) identifier/ID of this instance.
+     * The locally unique identifier of this conference (i.e. unique across the
+     * conferences on this bridge). It is locally generated and exposed via
+     * colibri, and used by colibri clients to reference the conference.
+     *
+     * The current thinking is that we will phase out use of this ID (but keep
+     * it for backward compatibility) in favor of {@link #gid}.
      */
     private final String id;
 
     /**
-     * The "global" id of this conference, set by the controller (e.g. jicofo)
-     * as opposed to the bridge. This is a 32-bit unsigned integer, or {@code -1}
-     * if it is not set.
+     * The "global" id of this conference, selected by the controller of the
+     * bridge (e.g. jicofo). The set of controllers are responsible for making
+     * sure this ID is unique across all conferences in the system.
+     *
+     * This is not a required field unless Octo is used, and when it is not
+     * specified by the controller its value is {@link #GID_NOT_SET}.
+     *
+     * If specified, it must be a 32-bit unsigned integer (i.e. in the range
+     * [0, 0xffff_ffff]).
+     *
+     * This ID is shared between all bridges in an Octo conference, and included
+     * on-the-wire in Octo packets.
      */
     private final long gid;
 
@@ -190,7 +209,7 @@ public class Conference
                       boolean enableLogging,
                       long gid)
     {
-        if (gid < -1 || gid > 0xffff_ffffl)
+        if (gid != GID_NOT_SET && (gid < 0 || gid > 0xffff_ffffL))
         {
             throw new IllegalArgumentException("Invalid GID:" + gid);
         }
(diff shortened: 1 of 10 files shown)


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -188,17 +188,14 @@ public Conference(Videobridge videobridge,
...
         this.videobridge = Objects.requireNonNull(videobridge, "videobridge");
         Level minLevel = enableLogging ? Level.ALL : Level.WARNING;
         Map<String, String> context = JMap.ofEntries(
             entry("confId", id)
         );
-        if (gid != null)
-        {
-            context.put("gid", gid);
-        }
+        context.put("gid", String.valueOf(gid));
```
bbaldino, line 198: You can movie this to the `JMap.ofEntries` chunk above now

src/main/java/org/jitsi/videobridge/health/Health.java
```diff
@@ -174,4 +176,15 @@ protected void performCheck()
...
+
+    /**
+     * Generate the next GID for a test conference.
+     */
+    private long getNextGid()
+    {
+        // Jicofo splits the 32bit GID field in two parts: 16bits to identify
+        // the jicofo instance, and 16 to identify the conference within that
+        // instance. Here we use 0xff as the ID of the jicofo instance.
+        return (0xff << 16) | (gidCounter++ & 0xff);
```
bbaldino, line 188: Should it be `0xffff` for the 'fake' jicofo id?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -21,22 +21,25 @@
...
  * <pre>{@code
  *  0                   1                   2                   3
  *  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
- * |R| M | S | res |              Conference ID                    |
+ * |                         Conference ID                         |
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * |                         Endpoint ID                           |
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+ * |MT|                     Reserved                              |
```
bbaldino, line 33: nit: one space short here

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -85,28 +88,23 @@ private static int getMediaTypeId(MediaType mediaType)
...
-        }
-        buf[off] |= (getMediaTypeId(mediaType) & 0x03) << 5;
-        buf[off] |= (s & 0x03) << 3;
-
-        writeConferenceId(conferenceId, buf, off, OCTO_HEADER_LENGTH);
-        writeEndpointId(endpointId, buf, off, OCTO_HEADER_LENGTH);
+        assertMinLen(buf, off, buf.length);
+        off += writeConferenceId(conferenceId, buf, off);
+        off += writeEndpointId(endpointId, buf, off);
+        buf[off] = (byte) (getMediaTypeId(mediaType) << 6);
```
bbaldino, line 104: Should we have `readMediaType` and `writeMediaType`?

bgrozev: We already have `readMediaType`. The `writeXXX` helpers are private, so I didn't think necessary to add one for `mediaType` (to make the semantics of `writeMediaType` nice we would have to take care of not replacing the extra bits of that byte, and then zero it out here, which is unnecessary complexity).

bbaldino: Oh, oops, I saw the read code below but didn't notice it was within `readMediaType`.

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -104,10 +104,10 @@
 
     /**
      * The "global" id of this conference, set by the controller (e.g. jicofo)
-     * as opposed to the bridge. This defaults to {@code null} unless it is
-     * specified.
+     * as opposed to the bridge. This is a 32-bit unsigned integer, or {@code -1}
```
bbaldino, line 107: Could we do a constant for the `-1` value, since we use it for health checks?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -21,22 +21,25 @@
...
  * <pre>{@code
  *  0                   1                   2                   3
  *  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
- * |R| M | S | res |              Conference ID                    |
+ * |                         Conference ID                         |
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
  * |                         Endpoint ID                           |
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+ * |MT|                      Reserved                              |
```
bbaldino, line 33: nit: there should be a space before the closing `|` of the `MT` chunk to align it with the boundary of the 3rd bit

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -85,28 +88,23 @@ private static int getMediaTypeId(MediaType mediaType)
...
-     * @param s the value of the {@code s} flag.
      * @param conferenceId the Octo ID of the conference.
      * @param endpointId the Octo ID of the endpoint.
      */
     public static void writeHeaders(
             byte[] buf, int off,
-            boolean r, MediaType mediaType, int s,
-            String conferenceId,
+            MediaType mediaType,
+            long conferenceId,
```
bbaldino, line 98: `conferenceId` -> `gId` ?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -21,22 +21,25 @@
...
 /**
  * A utility class which handles the on-the-wire Octo format. Octo encapsulates
- * its payload (RTP, RTCP, or anything else) in an 8-byte header:
+ * its payload (RTP, RTCP, or anything else) in an 12-byte header:
  * <pre>{@code
  *  0                   1                   2                   3
  *  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
  * +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
- * |R| M | S | res |              Conference ID                    |
+ * |                         Conference ID                         |
```
bbaldino, line 29: `conferenceId` -> `gId` ?

src/main/java/org/jitsi/videobridge/octo/OctoPacket.java
```diff
@@ -116,12 +114,11 @@ public static void writeHeaders(
      * @param len the length of the buffer.
      * @return the Octo conference ID read from the buffer.
      */
-    public static String readConferenceId(byte[] buf, int off, int len)
+    public static long readConferenceId(byte[] buf, int off, int len)
```
bbaldino, line 117: `readGid`?

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: I removed the use for health checks. It's used in Videobridge and VideobridgeShim.

src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -229,11 +229,29 @@ public Videobridge()
      * @param name world readable name of the conference to create.
      * @param enableLogging whether logging should be enabled or disabled for
      * the {@link Conference}.
-     * @param gid the optional "global" id of the conference.
+     */
+    public @NotNull Conference createConference(
+            String name, boolean enableLogging)
+    {
+        return createConference(name, enableLogging, -1);
```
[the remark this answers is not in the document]
bbaldino: ~Ok, I see that we end up with `GID_NOT_SET` if the `ColibriConferenceIQ` is missing the field...would we rather fail when parsing the IQ?  Or is there a use case for Jicofo not to set the field?  (Does Jicofo do that if Octo is disabled, maybe?  In that case we could probably get rid of `GID_NOT_SET` for `HEALTH_CHECK_GID` in the future when we get rid of conference ID and always use gid?~ EDIT: Ok, it looks like Jicofo does this if Octo is not configured, I think, in which case we can leave the naming as-is for now)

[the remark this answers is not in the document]
bgrozev: Jicofo does not use -1 (neither current master, nor my branch), but other colibri clients might. For now, GID is only necessary with Octo, so the health checks don't have to take it into account.

I'll use the constant.

PR #1315, opened 2020-06-17
Log when we send a large (potentially >MTU) packets.

bgrozev:


src/main/kotlin/org/jitsi/videobridge/transport/octo/BridgeOctoTransport.kt
```diff
@@ -170,6 +170,10 @@ class BridgeOctoTransport(
     @Suppress("DEPRECATION")
     fun sendString(msg: String, targets: Set<SocketAddress>, confId: String) {
         val msgData = msg.toByteArray(StandardCharsets.UTF_8)
+        if (msgData.size > 1500) {
+            logger.warn("Sending a large data packet over octo (${msgData.size} bytes): $msg")
```
JonathanLennox, line 174: Do we really want to dump the whole packet into the log?  Especially as verbatim bytes?

PR #1322, opened 2020-06-19
wip: Receiver video constraints for layer suspension.

gpolitis:


src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -318,13 +318,42 @@ public void describe(ColibriConferenceIQ.ChannelBundle channelBundle)
...
+    {
+        VideoConstraints oldReceiverMaxVideoConstraints
+            = this.maxReceiverVideoConstraints;
+
+        VideoConstraints newReceiverMaxVideoConstraints = newVideoConstraints
+            .stream()
+            .max(Comparator.comparingInt(VideoConstraints::getIdealHeight))
+            .orElse(defaultMaxReceiverVideoConstraints);
+
+        if (!newReceiverMaxVideoConstraints.equals(oldReceiverMaxVideoConstraints))
```
bgrozev, line 350: Same as before (`equals` takes into account more than `idealHeight`). I though that's why we planned to do the `VideoConstraints` refactor as a first step?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -354,8 +383,58 @@ public JSONObject getDebugState()
     public abstract void addRtpExtension(RtpExtension rtpExtension);
 
     /**
+     * @param videoConstraints
```
bgrozev, line 386: Can you please add a description?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -354,8 +383,58 @@ public JSONObject getDebugState()
...
+     * @param maxVideoConstraints the max video constraints that the bridge
+     * needs to receive from this endpoint
+     */
+    protected abstract void
+    maxReceiverVideoConstraintsChanged(VideoConstraints maxVideoConstraints);
+
+    /**
+     * The map of receiver endpoint id -> video constraints.
+     */
+    private final Map<String, VideoConstraints> receiverVideoConstraintsMap = new ConcurrentHashMap<>();
```
bgrozev, line 405: This should be on top with the other fields

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -109,28 +110,31 @@ private void onJSONData(
...
-                    onClientEndpointMessage(src, jsonObject);
+                    clientEndpointMessage(src, jsonObject);
                     break;
                 case COLIBRI_CLASS_LASTN_CHANGED:
-                    onLastNChangedEvent(src, jsonObject);
+                    lastNChangedEvent(src, jsonObject);
                     break;
                 case COLIBRI_CLASS_RECEIVER_VIDEO_CONSTRAINT:
-                    onReceiverVideoConstraintEvent(src, jsonObject);
+                    receiverVideoConstraintEvent(src, jsonObject);
```
bgrozev, line 134: Can you add a doc somewhere (either here, in the methods that handle each message, or in the description of the constants) that describe the semantics of `RECEIVER_VIDEO_CONSTRAINT` vs `RECEIVER_VIDEO_CONSTRAINTS_CHANGED`?

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -408,18 +412,21 @@ protected void onPinnedEndpointsChangedEvent(
...
         Set<String> newSelectedEndpoints)
     {
         videoConstraintsCompatibility.setSelectedEndpoints(newSelectedEndpoints);
-        onVideoConstraintsChangedEvent(
+        senderVideoConstraintsChangedEvent(
             videoConstraintsCompatibility.computeVideoConstraints());
     }
 
-    protected void onVideoConstraintsChangedEvent(Map<String, VideoConstraints> newVideoConstraints)
+    protected void senderVideoConstraintsChangedEvent(Map<String, VideoConstraints> videoAllocationPolicyMap)
```
bgrozev, line 424: Rename the variable to `videoConstraintsMap` and add a doc?

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -479,7 +499,7 @@ protected void onReceiverVideoConstraintEvent(
         }
 
         videoConstraintsCompatibility.setMaxFrameHeight(maxFrameHeight);
-        onVideoConstraintsChangedEvent(videoConstraintsCompatibility.computeVideoConstraints());
+        senderVideoConstraintsChangedEvent(videoConstraintsCompatibility.computeVideoConstraints());
```
bgrozev, line 502: Does that mean that we have two completely independent code paths that send the "sender video constraints" message -- one from selected/pinned/old receiver constraint events (that goes through the compatibility layer), and one from the new "receiver constraints" message (which does NOT go through the compatibility layer)?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: > Same as before (equals takes into account more than idealHeight).

I can certainly change this line to compare the ideal height only, but 1) in practice this won't fix nor change anything, and 2) when we add an `idealFps` field we'll have to to remember to come back here and augment the comparison. 

Eventually we will separate the video constraints (ideal fps, ideal resolution) from the video allocation policy (preferred fps, preferred resolution) and this equality comparison will become less awkward.

Let me know if you wish me to elaborate on 1) and/or if you feel this needs to compare the ideal height only, regardless. 

> I though that's why we planned to do the VideoConstraints refactor as a first step?

Yes.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
gpolitis: Yes

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
gpolitis: I tried adding as much details as possible here bf6d6616165cfeee6415e38db6fb2bf31d8b3775. Please let me know if further details are necessary.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
gpolitis: Thanks!

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
bgrozev: Makes sense for now, I just expected the other step first.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
bgrozev: Reading further, support for the new message is a no-op. Are you *planning* to make it go through the compatibility layer?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -354,8 +389,67 @@ public JSONObject getDebugState()
     public abstract void addRtpExtension(RtpExtension rtpExtension);
 
     /**
+     * Sets the map of endpoint id to {@link VideoConstraints} that contains the
```
bgrozev, line 392: Can you add something like "the video constraints for the streams that this endpoint wishes to receive"?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -638,9 +633,54 @@ public void addRtpExtension(RtpExtension rtpExtension)
     }
 
     @Override
-    public void setVideoConstraints(Map<String, VideoConstraints> newVideoConstraints)
+    public void setSenderVideoConstraints(ImmutableMap<String, VideoConstraints> newVideoConstraints)
```
bgrozev, line 636: I think the `sender/receiver` terminology makes things very confusing. After going through the code and sketching things on a piece of paper, I understand that:

`e.setSenderVideoConstraints()` sets the constraints that `e` wishes to receive from other senders. While `e.receiverVideoConstraintsChanged()` sets the constraints that endpoints receiving video from `e` wish to impose. I think that's consistent, but somewhat confusing and not documented.

I think using "local/remote" language may be clearer (along with docs that say the basics in a simple way). I.e. 
```
/*
Sets the constraints to be applied to the streams sent by this endpoint. Each entry in the map is a constraint imposed by a receiving endpoint.`
*/
setLocalVideoConstraints(Map<String, VideoConstraints>)

/**
Sets the constraints that this endpoint wishes to impose on streams sent by remote endpoints.
*/
setRemoveVideoConstraints()
```

If you don't like this language that's okay, it's just an example. But please add a simple description of the semantics in the docs.

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -74,6 +91,14 @@
      */
     private boolean expired = false;
 
+    /**
+     * The maximum set of constraints applied by all receivers of this endpoint
+     * in the conference. The client needs to send _at least_ this to satisfy
+     * all receivers.
+     */
+    private VideoConstraints maxReceiverVideoConstraints
+        = new VideoConstraints(BitrateControllerConfig.Config.onstageIdealHeightPx());
```
bgrozev, line 100: This starts as `onstageIdealHeightPx`,  but then defaults to `defaultMaxReceiverVideoConstraints = thumbnailVideoConstraints` (which is different). Is that on puspose?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -638,9 +633,54 @@ public void addRtpExtension(RtpExtension rtpExtension)
...
+        ImmutableMap<String, VideoConstraints>
+            oldVideoConstraints = bitrateController.getVideoConstraints();
+
         bitrateController.setVideoConstraints(newVideoConstraints);
+
+        Set<String> endpointsThatNoLongerCare
+            = new HashSet<>(oldVideoConstraints.keySet());
+        endpointsThatNoLongerCare.removeAll(newVideoConstraints.keySet());
+
+        // Endpoints that no longer care what they receive.
```
bgrozev, line 647: The method sets the constraints that `this` wishes to impose on other senders. So this must be the set of ID that `this` "doesn't care" about.

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
bgrozev: If I understand correctly, the expectation the current client code has is that endpoints that are not specified (either pinned or selected) default to the lowest layer. If we want to allow senders to completely stop sending video in the future, should we handle this elsewhere (i.e. have receivers add themselves with a default of `thumbnailVideoConstraints`) and default to `none` here?

src/main/java/org/jitsi/videobridge/Endpoint.java
gpolitis: I think of it in terms of rtp sender/receiver, i.e. conceptually the setSenderVideoConstraints sets the video constraints on the rtp sender and the setReceiverVideoConstraints sets the video constraints on the rtp receiver (we no longer have such a method because it got simplified in the previous pr). Does that make more sense?

That said, I'll admit I'm not very good at naming, so I'm happy to follow through your suggestion if you insist.

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
gpolitis: The method that this comment pertains to got renamed to `receiverVideoConstraintsChangedEvent ` but I suppose you're still talking about `COLIBRI_CLASS_RECEIVER_VIDEO_CONSTRAINTS_CHANGED`.

I wasn't planning to add support for the message in this PR because I don't want to delay things any further and it's not functionally necessary; I'm happy to work on it if you think it's better to have it as part of this PR. Its main usage will be for the octo part, at least that was how it was setup in the previous PR.

src/main/java/org/jitsi/videobridge/Endpoint.java
gpolitis: You're right, the comment in the code is wrong. How about this `// Endpoints that "this" no longer cares about what it receives.`.

And also rename the variable to `endpointsINoLongerCare`?

bgrozev: Makes sense. `removedEndpoints`?

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: > If we want to allow senders to completely stop sending video in the future

I'm thinking we can augment the video constraints with a max with the semantics: "The bridge will not exceed this resolution. If there aren't any encodings does provide resolution or less, the stream from this endpoint becomes suspended."

Much like in the webrtc constraints spec.

gpolitis: Something like this 592b92e4eab21a6f7a1d2027411cd617352d3c4f (nothing uses this atm and I should probably revert this, unless you think it's useful).

src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
gpolitis: I agree it would be nice to have it as a first step, but @bbaldino rejected my previous suggestion, so I'd leave this as a future step when we're ready to tackle it.

src/main/java/org/jitsi/videobridge/Endpoint.java
bgrozev: > I think of it in terms of rtp sender/receiver, i.e. conceptually the setSenderVideoConstraints sets the video constraints on the rtp sender and the setReceiverVideoConstraints sets the video constraints on the rtp receiver

This description makes it much easier to understand!

PR #1327, opened 2020-06-29
Refactor bridge channel messages.

bgrozev:


src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -0,0 +1,225 @@
...
+    fun toJson(): String = ObjectMapper().writeValueAsString(this)
+
+    companion object {
+        @JsonIgnoreProperties(ignoreUnknown = true)
+        private class EmptyBridgeChannelMessage : BridgeChannelMessage("")
+
+        @Throws(InvalidMessageTypeException::class, JsonProcessingException::class, JsonMappingException::class)
+        fun parse(string: String): BridgeChannelMessage {
+            // First parse as an empty message, ignoring any fields other than colibriClass, in order to get the type.
+            val colibriClass = jacksonObjectMapper().readValue<EmptyBridgeChannelMessage>(string).type
```
bbaldino: I think I can see why you didn't do a custom deserializer (looking at some examples, it looks like it would be overkill unless there's a good way to leverage existing code after doing the first bit), but I found another strategy that might fit better [here](https://stackoverflow.com/a/44123978/2734014).  This way is a bit confusing since creating an `EmptyBridgeChannelMessage` instantiates the parent `type` field with an empty string, but somehow it does magically get filled out--to be honest I'm not even sure how that field gets properly set considering it isn't even an argument to `EmptyBridgeChannelMessage`.

src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -75,24 +77,30 @@ sealed class BridgeChannelMessage(
...
-        is EndpointMessage -> endpointMessage(message)
-        is LastNMessage -> lastN(message)
-        is ReceiverVideoConstraintMessage -> receiverVideoConstraint(message)
-        is ReceiverVideoConstraintsMessage -> receiverVideoConstraints(message)
-        is DominantSpeakerMessage -> dominantSpeaker(message)
-        is EndpointConnectionStatusMessage -> endpointConnectionStatus(message)
-        is ForwardedEndpointMessage -> forwardedEndpoints(message)
-        is SenderVideoConstraintsMessage -> senderVideoConstraints(message)
+    fun handleMessage(message: BridgeChannelMessage): BridgeChannelMessage? {
+        receivedCounts.computeIfAbsent(message::class.java.simpleName) { AtomicLong()} .incrementAndGet()
```
bbaldino: nit: I think this space got shifted...should be before the `}`

src/main/kotlin/org/jitsi/videobridge/transport/octo/BridgeOctoTransport.kt
```diff
@@ -274,7 +274,7 @@ class BridgeOctoTransport(
         }
 
         fun largePacketSent(mediaType: MediaType) {
-            val value = largePacketsSent.getValue(mediaType).incrementAndGet()
+            val value = largePacketsSent.computeIfAbsent(mediaType) { AtomicLong() }.incrementAndGet()
```
bbaldino, line 277: Uh oh, why didn't the `withDefault` work? (arguably should be a different PR)

src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
bbaldino: This got resolved in a later commit

src/main/kotlin/org/jitsi/videobridge/transport/octo/BridgeOctoTransport.kt
bgrozev: `withDefault` returns a default value without storing it in the map (i.e. returns a new AtomicLong() every time). I'll open a separate PR.

bbaldino: 🤦  Ugh, ok that was my bad.  I can do the other PR if you want.  Let me know.

PR #1331, opened 2020-06-30
Enforce/fix line lengths for java

bbaldino:


jvb_pom.xml
```diff
@@ -100,6 +100,12 @@
         <artifactId>spotbugs-annotations</artifactId>
         <version>${spotbugs.version}</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/org.apache.maven.plugins/maven-checkstyle-plugin -->
```
bgrozev, line 103: Is this intentional?

bbaldino: When looking at spotbugs stuff I noticed you had left that line there, so I think I got it in my head that we should...though I haven't in the past and I'm not sure if there's much value?

bbaldino: OH, oops...sorry, misunderstood which line this was.  No, it's not.  I'll remove it.

PR #1333, opened 2020-07-01
Fixes an NullPointerException thrown in the video constraints compatibility module.

gpolitis:


gpolitis:
strange that the tests fail on this, I'll look into it tomorrow

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -470,6 +470,10 @@ public BridgeChannelMessage selectedEndpoints(SelectedEndpointsMessage message)
     {
         Set<String> newSelectedEndpoints = new HashSet<>(message.getSelectedEndpoints());
 
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("Selected " + newSelectedEndpoints);
```
bbaldino, line 475: You can now do `logger.debug(() -> "Selected " + newSelectedEndpoints);` and avoid the extra `if` check.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -488,6 +492,12 @@ public void setSenderVideoConstraints(
         // Don't "pollute" the video constraints map with constraints for this
         // endpoint.
         videoConstraintsMap.remove(endpoint.getID());
+
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("New video constraints map: " + videoConstraintsMap);
```
bbaldino, line 498: Same as above

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
gpolitis: Doesn't this particular syntax create an anonymous class type (once) and a new instance every time it runs? Makes me wonder if simply concatenating the two strings is more efficient.

PR #1334, opened 2020-07-01
Move jvb code into a subdir

bbaldino:
When creating the JVB API, I left the jvb source where it was in the interest of avoiding another big change.  Maven/intellij seemed happy with this, but @damencho and now myself have run into some issues with intellij with the current configuration and making the jvb a 'proper' submodule (in a subdirectory) appears to make intellij happy (along with being a more traditional model, from what I can tell).

This is a first shot at moving things around to achieve this--there may be some files I moved into `jvb/` that shouldn't be or some files that should be moved that I didn't, so let me know if there are any thoughts.


damencho:
Why do we need to move debian, resources, docs, config and lib folder?
For example debian is normally in the top folder, and same for docs, people tend to search for those there.

damencho:
I would say just to move pom file and sources and leave rest as it is.

damencho:
I had moved the branch to the jitsi repo, it is easier that way when two or more people are working together on a project.
I fixed the debian package, you can see the commits https://github.com/jitsi/jitsi-videobridge/commits/module_fix
I'm only not certain about api-common and api-client dependencies, are those required for jvb to run?
Haven't tested the resulting deb package as there is some problem with jitsi-meet-web-config package when installing on a clean VM. But will do at some point.

bbaldino:
> Why do we need to move debian, resources, docs, config and lib folder?
> For example debian is normally in the top folder, and same for docs, people tend to search for those there.

I moved all those things because they only apply to JVB, so it felt wrong to have them at the top level (since the jvb-api stuff is totally independent).  For example, the debian package will not include the jvb-api stuff.

damencho:
> > Why do we need to move debian, resources, docs, config and lib folder?
> > For example debian is normally in the top folder, and same for docs, people tend to search for those there.
> 
> I moved all those things because they only apply to JVB, so it felt wrong to have them at the top level (since the jvb-api stuff is totally independent). For example, the debian package will not include the jvb-api stuff.

Then why we just don't move this API out in a separate repo and don't bother with all these parent poms and stuff?

The debian and doc folder, in my opinion, should stay where they are, we should not move them.

bbaldino:
> > > Why do we need to move debian, resources, docs, config and lib folder?
> > > For example debian is normally in the top folder, and same for docs, people tend to search for those there.
> > 
> > 
> > I moved all those things because they only apply to JVB, so it felt wrong to have them at the top level (since the jvb-api stuff is totally independent). For example, the debian package will not include the jvb-api stuff.
> 
> Then why we just don't move this API out in a separate repo and don't bother with all these parent poms and stuff?
> 
> The debian and doc folder, in my opinion, should stay where they are, we should not move them.

Because we're already frustrated with the amount of different repos and have considered merging other ones here (JMT, RTP).  Also I think it makes sense for the API which controls the JVB to live alongside the JVB code.  I don't think the parent poms are that bad...multi-module/monorepo projects are pretty common from what I've seen.

Both those docs and the debian files are specific to JVB, so I think it makes sense to have them with the jvb code, as the top-level will now encompass other things to which they don't apply (the JVB API stuff).  I do think the top-level README should exist, but it should  be changed to give information about the individual projects (jvb and jvb-api) and include links to the docs in those submodules, but I didn't do that yet.

damencho:
So this is jvb repo, and we are building jvb. The debian folder should be top level.
We have millions of links scattered around pointing to that doc folder and people relay on that when reading the community forum or the issue tracker, so we should keep that one too. 
For config, lib and resources, I don't know .... resources keep the script for Jenkins building this project so why move them then ... 
If you are trying to create a mono repo of all dependencies, let's start clear in some other repo and not convert this one to such ... as it brings only problems and confusion ...

bbaldino:
> So this is jvb repo, and we are building jvb. The debian folder should be top level.
> We have millions of links scattered around pointing to that doc folder and people relay on that when reading the community forum or the issue tracker, so we should keep that one too.

doc we could keep for now, but I do think we should migrate it to live alongside the jvb code at some point.  I'm not sure what criteria we would apply as to when that would be any easier, though, so I'm not sure why we'd delay it.  Breaking links isn't great, sure, but it seems unavoidable...we never move anything?

> For config, lib and resources, I don't know .... resources keep the script for Jenkins building this project so why move them then ...

The jenkins scripts can stay (I left them where they were) as that's for building the projects in this repo, so it makes sense to have at the top level.  Config, lib and other resources are jvb-specific so I moved those.

> If you are trying to create a mono repo of all dependencies, let's start clear in some other repo and not convert this one to such ... as it brings only problems and confusion ...

This seems like a much more massive/breaking change?

damencho:
> doc we could keep for now, but I do think we should migrate it to live alongside the jvb code at some point. I'm not sure what criteria we would apply as to when that would be any easier, though, so I'm not sure why we'd delay it. Breaking links isn't great, sure, but it seems unavoidable...we never move anything?

That should go in the handbook, I guess. And the files should look like: https://github.com/jitsi/jitsi-meet/blob/master/doc/api.md for a while.

damencho:
Ok, I will move back config, lib and resources back to jvb and will fix the debian packaging ... 
I'm not sure how to proceed with this PR and the branch though ... will you just revert the debian and doc folder move and move my commits in this branch ... or ...

bbaldino:
> Ok, I will move back config, lib and resources back to jvb and will fix the debian packaging ...
> I'm not sure how to proceed with this PR and the branch though ... will you just revert the debian and doc folder move and move my commits in this branch ... or ...

Since the rebase is a nightmare, the easiest thing will probably be to just re-do the work on top of current master.  I can do that.

So we're saying leave doc and debian at the top level?  I don't agree but can compromise there for now.  FWIW I don't want to remove docs from the JVB repo in favor of the handbook--I don't know if anyone is arguing for that but I think it would be better to _link_ to JVB docs _from_ the handbook and have the docs still live here.

bbaldino:
I'll also make sure to cherry-pick your fix of the wrong relative-path parent thing in the jvb pom, nice catch.

PR #1342, opened 2020-07-06
Fix endpoint messaging with Octo (and refactor).

bgrozev:


src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
```diff
@@ -97,14 +96,15 @@ protected void notifyTransportChannelConnected()
...
      */
-    @SuppressWarnings("unchecked")
     @Override
     public BridgeChannelMessage endpointMessage(EndpointMessage message)
     {
         String to = message.getTo();
 
         // First insert/overwrite the "from" to prevent spoofing.
-        message.setFrom(getId());
+        String from = getId(message.getFrom());
```
bbaldino: side note: having the default behavior of `getId` be specific to `EndpointMessageTransport` really tripped me up here, I think we should make it abstract in this class and do the impls in `OctoEndpointMessageTransport` and `EndpointMessageTransport`, but not an issue for this PR.

bbaldino: Haha!  Clearly you felt the same, based on the next commit :)

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -535,4 +537,69 @@ public BridgeChannelMessage lastN(LastNMessage message)
...
+        if (conference == null || conference.isExpired())
+        {
+            logger.warn("Unable to send EndpointMessage, conference is null or expired");
+            return null;
+        }
+
+        boolean sendToOcto;
+
+        List<AbstractEndpoint> targets;
+        if ("".equals(to))
```
bbaldino, line 573: nit: use an `isEmpty` helper ?

src/main/java/org/jitsi/videobridge/AbstractEndpointMessageTransport.java
bgrozev: Yeah, I thought about it before, but decided against because I didn't want to duplicate code. And it bit me.

src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
bgrozev: I added a helper in EndpointMessage itself

PR #1344, opened 2020-07-08
Order last-n endpoints by video availability.

bgrozev:
When ordering the endpoints to forward, order not only by speech activity, but also whether endpoints are sending video (those sending video are put on the top of the list).

### DIFF
--- a/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/src/main/java/org/jitsi/videobridge/Conference.java
@@ -190,6 +190,12 @@ public class Conference
     private final LastNEndpoints lastNEndpoints = new LastNEndpoints();
 
     /**
+     * The task of updating the ordered list of endpoints in the conference. It runs periodically in order to adapt to
+     * endpoints stopping or starting to their video streams (which affects the order).
+     */
+    private ScheduledFuture<?> updateLastNEndpointsFuture;
+
+    /**
      * Initializes a new <tt>Conference</tt> instance which is to represent a
      * conference in the terms of Jitsi Videobridge which has a specific
      * (unique) ID.
@@ -232,6 +238,18 @@ public class Conference
         this.conferenceName = conferenceName;
 
         speechActivity = new ConferenceSpeechActivity(this);
+        updateLastNEndpointsFuture = TaskPools.SCHEDULED_POOL.scheduleAtFixedRate(() -> {
+            try
+            {
+                lastNEndpoints.update();
+            }
+            catch (Exception e)
+            {
+                logger.warn("Failed to update lastN endpoints:", e);
+            }
+
+        }, 3, 3, TimeUnit.SECONDS);
+
         audioLevelListener
             = (sourceSsrc, level) -> speechActivity.levelChanged(sourceSsrc, (int) level);
 
@@ -499,6 +517,13 @@ public class Conference
         }
 
         logger.info("Expiring.");
+
+        if (updateLastNEndpointsFuture != null)
+        {
+            updateLastNEndpointsFuture.cancel(true);
+            updateLastNEndpointsFuture = null;
+        }
+
         EventAdmin eventAdmin = getEventAdmin();
         if (eventAdmin != null)
         {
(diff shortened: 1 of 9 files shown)


src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1292,4 +1314,41 @@ public Object put(String key, Object value)
...
+        private void update(boolean force)
+        {
+            List<AbstractEndpoint> endpointsBySpeechActivity = speechActivity.getEndpoints();
+
+            Map<Boolean, List<AbstractEndpoint>> bySendingVideo
+                    = endpointsBySpeechActivity.stream()
+                        .collect(Collectors.groupingBy(AbstractEndpoint::isSendingVideo));
+
+            List<AbstractEndpoint> lastNEndpoints = new LinkedList<>();
+            lastNEndpoints.addAll(bySendingVideo.getOrDefault(true, Collections.emptyList()));
```
bbaldino, line 1341: I played with this code a bit and came up with a couple ideas.  Both are in Kotlin (I'd be all for changing `LastNEndpoints` to Kotlin, even if it meant it could no longer be a private inner class).

Option 1:
```
val (epsWithVideo, epsWithoutVideo) = endpointsBySpeechActivity.partition(AbstractEndpoint::isSendingVideo)
val lastNEndpointIds = (epsSending + epsNotSending).map(AbstractEndpoint::id)
```

Option 2: use a comparator
```
val lastNRankComparator = object : Comparator<Pair<Int, AbstractEndpoint>> {
    override fun compare(o1: Pair<Int, AbstractEndpoint>, o2: Pair<Int, AbstractEndpoint>): Int {
        // If both endpoints are sending/not sending video, compare by their original speaker rank
        if (o1.second.sendingVideo == o2.second.sendingVideo) {
            return o1.first.compareTo(o2.first)
        } else if (o1.second.sendingVideo) {
            return -1
        }
        return 1
    }
}
```
And then the code to sort:
```
val result = endpointsBySpeechActivity.mapIndexed { index, entry ->
    // Create an intermediary pair to store the original speaker rank
    Pair(index, entry)
}.sortedWith(lastNRankComparator)
```

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -73,7 +73,7 @@
  */
 public class Endpoint
     extends AbstractEndpoint implements PotentialPacketHandler,
-        EncodingsManager.EncodingsUpdateListener
+        EncodingsManager.EncodingsUpdateListener, TransceiverEventHandler
```
bbaldino, line 76: For implementations of interfaces that will never be called by an external user of the class, I think it's nice to do "private" implementations of the interface (e.g. via a private inner class).  It also moves the methods out of the 'top level' of the class which I think is cleaner.

src/main/java/org/jitsi/videobridge/octo/OctoTransceiver.java
```diff
@@ -54,18 +54,23 @@
     private final StreamInformationStore streamInformationStore
             = new StreamInformationStoreImpl();
 
-    private final OctoRtpReceiver octoReceiver;
+    final OctoRtpReceiver octoReceiver;
```
bbaldino, line 57: Can you keep this private and add a passthrough method in `OctoTransceiver`?

src/main/kotlin/org/jitsi/videobridge/octo/config/OctoRtpReceiver.kt
```diff
@@ -60,6 +61,7 @@ import java.util.concurrent.atomic.AtomicBoolean
  */
 class OctoRtpReceiver(
     val streamInformationStore: ReadOnlyStreamInformationStore,
+    val eventHandler: TransceiverEventHandler?,
```
bbaldino, line 64: Why nullable?  It wasn't for the non-octo classes in JMT.

bbaldino: If there is a reason to keep it nullable, I'd add a default value of `null`.

bgrozev: The instance created in OctoEndpoints doesn't need an event handler

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: I agree Option 1 looks nicer (though that would be offset by having to pull it out of Conference), but I don't want to re-write and test it just because of that. Do you see a problem with this code?

bbaldino: > Do you see a problem with this code?

Just that I think it's doing something simple but I find it hard to follow.  And I think being able to test it separately would be a good thing.

bgrozev: I'll add tests in a separate PR, and I'll make sure to simplify the code. Can we move forward and ship this?

bbaldino: Ok

PR #1348, opened 2020-07-14
Refactor speech activity and add tests

bgrozev:


src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -40,261 +41,135 @@
...
     }
 
     /**
-     * Retrieves a JSON representation of the dominant speaker
-     * for the purposes of the REST API of Videobridge.
-     *
-     * @return a <tt>JSONObject</tt> which represents
-     * <tt>dominantSpeakerIdentification</tt> for the purposes of the REST API
-     * of Videobridge
+     * Re-calculates the list of endpoints in LastN order ({@link #endpointsInLastNOrder}) based on the speech activity
```
bbaldino, line 149: Can you document the semantics of the return value?

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -40,261 +41,135 @@
...
-            // DominantSpeakerIdentification only.
-            jsonObject = null;
-        }
-        else
+        synchronized (syncRoot)
         {
-            Conference conference = this.conference;
+            Map<Boolean, List<AbstractEndpoint>> bySendingVideo
+                    = endpointsBySpeechActivity.stream()
+                    .filter(Objects::nonNull)
```
bbaldino, line 158: How would we get null objects on this list?

src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -237,11 +230,14 @@ public Conference(Videobridge videobridge,
...
 
-        speechActivity = new ConferenceSpeechActivity(this);
+        speechActivity = new ConferenceSpeechActivity(new SpeechActivityListener());
         updateLastNEndpointsFuture = TaskPools.SCHEDULED_POOL.scheduleAtFixedRate(() -> {
             try
             {
-                lastNEndpoints.update();
+                if (speechActivity.updateLastNEndpoints())
+                {
+                    lastNEndpointsChanged();
```
bbaldino, line 239: Why not have this be an event defined in `ConferenceSpeechActivity.Listener` ?

So here we'd just do `speechActivity.updateLastNEndpoints()` and, if something changed, the event would be fired (instead of return a bool?)

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -40,261 +41,135 @@
...
-                        "The dominant speaker in conference "
-                            + conference.getID() + " is now the SSRC " + ssrc
-                            + ".");
-            }
+            return;
+        }
 
-            AbstractEndpoint endpoint
-                = conference.findEndpointByReceiveSSRC(ssrc);
+        Objects.requireNonNull(id);
```
bbaldino, line 118: annotate the function argument with `@NotNull`?

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -303,12 +178,12 @@ void expire()
...
-                dominantSpeakerIdentification
-                        .removeActiveSpeakerChangedListener(
-                                activeSpeakerChangedListener);
+                dominantSpeakerIdentification.removeActiveSpeakerChangedListener(activeSpeakerChangedListener);
             }
-            this.conference = null;
+            this.listener = null;
             this.dominantSpeakerIdentification = null;
+            endpointsBySpeechActivity.clear();
+            endpointsInLastNOrder = new ArrayList<>();
```
bbaldino, line 186: `Collections.emptyList()` ?

src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -433,10 +433,9 @@ else if (PacketExtensionsKt.looksLikeRtcp(packetInfo.getPacket()))
     }
 
     /**
-     * Notifies this {@code Endpoint} that the list of {@code Endpoint}s ordered
-     * by speech activity (i.e. the dominant speaker history) has changed.
+     * Notifies this {@code Endpoint} that the ordered list of {@code Endpoint}s changed.
      */
-    void speechActivityEndpointsChanged(List<String> endpoints)
+    void lastNEndpointsChanged(List<String> endpoints)
```
bbaldino, line 438: nit: `endpoints` -> `endpointIds` (or maybe, `orderedEndpointIds` ?)

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -97,53 +97,50 @@ public ConferenceSpeechActivity(Conference conference)
      * changed to one identified by a specific synchronization source
      * identifier/SSRC.
      * 
-     * @param ssrc the synchronization source identifier/SSRC of the new
+     * @param id the synchronization source identifier/SSRC of the new
      * active/dominant speaker
      */
-    private void activeSpeakerChanged(long ssrc)
+    private void activeSpeakerChanged(Object id)
```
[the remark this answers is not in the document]
bgrozev: Yeah, fixed in the latest commit

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -68,7 +68,7 @@
      * {@link #conference} with the dominant (speaker) <tt>Endpoint</tt> at the
      * beginning of the list i.e. the dominant speaker history.
      */
-    private final List<AbstractEndpoint> endpoints = new ArrayList<>();
+    private final List<String> endpoints = new ArrayList<>();
```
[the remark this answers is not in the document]
bgrozev: Yeah, but I ended up bringing it back. Some part of this logic needs to be Endpoint-aware in order to check which ones are sending video, and having both sides of the logic (speech activity and video availability) in one places ultimately makes things simpler.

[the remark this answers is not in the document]
bbaldino: Yeah, I saw that in the later commits...I think it makes sense.

src/main/java/org/jitsi/videobridge/Conference.java
bgrozev: I want updating the bitrate controllers to run in a separate task (it may send endpoint messages), but avoid creating the task if the list hasn't changed. I see now that I have some cases that lead to unnecessary tasks, so I'm making some changes.

src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
bgrozev: We can't as far as I can tell. It would be an illegal state. I removed the filter.

PR #1349, opened 2020-07-17
ref: Extracts the VideoPolicy from the VideoConstraints.

gpolitis:
This commit separates the _video policy_ from the _video constraints_.

Video constraints express the physicall constraints of a video source.
These are used from the bridge in both its video sending and receiving
function: When the bridge acts as a sender they're used to limit the
choice of layers to consider when performing bandwidth distribution.
When the bridge acts as a receiver, it feeds these constraints back to
the client so that the client can reduce its bitrate.

The video policy defines preferences and carries instructions on how to
perform bandwdith distribution, so they are irrelevant in the video
receiving function of the bridge. These instructions are internally
translated as additional constraints by the bandwidth distribution algo.

The motivation behind this commit is to clean-up the `VideoConstraints`
class from information that is irrelevant for the receive-side aspect of
the bridge.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -17,7 +17,6 @@ package org.jitsi.videobridge;
 
 import org.jetbrains.annotations.*;
 import org.jitsi.nlj.*;
-import com.google.common.collect.*;
 import org.jitsi.nlj.format.*;
 import org.jitsi.nlj.rtp.*;
 import org.jitsi.nlj.util.*;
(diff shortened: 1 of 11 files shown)


bbaldino:
Hey George, I'm afraid I'm having a hard time following this--maybe because I'm still stuck in my own frame of mind for the approach to this problem.  The way I see it we have, logically, 2 policies we want to implement:

1. Prefer quality of video over the amount of video streams seen.  This involves taking each source, in order of active speaker, and maximizing their forwarded resolution before moving to the next source (akin to "depth first").
1. Prefer the amount of video streams seen over the quality.  This involves taking the lowest resolution of every source, and only after making space for all those, doing another pass and "upgrading" the resolution of each source (akin to "breadth first").

When I look through this I'm having a hard time finding this logic in here--again, probably because I'm just stuck in my own frame of mind.  But do you see this PR as the groundwork for those policies?  Or perhaps I've misunderstood what we're trying to achieve.

gpolitis:
Thanks for the review @bbaldino! I'll mark this as a draft for now until we have a clearer picture of what we want to achieve here. I'll come back with some thoughts and answers to your comments when time permits.

PR #1350, opened 2020-07-17
Catch all exceptions from executors

bbaldino:
This is my take on https://github.com/jitsi/jitsi-videobridge/pull/1338.

I did not override all task submission functions for the classes, just the ones we use in practice, but could change that.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/util/SafeExecutor.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/util/SafeExecutor.kt
@@ -19,17 +19,22 @@ package org.jitsi.videobridge.util
 import org.jitsi.utils.logging2.LoggerImpl
 import org.json.simple.JSONObject
 import java.util.concurrent.ExecutorService
+import java.util.concurrent.Future
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.ScheduledFuture
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.LongAdder
 
-abstract class SafeExecutorService(protected val name: String) {
+/**
+ * Tracks the number of uncaught exceptions from tasks
+ */
+abstract class TaskServiceWrapper<T : ExecutorService>(
+    val name: String,
+    protected val delegate: T
+) {
     private val logger = LoggerImpl(name)
     private val numExceptions = LongAdder()
-
-    abstract val innerExecutor: ExecutorService
 
     protected fun wrapTask(task: Runnable): Runnable {
         return Runnable {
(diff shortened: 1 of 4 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/util/SafeExecutor.kt
```diff
@@ -0,0 +1,96 @@
...
+) {
+    private val logger = LoggerImpl(name)
+    private val numExceptions = LongAdder()
+
+    protected fun wrapTask(task: Runnable): Runnable {
+        return Runnable {
+            try {
+                task.run()
+            } catch (t: Throwable) {
+                logger.warn("Uncaught exception ($name): $t")
```
bgrozev, line 44: I don't think we need `name` here, since it's already passed as the name of the logger.

jvb/src/main/kotlin/org/jitsi/videobridge/util/SafeExecutor.kt
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * An [ExecutorService] which catches and logs any uncaught exceptions from tasks
+ */
+class SafeExecutor(
+    name: String,
+    delegate: ExecutorService
+) : ExecutorService by delegate, TaskServiceWrapper<ExecutorService>(name, delegate) {
+    private val logger = LoggerImpl(name)
```
bgrozev, line 73: Are these two vals used anywhere? They are also declared in `TaskServiceWrapper`

jvb/src/main/kotlin/org/jitsi/videobridge/util/SafeExecutor.kt
```diff
@@ -0,0 +1,96 @@
...
+    private val numExceptions = LongAdder()
+
+    override fun submit(task: Runnable): Future<*> = delegate.submit(wrapTask(task))
```
bgrozev, line 76: Can we wrap the whole ExecutorService API?

jvb/src/main/kotlin/org/jitsi/videobridge/util/SafeExecutor.kt
bbaldino: Nope, looks like these were left over.  Removed.

jvb/src/main/kotlin/org/jitsi/videobridge/util/SafeExecutor.kt
bbaldino: I thought more about it and I don't think it actually makes sense to wrap the others: the other methods return a future which will 'catch' the exception.

bbaldino:
We'll be writing a new `ExecutorService` wrapper which exposes a safe and unsafe api, closing this.

PR #1351, opened 2020-07-20
Port JVB configs to new metaconfig lib

bbaldino:


pom.xml
```diff
@@ -62,7 +62,7 @@
                 <version>${jitsi.utils.version}</version>
             </dependency>
             <dependency>
-                <groupId>com.github.jitsi</groupId>
+                <groupId>${project.groupId}</groupId>
```
JonathanLennox, line 65: Hm, in this pom you have a direct dependency on jitsi-metaconfig, whereas in jmt you don't (you just let the jicoco dependency pull it in recursively).

Either approach is probably okay, but the inconsistency seems odd?

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -670,7 +652,8 @@ private void startIce4j(
             }
 
             // Reload for all the new system properties to be seen
-            JitsiConfig.Companion.reload();
+            // TODO(brian): I don't think this reload is necessary?
+//            JitsiConfig.Companion.reload();
```
JonathanLennox, line 656: Figure this out, and remove it if it's really not necessary?

jvb/src/main/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfig.kt
```diff
@@ -16,193 +16,120 @@
...
-                "org.jitsi.videobridge.ENABLE_STATISTICS",
-                "videobridge.stats.enabled",
-                readOnce = true
-            )
-
-            private val enabledProp = ResettableSingleton { EnabledProperty() }
-
-            @JvmStatic
-            fun enabled() = enabledProp.get().value
+class NewStatsManagerBundleActivatorConfig {
```
JonathanLennox, line 30: Does this need to be named "New" still?

jvb/src/main/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfig.kt
```diff
@@ -16,193 +16,120 @@
...
+            else -> null
         }
     }
 }
 
 /**
  * Helper classes which model the config parameters for each stats transport
  */
-sealed class StatsTransportConfig(
+sealed class NewStatsTransportConfig(
```
JonathanLennox, line 123: Likewise this?

jvb/src/main/kotlin/org/jitsi/videobridge/websocket/config/WebsocketServiceConfig.kt
```diff
@@ -16,119 +16,49 @@
...
-                        is ConditionalPropertyConditionNotMetException -> throw t
-                        else -> null
-                    }
-                }
-            }
+    /**
+     * The domain name used in URLs advertised for COLIBRI WebSockets.
+     */
+    val domain: String by config {
+        onlyIf("Websocket are enabled", ::enabled) {
```
JonathanLennox, line 39: Grammar: "Websockets are" or "Websocket is"

jvb/src/main/resources/reference.conf
```diff
@@ -153,8 +153,8 @@ videobridge {
     resolve-remote-candidates = false
 
     # The nomination strategy to use for ICE.  THe accepted values are defined in ice4j's NominationStrategy:
-    # "NominateFirstValid", "NominateHighestPriority", "NominateFirstHostOrReflexiveValid", or "NominateBestRTT".
-    nomination-strategy = "NominateFirstValid"
+    # "NOMINATE_FIRST_VALID", "NOMINATE_HIGHEST_PRIORITY", "NOMINATE_FIRST_HOST_OR_REFLEXIVE_VALID", or "NOMINATE_BEST_RTT".
+    nomination-strategy = "NOMINATE_FIRST_VALID"
```
JonathanLennox, line 157: Hm?  Don't these strings come from the `NominationStrategy` enum in ice4j?  What's converting this from CamelCase to ALL_CAPS?

jvb/src/main/kotlin/org/jitsi/videobridge/cc/config/BandwidthProbingConfig.kt
```diff
@@ -16,41 +16,17 @@
...
-                    readOnce()
-                    name("org.jitsi.videobridge.DISABLE_RTX_PROBING")
-                    deprecated("RTX probing is always used when RTX is supported.")
-                }
-            )
-        }
+    /**
+     * How often we check to send probing data
+     */
+    val paddingPeriodMs: Long by config {
```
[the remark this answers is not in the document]
bbaldino: Agree on both counts :)

pom.xml
bbaldino: Nice catch...both JVB and JMT use functions from metaconfig directly, so I think they should be explicitly set the dependency...I must have just missed it in JMT.

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
bbaldino: Yeah, still need to verify that.

jvb/src/main/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfig.kt
bbaldino: Nope, fixed.

jvb/src/main/resources/reference.conf
bbaldino: The reason is that the old code retrieved as a `String` and used a `fromString` helper.  In the new code I can retrieve the enum type directly, but that uses the built-in `enumValueOf` which expected the value to match the enum _name_ itself, which is in all caps.

JonathanLennox: Hm.  The problem with that is that it breaks old config files, of course (including ours).  How hard would it be to make it use the `fromString` helper instead?

bbaldino: Well...lucky for us I happen to know a lib which can accommodate exactly such a breaking change!  (Though I don't know why I didn't think to address that).  I can configure the property like so:

```
val nominationStrategy: NominationStrategy by config {
    retrieve("videobridge.ice.nomination-strategy".from(JitsiConfig.newConfig))
    retrieve("videobridge.ice.nomination-strategy".from(JitsiConfig.newConfig).asType<String>().andCovertyBy(NominationStrategy::FromString))
}
```

Will add a test to verify that works as expected.

bbaldino: (And same for `keep-alive-strategy`)

JonathanLennox: I'd be inclined to think we just need the `andType`...`andConvertBy` flavor, but if you want both, that's fine.

Doesn't the converted one need to be first, because the unconverted one will pull from the defaults file?

bbaldino: The merge of the defaults file and the overrides file is already done before this, so both of them are just trying to parse the 'final' config value in their own way.  

I did have to move the "parse as enum" one afterwards, though, because I give a rich error message when an enum value is bad that I bubble all the way up (i.e. it doesn't fall-through to another supplier) to help catch typos, etc, so if it was the `NominateFirstValid` style, then we'd throw an exception rather than fall through.

However, after changing the order, I encountered a strange situation (kotlin bug?) where, even though I have the type specified as a non-nullable one, the `KeepAliveStrategy` value, because it's a platform type coming from Java, allows 'null' to bubble all the way up, meaning if it's a new-style value (e.g. `NOMINATE_FIRST_VALID`), 
```kotlin
retrieve("videobridge.ice.nomination-strategy".from(JitsiConfig.newConfig).asType<String>().andCovertyBy(NominationStrategy::FromString))
```
will successfully return a value of `null` for the property (won't fall through).  I was relying on this throwing an exception, but weirdly it doesn't.  I think I'll revert back to the old style (parse as `String`, use `fromString`) only for this and track a bug in jitsi-metaconfig to look into the other issue further.

bbaldino:
jenkins, test this please

jvb/pom.xml
```diff
@@ -284,7 +288,7 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.jitsi</groupId>
+        <groupId>${project.groupId}</groupId>
```
bgrozev, line 291: nit: extra whitespace

jvb/src/main/java/org/jitsi/videobridge/ice/Harvesters.java
```diff
@@ -83,7 +81,7 @@ public static void initializeStaticConfiguration()
 
 
             singlePortHarvesters
-                    = SinglePortUdpHarvester.createHarvesters(Config.port());
+                    = SinglePortUdpHarvester.createHarvesters(IceTransport.config.getPort());
```
bgrozev, line 84: Feels weird that `Harvesters` accesses the config via `IceTransport`. Maybe move `IceTransport.config` to `IceConfig.config`?

jvb/src/main/kotlin/org/jitsi/videobridge/VideobridgeConfig.kt
```diff
@@ -0,0 +1,42 @@
...
+            .convertFrom<String> {
+                if (it.isNotBlank()) {
+                    try {
+                        Pattern.compile(it)
+                    } catch (e: PatternSyntaxException) {
+                        throw ConfigException.UnableToRetrieve.WrongType(
+                            "shutdownSourceRegex expected valid regex pattern: $e")
+                    }
+                } else {
+                    throw ConfigException.UnableToRetrieve.NotFound("not found")
```
bgrozev, line 38: Why throw an exception if the value is optional? I see the `optionalconfig` will catch it and return `null`, but if the value is blank it feels more natural to return null right here.

jvb/src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -16,151 +16,75 @@
...
+        // of validating the other properties in the legacy config
+        // file.  If neither property is present, we consider the field
+        // "not found" in the legacy config.
+        "Legacy Octo relay enabled" {
+            if (legacyBindAddress == null && legacyBindPort == null) {
+                throw ConfigException.UnableToRetrieve.NotFound("not found")
             }
+            legacyBindAddress != null && legacyBindPort?.isUnprivilegedPort() == true
+        }
+        "videobridge.octo.enabled".from(JitsiConfig.newConfig)
```
bgrozev, line 52: Just curious if I understand this correctly: if the legacy bind address or port are unset, the first block will throw `NotFound`, so we'll retrieve the new property?

jvb/src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -16,151 +16,75 @@
...
-            private val bindPortProp = BindPortProperty()
+    val region: String? by optionalconfig {
+        "org.jitsi.videobridge.REGION".from(JitsiConfig.legacyConfig)
+        "videobridge.octo.region".from(JitsiConfig.newConfig)
+    }
 
-            @JvmStatic
-            fun bindPort(): Int = bindPortProp.value
+    val bindAddress: String by config {
+        "bind address from legacy config" { legacyBindAddress!! }
```
bgrozev, line 61: And here, since we use `by config` rather than `by optionalconfig`, if `legacyBindAddress == null` the first retrieval will fail, so we fallback to the new property?

jvb/src/main/kotlin/org/jitsi/videobridge/VideobridgeConfig.kt
bbaldino: We can't return null from a supplier, they expect either a non-null value or an exception...the previous code specifically treated a blank string as the lack of a regex, so we do the same thing here by equating a blank string to 'not found'

jvb/src/main/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfig.kt
```diff
@@ -16,141 +16,97 @@
...
+    private fun Map<String, String>.toStatsTransportConfig(): List<StatsTransportConfig> {
+        val transportTypes =
+            this["org.jitsi.videobridge.STATISTICS_TRANSPORT"]?.split(",") ?: return listOf()
+        return transportTypes.mapNotNull { transportType ->
+            val interval = this["org.jitsi.videobridge.STATISTICS_INTERVAL.$transportType"]?.let {
+                Duration.ofMillis(it.toLong())
+            } ?: this@StatsManagerBundleActivatorConfig.interval
+            when (transportType) {
+                "muc" -> StatsTransportConfig.MucStatsTransportConfig(interval)
+                "callstats.io" -> StatsTransportConfig.CallStatsIoStatsTransportConfig(interval)
```
bgrozev, line 94: You use "callstats.io" here and "callstatsio" below. Is it intentional?

jvb/src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
bbaldino: Yeah...we didn't have an 'enabled' property before, the code itself checked that both bind address and port were set, so that's what's being replicated here.  (also, if the port is set but isn't an 'unprivileged port', it'll fall through)

jvb/src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
bbaldino: Right

jvb/src/test/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfigTest.kt
```diff
@@ -1,129 +1,123 @@
-/*
```
bgrozev, line 1: Removed the licence header?

jvb/src/main/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfig.kt
bbaldino: Looks like this is how it was in master...don't know why I changed it to `callstatsio` in new config but looks like I did.  I may have considered using the name as 'keys' at one point within an object, which would mean they couldn't have dots. 🤷‍♂️

jvb/src/main/java/org/jitsi/videobridge/ice/Harvesters.java
bbaldino: Sure

jvb/src/test/kotlin/org/jitsi/videobridge/stats/config/StatsManagerBundleActivatorConfigTest.kt
bbaldino: Weird. Fixed.

bbaldino:
The test failures are known issues with torture tests (which have passed previously), so merging anyway.

PR #1357, opened 2020-07-25
synchronize endpoint expiry and endpoint initialization

craigmj:
We're running into routine crashes where we get `null` for a localEndpoint in `ContentShim.createRtpChannel`. This patch ensures that `ConferenceShim.ensureEndpointCreated` does not overlap with endpoint expiry, which seems to be the cause of our issue.


craigmj:
Thanks, I'll have a look, but right now, I think this patch is too simplistic. Oh, I think I'm talking to Jenkins!

PR #1360, opened 2020-07-30
Use SafeExecutor.

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -418,24 +418,6 @@ public void describeShallow(ColibriConferenceIQ iq)
...
-    {
-        TaskPools.IO_POOL.submit(() ->
-        {
-            try
-            {
-                lastNEndpointsChanged();
-            }
-            catch (Exception e)
-            {
-                logger.warn("Failed to handle change in last N endpoints: ", e);
```
bbaldino, line 434: Do we want to get rid of this logging?  I thought this was the sort of stuff we planned on going through and _adding_ wherever it was missing?

bgrozev: I can add the logging back, but if all we do in case of an exception is log a message, then we'd get the same result with the "safe" wrapper.

bbaldino: Yeah but we lose the context in the log message, right? I.e. `Failed to handle change in last N endpoints` ?

PR #1361, opened 2020-07-30
Remove unused deps

bbaldino:
From what I can tell, these deps can be safely removed.  Some of them are no longer used at all, and some are still in the dependency tree but not used directly by JVB.


bgrozev:
LGTM

bbaldino:
Failure here is a spotty torture test, merging

PR #1368, opened 2020-08-03
Minor config clean up

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/cc/config/BitrateControllerConfig.kt
```diff
@@ -25,17 +25,18 @@ class BitrateControllerConfig {
...
-         * last bandwidth estimation.
+         * In order to limit the resolution changes due to bandwidth changes we only react to bandwidth changes greater
+         * than {@code bweChangeThreshold * last_bandwidth_estimation}.
          */
-        private val bweChangeThresholdPct: Int by config {
+        private val bweChangeThreshold: Double by config {
             "org.jitsi.videobridge.BWE_CHANGE_THRESHOLD_PCT".from(JitsiConfig.legacyConfig)
+                .convertFrom<Int> { it / 100.0 }
             "videobridge.cc.bwe-change-threshold-pct".from(JitsiConfig.newConfig)
+                .convertFrom<Int> { it / 100.0 }
```
JonathanLennox, line 35: Would it work to read the percentage as a double? It seems like it'd be a backward-compatible change to allow people to set their threshold to 2.5% or whatever.

bgrozev: I added a new property without the `-pct` (keeping the existing for backward compatibility). Unfortunately it looks like typesafe doesn't parse `5%` as a number:
`videobridge.cc.bwe-change-threshold-pct has type STRING rather than NUMBER`

JonathanLennox: Yeah, I meant that they could literally type "2.5" (without the % sign).

bgrozev: Now they can do either
`videobridge.cc.bwe-change-threshold-pct=2.5` or
`videobridge.cc.bwe-change-threshold=0.025`

JonathanLennox: Can they?  The code is still doing `convertFrom<Int>`

bgrozev: Ah, I missed your point. Fixed now.

PR #1369, opened 2020-08-03
Upgrade kotlintest -> kotest

bbaldino:


jvb/pom.xml
```diff
@@ -17,6 +17,7 @@
   <url>https://github.com/jitsi/jitsi-videobridge</url>
   <properties>
     <assembly.skipAssembly>false</assembly.skipAssembly>
+    <kotest.version>4.1.2</kotest.version>
```
bgrozev, line 20: Any reason not to use 4.1.3?

jvb/pom.xml
```diff
@@ -285,12 +286,26 @@
...
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>io.kotlintest</groupId>
-      <artifactId>kotlintest-runner-junit5</artifactId>
-      <version>3.4.2</version>
-      <scope>test</scope>
-    </dependency>
+	<dependency>
+		<groupId>io.kotest</groupId>
```
bgrozev, line 290: The rest of the file uses 2 spaces for indentation

jvb/pom.xml
bbaldino: I looked [here](https://github.com/kotest/kotest) and saw latest release as 4.1.2 so used that...I'll change to 4.1.3

jvb/pom.xml
bbaldino: Tab snuck in there.

PR #1372, opened 2020-08-05
update deps

bbaldino:


JonathanLennox:
For the dependencies you removed, you only removed them from the `dependencyManagement` section in the top-level pom, not from the individual dependencies in the sub-modules.  Thus the Maven failure.

JonathanLennox:
Looks like you still missed the jitsi-utils-kotlin ref in jvb-api-client.

pom.xml
```diff
@@ -78,7 +68,7 @@
             <dependency>
                 <groupId>${project.groupId}</groupId>
                 <artifactId>rtp</artifactId>
-                <version>1.0-46-gd3742d1</version>
+                <version> 1.0-49-gb097255</version>
```
JonathanLennox, line 71: Spurious space in xml.

PR #1374, opened 2020-08-06
port rest api to new config

bbaldino:
I bumped into this when updating AbstractJettyBundleActivator to new config.  RESTBundleActivator was still using an old config, so this PR removes that usage.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/rest/RESTBundleActivator.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/rest/RESTBundleActivator.java
@@ -15,8 +15,6 @@
  */
 package org.jitsi.videobridge.rest;
 
-import java.util.*;
-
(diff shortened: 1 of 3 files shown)


JonathanLennox:
The `System.setProperty(Videobridge.REST_API_PNAME)` in Main.java still works with this?  And is that still the way you want to implement this?

bbaldino:
> The `System.setProperty(Videobridge.REST_API_PNAME)` in Main.java still works with this? And is that still the way you want to implement this?

It does because typesafe config allows overriding keys with system properties.  The other option is one I pursued before which is where we have a command-line-args config source, but that's a lot of work and given this is our last command-line arg (which will go away eventually) I don't think it's worth it.  I probably wouldn't implement a property like this from scratch, but as a way to 'shim' this value into the new config I don't think it's terrible.

PR #1375, opened 2020-08-06
adapt to change in AbstractJettyBundleActivator

bbaldino:
See https://github.com/jitsi/jicoco/pull/99

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/rest/RESTBundleActivator.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/rest/RESTBundleActivator.java
@@ -54,7 +54,7 @@ public class RESTBundleActivator
      */
     public RESTBundleActivator()
     {
-        super(JETTY_PROPERTY_PREFIX);
+        super(JETTY_PROPERTY_PREFIX, "videobridge.rest.private");
     }
 
     /**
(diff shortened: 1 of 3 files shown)


JonathanLennox:
Do you want to update the docs to reference the new config syntax?

JonathanLennox:
Also I think this will need the corresponding jicofo change in order for the PR tests to pass.

bbaldino:
> Also I think this will need the corresponding jicofo change in order for the PR tests to pass.

Yeah I ended up adding a backwards-compatible constructor to AbstractJettyBundleActivator

bbaldino:
> Also I think this will need the corresponding jicofo change in order for the PR tests to pass.

I think I will wait on that until more of these are ported over.  The old values will still work.

JonathanLennox:
The backward-compatible constructor isn't sufficient to get jicofo to compile -- because you removed `AbstractJettyBundleActivator.getDefaultPort()` and 'getDefaultTlsPort()`, the overrides in `AuthBundleActivator` are no longer overrides and the compiler errors out.

bbaldino:
> The backward-compatible constructor isn't sufficient to get jicofo to compile -- because you removed `AbstractJettyBundleActivator.getDefaultPort()` and `getDefaultTlsPort()`, the overrides in `AuthBundleActivator` are no longer overrides and the compiler errors out.

Hmph, ok I guess I'll update Jicofo too then.

bbaldino:
jenkins, test this please

PR #1377, opened 2020-08-10
fix jetty config prefix

bbaldino:


bbaldino:
Nevermind, this isn't actually broken on the JVB side, it's just that I didn't have the config generated correctly.

PR #1380, opened 2020-08-11
Tweak debug API and add docs

bgrozev:


bgrozev:
I wonder whether it's too late to move `debug` out of `/colibri`. I think the only place this is used is in the PR testing scripts, and that's easy to update. @bbaldino thoughts?

bbaldino:
> I wonder whether it's too late to move `debug` out of `/colibri`. I think the only place this is used is in the PR testing scripts, and that's easy to update. @bbaldino thoughts?

I dunno, but it would definitely be nice.  Maybe we can move it and put a redirect in place of the old URL for a time?

bgrozev:
It's just the pre-join test that is failing

bbaldino:
> Above we test classes under `rest.root.colibri` (or am I missing something?).

You're right...but any reason not to test them all in one place?

bgrozev:
> You're right...but any reason not to test them all in one place?

Well, for `debug` I can verify the value of the annotation (that they use the`DEBUG` config property and not something else). For `colibri` I can't do the same, because `Shutdown` is under the `SHUTDOWN` property (which is only enabled if colibri is enabled). So for `colibri` I only check that an annotation exists.

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -745,7 +724,6 @@ public OrderedJsonObject getDebugState(String conferenceId, String endpointId, b
         debugState.put("time", System.currentTimeMillis());
 
         debugState.put("health", getHealthStatus());
-        debugState.put("e2e_packet_delay", Endpoint.getPacketDelayStats());
```
bbaldino, line 748: Did you mean to remove this?

bgrozev: Yes. It is still available under `/debug/stats/transit-stats`. I don't think this is a good place for it, because it makes the output of `/debug` quite long (and I use `/debug` all the time to get the conference IDs).

PR #1381, opened 2020-08-12
wip: Effective constraints & aggressive layer suspension.

gpolitis:


gpolitis:
@bbaldino @bgrozev I think you guys where the most familiar with the video constraints work. Could you please take a look?

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -170,19 +168,18 @@
...
      * The map of endpoint id to video constraints that contains the video
      * constraints to respect when allocating bandwidth for a specific endpoint.
      */
     private ImmutableMap<String, VideoConstraints> videoConstraintsMap = ImmutableMap.of();
 
+    /**
+     * A modified copy of the original video constraints map, augmented with video constraints for the endpoints that
+     * fall outside of the last-n set + endpoints not announced in the videoConstraintsMap.
+     */
+    private Map<String, VideoConstraints> effectiveConstraintsMap = Collections.EMPTY_MAP;
```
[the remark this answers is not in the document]
gpolitis: Sure, so all the client sends is who's selected and max resolution MAXp. In my past work I pass these messages through a compatibility layer to compute the video constraints: for every selected endpoint it creates a video constraints instance of max resolution MAXp (the actual computation is a tiny bit more complicated than that because the compat. layer also takes into account tile-view). The video constraints computed by the compatibility layer end-up in `videoConstraintsMap`.

Before the bitrate allocation algorithm runs, we need video constraints for all endpoints, i.e. contraints for the thumbnails (video constraints of 180p) and for the people that are outside of last-n (video constraints of 0p). These are the _effective_ video constraints.

gpolitis:
> George, I'm struggling to review these as I don't think I have a good picture of how the constraints stuff works. Could you write a little something up on the general design maybe?

Would a constraints.md in the docs folder work?

jvb/src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -616,20 +616,23 @@ public void addRtpExtension(RtpExtension rtpExtension)
...
 
-        // Endpoints that "this" no longer cares about what it receives.
+        // Sources that "this" endpoint no longer cares about what it receives.
         for (String id : removedEndpoints)
         {
             AbstractEndpoint senderEndpoint = getConference().getEndpoint(id);
             if (senderEndpoint != null)
             {
-                senderEndpoint.removeReceiver(getID());
+                senderEndpoint.addReceiver(getID(), VideoConstraints.thumbnailVideoConstraints);
```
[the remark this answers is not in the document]
gpolitis: I can no longer recall the reason why I changed this, it seems like it should be left as it was. I'll do some testing and either revert or document why it's needed.

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -686,9 +677,11 @@ private synchronized void update()
             }
         }
 
+        Map<String, VideoConstraints> copyOfVideoConstraintsMap = videoConstraintsMap;
```
bgrozev, line 680: Did you intend to make a copy of the map here?

gpolitis: Just the reference, not a deep copy.

jvb/src/main/java/org/jitsi/videobridge/Endpoint.java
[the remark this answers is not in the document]
gpolitis: Did some testing and I think this was a left-over of some other intermediate change that I had locally. I rolled this change back. Thanks for noticing.

jvb/src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -616,20 +616,23 @@ public void addRtpExtension(RtpExtension rtpExtension)
...
 
+    public void effectiveVideoConstraintsChanged(
+        ImmutableMap<String, VideoConstraints> oldVideoConstraints,
+        ImmutableMap<String, VideoConstraints> newVideoConstraints)
+    {
         Set<String> removedEndpoints = new HashSet<>(oldVideoConstraints.keySet());
         removedEndpoints.removeAll(newVideoConstraints.keySet());
 
-        // Endpoints that "this" no longer cares about what it receives.
+        // Sources that "this" endpoint no longer cares about what it receives.
```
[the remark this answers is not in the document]
gpolitis: Actually, these are the sources that destination endpoint no longer receives, based on the _effective_ video constraints. I've corrected the comment to reflect this.

gpolitis:
> > George, I'm struggling to review these as I don't think I have a good picture of how the constraints stuff works. Could you write a little something up on the general design maybe?
> 
> Would a constraints.md in the docs folder work?

Added.

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bgrozev: The only place you use the "copy" is on the line below. I think it should be removed.

Also the parameters for `allocate`, `prioritize` and `makeEndpointMultiRankList` are named `copyOfVideoConstraintsMap`, but it's not clear what it's a copy of (and indeed it isn't a copy of anything)

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -1064,6 +1058,27 @@ private synchronized void update()
...
+        List<EndpointMultiRank> endpointMultiRankList = new ArrayList<>(conferenceEndpoints.size());
+        for (int i = 0; i < conferenceEndpoints.size(); i++)
+        {
+            AbstractEndpoint endpoint = conferenceEndpoints.get(i);
+
+            VideoConstraints effectiveVideoConstraints = (i < adjustedLastN || adjustedLastN < 0)
+                ? copyOfVideoConstraintsMap.getOrDefault(endpoint.getID(), VideoConstraints.thumbnailVideoConstraints)
+                : VideoConstraints.disabledVideoConstraints;
+
+            int rank = conferenceEndpoints.indexOf(endpoint);
```
bgrozev, line 1075: `rank` is just `i`, isn't it?

doc/constraints.md
```diff
@@ -0,0 +1,88 @@
...
+a max resolution/frame-rate.
+
+###### Message format to the bridge
+
+An endpoint can request different qualities for different endpoints, so it sends
+a map of endpoint ids -> video constraints back to the bridge. In JSON notation
+the message can be represented like this:
+
+    [{
+      'id': endpoint-1,
```
bgrozev, line 15: Is this what the client sends? Where is the code?

In the bridge I see `ReceiverVideoConstraintsMessage` using a different format (without `preferred` properties)

doc/constraints.md
```diff
@@ -0,0 +1,88 @@
...
+      'constraints': {
+        'idealHeight': resolution,
+        'preferredHeight': resolution,
+        'preferredFps': frame-rate
+      },
+      ...
+    }]
+
+The preferredHeight and preferredFps properties are additional parameters that
```
bgrozev, line 24: I struggle to understand the semantics of `preferredHeight` and `preferredFps`. What does `idealHeight=720, preferredHeight=360` mean? Ideally I want 720, but I prefer 360?

Following the code I find that:
1. `prefferedHeight` is not set by the client using `ReceiverVideoConstraintsMessage` (which is in fact not handled at all?).
2. It is set to `VideoConstraintsCompatibility` to:
a) The value from the bridge config for selected endpoints when there's more than one selected endpoint
b) -1 for everything else

It is used for:
1. Ranking endpoints, with higher `preferredHeight` being ranked higher (it is essentially a boolean pushing the selected endpoints up the list (but only when more than one is selected))
2. Allocating bitrate where it sets the level of the first round of allocation (i.e. we first fill the `preferred` value for every endpoint before improving any above that).


I think this is a very confusing concept, and we should fix that while this is still only in the bridge. The compatibility layer attempts to hide the complexity of "stage view" vs "tile view", but instead of providing a simple concept ("these endpoints need to go in the top of the list because they are special (selected)") it sets this `preferredHeight` value that then gets interpreted as a boolean in the ranking.

I think we should:
1. Separate the use of `preferred` for ranking vs as a floor for the first stage of allocation.
2. Use meaningful names. 

Perhaps a model with "min", "max" and "selected" makes sense. I.e. the client provides a list of "selected" endpoints for which selection should be prioritized. For each endpoint it also gives a "max" resolution and a "min" resolution. The semantics of "max" is clear, while "min" means "satisfy the min of everyone before allocating above it".

All that being said, I don't expect any changes in this PR other than making the documentation accurate (`preferred` is very confusing, is not explained anywhere, and after carefully reading the code for a while I'm still not sure I understand it).

bgrozev:
OK, I think I understand this now. The main result should be that when deciding how to restrict a sender we take into account last-n (i.e. a sender that is not forwarded anywhere *due to last-n* will be sent a constraint with height=0). 

Is this "aggressive layer suspension"? Are there other changes that I missed?

LGTM apart from the minor comments I left

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: Ah, yes, good point!

doc/constraints.md
gpolitis: Thanks Boris, that's a good point, I'll make the JSON description consistent.

doc/constraints.md
gpolitis: I totally agree it's confusing and it therefore needs to be fixed while it's still contained in the bridge. Do you see this as part of this PR ?

I'd argue we should keep/have/continue this discussion in the video policy refactoring PR [here](https://github.com/jitsi/jitsi-videobridge/pull/1349) that I opened a few weeks ago to tackle exactly this. @bbaldino also has some interesting ideas how to model things and @JonathanLennox can probably chime in.

This PR only meant to provide "aggressive" layer suspension (i.e. disabling the video for participants that are outside of the last-n set). "Effective video constraints" is the means to achieve just that.

gpolitis:
> Is this "aggressive layer suspension"? Are there other changes that I missed?

That's all. "Effective video constraints" is the means to achieve just that.

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
gpolitis: That's a good point! I've simplified the code a bit.

doc/constraints.md
bgrozev: > I totally agree it's confusing and it therefore needs to be fixed while it's still contained in the bridge. Do you see this as part of this PR ?

No, as I said I think we should just fix the documentation. The situation right now is this:
1. We deprecated the selected/pinned/maxResolution signaling, and we support them via a "compatibility" layer.
2. We introduced a new way to describe the constraints with the "ideal/preferred", which we both agree is confusing and shouldn't be used.
3. The new way is only partially supported by the bridge (it handles only "ideal").
4. The new way is not supported by the client.

This PR add documentation about the new way, describing parts of it that are not implemented anywhere. I think this only adds to the confusion.

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -275,8 +275,8 @@ class ReceiverVideoConstraintMessage(val maxFrameHeight: Int) : BridgeChannelMes
  * {
  *   "colibriClass": "ReceiverVideoConstraintsChangedEvent",
  *   "videoConstraints": [
- *     { "id": "abcdabcd", "idealHeight": 180 },
- *     { "id": "12341234", "idealHeight": 360 }
+ *     { "id": "endpoint-1", "idealHeight": 180 },
+ *     { "id": "endpoint-2", "idealHeight": 360, "preferredHeight": 360, "preferredFps": 30.0 }
```
bgrozev, line 279: The existing code does not support this

doc/constraints.md
gpolitis: Thanks for explaining Boris, I've updated the documentation to reflect reality a bit better.

PR #1382, opened 2020-08-12
Port shutdown source pattern config to new config

bbaldino:


bbaldino:
Believe this is now obsolete.  https://github.com/jitsi/jitsi-videobridge/pull/1380 removed support for shutdown over XMPP.

PR #1383, opened 2020-08-12
change octorelayservice to not be an osgi service

bbaldino:
There wasn't much preventing `OctoRelayService` from becoming a 'regular' class instead of an OSGi service, so this PR moves it away from an OSGi service.

Although I didn't do it here, it wouldn't be too bad to plumb in an `OctoRelayServiceSupplier` instance to the classes that need it which would prevent them accessing a global singleton.  Neither of those classes currently have tests so I didn't bother doing that here.


bbaldino:
Failures here are spotty torture tests

PR #1384, opened 2020-08-13
Refactor EpConnectionStatus into non-osgi service

bbaldino:
This refactors/rewrites EpConnectionStatus and makes it a regular class, not an OSGi service.  I also thought this made more sense to have at the conference level anyway, so I scoped an instance to a conference.  A couple notes:

1. The old service used the PeriodicRunnable scheduler, the new one uses a normal scheduler.  It is worth noting that, since these are now per-conference, we'll have more of them running.  The current interval is 500ms, which seems more frequent than it need be, so we should perhaps consider changing that.
1. The old code handled a case where an endpoint was replaced (new Endpoint instance with an existing ID) and, if that endpoints was previously inactive, the code sent out notifications that it is active.  I didn't do that in the new code because I thought it made more sense to wait until the new endpoint became active to do so (we don't normally assume endpoints inactive when they first join, but in this cast the endpoint _was_ there and _was_ inactive, so it seems appropriate?). EDIT: I talked this over with @paweldomas and he thinks this new approach makes sense.


jvb/src/main/resources/reference.conf
```diff
@@ -29,6 +29,9 @@ videobridge {
     # How long an endpoint can be 'inactive' before it will
     # be considered disconnected
     max-inactivity-limit=3 seconds
+
+    # How often we check endpoint's connectivity status
+    check-interval=500 milliseconds
```
paweldomas, line 34: is this where the default value comes from?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitor.kt
```diff
@@ -0,0 +1,141 @@
...
+    private val taskHandle = AtomicReference<ScheduledFuture<*>>(null)
+
+    fun start() {
+        if (taskHandle.compareAndSet(
+                null,
+                executor.scheduleWithFixedDelay(::run, config.intervalMs, config.intervalMs, TimeUnit.MILLISECONDS))
+        ) {
+            logger.info("Starting connection status monitor")
+        } else {
+            logger.warn("Task already started, not starting again")
```
paweldomas, line 52: is this check just in case or is it normal to call start more than once?

jvb/src/main/resources/reference.conf
bbaldino: Yeah

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitor.kt
bbaldino: Not normal, just a just-in-case check

bbaldino:
> Have tried it with jitsi-meet?

No, not yet.  Do any torture tests validate this?

paweldomas:
> No, not yet. Do any torture tests validate this?

No, they don't. One way for testing it is to kill the participant or maybe have another laptop and disconnect it from the network.

bbaldino:
> > No, not yet. Do any torture tests validate this?
> 
> No, they don't. One way for testing it is to kill the participant or maybe have another laptop and disconnect it from the network.

Ok, thanks...I will do some testing there.

bbaldino:
> > > No, not yet. Do any torture tests validate this?
> > 
> > 
> > No, they don't. One way for testing it is to kill the participant or maybe have another laptop and disconnect it from the network.
> 
> Ok, thanks...I will do some testing there.

Did some testing with real calls and cutting wifi and things looked good.

PR #1395, opened 2020-08-21
Suppliers cleanup

bbaldino:
Cleans up the service suppliers to be more consistent and adapts to changes in https://github.com/jitsi/jicoco/pull/109


jvb/src/main/kotlin/org/jitsi/videobridge/service/ServiceSupplier.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jitsi.videobridge.service
+
+interface ServiceSupplier<T> {
```
bgrozev: Why is this needed in addition to `java.util.function.Supplier`?

jvb/src/main/kotlin/org/jitsi/videobridge/rest/binders/ServiceBinder.kt
```diff
@@ -29,5 +33,7 @@ class ServiceBinder : AbstractBinder() {
         bind(videobridgeSupplierSingleton).to(VideobridgeSupplier::class.java)
         bind(statsManagerSupplierSingleton).to(StatsManagerSupplier::class.java)
         bind(clientConnectionImplSupplierSingleton).to(ClientConnectionImplSupplier::class.java)
+        bind(healthCheckServiceSupplierSingleton).to(HealthCheckServiceSupplier::class.java)
+        bind(singleton).to(VersionServiceSupplier::class.java)
```
bgrozev, line 37: Rename `singleton` to `versionServiceSupplierSingleton`?

bbaldino: Yup, missed this one.  I've also since discovered that the bare name `singleton` isn't _great_ as you can then only have one per package in kotlin.  I'll either rename the import here or rename the singleton instances (the only reason I didn't rename the singleton instances from the get-go is that it's awkward from Java since you have to do `VersionServiceSupplierKt.versionServiceSupplierSingleton`)

jvb/src/main/kotlin/org/jitsi/videobridge/service/ServiceSupplier.kt
bbaldino: This was a leftover from a prior implementation and was removed in later commits

PR #1397, opened 2020-08-24
Enforce an interval at which BitrateController#update runs

bbaldino:
This PR adds a configurable interval at which we'll force a call to `BitrateController#update` if it has not run.  

I'll note that the scheduling of `update` tasks is not foolproof with regards to potential duplicates: to address this I think we'd probably implement a _minimum_ time between updates as well.  I can add that if we think it would be worthwhile, but we don't currently gate calls to `update` and this didn't seem like it would add a significant increase.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
@@ -220,7 +220,7 @@ public class BitrateController
     /**
      * The last time {@link BitrateController#update()} was called
      */
-    private Instant lastUpdateTime = Instant.MAX;
+    private Instant lastUpdateTime = Instant.MIN;
 
     /**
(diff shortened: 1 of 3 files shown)


jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -212,23 +215,40 @@
     private final AtomicInteger numDroppedPacketsUnknownSsrc =
         new AtomicInteger(0);
 
+    private final Clock clock;
+
+    /**
+     * The last time {@link BitrateController#update()} was called
+     */
+    private Instant lastUpdateTime = Instant.MAX;
```
JonathanLennox, line 223: Should be `Instant.MIN`, no?

bbaldino: Indeed. Thanks.

PR #1398, opened 2020-08-25
JVB-wide last-n value

bbaldino:
This is another step in survival-last-n: it implements a last-n setting on the JVB level which will be applied to all endpoints.  Endpoints do not just take this JVB value, but instead the minimum of any last-n setting they have at the endpoint level and this JVB setting.  Soon, the load manager will leverage this field to reduce load on the bridge when needed.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/JvbLastN.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/JvbLastN.kt
@@ -40,8 +40,6 @@ val jvbLastNSingleton: JvbLastN = JvbLastN()
 fun calculateLastN(lastN1: Int, lastN2: Int): Int {
     return if (lastN1 != -1 && lastN2 != -1) {
         minOf(lastN1, lastN2)
-    } else if (lastN1 == -1 && lastN2 == -1) {
-        -1
     } else {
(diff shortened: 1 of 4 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/JvbLastN.kt
```diff
@@ -0,0 +1,48 @@
...
+    override fun get(): Int = jvbLastN
+}
+
+@JvmField
+val jvbLastNSingleton: JvbLastN = JvbLastN()
+
+fun calculateLastN(lastN1: Int, lastN2: Int): Int {
+    return if (lastN1 != -1 && lastN2 != -1) {
+        minOf(lastN1, lastN2)
+    } else if (lastN1 == -1 && lastN2 == -1) {
```
bgrozev, line 43: It should give the same result without this line

bgrozev (dismissed):

jvb/src/main/kotlin/org/jitsi/videobridge/JvbLastN.kt
bbaldino: I think you're right. Will double check and remove it.

PR #1399, opened 2020-08-25
Attempt to fix MediaSourceFactoryTest flakiness

bbaldino:
`MediaSourceFactoryTest` has been flaky.  I suspect this may stem from our use of the `Whitebox` lib to modify a static member.  It turns out we don't need this static member, so this PR removes it and adapts the tests (as well as some other cleanup).

### DIFF
--- a/jvb/src/test/java/org/jitsi/videobridge/xmpp/MediaSourceFactoryTest.java
+++ b/jvb/src/test/java/org/jitsi/videobridge/xmpp/MediaSourceFactoryTest.java
@@ -61,7 +61,7 @@ public class MediaSourceFactoryTest
         verifyAll();
     }
 
-    // 1 video stream -> 1 source, 1 encoding
+    // 1 video stream -> 1 source, 1 layer
     @Test
     public void createMediaSource()
     {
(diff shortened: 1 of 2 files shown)


jvb/src/test/java/org/jitsi/videobridge/xmpp/MediaSourceFactoryTest.java
```diff
@@ -109,25 +99,20 @@ public void createMediaSources1()
...
+                Arrays.asList(videoSource, rtx), Collections.singletonList(rtxGroup));
 
         assertNotNull(sources);
         assertEquals(1, sources.length);
         MediaSourceDesc source = sources[0];
         assertEquals(1, source.numRtpLayers());
     }
 
-    // 3 sim streams, 3 rtx -> 1 source, 3 encodings
+    // 3 sim streams, 3 rtx -> 1 source, 9 encodings
```
JonathanLennox, line 110: Should say "layers" not "encodings"

bbaldino: Any other places that should be changed?

bbaldino: Should all uses of "encodings" be changed to "layers" in this file?

JonathanLennox: Yes, probably, unless they're specifically referring to the encoded stream (i.e. all packets with a given ssrc).

PR #1400, opened 2020-08-25
Conference expiration cleanup

bbaldino:
There was some weirdness with the use of `Expireable` and `ExpireableImpl` in Conference I noticed when adding support for firing conference creation/expiration events in `Videobridge`: `Videobridge#expireConference` calls `Conference#expire` which calls `Videobridge#expireConference`.  From what @bgrozev and I could figure out, there were some previous attempts to either prevent race conditions or prevent `VideobridgeExpireThread` from blocking, and the idea was maybe that `Conference#expire` should be private and everything should call `Conference#safeExpire`.  The blocking of `VideobridgeExpireThread` doesn't seem to be a problem (the thread's work isn't urgent and the block wouldn't be for long) so I think we can get rid of all this.  It should be noted, however, that all `Conference` expiration should be called on by `Videobridge`; anything else wanting to expire a conference should call `Videobridge#expireConference`.

I considered keeping the `Expireable` impl around, but it's `safeExpire` method didn't seem appropriate/necessary, and, we don't have any place where we expire a group of `Expireable`s or pass something as an `Expirable`, so we weren't getting much use out of it.


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -529,19 +518,15 @@ private double getMaxReceiverRtt(String excludedEndpointId)
      * respective <tt>Channel</tt>s. Releases the resources acquired by this
      * instance throughout its life time and prepares it to be garbage
      * collected.
+     *
+     * NOTE: this should _only_ be called by the Conference "manager" (in this
```
bgrozev, line 522: Can we make it package private?

bbaldino: Yeah good idea

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -864,7 +837,7 @@ public boolean isExpired()
     {
         // this.expired starts as 'false' and only ever changes to 'true',
         // so there is no need to synchronize while reading.
-        return expired;
+        return expired.get();
```
bgrozev, line 840: The comment above is now stale?

PR #1401, opened 2020-08-26
Support for RED streams.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/util/PayloadTypeUtil.kt
```diff
@@ -0,0 +1,115 @@
...
+            ext.parameters.forEach { parameter ->
+                // In SDP, format parameters don't necessarily come in name=value pairs (see e.g. the format used in
+                // RFC2198). However XEP-0167 requires a name and a value. Our SDP-to-Jingle implementation in
+                // lib-jitsi-meet translates a non-name=value SDP string into a parameter extension with a value but no
+                // name. Here we'll just ignore such parameters, because we don't currently support any and changing the
+                // implementation would be inconvenient (we store them mapped by name).
+                if (parameter.name != null) {
+                    parameters[parameter.name] = parameter.value
+                } else {
+                    logger.warn("Ignoring a format parameter with no name: " + parameter.toXML())
```
JonathanLennox, line 78: If this is something that happens frequently will this warning get too chatty?

bgrozev: No, it never happens with an unmodified jicofo. I only noticed it while looking at adding fmtps for RED, which I ended up not committing.

jvb/src/main/kotlin/org/jitsi/videobridge/util/PayloadTypeUtil.kt
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+
+            val rtcpFeedbackSet: MutableSet<String> = CopyOnWriteArraySet()
+            ext.rtcpFeedbackTypeList.forEach { rtcpExtension ->
+                var feedbackType = rtcpExtension.feedbackType
+                val feedbackSubtype = rtcpExtension.feedbackSubtype
+                if (feedbackSubtype != null && feedbackSubtype != "") {
```
bbaldino, line 86: There's `isNullOrEmpty` that you can call on a `String?`

bbaldino: Could also leverage `map`/`toCollection` here
```
ext.rtcpFeedbackTypeList.map { rtcpExt ->
    buildString {
      append(rtcpExt.feedbackType)
      if (!rtcpExt.feedbackSubtype.isNullOrEmpty()) {
          append(" ${rtcpExt.feedbackSubtype")
      }
    }
}.toCollection(rtcpFeedbackSet)
```

I wish there was an "append if not null or empty", but don't think one exists in stdlib from what I can tell.

bgrozev:
> Does it still need to be CopyOnWrite?

I don't understand why `CopyOnWrite` was used, but if it needs to be used it shoud be enforced from the class. I pushed a change to JMT to always use `CopyOnWriteArraySet`

JonathanLennox (approved):

PR #1405, opened 2020-09-02
FEAT: Optimize bwe calculations: only calculate bwe once per feedback message.

JonathanLennox:
(Rather than once per packet reported on by feedback.)

Updates jmt.


bgrozev:
I just merged jitsi-media-transform/287. Maybe bump the version to include both?

bbaldino (approved):

PR #1406, opened 2020-09-02
Public Jetty server rework

bbaldino:
Currently we have two Jetty server instances: one which is intended to be 'private' (only accessible from localhost) and another which is intended to be 'public' (accessible externally).  On each of those servers various services are run.  Right now the only service on the 'public' instance is the WebSocket servlet.  Because of this, the WebSocket service in the bridge "became" the public server and was responsible for its creation/startup.

This PR defines a new way to create a Jetty server and makes the server a first-class citizen which services (defined as servlets) are then added to.

The main idea here is that services should be defined as servlets and then components (like the bridge) can add them to a Jetty server instance it creates.  As a first phase, I just ported the public Jetty server and the services hosted there (WebSocketService).  The idea is that other services like About, Version, etc. defined in Jicoco will be refactored as servlets so they can be added to any arbitrary server instance.


jvb/src/main/kotlin/org/jitsi/videobridge/JettyHelpers.kt
```diff
@@ -0,0 +1,130 @@
...
+fun createSecureJettyServer(
+    port: Int,
+    keyStorePath: String,
+    host: String? = null,
+    keyStorePassword: String? = null,
+    needClientAuth: Boolean = false
+): Server {
+    val sslContextFactoryKeyStoreFile = Paths.get(keyStorePath).toFile()
+    val sslContextFactory = SslContextFactory().apply {
+        setIncludeProtocols("TLSv1.2")
```
bgrozev, line 64: jicoco adds "TLSv1.3" on java >= 11. Do we want the same here?

jvb/src/main/kotlin/org/jitsi/videobridge/JettyHelpers.kt
```diff
@@ -0,0 +1,130 @@
...
+            config.host
+        )
+    }
+}
+
+val Server.servletContextHandler: ServletContextHandler?
+    get() = handler as? ServletContextHandler
+
+fun Server.addServlet(servlet: ServletHolder, pathSpec: String) {
+    (handler as? ServletContextHandler)?.addServlet(servlet, pathSpec)
```
bgrozev, line 131: Can you use the val defined above here?

jvb/src/main/kotlin/org/jitsi/videobridge/JettyHelpers.kt
bbaldino: I left that out because it seemed unnecessary and determining the Java version was a little awkward, from what I remember.

jvb/src/main/java/org/jitsi/videobridge/Main.java
```diff
@@ -155,4 +170,23 @@ public void debug(@NotNull Function0<String> function0)
...
+    private static Server setupPublicHttpServer()
+    {
+        JettyBundleActivatorConfig publicServerConfig = new JettyBundleActivatorConfig(
+            "org.jitsi.videobridge.rest",
+            "videobridge.http-servers.public"
+        );
+
+        final Server publicServer = JettyHelpers.createServer(publicServerConfig);
+        ColibriWebSocketService colibriWebSocketService =
+            new ColibriWebSocketService(publicServerConfig.getKeyStorePath() != null);
```
bgrozev, line 183: I think the `keyStorePath == null` check is used in 3 places now (here, the new helper, and jicoco). I would push it down to `JettyBundleActivatorConfig`: `val isTls: Boolean = keystorePath != null`

bgrozev: We have configuration which enables/disables `ColibriWebSocketService`, which is currently in its constructor. Should we move it here? I.e. only create and register the service if it's enabled?

jvb/src/main/kotlin/org/jitsi/videobridge/websocket/ColibriWebSocketService.kt
```diff
@@ -0,0 +1,85 @@
...
+     * ID [conferenceId] to use to connect to the websocket with password [pwd] or null if the
+     * [ColibriWebSocketService] is not enabled.
+     */
+    fun getColibriWebSocketUrl(conferenceId: String, endpointId: String, pwd: String): String? {
+        if (!config.enabled) {
+            return null
+        }
+        // "wss://example.com/colibri-ws/server-id/conf-id/endpoint-id?pwd=123
+        return "$baseUrl/$conferenceId/$endpointId?pwd=$pwd".also {
+            logger.info("TEMP: returning ep websocket url: $it")
```
bgrozev, line 55: Temp log?

jvb/src/main/kotlin/org/jitsi/videobridge/JettyHelpers.kt
bgrozev: cc @JonathanLennox, he specifically added it recently

bbaldino: Thanks for pointing that out, I didn't realized it had been added recently.  It doesn't look like it's a Java 11 compatibility thing so much as opportunistically doing TLS 1.3.  My hope was that we could ignore this for now and, when it comes up, maybe be in a better place about java 11 support and therefore avoid the hacky java version check, but I dunno.  Will let Jonathan weigh in.

jvb/src/main/java/org/jitsi/videobridge/Main.java
bbaldino: > I think the keyStorePath == null check is used in 3 places now (here, the new helper, and jicoco). I would push it down to JettyBundleActivatorConfig: val isTls: Boolean = keystorePath != null

I think I was avoiding having to mess with Jicoco :/ But yeah I can add it there.

> We have configuration which enables/disables ColibriWebSocketService, which is currently in its constructor. Should we move it here? I.e. only create and register the service if it's enabled?

Yeah I think these are the two approaches for these sorts of services: either always "create" them and let the service itself decide whether or not it's enabled, or have the caller figure it out.  There are aspects of both solutions I don't like:

1. Having the "outside" code decide whether or not to start the service basically requires that the service's config be exposed, which I think it would be nice to avoid.  It also feels like the service itself should know best.

1. Having the service decide requires creating an instance of it (otherwise we have to use an outside config class instead of normal props, which I think is nicer; especially when no outside code _needs_ access to the config).

So here I ended up creating the service, but not registering the servlet if it isn't enabled.

jvb/src/main/kotlin/org/jitsi/videobridge/websocket/ColibriWebSocketService.kt
bbaldino: Oops, thanks. Removed.

jvb/src/main/kotlin/org/jitsi/videobridge/JettyHelpers.kt
JonathanLennox: TLS 1.3 is only supported on Java 11+, which is why the condition, and I believe the `setIncludeProtocols` call fails on older Java for unknown versions.

Getting the version is only hacky because of the syntax change ("1.8" -> "11") which technically you could ignore anyway since this only cares about 11+.

bbaldino: I think the question is: do we need to be opportunistically enabling TLS 1.3 for Java 11+ (which we don't currently use anywhere)?  What prompted adding that?

JonathanLennox: Just completeness -- I wanted to match the Mozilla Guidelines config recommendations as closely as possible.

bbaldino:
I changed websockets to be disabled by default, which is how it should have always been since a domain is required when websockets are enabled and we can't set that by default.  This wasn't necessary before because the websocket service was tied to the public http server, which was not enabled by default.  Seeing this, I also realized that the public server _shouldn't_ be enabled by default, so I'll revert another change I made there and use the port being -1 to decide whether or not the server should be started (which was the old behavior).

jvb/src/main/kotlin/org/jitsi/videobridge/JettyHelpers.kt
```diff
@@ -129,3 +134,12 @@ val Server.servletContextHandler: ServletContextHandler
 
 fun Server.addServlet(servlet: ServletHolder, pathSpec: String) =
     this.servletContextHandler.addServlet(servlet, pathSpec)
+
+// TLS 1.3 requires Java 11 or later.
+private fun supportsTls13(): Boolean {
+    return try {
+        getJavaVersion() >= 1
```
JonathanLennox, line 141: Should be 11 not 1

bbaldino: Oof. Fixed.

JonathanLennox (approved):

JonathanLennox (approved):

PR #1415, opened 2020-09-08
Ep connection monitor fixes

bbaldino:
This PR fixes a couple things from the ep connection status monitor refactoring done in https://github.com/jitsi/jitsi-videobridge/pull/1384:

1. Correctly stop the monitor when a conference expires.  This was not an issue before as we had a single, global ep connection status monitor
1. Don't perform ep connection status monitoring for health check conferences.  The old code didn't ignore health check conferences, but it's a waste to monitor endpoint connection status for them.

I also cleaned up a couple low-hanging-fruit IDE warnings.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -174,7 +174,7 @@ public class Conference
      */
     private ScheduledFuture<?> updateLastNEndpointsFuture;
 
-    private EndpointConnectionStatusMonitor epConnectionStatusMonitor;
+    private final EndpointConnectionStatusMonitor epConnectionStatusMonitor;
 
     /**
      * Initializes a new <tt>Conference</tt> instance which is to represent a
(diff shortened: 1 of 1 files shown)


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -179,7 +179,7 @@
      */
     private ScheduledFuture<?> updateLastNEndpointsFuture;
 
-    private final EndpointConnectionStatusMonitor epConnectionStatusMonitor;
+    private EndpointConnectionStatusMonitor epConnectionStatusMonitor;
```
JonathanLennox, line 177: Can this still be `final` if we explicitly initialize it to `null` in the `!enableLogging` case?  (Would it be worth it?)

bbaldino: Looks like it can.  It probably doesn't make a huge difference but `final` is always nice when we can.

JonathanLennox (approved):

PR #1418, opened 2020-09-08
Make load management logs less chatty

bbaldino:


jvb/src/main/kotlin/org/jitsi/videobridge/load_management/JvbLoadManager.kt
```diff
@@ -39,17 +39,34 @@ class JvbLoadManager<T : JvbLoadMeasurement> @JvmOverloads constructor(
...
             state = State.NOT_OVERLOADED
             if (loadMeasurement.getLoad() < jvbRecoveryThreshold.getLoad()) {
-                logger.info("Load measurement $loadMeasurement is below threshold of $jvbLoadThreshold, " +
-                        "maybe running recovery")
-                maybeRun("recovery") { recover() }
+                if (canRunReducer(now)) {
+                    if (loadReducer.recover()) {
+                        logger.info("Recovery ran after a load measurement of $loadMeasurement (which was below " +
+                            "threshold of $jvbLoadThreshold) was received")
+                        lastReducerTime = now
```
bbaldino, line 61: Note that I only set `lastReducerTime` in the `recover` case when there was actually work done: otherwise all normal measurements will register a run of the load reducer and we would have to wait longer to run `reduceLoad`.

jvb/src/main/kotlin/org/jitsi/videobridge/load_management/JvbLoadManager.kt
```diff
@@ -39,17 +39,34 @@ class JvbLoadManager<T : JvbLoadMeasurement> @JvmOverloads constructor(
...
         } else {
             state = State.NOT_OVERLOADED
             if (loadMeasurement.getLoad() < jvbRecoveryThreshold.getLoad()) {
-                logger.info("Load measurement $loadMeasurement is below threshold of $jvbLoadThreshold, " +
-                        "maybe running recovery")
-                maybeRun("recovery") { recover() }
+                if (canRunReducer(now)) {
+                    if (loadReducer.recover()) {
+                        logger.info("Recovery ran after a load measurement of $loadMeasurement (which was below " +
+                            "threshold of $jvbLoadThreshold) was received")
```
JonathanLennox, line 60: Should be `$jvbRecoveryThreshold` in log message

JonathanLennox: (I wonder if `JvbLoadMeasurement`'s definition should also comment that it needs a useful `toString()`.)

bbaldino: Will fix and I'll add the comment.  I wondered if we should do a separate 'description' method but that's probably overkill.

jvb/src/main/kotlin/org/jitsi/videobridge/load_management/JvbLoadMeasurement.kt
```diff
@@ -16,6 +16,10 @@
 
 package org.jitsi.videobridge.load_management
 
+/**
+ * A measurement of load on the JVB.  Implementations should implement a description of the
+ * measurement in the [toString] method.
+ */
```
JonathanLennox, line 22: You should probably follow your own advice in `MockLoadMeasurement`.

bbaldino: Don't think it's really necessary there, but ok.

JonathanLennox (approved):

PR #1423, opened 2020-09-09
add support for new config file in debian installs

bbaldino:
The intent here is to generate an empty `jvb.conf` if one doesn't exist, and pass this config file (via `-Dconfig.file`) to the bridge when starting up so that debian installs can support the new configuration.

I didn't go so far as to migrate any configurations put in `sip-communicator.properties` into the new config, as the logic in this script is already pretty complex (already supports migrating from an old, _old_ config file). Maybe there's some code that can be cleaned up/removed there for the older cases?

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -71,6 +71,8 @@ case "$1" in
 
         OLD_JITSI_CONFIG="/usr/share/jitsi-videobridge/.sip-communicator/sip-communicator.properties"
         NEW_JITSI_CONFIG="/etc/jitsi/videobridge/sip-communicator.properties"
+        HOCON_CONFIG="/etc/jitsi/videobridge/jvb.conf"
+
         if [ -f $OLD_JITSI_CONFIG ]; then
             mv $OLD_JITSI_CONFIG $NEW_JITSI_CONFIG
             echo "# Config has moved to /etc/jitsi/videobridge/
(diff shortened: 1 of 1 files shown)


debian/postinst
```diff
@@ -117,6 +120,12 @@ case "$1" in
             echo "org.jitsi.videobridge.xmpp.user.shard.MUC_NICKNAME=$(uuidgen)" >> $NEW_JITSI_CONFIG
         fi
 
+        if [ ! -f $HOCON_CONFIG ]; then
+            echo "Generating an empty hocon config"
+            echo "videobridge {
+            }" >> $HOCON_CONFIG
```
JonathanLennox, line 126: If the file doesn't exist you don't need `>>`, just `>`, though I suppose it doesn't matter.

Should the file have a comment at the top giving a ref to the appropriate point in the JVB docs for its contents and syntax?

bbaldino: I think it'll be the other way around: docs will point to this file.

damencho:
@bbaldino The parameter `-Dconfig.file...` will be added only for new install and the empty file will be created for old and new installs, is this the intended behavior?

bbaldino:
> @bbaldino The parameter `-Dconfig.file...` will be added only for new install and the empty file will be created for old and new installs, is this the intended behavior?

Nope, that wasn't intended.  I think what we want to do both of those things for both old and new.

damencho:
I see. So you can be checking the config file for existing of `-Dconfig.file=` and then replace `JAVA_SYS_PROPS="` with `JAVA_SYS_PROPS ="-Dconfig.file= ` (with extra space at the end).
And that will cover old and new, I think.

damencho:
I hope sed with `|` as a delimiter is available on all modern distros

damencho (approved):

PR #1427, opened 2020-09-10
Add VP8 simulcast stream bitrate calculator

dsmeytis:
If WebRTC client's uplink is bad to a degree that encoder periodically sends key frames
of the higher spatial layer but it is not enough for stable playback it leads to a situation
when higher layer keyframe switches the stream however no packets are sent afterwards
and picture freezes until the next keyframe appears instead of keep on sending stable lower layer stream.

**Steps to reproduce:**
1. Start video conference in 720p highest spatial layer
2. Limit uplink for some webrtc client to 512kbit/s
3. Observe "slideshow"-like video from that peer on other participants' screens.

In order to solve this issue I would like to introduce simple bitrate calculator for "middle" and "high" spatial layers.
So it would upscale only if such layer is received at least at minimal bitrate (thresholds are got from webrtc source code, see [simulcast.cc](https://chromium.googlesource.com/external/webrtc/+/master/media/engine/simulcast.cc#80))

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveSourceProjectionContext.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/cc/vp8/VP8AdaptiveSourceProjectionContext.java
@@ -27,6 +27,7 @@ import org.jitsi.utils.logging2.Logger;
 import org.jitsi.videobridge.cc.*;
 import org.json.simple.*;
 
+import java.time.Clock;
 import java.util.*;
 
 /**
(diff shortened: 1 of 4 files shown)


dsmeytis:
Hi @bgrozev , thank you for the quick response.
I guess it would be better to show an example. Please, take a look at the screenshot
Screenshot
Red line represents the traffic from JVB to some peer and other lines are simulcast layer streams from another peer.
So between 55 and 80 seconds I limited uplink to 512 kbit/s. And as you can see the lowest layer is transmitted fine and sometimes there are some packets from the middle layer. As far as I understood the root cause, WebRTC client's bandwidth estimator tries increase bitrate and it gets enough to send next layer, however when it starts doing that packet losses gets more than 10% and it falls back to the point where it can't encode the layer. So JVB receives only few packets, especially keyframes and tries to switch the layer for its receivers.

Thank you for your hint about BitrateController, I have somehow missed that point. I agree it is better place for that I will take a look at that piece of code.

dsmeytis:
Better solution proposed in [PR#1439](https://github.com/jitsi/jitsi-videobridge/pull/1439)

PR #1429, opened 2020-09-10
Remove EventAdmin

bbaldino:
We no longer use this in the bridge.  If there's need we may bring back a different form of events in the future.


bgrozev (approved):

bgrozev (changes requested):
Can we also remove EventFactory.java?

bbaldino:
> Can we also remove EventFactory.java?

Ah good catch, wasn't aware of that. Yeah will remove that as well.

bgrozev (approved):

PR #1430, opened 2020-09-10
ref: Simplify code.

bgrozev:


jvb/src/test/kotlin/org/jitsi/videobridge/ConferenceTest.kt
```diff
@@ -17,53 +17,33 @@ package org.jitsi.videobridge
...
-    }
-    private val videobridge = mockk<Videobridge> {
-        every { bundleContext } returns this@ConferenceTest.bundleContext
-    }
-
-    init {
-        mockkStatic("org.jitsi.videobridge.octo.OctoRelayServiceProviderKt")
-        every { singleton() } returns octoRelaySupplier
-    }
+    private val videobridge = Videobridge()
```
bbaldino, line 30: I've got mixed feelings on using a real `Videobridge` here as it will make the test more brittle.  I know there's a counter argument that the test may be less accurate with a mock version, but I think it's better to mock the dependencies so they are better controlled for the test.

jvb/src/test/kotlin/org/jitsi/videobridge/ConferenceTest.kt
```diff
@@ -17,53 +17,33 @@ package org.jitsi.videobridge
...
 
     init {
         val name = JidCreate.entityBareFrom("roomName@somedomain.com")
-        context("Adding local endpoints should work") {
-            withNewConfig(newConfigOctoEnabled, loadDefaults = true) {
+
+        withNewConfig(newConfigOctoEnabled, loadDefaults = true) {
```
bbaldino, line 35: Hmm, from what I've seen, using `withNewConfig` at this level will not work correctly because `context` merely registers a test, so the context block is actually run _after_ the `withNewConfig` block finishes, meaning it restores the old config.

bgrozev: I don't understand how that even worked before. Fixed now

jvb/src/test/kotlin/org/jitsi/videobridge/ConferenceTest.kt
```diff
@@ -16,74 +16,55 @@
...
-        every { get() } returns octoRelayService
-    }
-
-    private val bundleContext = mockk<BundleContext> {
-        every { registerService(any() as String, any(), any()) } returns null
-    }
-    private val videobridge = mockk<Videobridge> {
-        every { bundleContext } returns this@ConferenceTest.bundleContext
-    }
+    // The octo relay binds on a port
```
bbaldino, line 32: Wouldn't it be easier to keep the mock for octo?

bgrozev: It was a mock for the `OctoRelay*Supplier*` with a real `OctoRelayService` instance. The tentacle needs the UDP and Conf transports, so there would be a whole bunch of things to mock.

jvb/src/test/kotlin/org/jitsi/videobridge/ConferenceTest.kt
```diff
@@ -16,74 +16,55 @@
...
-            withNewConfig(newConfigOctoEnabled, loadDefaults = true) {
-                octoRelayService.start()
-                with(Conference(videobridge, "id", name, false, 1234)) {
-                    isOctoEnabled shouldBe false
-                    tentacle
-                    isOctoEnabled shouldBe true
-                    tentacle.setRelays(listOf("127.0.0.1:4097"))
+            with(Conference(videobridge, "id", name, false, 1234)) {
+                isOctoEnabled shouldBe false
+                tentacle
```
bbaldino, line 63: Not for this PR, but reading this test flow makes me wonder if the API we have set up in Conference makes sense (`isOctoEnabled` checks if `tentacle != null`, but, even if Octo _is_/_will be_ enabled, it won't be reflected until `getTentacle` has been called)

jvb/src/test/kotlin/org/jitsi/videobridge/ConferenceTest.kt
```diff
@@ -94,7 +75,7 @@ private val newConfigOctoEnabled = """
         octo {
             enabled = true
             bind-address = 127.0.0.1
-            bind-port = 4096
+            bind-port = 54096
```
bbaldino, line 78: Should we use a random port here?

bgrozev: I don't know of an easy way to get a random free port without binding to it. Are you worried about this failing if 54096 is  not free?

bbaldino: That and running tests in parallel.  I wasn't thinking we needed to get too fancy...just literally using a random (valid) number.

bgrozev: The more tests we run in parallel the higher the probability we hit a used port.

bbaldino: Yeah, I think the ideal thing would be to add a hook in OctoRelayService for a "UdpTransportFactory": a function it uses to create a UDP transport so we could have it create a fake which doesn't bind at all.

bbaldino (approved):

PR #1443, opened 2020-09-15
OctoRelayService cleanup

bbaldino:
There were some awkward flows in `OctoRelayService` where we didn't initialize members until `start`, so they had to be nullable.  This moves the initialization to the constructor so we can avoid that.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/octo/OctoRelayService.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/octo/OctoRelayService.kt
@@ -22,6 +22,8 @@ import org.jitsi.videobridge.transport.octo.BridgeOctoTransport
 import org.jitsi.videobridge.transport.udp.UdpTransport
 import org.jitsi.videobridge.util.TaskPools
 import java.net.SocketAddress
+import java.net.SocketException
+import java.net.UnknownHostException
 import java.time.Instant
 
 /**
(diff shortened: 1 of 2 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/octo/OctoRelayService.kt
```diff
@@ -22,84 +22,79 @@ import org.jitsi.videobridge.transport.octo.BridgeOctoTransport
...
-                is UnknownHostException, is SocketException -> {
-                    logger.error("Failed to initialize Octo UDP transport with " +
-                            "address " + address + ":" + port + ".", e)
-                    return
-                }
-                else -> throw e
-            }
-        }
+        udpTransport = UdpTransport(address, port, logger, OCTO_SO_RCVBUF, OCTO_SO_SNDBUF)
+        logger.info("Created Octo UDP transport")
```
JonathanLennox, line 54: Can this no longer throw?  Or is something outside this constructor catching it?

bbaldino: It can, but I wanted it to bubble up.  I forgot we still create this in the singleton, though, so I think we will still need some form of this + a try catch in the singleton

JonathanLennox (approved):

jvb/src/main/kotlin/org/jitsi/videobridge/octo/OctoRelayServiceProvider.kt
```diff
@@ -21,7 +21,11 @@ import org.jitsi.videobridge.octo.config.OctoConfig
 class OctoRelayServiceProvider {
     private val octoRelayService: OctoRelayService? by lazy {
         if (OctoConfig.config.enabled) {
-            OctoRelayService()
+            try {
+                OctoRelayService()
+            } catch (t: Throwable) {
+                null
```
bgrozev, line 27: Should we log an error?

bbaldino: We don't have a logger here, so I logged it in `OctoRelayService`

bgrozev (approved):

PR #1447, opened 2020-09-16
Fix: don't over-send bandwidth probing, and send bandwidth probing even when no sources can be sent.

JonathanLennox:
Should fix observed packet losses.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
@@ -54,6 +54,13 @@ import java.util.*;
       */
      public boolean enabled = false;
 
+     /**
+      * The number of bytes left over from one run of probing to the next.  The
+      * primary use case for this is when the number of bytes we want to send is
+      * less than the size of an RTP header extension.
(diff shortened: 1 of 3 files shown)


JonathanLennox:
Depends on https://github.com/jitsi/jitsi-media-transform/pull/297 .

jvb/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
```diff
@@ -151,28 +159,15 @@ public void run()
          if (paddingBps < 1)
          {
              // Not much.
+             bytesLeftOver = 0;
              return;
          }
 
+         long bytes = (config.getPaddingPeriodMs() * paddingBps / 1000 / 8) + bytesLeftOver;
```
bbaldino, line 166: Merely in the interest of not changing too many things at once, I'm wondering if we should hold off on the `bytesLeftOver` as part of this fix.

JonathanLennox: Ok, I can separate that.

bbaldino (approved):

Issue #1444, opened 2020-09-15
JVB "top level services" cleanup


bbaldino:
As a transitionary step to migrating away from OSGi, many "top level services" in the JVB were shuffled into global singletons so we could maintain similar accessibility to before (with OSGi) but work on removing the OSGi framework itself.  With that done, now we can do some cleanup of these services and hopefully get rid of most/all of the static singletons.

Unfortunately, the dependency graph is fairly complex:
![ZCkLW5YHOnqi](https://user-images.githubusercontent.com/4016469/93270771-a084bd00-f766-11ea-991c-befec510accb.jpg)

So untangling things will take time.

Steps (will be updated over time)

- [ ] Remove uses of singletons inside `VideobridgeStatistics` and pass the dependencies in instead.

PR #1452, opened 2020-09-18
Refactor JVB Client Connection code

bbaldino:
This PR refactors the XMPP client connection code in the bridge to address a couple things:

1. XMPPCommon is leftover from when we supported both Component and Client connections and is no longer needed
2. Address a "deep" access of the ClientConnection singleton and move it to a higher layer as part of cleaning up the singleton code (see https://github.com/jitsi/jitsi-videobridge/issues/1444)


jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/ClientConnection.kt
```diff
@@ -0,0 +1,199 @@
...
+        if (running.compareAndSet(true, false)) {
+            mucClientManager.stop()
+        }
+    }
+
+    fun setPresenceExtension(extension: ExtensionElement) {
+        mucClientManager.setPresenceExtension(extension)
+    }
+
+    fun addMucClient(jsonObject: JSONObject): Boolean {
```
bgrozev, line 82: Can you bring back the docs from the original? They describe the json format

bgrozev: Other docs in this class are also useful

bbaldino: Oops, definitely.

jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/ClientConnection.kt
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Process an incoming IQ
+     */
+    override fun handleIq(iq: IQ?): IQ? {
+        if (iq == null) {
+            return null
+        }
+        logger.cdebug { "RECV: ${iq.toXML()}" }
+
+        val response = when (iq.type) {
```
bgrozev, line 122: IQ types are `get`, `set`, `result` and `error`. The latter two are responses, and we wouldn't send anything back for them, so there's not need to log "SENT null".

The code could be simplified some, too: 
```
return when (iq.type) {
  IQ.Type.get, IQ.Type.set -> handleIqRequest(iq).also { log() }
  else -> null
}

File: jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/ClientConnection.kt [hunk 3]
```diff
@@ -0,0 +1,199 @@
...
+    companion object {
+        private val FEATURES = arrayOf(
+            ColibriConferenceIQ.NAMESPACE,
+            HealthCheckIQ.NAMESPACE,
+            "urn:xmpp:jingle:apps:dtls:0",
+            "urn:xmpp:jingle:transports:ice-udp:1",
+            Version.NAMESPACE
+        )
+
+        private val delayThresholds = longArrayOf(5, 50, 100, 1000)
```
bgrozev, line 186: Note: these were static to accomodate both XMPP client and component modes. They don't need to be static anymore (but that doesn't need to change in this PR).

bgrozev:
The name `ClientConnection` was used to distinguish from the XMPP component connection. I think this is the right time to rename it to `XmppConnection` or something that makes it more obvious what it is.

jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/ClientConnection.kt
bbaldino: Ok, yeah I think I'll leave those as-is for now.

PR #1455, opened 2020-09-22
feat: Add loss stats based on total number of packets.

bgrozev:
This is on top of #1453, so the first two commits are not relevant. I'll rebase when #1453 is merged.

It also depends on jitsi/jitsi-media-transform#302


jvb/src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -197,15 +191,20 @@ private void generate0()
...
-        long packetsReceivedLost = 0; // TODO verify
-        long bitrateDownloadBps = 0;
-        long bitrateUploadBps = 0;
+        double bitrateDownloadBps = 0;
+        double bitrateUploadBps = 0;
         long packetRateUpload = 0;
         long packetRateDownload = 0;
 
+        // Packets we received
+        long packetsReceivedReceived = 0;
```
bbaldino, line 200: How about `incomingPacketsReceived`, `incomingPacketsLost`, `outgoingPacketsReceived`, and `outgoingPacketsLost`?

jvb/src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -314,21 +297,24 @@ private void generate0()
...
-                        = transceiverStats.getEndpointConnectionStats().getRtt();
+                EndpointConnectionStats.Snapshot endpointConnectionStats = transceiverStats.getEndpointConnectionStats();
+                double endpointRtt = endpointConnectionStats.getRtt();
                 if (endpointRtt > 0)
                 {
                     rttSumMs += endpointRtt;
                     rttCount++;
                 }
+
+                packetsReceivedReceived += endpointConnectionStats.getReceiveLoss().getPacketsReceived();
```
bbaldino, line 314: Reading this, it feels like the endpoint connection stats method should be `getReceiveLossStats()`/`getSendLossStats()`.

bbaldino (dismissed):

JonathanLennox:
Are any of the other jmt 190-195 changes worth mentioning in the commit message?

bgrozev:
Yes, I think these are:
`Handle packets initially reported lost, and later reported received, in bwe. (#238)`
`feat: Increase bucket size for rate trackers (decrease memory use).' (#305)`

We can do it with the squash-merge

JonathanLennox (approved):

PR #1466, opened 2020-09-25
Refactor uses of System.currentTimeMillis to use Clock, plumb clock where needed

bbaldino:
This turned into a bit of a yak-shave...I saw some uses of `System.currentTimeMillis` and wanted to change them to use `Clock`.  Then I hit a case where I had to compare `Duration` instances, which I find awkward in java (you have to use `compareTo`, whereas Kotlin just supports the comparison opereators), so I ended up moving some methods to a Kotlin subclass of Conference (which I've thought would be nice in the past, but not strictly necessary).  Then I wanted to clean up things to consistently using the same `Clock` instance throughout  which I think would open some interesting doors for videobridge testing.  I wasn't able to finish that 100% though, some places need to have the singleton cleanup done first.

I didn't plan on doing this much, but ended up getting sucked into it and figured I'd put it up.  Not urgent.  Going commit-by-commit should keep things pretty simple.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -163,7 +163,6 @@ public abstract class Conference
      */
     protected ScheduledFuture<?> updateLastNEndpointsFuture;
 
-    protected final EndpointConnectionStatusMonitor epConnectionStatusMonitor;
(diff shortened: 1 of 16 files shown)


bbaldino:
Meh. Might revisit this later in smaller pieces.

PR #1471, opened 2020-09-30
set the websocket service instance in the singleton

bbaldino:
This was a regression from https://github.com/jitsi/jitsi-videobridge/pull/1468


damencho (approved):

damencho:
ActiveSpeaker test was not passing before this commit and speaker stats/ the websocket was broken. After the change, ActiveSpeaker test passed on chrome so will merge and will be tracing the other problems as the grid was updated ...

PR #1473, opened 2020-10-02
Remove use of global StatsManager singleton

bbaldino:
Previously, `Conference` would grab a global `StatsManager` singleton to fire conference created/expired events.  This PR adds events fired by `Videobridge` to expose these, and wires those events into the `StatsManager` to avoid the global singleton usage.


jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -305,24 +316,23 @@ private void enableGracefulShutdownMode()
...
             if (conference.equals(conferencesById.get(id)))
             {
                 conferencesById.remove(id);
-                expireConference = true;
-            }
-            else
-            {
-                expireConference = false;
+                conference.expire();
+                if (conference.includeInStatistics())
```
JonathanLennox, line 326: I'm feeling like `includeInStatistics` is being processed at the wrong level here - there's nothing about the `conferenceExpired` event that's intrinsically related to statistics.  I'd say that either it should be processed in the handler, or else the statistics-related handler shouldn't be installed for non-`includeInStatistics` conferences.

jvb/src/main/kotlin/org/jitsi/videobridge/util/EventEmitter.kt
```diff
@@ -14,23 +14,20 @@
  * limitations under the License.
  */
 
-package org.jitsi.videobridge.stats
+package org.jitsi.videobridge.util
 
-import java.util.function.Supplier
+class EventEmitter<EventHandlerType> {
+    private val eventHandlers = mutableListOf<EventHandlerType>()
```
JonathanLennox, line 20: Does this list need synchronization?

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
bbaldino: I agree that I find this aspect (and the general special-casing for 'health check conferences') to be awkward in general.

>  I'd say that either it should be processed in the handler

I can, but I think there's an argument that this flag is already more exposed than we want it, and the stat handler shouldn't have to care about this.  Since the conference really is basically an "internal, videobridge health check tool", I thought it made sense that the bridge be the one who "hides" it.

> or else the statistics-related handler shouldn't be installed for non-includeInStatistics conferences.

The subscription is done at the "Videobridge events" level, not per conference so I couldn't do this (the way it's currently implemented, at least).

jvb/src/main/kotlin/org/jitsi/videobridge/util/EventEmitter.kt
bbaldino: Yeah that's probably a good idea.

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
bbaldino: I'm not necessarily against changing it (of your suggestions, I think the handler doing it probably makes the most sense), but if my reasoning is convincing I'll leave it.

JonathanLennox: If we're going to leave it, there should at least be a comment at this point in the code that at present this event is only used for statistics, and if it were to be used for something else the code would need to be changed...

bbaldino: Ended up moving it to the handler.

JonathanLennox (approved):

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
bbaldino: Changed my mind on this again: I think this is something that is internal to Videobridge and we wouldn't ever expose this conference as part of the events interface.  I do agree the health check conference stuff could use some cleanup.

JonathanLennox (approved):

PR #1475, opened 2020-10-05
Fix NullPointerException if conferenceName is not defined.

dsmeytis:
This null-check is present in constructor but missing in getDebugState


damencho (approved):

damencho:
Jenkins test this please.

PR #1478, opened 2020-10-05
Fire events from Endpoint

bbaldino:
This is the start of some work to get rid of all the dependencies on `Conference` we have in `AbstractEndpoint` and its subclasses.  It uses the `EventEmitter` class introduced in https://github.com/jitsi/jitsi-videobridge/pull/1473, so that PR will need to be merged first.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -25,6 +25,7 @@ import org.jitsi.utils.*;
 import org.jitsi.utils.logging2.*;
 import org.jitsi.videobridge.message.*;
 import org.jitsi.videobridge.rest.root.debug.*;
+import org.jitsi.videobridge.util.*;
 import org.jitsi.xmpp.extensions.colibri.*;
 import org.json.simple.*;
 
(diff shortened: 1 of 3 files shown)


jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -444,4 +457,11 @@ public void removeReceiver(String receiverId)
             receiverVideoConstraintsChanged(receiverVideoConstraintsMap.values());
         }
     }
+
+    interface EventHandler {
```
bgrozev, line 461: Can this be in `Endpoint` instead?

bbaldino: Why there instead of here?  Then everywhere that wants to use it has to care about what the real type of an `AbstractEndpoint` is.  It's ok for subclasses not to fire some events (e.g. `OctoEndpoint` never firing `iceSucceeded`)

bgrozev (approved):

bbaldino:
I was a bit worried about the performance of going through this list-based event handler for the audio level (as I also planned to make the incoming-packet-ready an event as well), and it looks like the overhead of the event emitter (which is basically iterating through the loop) is enough that, to be safe, we probably shouldn't use it for per-packet events.  I'm going to revert the change to use audio level for this, and my plan will be to use a higher-performance handler for the per-packet events.

bgrozev (approved):

PR #1479, opened 2020-10-06
Remove conf/ep creation from health checks

bbaldino:
After discussing with @bgrozev, this aspect of the check was adding very little value and leads to some awkwardness with having to deal with these "special" conferences, so removing this logic.  It's possible we'll bring it back in some other form in the future.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -28,7 +28,6 @@ import org.jitsi.utils.logging2.*;
 import org.jitsi.videobridge.message.*;
 import org.jitsi.videobridge.octo.*;
 import org.jitsi.videobridge.shim.*;
-import org.jitsi.videobridge.stats.*;
 import org.jitsi.videobridge.util.*;
 import org.jitsi.xmpp.extensions.colibri.*;
 import org.json.simple.*;
@@ -40,7 +39,6 @@ import java.io.*;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.*;
-import java.util.logging.*;
 import java.util.stream.*;
 
 import static org.jitsi.utils.collections.JMap.*;
@@ -115,6 +113,7 @@ public class Conference
     /**
      * The world readable name of this instance if any.
      */
+    @NotNull
     private final EntityBareJid conferenceName;
 
     /**
@@ -140,11 +139,6 @@ public class Conference
     private final Logger logger;
 
     /**
-     * Whether this conference should be considered when generating statistics.
-     */
-    private final boolean includeInStatistics;
-
-    /**
      * The time when this {@link Conference} was created.
      */
     private final long creationTime = System.currentTimeMillis();
@@ -168,7 +162,7 @@ public class Conference
      */
     private ScheduledFuture<?> updateLastNEndpointsFuture;
 
-    private final EndpointConnectionStatusMonitor epConnectionStatusMonitor;
+    @NotNull private final EndpointConnectionStatusMonitor epConnectionStatusMonitor;
 
     /**
      * Initializes a new <tt>Conference</tt> instance which is to represent a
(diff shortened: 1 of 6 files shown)


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -115,6 +113,7 @@
     /**
      * The world readable name of this instance if any.
```
JonathanLennox, line 114: Should the "if any" be removed from the comment now that this is `@NotNull`?

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -179,36 +173,29 @@
      * <tt>Conference</tt> instance is to be initialized
      * @param id the (unique) ID of the new instance to be initialized
      * @param conferenceName world readable name of this conference
-     * @param enableLogging whether logging should be enabled for this
      * {@link Conference} and its sub-components, and whether this conference
```
JonathanLennox, line 177: The `@param` doc was more than one line, you only removed one of them.

JonathanLennox (commented):
This will need a matching Jicofo change.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -175,7 +175,6 @@
      * @param id the (unique) ID of the new instance to be initialized
      * @param conferenceName world readable name of this conference
      * {@link Conference} and its sub-components, and whether this conference
```
JonathanLennox, line 177: Nope, still one more line. 🙂

bbaldino: Oof, I swear I deleted them both.

JonathanLennox (dismissed):

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -179,36 +173,29 @@
...
-     * @param enableLogging whether logging should be enabled for this
      * {@link Conference} and its sub-components, and whether this conference
      * should be considered when generating statistics.
      * @param gid the optional "global" id of the conference.
      */
     public Conference(Videobridge videobridge,
                       String id,
-                      EntityBareJid conferenceName,
-                      boolean enableLogging,
+                      @NotNull EntityBareJid conferenceName,
```
bgrozev, line 181: Now that I think about it, do we want to force this to be present? It's not essential, and non-jicofo controllers might not use it

bbaldino: I thought we were relying on this for some things now? (though maybe that would still fall under just the Jicofo case).  @damencho ?

bbaldino: I know he had originally made it a String, which I think would be nice, but we require having the domain and stuff in there so he did the EntityBareJid to enforce that.

bgrozev: We require it to have a JID format if it's present. Otherwise, we only really use it if callstats is enabled. It feels wrong to require a JID even when the bridge doesn't use XMPP

bbaldino: A name seems like it would be useful even in the non-Jicofo case (we use it in debug state and in log context), so it seems likely they'd fake an `EntityBareJid` anyway.  I really liked the String idea, but then we had to do all this validation on it since we _need_ it to be a valid `EntityBareJid`, so enforcing it in the type was better.

I agree an actual `EntityBareJid` is not always needed, but I think the question is would a non-Jicofo user be more likely to want to use `null`, or a fake `EntityBareJid` to get more context?

After looking into it some more, it looks like the only place the bridge cares about it being an _actual_ Jid is in `describeShallow`--and that already handles the case where the conversion may fail, so maybe we can just use a `String` at this layer?

bgrozev (approved):

PR #1481, opened 2020-10-07
Refactor Callstats and stats transport code.

bgrozev:
This is ready for review, but not merging (I want to add tests and run some manual tests).

It untangles the per-conference and global stats reported to callstats, and otherwise cleans up the callstats code. It also removes the "stat transports" config and replaces it with configuration in the modules that subscribe for the stats (the XMPP code which publishes to a MUC, and the callstats code which publishes to callstats).

Viewing commit by commit will probably be best


jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
```diff
@@ -86,23 +84,26 @@ class CallstatsService(
 
         statsManager?.let { statsManager ->
             logger.info("Subscribing to global stats with interval ${config.interval}")
-            statsTransport = CallStatsIOTransport(statsService).also { statsTransport ->
+            statsTransport = CallstatsTransport(statsService)
+                .also { statsTransport ->
```
bbaldino, line 88: Was this newline on accident? The previous formatting looks like it made more sense

jvb/src/main/resources/reference.conf
```diff
@@ -192,6 +192,32 @@ videobridge {
       #}
     ]
   }
+  # Configuration related to pushing statistics to callstats.io.
+  callstats {
```
bbaldino, line 187: Seems like this should be scoped under `stats`?  (Though not under `transports`, as I know this isn't the transport config but the general CallStats config)

jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/XmppConnection.kt
```diff
@@ -52,7 +54,7 @@ public class XmppConnection : IQListener {
 
     var eventHandler: EventHandler? = null
 
-    fun start() {
+    fun start(statsManager: StatsManager?) {
```
bbaldino, line 57: I think I prefer this the way it was before: `XmppConnection` is just a transport which doesn't need to care about what it's carrying.  A layer above decides we want to transmit stats over this transport and creates an appropriate stats transport (`MucStatsTransport`) and passes it the transport to use (the `XmppConnection`) without `XmppConnection` knowing or caring.

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
```diff
@@ -0,0 +1,152 @@
...
+import java.util.concurrent.CopyOnWriteArrayList
+
+/**
+ * A class that manages the statistics. Periodically calls the
+ * passes them to <tt>StatsTransport</tt> instances that sends them.
+ *
+ * @author Hristo Terezov
+ * @author Lyubomir Marinov
+ */
+class StatsManager(statistics: Statistics) {
```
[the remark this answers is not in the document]
bgrozev: No, it's used in `init`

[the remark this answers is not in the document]
bgrozev: I actually made it a `val` and removed the extra `val statistics`

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
```diff
@@ -59,20 +60,24 @@ class StatsManager(statistics: Statistics) {
...
      * @param period the internal/period in milliseconds at which this [StatsManager] is to repeatedly send statistics
      * to the specified [transport].
      */
     fun addTransport(transport: StatsTransport, period: Long) {
         require(period >= 1) { "period $period" }
 
-        // XXX The field transport is a CopyOnWriteArrayList in order to avoid synchronization here.
-        transportRunnables.add(TransportPeriodicRunnable(transport, period))
+        TransportPeriodicRunnable(transport, period).also {
+            transportRunnables.add(it)
```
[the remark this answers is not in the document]
bgrozev: We do now (see `Main` in the last commit): we start the manager early and pass it to other modules (i.e. callstats and XMPP), which add themselves. I guess we could change this to delay the `start`, but this seems cleaner

jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
bgrozev: yep, I don't know how that happened

jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
```diff
@@ -0,0 +1,183 @@
...
+        "io.callstats.sdk.CallStats.conferenceIDPrefix".from(JitsiConfig.legacyConfig)
+    }
+
+    /**
+     * This is the interval at which stats are pushed to callstats. It affects both global and per-conference stats.
+     *
+     * For backwards compatibility, we read it from the stats manager "callstatsio" transport, if present, and
+     * default to the stats manager interval.
+     */
+    val interval: Duration = StatsManager.config.transportConfigs.stream()
```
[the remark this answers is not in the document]
bgrozev: Yes, this is still the case and implemented in the transport configs. But if none of these exist, there's a new property in the `videobridge.callstats` section to configure it.

jvb/src/main/resources/reference.conf
bgrozev: I thought about this, but wasn't sure. I don't mind changing it, will take a look in the morning.

jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/XmppConnection.kt
bgrozev: OK. My concern was that `Main` was becoming a bit of a mess. I'll think about it and refactor in the morning (same applies to the one for callstats)

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
```diff
@@ -28,9 +28,14 @@ import java.util.concurrent.atomic.AtomicBoolean
...
  * @author Lyubomir Marinov
  */
-class StatsManager(statistics: Statistics) {
+class StatsManager(
     /**
-     * The periodic runnable which gathers statistics.
+     * The instance which can gather/generate statistics via [Statistics.generate]. The [StatsManager] invokes this
+     * periodically.
+     */
+    val statistics: Statistics) {
```
bbaldino: nit: closing paren/opening brace on new line (I would've thought ktlint would complain?)

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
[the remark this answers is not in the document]
bbaldino: Ok, passing them in on construction would probably simplify things a bit (no need for this logic and `transportRunnables` could be immutable and not need to be threadsafe).  Is it awkward in Main to pass them in when creating it?

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
```diff
@@ -151,11 +143,19 @@ class StatsManager(
...
-    }
 
-    init {
-        val period = config.interval.toMillis()
-        require(period >= 1) { "period $period" }
-        statisticsRunnable = StatisticsPeriodicRunnable(statistics, period)
+        /**
+         * The [RecurringRunnableExecutor] which periodically invokes [statisticsRunnable].
+         */
+        private val statisticsExecutor = RecurringRunnableExecutor(
```
bbaldino, line 151: We'll only have a single instance of `StatsManager` anyway, right?

jvb/src/main/resources/reference.conf
```diff
@@ -177,37 +177,36 @@ videobridge {
...
-    # The interval at which statististics will be published to callstats. This affects both per-conference and global
-    # statistics.
-    # Note that this value will be overriden if a "callstatsio" transport is defined in the "stats" sections above.
-    interval = ${videobridge.stats.interval}
+      # TODO: document
+      //conference-id-prefix = "abcd"
 
+      # The interval at which statististics will be published to callstats. This affects both per-conference and global
+      # statistics.
+      # Note that this value will be overriden if a "callstatsio" transport is defined in the parent "stats" section.
```
bbaldino, line 207: Will the transport still be scoped under a `transports` block? I.e.:
```
videobridge {
    stats {
        transports {
            callstatsio {
            }
        }
    }
}
```

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
```diff
@@ -33,7 +33,8 @@ class StatsManager(
      * The instance which can gather/generate statistics via [Statistics.generate]. The [StatsManager] invokes this
      * periodically.
      */
-    val statistics: Statistics) {
+    val statistics: Statistics
+) {
```
bbaldino, line 37: Ah, already fixed :)

jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
```diff
@@ -48,14 +52,21 @@ class CallstatsService(
...
+        * Function to call if and when the service successfully initializes.
+        */
+        initializedCallback: () -> Unit = {}
+    ) {
+        logger.info("Starting CallstatsService with config: $config")
 
         // as we create only one instance of StatsService
         StatsServiceFactory.getInstance().createStatsService(
-            videobridge.versionService.currentVersion,
+            version,
```
bbaldino, line 69: 🥳

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
bgrozev: Yes

bbaldino: I'd leave them as members (we'd also want a hook in ctor to override them if we want to test this)

jvb/src/main/resources/reference.conf
bgrozev: No, no, no. The "transports" are gone. We just read them in the old format (`transports: [ ... ]`) for backward compat.

To configure callstats use the `callstats` section. To configure the interval for MUC presence updates use `videobridge.apis.xmpp-client.presence-interval`. The fact that under the hood these are implemented as "transports" in a stats manager is irrelevant and isn't exposed in the config file anymore.

bbaldino: OK, and I see we no longer explicitly enable the muc stats transport, the code just always initialize it?  And uses the xmpp connection's presence interval for the stats interval instead of allowing it to be configured?  What if someone isn't using the XMPP connection?

jvb/src/main/kotlin/org/jitsi/videobridge/stats/StatsManager.kt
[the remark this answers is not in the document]
bgrozev: There's no good way to do this, because the callstats lib initializes async. The old code creates a `StatsTransport` (that was also serving a bunch of other purposes) early, but this also means it's receiving and ignoring stats until the library loads. If we made the field immutable we would also be unable to remove that transport if callstats fails to load.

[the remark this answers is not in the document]
bbaldino: Ah ha, ok...if this is tied to the CallStats init async callback then I think that makes perfect sense.

jvb/src/main/resources/reference.conf
bgrozev: > OK, and I see we no longer explicitly enable the muc stats transport, the code just always initialize it?

Yes. The XMPP connection without presence updates is not very useful.

> And uses the xmpp connection's presence interval for the stats interval instead of allowing it to be configured?

No. The "stats interval" is the interval at which stats are gathers. It's controlled through `stats.interval` (I haven't changed that). 

There used to be per-transport intervals (meaning how often stats are pushed to that transport). These are now configurable through the module that uses them (`videobridge.apis.xmpp-client.presence-interval` and `videobridge.stats.callstats.interval`), and for backwards-compat they default to the stats interval.

>  What if someone isn't using the XMPP connection?

I think if the XMPP connection is not used, then it shouldn't subscribe to stats, though stats should still be generated as configured in `videobridge.stats` to e.g. expose via HTTP. I now see that `XmppConnection` is created and started unconditionally, so that won't work. I'll make a change.

bgrozev: Actually, that won't work, since MUC clients can be added dynamically over HTTP. The answer is that the current code does not allow you to not have an XMPP connection. I think this should change, but not in this PR.

bbaldino: > There used to be per-transport intervals (meaning how often stats are pushed to that transport). These are now configurable through the module that uses them (videobridge.apis.xmpp-client.presence-interval and videobridge.stats.callstats.interval), and for backwards-compat they default to the stats interval.

Ok, this is what I meant.  I do think it's a little implicit that the xmpp stats publish interval is configured via the presence interval, but probably not terrible.

> Actually, that won't work, since MUC clients can be added dynamically over HTTP. The answer is that the current code does not allow you to not have an XMPP connection. I think this should change, but not in this PR.

Ok

bgrozev: >  I do think it's a little implicit that the xmpp stats publish interval is configured via the presence interval

XMPP stats publish *is* the presence update

bbaldino: Yes but presence could be sent for other reasons and not contain any stats as well.  And it's not inherent that presence contains stats, we just happen to implement it that way, no?

bbaldino: At the XMPP level, I mean.  I'm not saying it's wrong, it's just an implicit implementation detail that I don't think would be obvious.  If I wanted to know how often stats were sent, I wouldn't think to look for the presence update interval.

bbaldino: The other reason I'm thinking this is: imagine we add another stats service here (RTCStats, maybe): we'd configure that interval within the RTCStats configuration, so it seems like these intervals make sense to be configured alongside the stats service, not whatever transport they may be using.

I agree this is potentially awkward for the 'presence interval' config value: as it would effectively be controlled by 2 config values (one in the client and one here in stats).  I'm not sure what the best course of action would be there.

bgrozev: > At the XMPP level, I mean. I'm not saying it's wrong, it's just an implicit implementation detail that I don't think would be obvious. If I wanted to know how often stats were sent, I wouldn't think to look for the presence update interval.

Would it be more clear to name the property `stats-interval` instead of `presence-interval`?

bgrozev: > we'd configure that interval within the RTCStats configuration, so it seems like these intervals make sense to be configured alongside the stats service, not whatever transport they may be using.

I don't follow. Why does it make sense to configure the interval alongside the stats service? How does it make sense for the stats manager, which now just generates stats and notifies entities that are subscribed to it, to know that a subscriber is RTCStats and therefore it should use a particular interval for it? That's exactly what I'm trying to get away from.

bbaldino: > I don't follow. Why does it make sense to configure the interval alongside the stats service? How does it make sense for the stats manager, which now just generates stats and notifies entities that are subscribed to it, to know that a subscriber is RTCStats and therefore it should use a particular interval for it? That's exactly what I'm trying to get away from.

Sorry, I'm not suggesting that.  I'm suggesting the stats publish interval to some stats service be configured alongside the rest of the configuration for that stats service.

bgrozev: That's exactly what this PR does

bbaldino: Sorry, I do think I forgot that the Callstats interval is alongside the Callstats config and that the muc stats don't have any explicit config, but I was referring to the fact that the MUC stats interval is configured via `videobridge.apis.xmpp-client.presence-interval`, as opposed to something like: `videobridge.stats.muc.interval`.

bbaldino (approved):

Issue #1483, opened 2020-10-09
Increasing memory comsumption in Videobridge


ramon-garcia:
In a production run of Videobridge, we observe an increasing consumption of memory. Without any activity, the instance that started with 250 Mb of heap ends with about 500 Mb at the end of the day.

After obtaining a heap dump with jmap and examining it with visualvm, we find that memory is retained in the buffer pool (PartitionedByteBufferPool) in large amounts, memory is increased and never released.

The reason appears to be a small mistake in the size of the byte arrays. Visualvm shows that the byte arrays stored are frequently of size 1244, 1243 and 1242, but the maximum expected byte array in src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java is 1240. 
```
    private static int T1 = 220;
    private static int T2 = 775;
    private static int T3 = 1240;
```

As a quick fix, please replace 1240 by 1244. 

![javavm_screenshot](https://user-images.githubusercontent.com/1418033/95525138-9e172b00-09d3-11eb-92c4-9c57887c7c93.png)

bbaldino:
Thanks for this report, we recently saw we're encountering the same issue.  A fix will likely be merged today (slightly different than what you have in #1484 , but thank you for the submission)

bbaldino:
Fix is in https://github.com/jitsi/jitsi-videobridge/pull/1485

ramon-garcia:
I can tell you that our problem in production was gone with the threshold of 1244.

I would have place a threshold of 1244 and perhaps an additional threshold of 2000, just in case some large buffer appears. Otherwise there is some waste of memory.

Just out of curiosity, we observe this issue only after the last upgrade. When was it introduced?

JonathanLennox:
We didn't make any change in our system - our current hypothesis is that Chrome started sending bigger packets with the release of Chrome 86.

We decided to make the threshold bigger than the 1244 minimum so that we'd robust against future further increases in the packet sizes that browsers send.

0xnor0:
Seems to be fixed in latest stable release.

bbaldino:
> during our test, I can see ByteBufferPool upto 3000+... is this a memory leak? meaning jvb won't be able to recover the heap memory during garbage collection?
> @bbaldino

[...]

PR #1484, opened 2020-10-09
Fix for issue #1483

ramon-garcia:
Issue #1483  is fixed by increasing the largest buffer size to 1244


bbaldino:
Thanks for this, we're going to fix this issue but in a slightly different way.  It should land today.

bbaldino:
https://github.com/jitsi/jitsi-videobridge/pull/1485

bgrozev:
Thanks for the report and proposed fix! We did basically the same, but increased the size further.

PR #1486, opened 2020-10-10
feat: Port AWS harvester config (update ice4j).

bgrozev:


damencho:
What about these? https://github.com/jitsi/jitsi-videobridge/blob/master/debian/postinst#L106

bgrozev:
We should migrate them, but I think that's independent

bbaldino (approved):

PR #1497, opened 2020-10-19
ref: Refactor BitrateController to make testing easier.

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -272,26 +278,37 @@ private static boolean changeIsLargerThanThreshold(
      * belong to a particular {@link Endpoint}.
      */
     public BitrateController(
-            Endpoint destinationEndpoint,
+            String destinationEndpointId,
+            EventHandler eventHandler,
+            Supplier<List<T>> endpointsSupplier,
```
bbaldino, line 283: There's a bit of inconsistency with the naming around 'endpoint' and 'media source container'

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -94,7 +96,7 @@
  * @author George Politis
  */
 @SuppressWarnings("JavadocReference")
-public class BitrateController
+public class BitrateController<T extends BitrateController.MediaSourceContainer>
```
bbaldino, line 99: Do we need to be generic around this?  Could it just always use the `MediaSourceContainer` type, rather than tying it to some specific type `T`?  I thought this might be there because we need to get the exact type `T` _out_, but I didn't see that (though I may have missed it).  I also see some places that do just use `MediaSourceContainer` instead of `T` (not just the static ones, but members as well I think?)

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bgrozev: Yeah, I left the variable naming as it is, because I imagine changing this in the future, and I don't know what the best option would be.

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bgrozev: The reason for the generics is to allow `Endpoint` to use `conference::getEndpoints` directly. It returns a `List<AbstractEndpoint>` and we need a `List<MediaSourceContainer>`.

bbaldino: Ok, I get that...but the mixing and matching of `List<T>` and `List<MediaSourceContainer>` feels a bit weird, could it all use `T`?

bgrozev: Yes, but it gets hairy. `EndpointMultiRank` becomes generic, so `makeEndpointMultiRankList()` can no longer be static...

bgrozev: Actually, it can be static and generic. I'll push that change.

bbaldino (approved):

PR #1506, opened 2020-10-22
Do Not Merge (yet): Unit test for low-bandwidth-mode screenshare

JonathanLennox:
This is a unit test replicating how the client configures its video constraints for low-bandwidth-mode screenshares.

Do not merge this yet - the test fails, we need to fix the bug.


JonathanLennox:
Attn @gpolitis

Issue #1503, opened 2020-10-22
Parameter specified as non-null is null: parameter statsCollector


pat-s:
After updating my Jitsi instance (v2.0.5142) I see the following in `/var/log/jitsi/jvb.log`.

```
Exception in thread "main" java.lang.IllegalArgumentException: Parameter specified as non-null is null: method org.jitsi.videobridge.rest.binders.ServiceBinder.<init>, parameter statsCollector
        at org.jitsi.videobridge.rest.binders.ServiceBinder.<init>(ServiceBinder.kt)
        at org.jitsi.videobridge.rest.root.Application.<init>(Application.java:37)
        at org.jitsi.videobridge.MainKt.main(Main.kt:129)
```

Looks like a parameter is not set either on my side or upstream. 
Does someone have an how I can solve this to get the videobridge back up?

bbaldino:
Can you post the full jvb log?

pat-s:
<details>
<summary>Details</summary>

```
OpenJDK 64-Bit Server VM warning: Ignoring option UseConcMarkSweepGC; support was removed in 14.0
2020-10-22 13:52:10.456 INFO: [1] JitsiConfig.<clinit>#46: Initialized newConfig: merge of system properties,/etc/jitsi/videobridge/jvb.conf: 1,system properties,reference.conf @ jar:file:/usr/share/jitsi-videobridge/jitsi-videobridge.jar!/reference.conf: 1,reference.conf @ jar:file:/usr/share/jitsi-videobridge/lib/jitsi-media-transform-1.0-198-g1babb83.jar!/reference.conf: 1,reference.conf @ jar:file:/usr/share/jitsi-videobridge/lib/ice4j-3.0-22-g67ffceb.jar!/reference.conf: 1
2020-10-22 13:52:10.491 INFO: [1] ReadOnlyConfigurationService.reloadConfiguration#51: loading config file at path /etc/jitsi/videobridge/sip-communicator.properties
2020-10-22 13:52:10.493 INFO: [1] JitsiConfig.<clinit>#67: Initialized legacyConfig: sip communicator props (no description provided)
2020-10-22 13:52:10.496 INFO: [1] JitsiConfig$Companion.reloadNewConfig#83: Reloading the Typesafe config source (previously reloaded 0 times).
[...]

PR #1507, opened 2020-10-22
fix: Avoid IllegalStateException when StatsCollector is disabled

bbaldino:
Stats can be disabled, and this was causing an issue when crossing a Java -> Kotlin boundary as Kotlin was expecting a non-null type, but this was from java so it didn't get enforced (hence why we didn't see this at compile time).  Jersey/hk2 didn't play too nicely with injecting a nullable type, so I had to play with that a bit and only conditionally inject in the non-null case and mark the injection as `Optional`.  

Fixes https://github.com/jitsi/jitsi-videobridge/issues/1503

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/rest/root/Application.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/rest/root/Application.java
@@ -31,14 +31,14 @@ public class Application extends ResourceConfig
     public Application(
             Videobridge videobridge,
             XmppConnection xmppConnection,
-            StatsCollector statsManager)
+            StatsCollector statsCollector)
 
     {
         register(
             new ServiceBinder(
                 videobridge,
                 xmppConnection,
-                statsManager
+                statsCollector
             )
         );
         // Filters
--- a/jvb/src/main/java/org/jitsi/videobridge/rest/root/colibri/stats/Stats.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/rest/root/colibri/stats/Stats.java
@@ -20,6 +20,7 @@ import org.jitsi.videobridge.rest.*;
 import org.jitsi.videobridge.rest.annotations.*;
 import org.jitsi.videobridge.stats.*;
 import org.json.simple.*;
+import org.jvnet.hk2.annotations.*;
 
 import javax.inject.*;
 import javax.ws.rs.*;
@@ -30,6 +31,7 @@ import javax.ws.rs.core.*;
 public class Stats
 {
     @Inject
+    @Optional
     protected StatsCollector statsManager;
 
     @GET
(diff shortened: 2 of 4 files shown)


JonathanLennox (approved):

avanhoveceovision:
Hello,
We really need a new package with this fix for the new version of our software because the one is promised to many customers.
The last build is not usable and i think he may needs to be remove from the "stable" repo
Kind regards,

VANHOVE Alexis

jlemangarin:
@bbaldino @damencho We can't use the latest stable build of Jitsi Videobridge because of this issue ?

Could you give us a new build or a workaround (how to activate statsCollector) ?

Also is there release notes for your builds ? For exemple we wasn't aware of the main class name change.

Thanks in advance !

damencho:
https://github.com/jitsi/jitsi-meet-release-notes

damencho:
You can use the build from the unstable repo.

jlemangarin:
Thank you for the release notes link, that's really helpful ! :+1: 

I can use the unstable repo for testing purpose for sure but I can't integrate builds marked as 'unstable' in our software as it will be used by French town halls and ministries. 

Security and reliability is a main point for them, especially since the Zoom "polemic" :confused: 

Thanks in advance for your help !

bgrozev:
As a workaround, you should be able to enable the stats collector by adding `videobridge.stats.enabled=true` in jvb.conf.

jlemangarin:
Yes I finally found this configuration, jvb starts but won't register to prosody without error messages (Also it tries to bind to port 8080 which is already in used in my server so I had to shut off the service binded to this port).

I also tried the unstable repo packages but even if jicofo seems to see videobridge

`Jicofo 2020-11-16 17:39:40.806 INFOS: [16] org.jitsi.jicofo.ComponentsDiscovery.log() New component discovered: videobridge.gofast-preprod-comm.ceo-vision.com, JVB(2.1.183-gdbddd169,Linux)`

It's tagged as unavailable : 

```
Jicofo 2020-11-16 17:40:08.288 AVERTISSEMENT: [28] org.jitsi.jicofo.bridge.BridgeSelectionStrategy.log() Failed to select initial bridge for participantRegion=null
Jicofo 2020-11-16 17:40:08.288 GRAVE: [28] org.jitsi.jicofo.JitsiMeetConferenceImpl.log() Can not invite participant -- no bridge available.
```

jlemangarin:
FYI if it can help, last versions working for us are : 

jicofo-1.0_549-2
jitsi-meet-xxxx-1.0.4025-2
jitsi-videobridge2-2.1_183_gdbddd169-2

jlemangarin:
@damencho Could you get back to me about that please ? Thank you !

bgrozev:
@jlemangarin you can apply the `videobridge.stats.enabled=true` workaround to the stable version.



> Jicofo 2020-11-16 17:39:40.806 INFOS: [16] org.jitsi.jicofo.ComponentsDiscovery.log() New component discovered: videobridge.gofast-preprod-comm.ceo-vision.com, JVB(2.1.183-gdbddd169,Linux)
Note that this is not what you might expect: it only indicates that prosody's configuration includes this component (not that it is connected). The component code in jvb has long been gone, so nothing will actually connect there.

jlemangarin:
@bgrozev Yes it works, also the old way to connect videobridge (through components) is not supported anymore so I had to make this change.

PR #1508, opened 2020-10-23
Properly prioritize screen share low bandwidth mode

gpolitis:
Fixes https://github.com/jitsi/jitsi-videobridge/pull/1506


JonathanLennox (approved):
I don't love this fix, but I believe this code is the code that'll get cleaned up in the refactoring that @bgrozev is doing, so it'll do for now.

PR #1510, opened 2020-10-28
Chore: Update ktlint version

JonathanLennox:


jvb/src/main/kotlin/org/jitsi/videobridge/cc/config/BitrateControllerConfig.kt
```diff
@@ -105,7 +105,7 @@ class BitrateControllerConfig {
          * participant.
          */
         private val onstageIdealHeightPx: Int by
-            config("videobridge.cc.onstage-ideal-height-px".from(JitsiConfig.newConfig))
+        config("videobridge.cc.onstage-ideal-height-px".from(JitsiConfig.newConfig))
```
bbaldino, line 108: Hm, these don't feel right

bbaldino: If this is the rule, then we should probably move the `config` to the same line after the `by` and do:

```
private val onstageIdealHeightPx: Int by config(
    "videobridge.cc.onstage-ideal-height-px".from(JitsiConfig.newConfig)
)
```

JonathanLennox: Done, take a look?

bbaldino (approved):

PR #1512, opened 2020-10-28
fix: Fix raw type warnings (reduce use of generics, use List instead of array).

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -823,9 +823,9 @@ private AdaptiveSourceProjection lookupOrCreateAdaptiveSourceProjection(
             System.arraycopy(newRatedTargetIndices, 0, oldRatedTargetIndices, 0, oldRatedTargetIndices.length);
 
             int newStateLen = 0;
-            for (int i = 0; i < sourceBitrateAllocations.length; i++)
+            for (int i = 0; i < sourceBitrateAllocations.size(); i++)
             {
-                SourceBitrateAllocation sourceBitrateAllocation = sourceBitrateAllocations[i];
+                SourceBitrateAllocation sourceBitrateAllocation = sourceBitrateAllocations.get(i);
```
bbaldino, line 828: I think there's a perf hit here (vs the array access).  This code gets called often but not sure if it's concerning or not.

bbaldino (approved):

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bgrozev: It's an `ArrayList`, so accessing by index only adds a little overhead over using the array directly.

bbaldino: I was thinking about the extra bounds checking it does, but looks like it's just an if so it's probably fine.  Will be nice to get rid of the array here.

PR #1515, opened 2020-11-03
Update: Use latest jicoco.

gpolitis:


bgrozev (approved):

JonathanLennox:
When you merge, be sure to mark the commit message with what changed in jicoco?

It looks like this is "ref: Disable smack reconnects & solely rely on periodic checks".

PR #1517, opened 2020-11-05
Bitrate allcation tests

bgrozev:
This adds tests for the allocations that BitrateController performs -- which
layers are selected for each endpoint in different conditions (selections from the
client and available bandwidth).

Some of the internals of BitrateController are exposed to help with the tests. The aim
is to have tests, so we can make more substantial refactoring (coming in another PR),
with more confidence.

- test: Add tests for bitrate allocation.
- fix: Fix debug JSON format.
- ref: Expose SourceBitrateAllocation and fire an event when the allocation changes.
- test: Add tests for bitrate allocation.
- ref: Remove redundant code.
- ref: Add @NotNull annotations.
- test: Add tests for VideoConstraintsCompatibility.
- ref: Move VideoConstraintsCompatibility to the cc package.
- ref: Remove redundant null check.
- ref: Make OversendingTimeTracker generic, remove redundant field.


jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -418,8 +415,8 @@ public JSONObject getDebugState()
         debugState.put("effectiveVideoConstraints", effectiveConstraintsMap);
         debugState.put("lastN", lastN);
         debugState.put("supportsRtx", supportsRtx);
-        debugState.put("oversending", oversending);
-        debugState.put("total_oversending_time_secs", oversendingTimeTracker.totalOversendingTime().getSeconds());
+        debugState.put("oversending", oversendingTimeTracker);
```
bbaldino: Don't think this is going to render anything useful.

jvb/src/test/kotlin/org/jitsi/videobridge/cc/BitrateControllerTest.kt
```diff
@@ -82,13 +103,206 @@ class BitrateControllerTest : ShouldSpec() {
...
+    val time: Instant,
+    val bwe: Bandwidth,
+    val event: T) {
+    override fun toString(): String {
+        return "\n[time=${time.toEpochMilli()} bwe=$bwe] $event"
+    }
+}
+
+/**
+ * List the normal List<T>.shouldContainInOrder, but compare elements' contents.
```
JonathanLennox, line 472: Grammar? Do you mean "Like the normal..."?

jvb/src/test/kotlin/org/jitsi/videobridge/cc/BitrateControllerTest.kt
```diff
@@ -82,13 +103,206 @@ class BitrateControllerTest : ShouldSpec() {
                 )
             }
         }
+
+        context("Allocation") {
```
JonathanLennox, line 107: I suggest this should have tests for Last-N, audio-only, and pinning modes, too?

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
```diff
@@ -410,7 +410,7 @@ public JSONObject getDebugState()
         debugState.put("effectiveVideoConstraints", effectiveConstraintsMap);
         debugState.put("lastN", lastN);
         debugState.put("supportsRtx", supportsRtx);
-        debugState.put("oversending", oversendingTimeTracker);
+        debugState.put("oversending", oversendingTimeTracker.getState());
```
bbaldino, line 413: Oops, you caught it already :)

jvb/src/test/kotlin/org/jitsi/videobridge/cc/BitrateControllerTest.kt
```diff
@@ -82,13 +103,206 @@ class BitrateControllerTest : ShouldSpec() {
...
+    val event: T) {
+    override fun toString(): String {
+        return "\n[time=${time.toEpochMilli()} bwe=$bwe] $event"
+    }
+}
+
+/**
+ * List the normal List<T>.shouldContainInOrder, but compare elements' contents.
+ */
+fun <T> List<Collection<T>>.shouldContainInOrder(vararg ts: Collection<T>) {
```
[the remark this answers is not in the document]
bbaldino: Sorry, I missed the comment and see the difference now.

jvb/src/main/java/org/jitsi/videobridge/cc/BitrateController.java
bgrozev: It got caught by unit tests :)

bbaldino: 👍  I got saved by those recently too

jvb/src/test/kotlin/org/jitsi/videobridge/cc/VideoConstraintsCompatibilityTest.kt
```diff
@@ -0,0 +1,28 @@
+package org.jitsi.videobridge.cc
+
+import io.kotest.core.spec.style.ShouldSpec
+import io.kotest.matchers.maps.shouldContainExactly
+import org.jitsi.videobridge.VideoConstraints
+
+class VideoConstraintsCompatibilityTest : ShouldSpec() {
```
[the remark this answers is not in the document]
bgrozev: I didn't see any difference in signaling selected/pinnned endpoints and maxFrameHeight with screensharing. The only difference seems to be that the selected endpoint is not always the dominant speaker. Am I wrong? If not, I'll add tests for the case with a selected endpooint which is not dominant speaker.

I'm actually not sure how audio-only mode is implemented. I'll check and add tests.

I am not aware of any case in which the client uses pinning. Am I wrong?

@jallamsetty1 I think you may be better able to help with some of those questions.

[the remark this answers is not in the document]
JonathanLennox: I meant "pinned" in the UI sense, i.e. when you explicitly click on someone in stage view rather than letting it be by dominant speaker, rather than meaning the signaling message.

Probably this isn't different for the CompatibilityTest, but it is for the BitrateController proper.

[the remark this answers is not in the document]
jallamsetty1: For client-only mode, we set lastN to 0. When a share is detected while the client is in audio-only mode, lastN is set to 1 and the share is automatically pinned.
I think desktop share and youtube share are the only sources which are automatically pinned by the client.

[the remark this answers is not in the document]
bgrozev: Great, thanks for the input! I added:
1. Tests for the "pinned in the UI" case (i.e. non-dominant speaker is on stage)
2. Tests for audio-only mode (i.e. LastN=0)
3. Tests for audio-only + screenshare (i.e. LastN=1 + a selected endpoint, possibly not the dominant speaker).

I also discovered I was not setting the `eid` of layers correctly, which explains at least some of the weird behavior.

JonathanLennox (approved):

PR #1521, opened 2020-11-16
Update docs, add hint about which file to edit

andreasnuesslein:
I was looking for quite some time until I realized that the relevant file was `sip-communicator.properties`. I figure others might have this problem as well.

### DIFF
--- a/doc/rest.md
+++ b/doc/rest.md
@@ -18,7 +18,7 @@ private interface, but for the public interface additional properties are requir
 Configuration
 ==============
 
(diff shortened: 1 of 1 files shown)


bbaldino:
Hm, I swear I updated this doc to refer to the new config file (jvb.conf), but it looks like I still have it in a branch.  I'll try and get a PR open with that.

bbaldino:
Updated the doc in https://github.com/jitsi/jitsi-videobridge/pull/1523, thanks for the reminder!

andreasnuesslein:
hehe okay that works too :)

thanks

PR #1526, opened 2020-11-17
Split BitrateController into an allocator and packet handler.

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -289,8 +290,15 @@ public void trace(@NotNull Function0<Unit> f)
                 }
             });
 
-        BitrateController.EventHandler bcEventHandler = new BitrateController.EventHandler()
+        org.jitsi.videobridge.cc.allocation.EventHandler bcEventHandler
+                = new org.jitsi.videobridge.cc.allocation.EventHandler()
```
bbaldino, line 294: I think making the event handler a static inner class of the class firing the events is what we've done elsewhere and makes the scope of it a bit more obvious.

bbaldino (commented):
There's a lot here and it was hard to tell what's new and what was moved, but generally seems ok to me (1 minor comment)

bbaldino (approved):

PR #1563, opened 2021-02-02
Allow enabling/disabling buffer pool bookkeeping via (private) REST API

bbaldino:
This PR implements buffer pool "bookkeeping" as a "debug feature" which can be turned on and off using the existing private rest api.  The "get stats" path does not apply for this feature, as this feature doesn't maintain stats, just logs issues.  We could look at changing that, but I think that should probably be in a different PR anyhow.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
@@ -284,6 +284,11 @@ public class ByteBufferPool
     public static void enableBookkeeping(boolean enable)
     {
         bookkeepingEnabled = enable;
+        if (!enable)
+        {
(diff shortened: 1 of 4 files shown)


jvb/src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
```diff
@@ -282,4 +281,14 @@ public static boolean statisticsEnabled()
         return enableStatistics;
     }
 
+    public static void enableBookkeeping(boolean enable)
+    {
+        bookkeepingEnabled = enable;
```
JonathanLennox, line 286: I suggest that disabling bookkeeping should clear the `bookkeeping` and `returnedBookkeeping` maps

bgrozev (approved):

PR #1564, opened 2021-02-08
Send sender constraints when endpoint message transport first connects.

JonathanLennox:


jvb/src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1123,13 +1128,18 @@ public void requestKeyframe()
...
     {
+        boolean wasEmpty = transceiver.getMediaSources().length == 0;
         if (transceiver.setMediaSources(mediaSources))
         {
             eventEmitter.fireEventSync(handler -> {
                 handler.sourcesChanged();
                 return Unit.INSTANCE;
             });
         }
+        if (wasEmpty)
```
bgrozev, line 1139: Does this need to execute when there was no change to the sources (i.e. shoul it be under the `if` above)|?

JonathanLennox: If the media sources is still empty, this will be a no-op when it hits the empty media source list in `sendVideoConstraints`, but I could move it inside if you think it'd be cleaner.

bgrozev (approved):

PR #1566, opened 2021-02-12
Change BufferPool bookkeeping

bbaldino:
While trying to track down some buffer issues, I found that we were hitting hash collisions using `System.identityHashCode` on buffers as a key in our bookkeeping.  @JonathanLennox found `IdentityHashMap` which uses a combination of `identityHashCode` and equality to compare and that has fared better, so this PR changes the BufferPool bookkeeping structures to use `IdentityHashMap`.

I also found that we can just store an `Exception` of the stacktraces and expand it into a `String` only when needed, which appears to give a performance boost, so this PR changes the tracking to use `Exception` instead of `String`.

We also used to log on every buffer request and return, but that didn't give a complete picture and was very chatty.  I changed it to instead track a timeline per buffer of all requests and returns which makes tracking down issues easier.

The intermittent commits here were due to some debugging and are completely "undone", so the best way to view this would be to view the entire diff (not commit by commit).  I'll be getting rid of those commit messages when merging.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
@@ -72,7 +72,8 @@ public class ByteBufferPool
     private static final Map<byte[], Exception> bookkeeping
             = Collections.synchronizedMap(new IdentityHashMap<>());
 
-    private static final Map<byte[], Queue<BufferEvent>> bufferEvents = Collections.synchronizedMap(new IdentityHashMap<>());
+    private static final Map<byte[], Queue<BufferEvent>> bufferEvents =
+        Collections.synchronizedMap(new IdentityHashMap<>());
 
     private static class ReturnedBufferBookkeepingInfo
     {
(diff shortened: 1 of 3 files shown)


jvb/src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
```diff
@@ -68,22 +69,25 @@
...
         {
             allocTrace = a;
             deallocTrace = d;
         }
     }
 
-    private static final Map<Integer, ReturnedBufferBookkeepingInfo> returnedBookkeeping
-        = new ConcurrentHashMap<>();
+    private static final Map<byte[], ReturnedBufferBookkeepingInfo> returnedBookkeeping
+        = Collections.synchronizedMap(new IdentityHashMap<>());
```
JonathanLennox, line 90: It looks like you're now just using the returnedBookkeeping as a Set to remember what's been returned - does it need to be a map at all?  I think you could do 
```
private static final Set<byte[]> returnedBookkeeping = Collections.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
```

And eliminate `ReturnedBufferBookkeepingInfo`.

bbaldino: Yeah I think I can do the same thing for `bookkeeping` as well.

jvb/src/main/java/org/jitsi/videobridge/util/ByteBufferPool.java
```diff
@@ -68,22 +69,14 @@
      * A debug data structure which tracks outstanding buffers and tracks from where (via
      * a stack trace) they were requested and returned.
```
JonathanLennox, line 70: This comment should be updated now.

JonathanLennox (approved):

PR #1567, opened 2021-02-13
Fix scenario where endpoint connection status can get stuck on 'inactive'

bbaldino:
@paweldomas noticed a scenario where, if the first time the `EndpointConnectionStatusMonitor` checked a given endpoint's status and it had no activity at that time, we would notify endpoints that it was inactive but not _track_ it as having been marked as inactive in the bridge, so we'd never send an update when it went active (since the bridge didn't remember it had been marked as inactive).

I still need to look over this a bit more, but so far the unit test seems to confirm the issue and the fix.  Putting it up so @paweldomas can take a look.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitor.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitor.kt
@@ -85,6 +85,9 @@ class EndpointConnectionStatusMonitor @JvmOverloads constructor(
                     "${endpoint.id} is having trouble establishing the connection " +
                         "and will be marked as inactive"
                 }
+                synchronized(inactiveEndpointIds) {
+                    inactiveEndpointIds += endpoint.id
+                }
                 notifyStatusChange(endpoint.id, false, null)
                 return
             } else {
--- a/jvb/src/test/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitorTest.kt
+++ b/jvb/src/test/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitorTest.kt
@@ -104,9 +104,14 @@ class EndpointConnectionStatusMonitorTest : ShouldSpec({
                     broadcastCalls shouldHaveSize 2
                     broadcastCalls.forAny { (msg, sendToOcto) ->
                         sendToOcto && msg.endpoint == "1" && msg.active == "false"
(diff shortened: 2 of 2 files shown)


paweldomas (approved):

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitor.kt
```diff
@@ -92,6 +92,11 @@ class EndpointConnectionStatusMonitor @JvmOverloads constructor(
                 return
             } else {
                 logger.cdebug { "${endpoint.id} not ready for activity checks yet" }
+                // It's possible this endpoint was previously on the inactive list but has
+                // been recreated, so try removing it from the inactive list just in case.
+                synchronized(inactiveEndpointIds) {
+                    inactiveEndpointIds -= endpoint.id
```
bgrozev, line 98: Okay, I'm rethinking this now...if we removed it from the list we should notify the other endpoints (otherwise the update will not be triggered when it connects). Maybe we should keep it in the list until it either connects or times out?

bbaldino: When it reconnects, we won't have sent any status update for it to the other endpoints, right?  And they assume 'active' by default?  If this is too sticky, maybe it's worth doing this fix separately.

bgrozev: I'm thinking: 
1. endpoint connects
2. it loses connectivity, we send "inactive" to other endpoints
3. it signals to jicofo that ice failed, jicofo re-creates it
4. it connects again (as a new Endpoint instance, but already in the `inactiveEndpointIds` list

bbaldino: Ah ok, I thought endpoints were active by default until the bridge signaled otherwise to the clients so merely removing it from the `inactiveEndpoints` list would be sufficient, but yes I see now we query `inactiveEndpoints` when an EP joins.  How about I take this commit off now so we can merge the first fix and we can think about how we want to handle this case?

bgrozev: Sure.

bgrozev (approved):

PR #1571, opened 2021-02-18
feat: Add debug logs for bandwidth allocation.

bgrozev:


JonathanLennox (commented):
I believe debug logs are enabled on PR testing - are we sure this won't be too noisy?

bgrozev:
Good point (the test run will probably fail because of that). I'll move to TRACE

JonathanLennox (approved):

PR #1576, opened 2021-02-23
update jmt to get dtls buffer leak fix

bbaldino:
See https://github.com/jitsi/jitsi-media-transform/pull/330


bbaldino:
No longer needed, used `java.util.ArrayDeque` explicitly to avoid the need for the experimental stdlib annotation in jmt.

PR #1577, opened 2021-02-23
perf: Batch Octo endpoint creation (do not call addEndpoint for each one individually).

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -831,30 +831,29 @@ void endpointExpired(AbstractEndpoint endpoint)
...
-                "conference = " + endpoint.getConference());
-        }
+        endpoints.forEach(endpoint -> {
+            if (endpoint.getConference() != this)
+            {
+                throw new IllegalArgumentException("Endpoint belong to other " +
+                        "conference = " + endpoint.getConference());
+            }
+            AbstractEndpoint replacedEndpoint = endpointsById.put(endpoint.getId(), endpoint);
+            endpointsById.put(endpoint.getId(), endpoint);
```
JonathanLennox, line 845: You're calling `put()` twice here.

JonathanLennox (approved):

PR #1588, opened 2021-02-26
fix: Re-use the ObjectMapper when serializing.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -298,16 +286,6 @@ class ForwardedEndpointsMessage(
      */
     val forwardedEndpoints: Collection<String>
 ) : BridgeChannelMessage(TYPE) {
-    /**
-     * Serialize using json-simple because it's faster.
-     */
-    override fun toJson(): String = JSONObject().apply {
```
bbaldino, line 304: So for these we're ok with now falling back to the jackson serialization instead of providing custom string concat implementations?

JonathanLennox: Yeah, they were the ones which were too complex for string concatenation, so we used json-simple, and it looks like with proper mapper re-use, Jackson is just about as fast as json-simple.

This also lets us avoid the hacks like having to convert `Set` to `List` to make json-simple happy.

JonathanLennox:
Jenkins, test this please

JonathanLennox (approved):

PR #1590, opened 2021-02-26
feat: Handle Colibri requests asynchronously.

bgrozev:
This is somewhat awkward for two reasons:
1. We need to maintain the synchronous handling because of [COLIBRI over REST](https://github.com/jitsi/jitsi-videobridge/blob/master/jvb/src/main/java/org/jitsi/videobridge/rest/root/colibri/conferences/Conferences.java#L171)
2. The existance of VideobridgeShim. I think we can just merge the remainder of it in ConferenceShim, but I'd rather not do the change right now.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -236,17 +236,16 @@ public class Conference
                     // TODO: we can avoid reaching into Videobridge here by merging VideobridgeShim into ConferenceShim
                     long start = System.currentTimeMillis();
                     IQ response = videobridge.handleColibriConferenceIQ(this, request.getRequest());
-                    long end = System.currentTimeMillis() - request.getReceiveTime();
+                    long end = System.currentTimeMillis();
                     long processingDelay = end - start;
-                    request.getProcessingDelayStats().addDelay(processingDelay);
-
                     long totalDelay = end - request.getReceiveTime();
-                    if (totalDelay > 100)
-                    {
-                        logger.warn("Took " + totalDelay + " ms to handle IQ (processing took "
-                                + processingDelay + " ms): " + request.getRequest().toXML());
-                    }
+                    request.getProcessingDelayStats().addDelay(processingDelay);
                     request.getTotalDelayStats().addDelay(totalDelay);
+                    if (processingDelay > 100)
+                    {
+                        logger.warn("Took " + processingDelay + " ms to process an IQ (total delay "
+                                + totalDelay + " ms): " + request.getRequest().toXML());
+                    }
                     request.getCallback().invoke(response);
                 }
                 catch (Throwable e)
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -17,7 +17,6 @@ package org.jitsi.videobridge;
 
 import org.jetbrains.annotations.*;
 import org.jitsi.nlj.*;
-import org.jitsi.rtp.*;
 import org.jitsi.rtp.Packet;
 import org.jitsi.rtp.rtcp.rtcpfb.payload_specific_fb.*;
 import org.jitsi.rtp.rtp.*;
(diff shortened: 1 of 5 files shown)


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -219,6 +225,35 @@ public Conference(Videobridge videobridge,
         videobridgeStatistics.totalConferencesCreated.incrementAndGet();
         epConnectionStatusMonitor = new EndpointConnectionStatusMonitor(this, TaskPools.SCHEDULED_POOL, logger);
         epConnectionStatusMonitor.start();
+
+        colibriQueue= new PacketQueue<>(
```
JonathanLennox, line 229: Formatting

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -219,6 +225,35 @@ public Conference(Videobridge videobridge,
...
+                        logger.warn("Took " + delay + " ms to handle IQ: " + request.getRequest().toXML());
+                    }
+                    request.getDelayStats().addDelay(delay);
+                    request.getCallback().invoke(response);
+                }
+                catch (Throwable e)
+                {
+                    logger.warn("Failed to handle colibri request: ", e);
+                    request.getCallback().invoke(
+                            IQUtils.createError(request.getRequest(), XMPPError.Condition.internal_server_error));
```
JonathanLennox, line 251: Should pass `e.getMessage()` to `createError`, as in jicoco.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -219,6 +225,35 @@ public Conference(Videobridge videobridge,
...
+            true,
+            "colibri-queue-" + id,
+            request ->
+            {
+                try
+                {
+                    // TODO: we can avoid reaching into Videobridge here by merging VideobridgeShim into ConferenceShim
+                    IQ response = videobridge.handleColibriConferenceIQ(this, request.getRequest());
+                    long delay = System.currentTimeMillis() - request.getReceiveTime();
+                    if (delay > 100)
```
JonathanLennox, line 240: Current delay is just the processing time of the IQ itself, not the time since the request was received - i.e. doesn't include the time the request was queued. I think we still want that.

bgrozev: Why do we want just the processing and not the queueing delay? I can see it can be useful when we're profiling, but the overall delay in the response seems like a more useful metric to have in production.

JonathanLennox: Well, the existing logs and stats show just the processing delay, no?  Since any queueing happens inside the XMPP stack.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -235,13 +234,19 @@ public Conference(Videobridge videobridge,
                 try
                 {
                     // TODO: we can avoid reaching into Videobridge here by merging VideobridgeShim into ConferenceShim
+                    long start = System.currentTimeMillis();
                     IQ response = videobridge.handleColibriConferenceIQ(this, request.getRequest());
-                    long delay = System.currentTimeMillis() - request.getReceiveTime();
-                    if (delay > 100)
+                    long end = System.currentTimeMillis() - request.getReceiveTime();
```
JonathanLennox, line 239: Subtracting `getReceiveTime()` here doesn't look right.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -235,13 +234,19 @@ public Conference(Videobridge videobridge,
...
+                    long processingDelay = end - start;
+                    request.getProcessingDelayStats().addDelay(processingDelay);
+
+                    long totalDelay = end - request.getReceiveTime();
+                    if (totalDelay > 100)
                     {
-                        logger.warn("Took " + delay + " ms to handle IQ: " + request.getRequest().toXML());
+                        logger.warn("Took " + totalDelay + " ms to handle IQ (processing took "
```
JonathanLennox, line 246: I'd argue that the warning should be based on processing delay, not total delay?

JonathanLennox (approved):

JonathanLennox (approved):

PR #1593, opened 2021-03-01
Partial refactor of Endpoint (and some associated classes) to Kotlin

bbaldino:
We ran into this when trying to make `Bandwidth` an inline class, which causes some issues when accessed from Java.  Here I tried to keep things to mostly just what was needed to support this change to keep the PR small.  Going through commit-by-commit would be easiest.


jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -126,6 +126,28 @@ class EndpointK @JvmOverloads constructor(
         }
     }
 
+    override fun send(packetInfo: PacketInfo) {
+        when (val packet = packetInfo.packet) {
+            is VideoRtpPacket -> {
+                if (bitrateController.transformRtp(packetInfo)) {
+                    // The original packet was transformed in place.
+                    transceiver.sendPacket(packetInfo)
```
bgrozev: This sends the packet twice? The original has a `return`

bbaldino: Yeah, I found this and fixed it in a later commit.  You could've saved me some debugging time ;)

jvb/src/main/java/org/jitsi/videobridge/Endpoint.java
```diff
@@ -1118,22 +1118,8 @@ public JSONObject getDebugState()
...
-    public boolean isSendingAudio()
-    {
-        // The endpoint is sending audio if we (the transceiver) are receiving audio.
-        return transceiver.isReceivingAudio();
-    }
 
-    @Override
-    public boolean isSendingVideo()
-    {
-        // The endpoint is sending video if we (the transceiver) are receiving video.
```
bgrozev, line 1468: I find the comment useful to remind of the perspective

bbaldino: Yeah I wondered about those, too.  Brought them back.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -37,6 +46,56 @@ class EndpointK @JvmOverloads constructor(
...
+        iceTransport.incomingDataHandler = object : IceTransport.IncomingDataHandler {
+            override fun dataReceived(data: ByteArray, offset: Int, length: Int, receivedTime: Instant) {
+                // DTLS data will be handled by the DtlsTransport, but SRTP data can go
+                // straight to the transceiver
+                if (looksLikeDtls(data, offset, length)) {
+                    // DTLS transport is responsible for making its own copy, because it will manage its own
+                    // buffers
+
+                    // DTLS transport is responsible for making its own copy, because it will manage its own
+                    // buffers
```
JonathanLennox, line 131: Double comment.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -96,6 +98,34 @@ class EndpointK @JvmOverloads constructor(
...
+        dtlsTransport.incomingDataHandler = object : DtlsTransport.IncomingDataHandler {
+            override fun dtlsAppDataReceived(buf: ByteArray, off: Int, len: Int) {
+                this@EndpointK.dtlsAppPacketReceived(buf, off, len)
+            }
+        }
+        dtlsTransport.outgoingDataHandler = object : DtlsTransport.OutgoingDataHandler {
+            override fun sendData(buf: ByteArray, off: Int, len: Int) {
+                iceTransport.send(buf, off, len)
+            }
+        }
```
JonathanLennox, line 181: Hm, I'm surprised the Kotlin syntax for these two handlers is so much heavier than the Java syntax, that's the opposite of what I'd usually expect.  There's not simpler Kotlin syntax for this?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -163,6 +165,18 @@ class EndpointK @JvmOverloads constructor(
...
+    override fun setFeature(feature: EndpointDebugFeatures, enabled: Boolean) {
+        when (feature) {
+            EndpointDebugFeatures.PCAP_DUMP -> transceiver.setFeature(Features.TRANSCEIVER_PCAP_DUMP, enabled)
+        }
+    }
+
+    override fun isFeatureEnabled(feature: EndpointDebugFeatures): Boolean {
+        return when (feature) {
+            EndpointDebugFeatures.PCAP_DUMP -> transceiver.isFeatureEnabled(Features.TRANSCEIVER_PCAP_DUMP)
+        }
```
JonathanLennox, line 252: You're no longer throwing an exception for an unsupported feature - is this deliberate?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -361,4 +361,46 @@ class EndpointK @JvmOverloads constructor(
...
+                }
+            }
+            updateStatsOnExpire()
+            transceiver.stop()
+            logger.cdebug { transceiver.getNodeStats().prettyPrint(0) }
+            logger.cdebug { bitrateController.debugState.toJSONString() }
+            logger.cdebug { iceTransport.getDebugState().toJSONString() }
+            logger.cdebug { dtlsTransport.getDebugState().toJSONString() }
+
+            logger.info("Spent ${bitrateController.getTotalOversendingTime().seconds} seconds oversending")
```
JonathanLennox, line 452: Could this use `cinfo`, for parallel to the above logs?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -0,0 +1,490 @@
...
+        }
+    }
+
+    override fun isSendingAudio(): Boolean {
+        // The endpoint is sending audio if we (the transceiver) are receiving audio.
+        return _transceiver.isReceivingAudio()
+    }
+    override fun isSendingVideo(): Boolean {
+        // The endpoint is sending video if we (the transceiver) are receiving video.
+        return _transceiver.isReceivingAudio()
```
JonathanLennox, line 261: Copy & paste error - this should be `isReceivingVideo()`.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: That was just for safety as a future fall through since there's only one enum value, but kotlin `when` errors when it isn't exhaustive so we don't need the fall through anymore.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: SAM conversions for kotlin interfaces isn't in until 1.4 :/

bbaldino: We could define a typealias for the handler instead of an interface, but these are annoying because they can't be scoped.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: 😅  nice catch

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: There isn't a `cinfo` defined since we don't ever worry about guarding `INFO` logs with checks.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
JonathanLennox: Another reason to get around to the 1.4 update...

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
JonathanLennox: ??? https://github.com/jitsi/jitsi-utils/blob/5d25e9b4a5c19f52dba48fdc3e7804a68a1eaf29/src/main/kotlin/org/jitsi/utils/logging2/LoggerExtensions.kt#L58

bbaldino: Ah, weird...it didn't auto complete for me.  Still we haven't really used `cinfo` anywhere, I think I'd rather leave that for a single, focused effort to make all info logs consistent if we want that.

JonathanLennox: Ok.

`getTotalOversendingTime()` doesn't look to be too expensive, either, even if `INFO` logs were turned off.

jvb/src/main/java/org/jitsi/videobridge/cc/BandwidthProbing.java
```diff
@@ -30,27 +30,27 @@
 /**
   * @author George Politis
   */
- public class BandwidthProbing
+ public abstract class BandwidthProbing
```
JonathanLennox, line 33: `BandwidthProbing` isn't that big a class - would it be easier/cleaner just to convert the whole thing to Kotlin?

bbaldino: Sure, I can.  I was just trying to keep things to the bare minimum needed for the `Bandwidth` class change.  Think I should do the rest as part of this? If not I can do it in the follow up (of which there will probably be a couple, I'm trying to keep diff sizes reasonable for PRs)

JonathanLennox: For `BandwidthProbing` specifically, I think that moving it in its entirety isn't that much bigger than this change, because of all the `protected` changes and the like that the Java/Kotlin split implies.

bbaldino: The remaining changes weren't trivial and I'm a bit nervous given how critical BandwidthProbing is, but I pushed the change and we can decide to keep it here or split it out.

jvb/src/main/kotlin/org/jitsi/videobridge/cc/BandwidthProbing.kt
```diff
@@ -0,0 +1,152 @@
...
+            return
+        }
+
+        val latestBweCopy = latestBwe
+        if (bitrateControllerStatus.currentIdealBps <= latestBweCopy) {
+            // it seems like the ideal bps fits in the bandwidth estimation,
+            // let's update the bitrate controller.
+            // TODO(brian): this trigger for a bitratecontroller update seems awkward and may now be obsolete
+            //  since i now update it every time we get an updated estimate from bandwidth estimator
+//             dest.getBitrateController().update(bweBps);
```
JonathanLennox, line 91: I guess update the commented-out code to Kotlin?

bbaldino: It looks like this wasn't even valid in java anymore, as `dest` no longer exists.

bbaldino: We can probably just remove this altogether?

JonathanLennox: Yeah, that makes sense.  The place for code history is version control, not comments.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -126,6 +128,15 @@ class EndpointK @JvmOverloads constructor(
         }
     }
 
+    override fun addPayloadType(payloadType: PayloadType) {
+        transceiver.addPayloadType(payloadType)
+        bitrateController.addPayloadType(payloadType)
+    }
+
+    override fun addRtpExtension(rtpExtension: RtpExtension) {
+        transceiver.addRtpExtension(rtpExtension)
```
JonathanLennox: You didn't want to use `=` notation for this?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -362,6 +363,20 @@ class EndpointK @JvmOverloads constructor(
         }
     }
 
+    override fun getDebugState(): JSONObject {
+        return super.getDebugState().apply {
+            // debugState.put("sctpManager", sctpManager.getDebugState());
```
JonathanLennox, line 420: Either convert the commented-out code to Kotlin, or remove it.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -362,6 +363,20 @@ class EndpointK @JvmOverloads constructor(
...
+            put("bitrateController", bitrateController.debugState)
+            put("bandwidthProbing", bandwidthProbing.debugState)
+            put("iceTransport", iceTransport.getDebugState())
+            put("dtlsTransport", dtlsTransport.getDebugState())
```
JonathanLennox, line 424: I see we're also inconsistent between `val debugState` and `fun getDebugState()`; we should regularize these too, but not for this PR.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -58,14 +63,47 @@ class EndpointK @JvmOverloads constructor(
...
+    // TODO: this naming is to avoid conflicts with getTransceiver in Endpoint.  It will change back
+    // once Endpoint.java goes away
+    val _transceiver = Transceiver(
+        id,
+        TaskPools.CPU_POOL,
+        TaskPools.CPU_POOL,
+        TaskPools.SCHEDULED_POOL,
+        diagnosticContext,
+        logger,
+        TransceiverEventHandlerImpl(),
```
JonathanLennox, line 75: `Endpoint` used to directly hold a reference to the `TransceiverEventHandlerImpl` instance - you decided this was not needed?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -84,9 +85,9 @@ class EndpointK @JvmOverloads constructor(
         })
     }
 
-    private val bandwidthProbing = BandwidthProbing(
-        _transceiver::sendProbing,
-        bitrateController::getStatusSnapshot
+    private val bandwidthProbing = BandwidthProbingK(
+        BandwidthProbing.ProbingDataSender { mediaSsrcs, numBytes -> _transceiver.sendProbing(mediaSsrcs, numBytes) },
+        java.util.function.Supplier { bitrateController.getStatusSnapshot() }
```
JonathanLennox, line 93: Now that this is Kotlin can it use a Kotlin idiom here rather than `java.util.function.Supplier`?

jvb/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
```diff
@@ -557,13 +557,13 @@ OrderedJsonObject toJson()
...
             debugState.put("bytes_received", bytesReceived);
-            debugState.put("receive_bitrate_bps", receiveBitRate.getRate().getBps());
+            debugState.put("receive_bitrate_bps", receiveBitRate.getRate());
             debugState.put("incoming_packets_dropped", incomingPacketsDropped);
 
             debugState.put("packets_sent", packetsSent.sum());
             debugState.put("send_packet_rate_pps", sendPacketRate.getRate());
             debugState.put("bytes_sent", bytesSent.sum());
-            debugState.put("send_bitrate_bps", sendBitRate.getRate().getBps());
+            debugState.put("send_bitrate_bps", sendBitRate.getRate());
```
JonathanLennox, line 566: These changes are taking advantage of the fact that the inlined `Bandwidth` looks like a `double` to Java - we're okay with that?

jvb/src/main/kotlin/org/jitsi/videobridge/cc/BandwidthProbing.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * The sequence number to use if probing with the JVB's SSRC.
+     */
+    private var seqNum = Random.nextInt(0xFFFF)
+
+    /**
+     * The RTP timestamp to use if probing with the JVB's SSRC.
+     */
+    private var ts: Long = Random.nextLong().and(0xFFFFFFFFL)
```
JonathanLennox, line 43: Java code did `nextInt()` not `nextLong()` - since this is only a 32-bit value, `nextInt()` should be sufficient, and the `and` will force it to be a positive `Long`.

jvb/src/main/kotlin/org/jitsi/videobridge/cc/BandwidthProbing.kt
```diff
@@ -0,0 +1,152 @@
...
+
+        // We calculate how much to probe for based on the total target bps
+        // (what we're able to reach), the total ideal bps (what we want to
+        // be able to reach) and the total current bps (what we currently send).
+        val bitrateControllerStatus = statusSnapshotSupplier.get()
+
+        // How much padding do we need?
+        val totalNeededBps = bitrateControllerStatus.currentIdealBps - bitrateControllerStatus.currentTargetBps
+        if (totalNeededBps < 1) {
+            // Don't need to send any probing
```
JonathanLennox, line 80: You removed the trailing period from this comment.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -75,4 +81,36 @@ class EndpointK @JvmOverloads constructor(
      * Set the local SSRC for [mediaType] to [ssrc] for this endpoint.
      */
     override fun setLocalSsrc(mediaType: MediaType, ssrc: Long) = transceiver.setLocalSsrc(mediaType, ssrc)
+
+    /**
+     * Returns true if this endpoint's transport is 'fully' connected (both ICE and DTLS), false otherwise
+     */
+    private fun isTransportConnected(): Boolean = iceTransport.isConnected() && dtlsTransport.isConnected
```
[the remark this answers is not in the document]
bbaldino: Yeah I think we've probably got some inconsistencies around what we expose as properties without backing fields and what are methods, would be nice to set up some guidelines around that.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Sure, fixed.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Looks like `SctpManager` doesn't even have `getDebugState`, so removed it.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Yeah, agreed.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: It wasn't referenced anywhere else so I just got rid of the reference.  I don't think there's any concern about Endpoint not holding a strong reference there (Transceiver holds onto that reference, and Endpoint holds the Transceiver reference throughout its lifetime).

jvb/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
bbaldino: I'm not sure I follow, what's the concern?

jvb/src/main/kotlin/org/jitsi/videobridge/cc/BandwidthProbing.kt
bbaldino: Yeah, but then I can't `and` that value with `0xFFFFFFFFL` :/.  But I just remembered we have an `Int.toPositiveLong()` helper, so I'll use `Random.nextInt().toPositiveLong()`

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: I actually think `Supplier` still works pretty well here.  If we change it, `BandwidthProbing` would just take: `() -> BitrateControllerStatusSnapshot`, but I find the naming gets a little awkward.  Currently the member is `statusSnapshotSupplier` and we call `statusSnapshotSupplier.get()`, if it was just a function it would just be `statusSnapshotSupplier`, so I think we'd want to rename the member to something like `getStatusSnapshot` which looks better when called, but is a bit weird when passing it in.  Either way I don't think I feel strongly so I can change it if you want.

jvb/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
JonathanLennox: This is code that is *dependent* on the fact that `Bandwidth` is an inline class; if it were changed to not be inlined, I believe either it wouldn't work or you'd end up with the non-numeric output of `Bandwidth#toString()` in the stats.

We could either add a `sendBitRate.getRateBps()` method, or decide we don't care.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
JonathanLennox: Ok, this is probably fine then.

Is this something that Kotlin 1.4's changes would improve?

jvb/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
bbaldino: Ah gotcha.  It didn't occur to me that this was something we'd worry about, but I'm fine either way.

JonathanLennox: I'm inclined to say let's not worry about it, as long as it's a conscious decision.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: I don't think this would change, because SAM conversion for _Java_ interfaces is already supported it's the support for Kotlin interfaces that isn't in yet.  Do you just not like the `java.util.function.Supplier` part?  I just realized I can get rid of the package name, so it'll just be `Supplier { ... }` at least.

JonathanLennox: It's mostly just that it was wordier than the equivalent Java code, which tends to annoy me.

Yeah, losing the package name would help.

JonathanLennox (approved):
Approved, though it'll need re-approval once the jmt update goes in.

JonathanLennox (approved):

PR #1594, opened 2021-03-02
fix: Remove the code clearing ice4j properties.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/Main.kt
```diff
@@ -236,10 +236,4 @@ private fun startIce4j() {
 private fun stopIce4j() {
     // Shut down harvesters.
     Harvesters.closeStaticConfiguration()
-
-    System.getProperties().keys.forEach { key ->
```
bbaldino, line 240: Do we know if these properties are scoped to a VM instance?  Or will they potentially persist across VM restarts if we don't clear them?

JonathanLennox: Scope of `System.getProperties()` is a VM instance, yes -- see https://stackoverflow.com/questions/908903/scope-of-the-java-system-properties .  So if we were ever loading JVB as a servlet or something there might be an issue, but for our current usage where we're executing it from `main` there's no problem.

bbaldino (approved):

PR #1599, opened 2021-03-04
feat: Read and expose meetingId.

bgrozev:


bbaldino:
It feels like there must be some redundancy now that we have `id`, `gid` and `meetingId` all for a conference.  Do you think there's a path to eliminating any of these in the future?

bgrozev:
Not sure. We could use `meetingId` for the ID, but that requires the user to specify it, and we have to trust that different users will not use the same one. `gid` is essentially `octoId`, and we could use `meetingId`, except that it's present in every packet so we want a small, fixed size and we currently use UUIDs for `meetingId`.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -164,6 +164,11 @@
     @NotNull
     private final EndpointConnectionStatusMonitor epConnectionStatusMonitor;
 
+    /**
+     * A unique meeting ID set by the signaling server. It is exposed via ({@link #getDebugState()} for outside use.
```
bbaldino, line 168: Could you add a comment about the fact that it's nullable, and what it means for it to be null vs not?

bbaldino (approved):

PR #1601, opened 2021-03-05
Refactor endpoint to Kotlin part 2

bbaldino:
Continuation of the work started in https://github.com/jitsi/jitsi-videobridge/pull/1593

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
@@ -301,6 +301,24 @@ class EndpointK @JvmOverloads constructor(
         }
     }
 
+    override fun effectiveVideoConstraintsChanged(
+        oldEffectiveConstraints: MutableMap<String, VideoConstraints>,
+        newEffectiveConstraints: MutableMap<String, VideoConstraints>
+    ) {
+        val removedEndpoints = oldEffectiveConstraints.keys.filter { it in newEffectiveConstraints.keys }
+
+        // Sources that "this" endpoint no longer receives.
+        for (removedEpId in removedEndpoints) {
+            // Remove ourself as a receiver from that endpoint
+            conference.getEndpoint(removedEpId)?.removeReceiver(id)
+        }
+
+        // Added or updated
+        newEffectiveConstraints.forEach { (epId, effectiveConstraints) ->
+            conference.getEndpoint(epId)?.addReceiver(id, effectiveConstraints)
+        }
+    }
+
     override fun sendVideoConstraints(maxVideoConstraints: VideoConstraints) {
         // Note that it's up to the client to respect these constraints.
         if (mediaSources.isEmpty()) {
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
@@ -51,6 +51,7 @@ import org.jitsi.videobridge.transport.ice.IceTransport
 import org.jitsi.videobridge.util.ByteBufferPool
 import org.jitsi.videobridge.util.TaskPools
 import org.jitsi.videobridge.util.looksLikeDtls
+import org.jitsi.videobridge.xmpp.MediaSourceFactory
 import org.json.simple.JSONObject
 import java.lang.IllegalArgumentException
 import java.time.Clock
@@ -314,6 +315,29 @@ class EndpointK @JvmOverloads constructor(
             else -> {
                 throw IllegalArgumentException("Media direction unknown: $direction")
             }
+        }
+    }
+
+    /**
+     * Re-creates this endpoint's media sources based on the sources
+     * and source groups that have been signaled.
+     */
+    override fun recreateMediaSources() {
+        val videoChannels = channelShims.filter { c -> c.mediaType == MediaType.VIDEO }
+
+        val sources = videoChannels
+            .mapNotNull(ChannelShim::getSources)
+            .flatten()
+            .toList()
+
+        val sourceGroups = videoChannels
+            .mapNotNull(ChannelShim::getSourceGroups)
+            .flatten()
+            .toList()
+
+        if (sources.isNotEmpty() || sourceGroups.isNotEmpty()) {
+            val mediaSources = MediaSourceFactory.createMediaSources(sources, sourceGroups)
+            setMediaSources(mediaSources)
         }
     }
 
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
@@ -43,6 +43,7 @@ import org.jitsi.utils.MediaType
 import org.jitsi.utils.logging2.Logger
 import org.jitsi.utils.logging2.cdebug
 import org.jitsi.videobridge.cc.BandwidthProbing
+import org.jitsi.videobridge.message.ForwardedEndpointsMessage
 import org.jitsi.videobridge.message.ReceiverVideoConstraintsMessage
 import org.jitsi.videobridge.rest.root.debug.EndpointDebugFeatures
 import org.jitsi.videobridge.shim.ChannelShim
@@ -58,7 +59,6 @@ import org.jitsi.xmpp.extensions.colibri.WebSocketPacketExtension
 import org.jitsi.xmpp.extensions.jingle.DtlsFingerprintPacketExtension
 import org.jitsi.xmpp.extensions.jingle.IceUdpTransportPacketExtension
 import org.json.simple.JSONObject
-import java.lang.IllegalArgumentException
 import java.time.Clock
 import java.time.Duration
 import java.time.Instant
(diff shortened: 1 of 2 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -268,6 +269,20 @@ class EndpointK @JvmOverloads constructor(
 
     override fun requestKeyframe(mediaSsrc: Long) = _transceiver.requestKeyFrame(mediaSsrc)
 
+    override fun isOversending(): Boolean = bitrateController.isOversending()
+
+    override fun setSelectedEndpoints(selectedEndpoints: MutableList<String>) {
```
JonathanLennox, line 684: Why is this `MutableList` now rather than `List`?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -268,6 +269,20 @@ class EndpointK @JvmOverloads constructor(
...
+    }
+
+    override fun setMaxFrameHeight(maxFrameHeight: Int) {
+        bitrateController.setMaxFrameHeight(maxFrameHeight)
+    }
+
+    override fun setBandwidthAllocationSettings(message: ReceiverVideoConstraintsMessage) {
+        bitrateController.setBandwidthAllocationSettings(message)
+    }
+
```
JonathanLennox, line 700: These three could use Kotlin `=` notation.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -261,6 +262,39 @@ class EndpointK @JvmOverloads constructor(
         return _transceiver.isReceivingVideo()
     }
 
+    /**
+     * Update media direction of {@link ChannelShim}s associated
```
JonathanLennox, line 579: Is this the right syntax for KDoc?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -261,6 +262,39 @@ class EndpointK @JvmOverloads constructor(
...
+     * accept incoming media from this endpoint, but will not forward
+     * other endpoint's media to this endpoint.
+     * When media direction is set to 'inactive' JVB will
+     * neither accept incoming media nor forward media from other endpoints.
+     *
+     * @param type media type.
+     * @param direction desired media direction:
+     *                       'sendrecv', 'sendonly', 'recvonly', 'inactive'
+     */
+    @SuppressWarnings("unused") // Used by plugins (Yuri)
```
JonathanLennox, line 599: Do we actually still have plugins?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -261,6 +262,39 @@ class EndpointK @JvmOverloads constructor(
...
+    override fun updateMediaDirection(type: MediaType, direction: String) {
+        when (direction) {
+            "sendrecv", "sendonly", "recvonly", "inactive" -> {
+                channelShims.firstOrNull { it.mediaType == type }?.setDirection(direction)
```
JonathanLennox, line 603: I think this should be `foreach`, not `firstOrNull`, at least to have the same semantic as the Java code.

I think we never currently have more than one channel shim with the same media type, but if we ever did this code would only set the first one.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -290,6 +290,29 @@ class EndpointK @JvmOverloads constructor(
...
     }
 
+    /**
+     * @return the password of the ICE Agent associated with this
+     * {@link Endpoint}.
+     */
+    override fun getIcePassword(): String = iceTransport.icePassword
+
+    /**
+     * Sends a message to this {@link Endpoint} in order to notify it that the set of endpoints for which the bridge
```
JonathanLennox, line 519: KDoc

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -290,6 +290,29 @@ class EndpointK @JvmOverloads constructor(
...
+     * is sending video has changed.
+     *
+     * @param forwardedEndpoints the collection of forwarded endpoints.
+     */
+    override fun sendForwardedEndpointsMessage(forwardedEndpoints: MutableCollection<String>) {
```
JonathanLennox, line 524: Why has this become `MutableCollection`?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -292,6 +299,83 @@ class EndpointK @JvmOverloads constructor(
...
+                    // This logic is for opening the data channel locally
+                    logger.info("Will open the data channel.")
+                    val dataChannel = dataChannelStack.createDataChannel(
+                        DataChannelProtocolConstants.RELIABLE,
+                        0,
+                        0,
+                        0,
+                        "default"
+                    )
+                    messageTransport.setDataChannel(dataChannel)
```
JonathanLennox, line 416: Java scoped this with `Endpoint.this` - we're sure this doesn't pick up a bad definition?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: This was probably an auto-convert, but `MutableList` is the Kotlin equivalent of Java's list.  I agree it can be `List` though.

bbaldino: Ah, no actually it's because it's an override of a java method which takes a Java `List`, but weirdly it does seem to let me override it with a Kotlin `List`.  I think I will leave it as a `MutableList` to be consistent with the Java parent, but add a TODO that it should become `List` once the java parent goes away.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: I changed `setMaxFrameHeight` and `setBandwidthAllocationSettings` but not sure which third one you meant.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: I think this style is still supported, but I'll change it.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Not sure.  @mstyura are you still using this?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Although that would match the old java behavior, I'm not sure the old behavior really makes sense.  I think we'd have a lot of broken code if we had more than one channel per media type, so I dunno that being flexible here adds much value.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Same as above, I'll add a TODO

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Yeah there's no other `messageTransport`

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -301,6 +301,20 @@ class EndpointK @JvmOverloads constructor(
         }
     }
 
+    // TODO: this should be part of an EndpointMessageTransport.EventHandler interface
+    override fun endpointMessageTransportConnected() {
+        sendVideoConstraints(this.maxReceiverVideoConstraints)
```
JonathanLennox, line 333: Is the `this` needed?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -301,6 +301,20 @@ class EndpointK @JvmOverloads constructor(
...
+    }
+
+    /**
+     * Handle a DTLS app packet (that is, a packet of some other protocol sent
+     * over DTLS) which has just been received.
+     */
+    fun dtlsAppPacketReceived(data: ByteArray, off: Int, len: Int) {
+        // TODO(brian): change sctp handler to take buf, off, len
+        sctpHandler.processPacket(PacketInfo(UnparsedPacket(data, off, len)))
+    }
```
JonathanLennox, line 343: Can these two use Kotlin `=`?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
JonathanLennox: `setSelectedEndpoints`.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
JonathanLennox: Ok.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: K done

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Nope, changed this to an expression body as well.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
mstyura: @bbaldino thanks for asking! I'm not really sure which version used in project I've worked on is such fresh. There will be some work to update, but I'm curios is there alternative way to tell bridge to not to route media to endpoint? Use case: endpoint is put on hold due to concurrent call, so not media need to be routed to it to save some bandwidth.

bbaldino: I don't think we have any other method for it.  I'm fine with keeping it.

JonathanLennox (approved):

PR #1603, opened 2021-03-05
fix: Fix previousSpeakers message.

bgrozev:


bbaldino:
Was the fix here the synchronization?

bgrozev:
No, it's using the `endpointsBySpeechActivity` list instead of the `endpointsInLastNOrder` list (via `getOrderedEndpoints`).

bbaldino (approved):

PR #1606, opened 2021-03-11
Refactor Endpoint to Kotlin part 3 (final)

bbaldino:


jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -469,7 +471,7 @@ class EndpointK @JvmOverloads constructor(
                 // holding a lock inside the SctpSocket which can cause a deadlock
                 // if two endpoints are trying to send datachannel messages to one
                 // another (with stats broadcasting it can happen often)
-                TaskPools.IO_POOL.execute { dataChannelHandler.consume(PacketInfo(dataChannelPacket)) }
+                TaskPools.IO_POOL.execute { dataChannelHandler.processPacket(PacketInfo(dataChannelPacket)) }
```
JonathanLennox: Why did this change from `consume` to `processPacket`?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -114,6 +116,43 @@ class EndpointK @JvmOverloads constructor(
         put("endpoint_id", id)
     }
 
+    /**
+     * Measures the jitter introduced by the bridge itself (i.e. jitter calculated between
+     * packets based on the time they were received by the bridge and the time they
+     * are sent).  This jitter value is calculated independently, per packet, by every
+     * individual {@link Endpoint} and their jitter values are averaged together
+     * in this static member.
```
JonathanLennox, line 131: Comment is wrong (also in original), because "this" isn't the static member.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -114,6 +116,43 @@ class EndpointK @JvmOverloads constructor(
         put("endpoint_id", id)
     }
 
+    /**
+     * Measures the jitter introduced by the bridge itself (i.e. jitter calculated between
+     * packets based on the time they were received by the bridge and the time they
+     * are sent).  This jitter value is calculated independently, per packet, by every
+     * individual {@link Endpoint} and their jitter values are averaged together
```
JonathanLennox, line 130: Update to KDoc.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -114,6 +116,43 @@ class EndpointK @JvmOverloads constructor(
...
+     */
+    private val channelShims = mutableSetOf<ChannelShim>()
+
+    /**
+     * Whether this endpoint should accept audio packets. We set this according
+     * to whether the endpoint has an audio Colibri channel whose direction
+     * allows sending.
+     */
+    private var acceptAudio = false
+    /**
```
JonathanLennox, line 162: Blank line between vars with docs?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -460,10 +499,10 @@ class EndpointK @JvmOverloads constructor(
             },
             logger
         )
-        sctpHandler.setSctpManager(sctpManager)
+        sctpHandler.setSctpManager(sctpManager!!)
```
JonathanLennox, line 516: Annoying we need `!!` here, but I guess Kotlin can't prove nothing resets it to `null`?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
```diff
@@ -71,24 +83,144 @@ import org.jitsi_modified.sctp4j.SctpDataCallback
...
+     * or web sockets).
+     */
+    private val _messageTransport = EndpointMessageTransport(
+        this,
+        Supplier { conference.videobridge.statistics },
+        conference,
+        logger
+    )
+
+    override fun getMessageTransport(): EndpointMessageTransport = _messageTransport
```
JonathanLennox, line 215: Hm, you can't override `getMessageTransport()` with a `val`?

JonathanLennox (commented):
`EndpointK` can be renamed back to being `Endpoint` now, yes?

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: `consume` is the protected method defined in `ConsumerNode`, it's not meant to be called directly.  Previously the code had changed the visibility and exposed it as public, I fixed it back to the top-level method on Node because calling `consume` directly prevents some of the stats from being calculated correctly.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Yeah this stat was never wired up, apparently.  I'll tweak the comment for now and will probably try and bring this stat back in another PR.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: Yeah, I think I originally considered nesting all this inside of an `.also` block after creating `SctpManager`, but there's quite a bit so that didn't feel right.  The other option is to create the `SctpManager` and hold it in a local (non-nullable) val which could be used for all this code and also assigned to the member.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
bbaldino: No, unfortunately that doesn't seem to work.

bbaldino:
> `EndpointK` can be renamed back to being `Endpoint` now, yes?

Yup, just forgot.  Will do that.

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointK.kt
JonathanLennox: Annoying, okay.

JonathanLennox (approved):

PR #1609, opened 2021-03-18
Create and use a singleton VideoConstraints.NOTHING object.

JonathanLennox:
Prioritize#getEffectiveConstraints creates lots of these for endpoints outside lastN, so let's save some GC.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/VideoConstraints.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/VideoConstraints.kt
@@ -29,7 +29,7 @@ data class VideoConstraints(
     }.toJSONString()
 
     companion object {
-        val ZERO = VideoConstraints(0)
+        val NOTHING = VideoConstraints(0)
     }
 }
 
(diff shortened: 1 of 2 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/VideoConstraints.kt
```diff
@@ -27,6 +27,10 @@ data class VideoConstraints(
         this["maxHeight"] = maxHeight
         this["maxFrameRate"] = maxFrameRate
     }.toJSONString()
+
+    companion object {
+        val ZERO = VideoConstraints(0)
```
bbaldino, line 32: I wonder if there's a better name.  "zero" reads as "zero constraints" which sounds like 'unconstrained', but really it's more like "I want nothing at all", right?

JonathanLennox: `ZERO_HEIGHT`?

JonathanLennox: Or `NOTHING`?

bbaldino: I'm not sure I have a good idea, it's hard to pair something with "constraints" that conveys "constrained in such a way that I know it will result in nothing at all".  `OFF`? `NOT_DISPLAYED`?

bbaldino: I'm good with `NOTHING`

bbaldino (approved):

PR #1610, opened 2021-03-18
Improve performance of determining max height of receiver constraints

bbaldino:
When doing large call testing, we noticed a very hot call path where we figure out what the maximum height of a set of constraints is.  This change keeps track of the max height as constraints are added and removed from the map so we don't have to walk the entire thing to find it.

Note: `AbstractEndpoint#receiverVideoConstraintsChanged` should probably take a some structure which has max values for various constraints, but currently we only look at maxHeight so rather than define that new class now, I changed the function to just take `int maxHeight`.  It won't be difficult to expand that later when we use other constraints.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/ReceiverConstraintsMap.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/ReceiverConstraintsMap.kt
@@ -1,0 +1,47 @@
+/*
+ * Copyright @ 2018 - present 8x8, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jitsi.videobridge.cc.allocation
+
+/**
+ * A Map of Endpoint IDs to their receiver video constraints.  Tracks the max height of all
+ * of those constraints as new constraints are added and removed.
+ */
+class ReceiverConstraintsMap {
+    private val map = mutableMapOf<String, VideoConstraints>()
+    private val lock = Any()
+    var maxHeight: Int = 0
+        private set
+
+    fun put(key: String, value: VideoConstraints): VideoConstraints? {
+        synchronized(lock) {
+            if (maxHeight < value.maxHeight) {
+                maxHeight = value.maxHeight
+            }
+            return map.put(key, value)
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/ReceiverConstraintsMap.kt
```diff
@@ -39,9 +39,22 @@ class ReceiverConstraintsMap {
...
                 if (removed.maxHeight == maxHeight) {
-                    maxHeight = map.values.map { it.maxHeight }.max() ?: 0
+                    maxHeight = findNewMax(maxHeight)
                 }
             }
         }
     }
+
+    // Note: this method must be called with the lock held
+    private fun findNewMax(oldMax: Int): Int {
```
JonathanLennox, line 49: Maybe name this so it's clear it's only applicable for `remove`?

JonathanLennox:
The alternative to defining another new object for `AbstractEndpoint#receiverVideoConstraintsChanged` would be to pass it the new `ReceiverConstraintsMap` itself.

jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/ReceiverConstraintsMap.kt
bbaldino: Well, technically it doesn't care when it's used.  I would've done it in `remove` itself but with the early returns it's tricky.

bbaldino:
> The alternative to defining another new object for `AbstractEndpoint#receiverVideoConstraintsChanged` would be to pass it the new `ReceiverConstraintsMap` itself.

Yeah, my thought was to just take what was needed.  Even if/when we do need more things, it might be nice to minimize the places where `ReceiverConstraintsMap` is passed, though, to minimize dependencies on that API.

jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/ReceiverConstraintsMap.kt
JonathanLennox: Well, it's not correct for `add`, but fortunately `add`'s calculation is easy.

bbaldino: Yeah, actually you're right.  Will rename.

jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/ReceiverConstraintsMap.kt
```diff
@@ -57,4 +59,9 @@ class ReceiverConstraintsMap {
         }
         return maxSeen
     }
+
+    fun getDebugState() = OrderedJsonObject().apply {
+        put("maxHeight", maxHeight)
+        put("constraints", map)
```
JonathanLennox, line 65: Does this leak the map object outside the synchronization?

bbaldino: Technically, yes.  I can do `map.toMap()` to make a copy there.

JonathanLennox (approved):

PR #1612, opened 2021-03-19
Use local participant count (not conference count) for graceful shutdown.

JonathanLennox:
We have a known bug where remote participant counts, and thus conferences, can
get stuck in some cases; don't let that block graceful shutdown.

### DIFF
--- a/resources/graceful_shutdown.sh
+++ b/resources/graceful_shutdown.sh
@@ -60,7 +60,7 @@ if ! [[ $pid =~ $re ]] ; then
 fi
 
 # Returns local participant count by calling JVB REST statistics API and extracting
-# participant count count from JSON stats text returned.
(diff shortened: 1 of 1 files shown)


bbaldino:
Does this make sense long term, or is it just a workaround? Should we be trying to fix the remote participant count issue instead?

JonathanLennox:
I think it probably makes sense long term as well?

If a conference is continuing but all the members are off this bridge, it's fine to shut it down.

bbaldino:
> I think it probably makes sense long term as well?
> 
> If a conference is continuing but all the members are off this bridge, it's fine to shut it down.

Yeah, this does sort of mask a bug though, since if a bridge has no participants in a conference the conference should've been removed/expire.

bgrozev (approved):

PR #1614, opened 2021-03-22
Support disabling websocket compression on the bridge channel.

JonathanLennox:
It uses a surprisingly large amount of CPU.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/websocket/config/WebsocketServiceConfig.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/websocket/config/WebsocketServiceConfig.kt
@@ -53,6 +53,15 @@ class WebsocketServiceConfig {
     }
 
     /**
+     * Whether compression (permessage-deflate) should be used for websockets
+     */
+    val useCompression: Boolean by config {
+        onlyIf("Websockets are enabled", ::enabled) {
+            "videobridge.websockets.compression".from(JitsiConfig.newConfig)
+        }
+    }
+
+    /**
      * The server ID used in URLs advertised for COLIBRI WebSockets.
      */
     val serverId: String by config {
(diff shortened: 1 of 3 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/websocket/config/WebsocketServiceConfig.kt
```diff
@@ -52,6 +52,15 @@ class WebsocketServiceConfig {
         }
     }
 
+    /**
+     * Whether compression (permessage-deflate) should be used for websockets
+     */
+    val useCompression: Boolean by config {
```
bbaldino, line 58: I think if you use `@JvmField` here you'll get a nicer name when calling from java.

jvb/src/main/resources/reference.conf
```diff
@@ -229,6 +229,8 @@ videobridge {
   websockets {
     enabled=false
     server-id="default-id"
+    # Whether to use WebSocket compression (permessage-deflate)
+    compression = true
```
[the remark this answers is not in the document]
JonathanLennox: In terms of what to call the parameter, you mean?

I was modeling it on the adjacent `tls` parameter, but I can use something else. (Probably `dash-separated` rather than `interCap` though.)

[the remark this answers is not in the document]
bbaldino: Yes and yes, good point.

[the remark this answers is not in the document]
bbaldino: As far as tls, I think that should probably be named `enable-tls` or something...so I'd say let's make this one better and we can fix that someday.

jvb/src/main/kotlin/org/jitsi/videobridge/websocket/config/WebsocketServiceConfig.kt
JonathanLennox: `JvmField cannot be applied to delegated property`

bbaldino: Ah, damn.  Usually when we are in this situation we make the field private and expose a better-named getter.  Like `fun shouldUseCompression() = useCompression`

JonathanLennox: Changed to `fun shouldEnableCompression()` (rather than it being `getEnableCompression()` in Java).

bbaldino (approved):

PR #1618, opened 2021-03-23
New (filterable) EndpointStats JVB channel message.

JonathanLennox:


jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
```diff
@@ -147,6 +148,46 @@ public BridgeChannelMessage endpointMessage(EndpointMessage message)
...
+            logger.warn("Unable to send EndpointStats, unknown endpoint " + from);
+            return null;
+        }
+
+        List<AbstractEndpoint> targets =
+            conference.getLocalEndpoints().stream().
+                filter((ep) -> ep.wantsStatsFrom(from)).
+                collect(Collectors.toList());
+
+        conference.sendMessage(message, targets, true);
```
bgrozev, line 187: Infinite octo loop :)

JonathanLennox: 🤦

jvb/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -228,6 +233,32 @@ public AbstractEndpoint getDominantEndpoint()
...
+     */
+    public List<String> getRecentSpeakers()
+    {
+        return getSpeakerHistory(1, NUM_RECENT_SPEAKERS);
+    }
+
+    /**
+     * Query whether an endpoint is a recent speaker.
+     */
+    public boolean isRecentSpeaker(AbstractEndpoint endpoint)
```
bbaldino, line 247: I think we could also do this here:

```
endpointsBySpeechActivity.stream()
    .limit(NUM_RECENT_SPEAKERS)
    .filter(ep -> ep == endpoint)
    .findFirst().orElse(false)
```

jvb/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -531,4 +532,34 @@ else if (targetEndpoint != null)
...
+        if (conference == null || conference.isExpired())
+        {
+            logger.warn("Unable to send EndpointStats, conference is null or expired");
+            return null;
+        }
+
+        List<AbstractEndpoint> targets =
+            conference.getLocalEndpoints().stream().
+            filter((ep) -> ep != endpoint && ep.wantsStatsFrom(endpoint)).
+                collect(Collectors.toList());
```
bbaldino, line 560: nit: I think we've generally formatted `stream` chains like:

```
List<AbstractEndpoint> targets = conference.getLocalEndpoints().stream()
    .filter((ep) -> ep != endpoint && ep.wantsStatsFrom(endpoint))
    .collect(Collectors.toList());
```

jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
```diff
@@ -357,6 +357,14 @@ synchronized void update()
                 oversending);
     }
 
+    /**
+     * Query whether this allocator is forwarding a source from a given endpoint, as of its
+     * most recent allocation decision.
+     */
+    public boolean isForwarding(T endpoint)
```
bbaldino, line 364: should this just take in a `String endpointId` instead of a `T` ?

jvb/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
bbaldino: Oh interesting, I thought stream was specifically supposed to work that way.  I know Kotlin's `asSequence` does and I thought it was the same thing, maybe not.

jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
JonathanLennox: We need to do the `getId()` at some level - I figured I'd push it downward as far as reasonable, but I could pull it upward instead.  Any reason for the higher level?

jvb/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
JonathanLennox: Maybe it was just the overhead of creating the relevant objects, I didn't investigate it that closely.

jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
bbaldino: My thought on things like this is to raise it up, such that methods take only what they actually need/use which reduces coupling.  It also makes testing a bit easier.

JonathanLennox: Ok.

bbaldino (approved):

PR #1624, opened 2021-04-02
Signal video type (camera, screenshare, none)

bgrozev:
Allow endpoints to signal the type of video steam they have available, or the lack thereof. Use the signaled video type, instead of the bitrate of the incoming stream when prioritizing endpoints for LastN.

Fixes an issue where an endpoint's video being suspended causes the endpoint to be de-prioritized, and thus video is never unsuspended.

This is a draft, because I want to test some more and add more unit tests.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -111,6 +111,13 @@ public abstract class AbstractEndpoint
         this.id = Objects.requireNonNull(id, "id");
     }
 
+    /**
+     * Return the {@link VideoType} that the endpoint has advertised as available. This reflects the type of stream
+     * even when the stream is suspended (e.g. due to no receivers being subscribed to it).
+     *
+     * In other words, CAMERA or SCREENSHARE means that the endpoint has an available stream, which may be suspended.
+     * NONE means that the endpoint has signaled that it has no available streams.
+     */
     @NotNull
     public VideoType getVideoType()
     {
--- a/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -122,7 +122,7 @@ public abstract class AbstractEndpoint
      * In other words, CAMERA or SCREENSHARE means that the endpoint has an available stream, which may be suspended.
      * NONE means that the endpoint has signaled that it has no available streams.
      *
-     * Note that when the endpoint has not advertised any sources, the video type is necessarily {@code NONE}.
+     * Note that when the endpoint has not advertised any video sources, the video type is necessarily {@code NONE}.
      */
     @NotNull
     public VideoType getVideoType()
--- a/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
@@ -25,6 +25,7 @@ import org.jitsi.utils.event.*;
 import org.jitsi.utils.logging2.*;
 import org.jitsi.videobridge.cc.allocation.*;
 import org.jitsi.videobridge.message.*;
+import org.jitsi.videobridge.util.*;
 import org.jitsi.xmpp.extensions.colibri.*;
 import org.json.simple.*;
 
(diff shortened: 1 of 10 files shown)


jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -93,6 +94,8 @@
 
     protected final EventEmitter<EventHandler> eventEmitter = new EventEmitter<>();
 
+    private VideoType videoType = VideoType.CAMERA;
```
bbaldino, line 97: I have a vague memory of a 'start with screenshare' option in the client...if so, is it problematic to set this to camera by default?  Will the client update with screenshare right away in that case?

If that feature exists then ignore the comment :)

jvb/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -160,13 +160,13 @@ boolean updateLastNEndpoints()
     {
         synchronized (syncRoot)
         {
-            Map<Boolean, List<AbstractEndpoint>> bySendingVideo
+            Map<Boolean, List<AbstractEndpoint>> byVideoAvailable
                     = endpointsBySpeechActivity.stream()
-                        .collect(Collectors.groupingBy(AbstractEndpoint::isSendingVideo));
+                        .collect(Collectors.groupingBy(ep -> ep.getVideoType() != VideoType.NONE));
```
bbaldino, line 165: Kotlin has a nice [`partition`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/partition.html) function that splits a collection into 2 based on a predicate which would be nice here.  Maybe someday when `ConferenceSpeechActivity` gets ported.

jvb/src/test/kotlin/org/jitsi/videobridge/SpeechActivityTest.kt
```diff
@@ -77,17 +79,17 @@ class SpeechActivityTest : ShouldSpec() {
 
             conferenceSpeechActivity.orderedEndpoints shouldContainExactly listOf(a, b, c, d)
 
-            every { a.isSendingVideo } returns false
+            every { a.videoType } returns NONE
```
bbaldino, line 82: `isSendingVideo` still exists, right?  It seems like `SpeechActivityTest` would only care about it sending/not sending rather than the specific video type (or maybe it _does_ care for a reason I'm missing?).

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
bgrozev: Well, the client could start with any of them. My thinking was to use the most common case as a default, and assume that the client will send updates at least any time it uses a non-default.

@jallamsetty1 does that seem reasonable?

jvb/src/test/kotlin/org/jitsi/videobridge/SpeechActivityTest.kt
bgrozev: It cares about whether video is available or not, which is equivalent to `videoType == NONE`. I can expose a boolean from Endpoint, but I'm planning to use the exposed `videoType` elsewhere too (bandwidth allocation), so it wouldn't be a huge win.

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
jallamsetty1: The `startScreenSharing` option still exists but yes client can always send an update when the video source type is non-default.

jvb/src/test/kotlin/org/jitsi/videobridge/SpeechActivityTest.kt
bbaldino: Yeah, my thought was that code that only cares about whether or not the endpoint is sending anything (which this test appears to) should use `isSendingVideo`.  Code that cares about what _kind_ of video the endpoint is sending would use `videoType`.

bgrozev:
jenkins please test this again please

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -108,6 +115,30 @@ protected AbstractEndpoint(Conference conference, String id, Logger parentLogger
...
     }
 
+    /**
+     * Return the {@link VideoType} that the endpoint has advertised as available. This reflects the type of stream
+     * even when the stream is suspended (e.g. due to no receivers being subscribed to it).
+     *
+     * In other words, CAMERA or SCREENSHARE means that the endpoint has an available stream, which may be suspended.
+     * NONE means that the endpoint has signaled that it has no available streams.
+     *
+     * Note that when the endpoint has not advertised any sources, the video type is necessarily {@code NONE}.
```
JonathanLennox, line 125: "Sources" are only video sources, not audio?

JonathanLennox (approved):

PR #1629, opened 2021-04-08
Fix: Shut down ice4j harvester thread pool properly. (Bump ice4j).

JonathanLennox:


JonathanLennox:
Other ice4j changes don't need noting here; just test and Maven deployment changes.

bgrozev (approved):

PR #1636, opened 2021-04-20
Changes for queue stats enhancements.

JonathanLennox:


JonathanLennox:
Depends on https://github.com/jitsi/jitsi-utils/pull/97; see also https://github.com/jitsi/jitsi-media-transform/pull/345.

bgrozev:
LGTM

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -639,31 +639,41 @@ public JSONObject getQueueStats()
...
+    private OrderedJsonObject getJsonFromQueueStatisticsAndErrorHandler(
+            CountingErrorHandler countingErrorHandler,
+            String queueName)
     {
-        JSONObject json = new JSONObject();
-        json.put("dropped_packets", countingErrorHandler.getNumPacketsDropped());
+        OrderedJsonObject json = (OrderedJsonObject)QueueStatistics.Companion.getStatistics().get(queueName);
+        if (json == null) {
+            json = new OrderedJsonObject();
+            json.put("dropped_packets", countingErrorHandler.getNumPacketsDropped());
```
ibauersachs, line 673: @JonathanLennox  Looking at the jvb.log output for #1760, I'm seeing the following output. I think putting `dropped_packets` should be outside the null check?

```
/home/jenkins/jitsi-pr-testing-scripts-master/verify-jvb-stats.js:92
        if (queue_stats[queue].dropped_packets > 0) {
                               ^

TypeError: Cannot read property 'dropped_packets' of null
    at verify_queue_stats (/home/jenkins/jitsi-pr-testing-scripts-master/verify-jvb-stats.js:92:32)
```

JonathanLennox: If the stats are not null, then there should be a `dropped_packets` in the JSON object that comes from the queue stats, so `dropped_packets` should be there either way.  I'm not sure why this is failing?

bgrozev: This was [here](http://pr-reports.jitsi.org/jitsi-videobridge-PR-test/2986/jvb-stats/) and it happened because [`colibri_queue` and `bridge-channel-message-incoming-queue`](https://github.com/jitsi/jitsi-videobridge/blob/master/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java#L708) were `null`, probably because the bridge was never used (jicofo failures?).

JonathanLennox: Ah, yes, those are the queue stats that don't go through `getJsonFromQueueStatisticsAndErrorHandler`, because they're not `PacketQueue` instances but rather queues of a different object type.

My thought would be that we should probably change `verify-jvb-stats.js` to handle those stats being missing.

ibauersachs: > If the stats are not null, then there should be a dropped_packets in the JSON object that comes from the queue stats, so dropped_packets should be there either way. I'm not sure why this is failing?

I was just wondering if dropped_packets would need updating if it's already there. But I really didn't follow the code flow much, so my assumption can be totally wrong. I don't know what to do here, if anything. Just an observation from the failing tests.

bgrozev: Making it not fail when there are missing stats that it expects seems wrong. I improved the message here:
https://github.com/8x8Cloud/jitsi-pr-testing-scripts/pull/92

PR #1641, opened 2021-04-27
fix: Bring back logs for SENT colibri responses.

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/shim/ConferenceShim.java
```diff
@@ -92,6 +92,7 @@ public ConferenceShim(Conference conference, Logger parentLogger)
                                     + totalDelay + " ms): " + request.getRequest().toXML());
                         }
                         request.getCallback().invoke(response);
+                        logger.debug(() -> "SENT Colibri response: " + response.toXML());
```
JonathanLennox, line 95: This moves the SENT logs from org.jitsi.videobridge.xmpp.XmppConnection to org.jitsi.videobridge.shim.ConferenceShim, which breaks existing log configs and is confusingly separated from RECV and synchronous SENT.  Can we put the log message inside the request callback instead?  (This would also guarantee that we log for the various asynchronous error codepaths.)

JonathanLennox (approved):

bgrozev:
The failed tests are unrelated

PR #1642, opened 2021-04-29
Use G1GC by default on Java 11+.

JonathanLennox:


bgrozev (approved):

damencho:
Do we need awk as a dependency or it is available in debian minimal distribution?
I don't remember whether we use it in the other deb packages...

damencho:
Ignore me 
```
Awk is part of POSIX and Single Unix (and has been since they exist). It has also been a part of classical Unix for a very long time, and is installed by default on every modern full-fledged Unix.
```

damencho (approved):

PR #1654, opened 2021-05-12
Do not oversend camera streams (only screensharing).'

bgrozev:


JonathanLennox:
I wonder if we want to rename the `allow-oversending-onstage` config parameter to make it clearer that it's only for desktop?

Also, do you have any way to test to make sure bwe probing works properly to recover when all video sources are turned off?

bgrozev:
> I wonder if we want to rename the `allow-oversending-onstage` config parameter to make it clearer that it's only for desktop?

I'm reluctant to change the name of a config propety, but I added docs. IANAL, but just because oversending is allowed doesn't mean it has to be used, so to me the code reads okay :)
I re-ordered the conditio to make it easier to understand. If you disagree though, I don't mind renaming the function.



> Also, do you have any way to test to make sure bwe probing works properly to recover when all video sources are turned off?

None of the unit tests cover this case. But the same thing happens when lastN is 0 and then increases (and that's not new behavior), so I think it's getting some testing in practice.

JonathanLennox (approved):

PR #1665, opened 2021-05-30
Process data channel messages sequentially

bertm:
This prevents SCTP DataChannel messages from being processed out of order, causing warnings on channel initialization and potential issues with Last-N selection.

I originally proposed fixing this with a different implementation in #1309. A simpler solution (reusing `PacketInfoQueue`) was then proposed by @mstyura in #1318.
Neither of the original PRs are in a mergeable state now `Endpoint` is re-implemented in Kotlin.

This change takes the approach from #1318, and applies it to the new Kotlin state of affairs.


bertm:
Signed on 2020-06-12.

bgrozev:
Jenkins please test this please

bgrozev (approved):

PR #1666, opened 2021-06-01
Fix "unknown endpoint" warning sending EndpointStats.

JonathanLennox:


JonathanLennox:
The code was printing the result of `conference.getEndpoint()` (which by definition, if it got here, was `null`) rather than the value passed to it.

damencho (approved):

PR #1669, opened 2021-06-15
Octo is now disable by default, need to enable it.

mferstl:
Since octo is disabled by default (see https://github.com/jitsi/jicofo/pull/712 ) one needs to enable it to get it running.

### DIFF
--- a/doc/octo.md
+++ b/doc/octo.md
@@ -18,10 +18,29 @@ You need to make sure that all of the bridges can communicate via the socket
 addresses described in the properties above, and that the network is secure.
 
 ## Jicofo configuration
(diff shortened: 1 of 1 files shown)


bgrozev:
Thanks for the contrribution. Can you please also update the selection strategy property with the new one:
```
jicofo {
  bridge {
    selection-strategy = RegionBasedBridgeSelectionStrategy
  }
}
```

mferstl:
Hmm, not so fond of the phrasing, but as non-native speaker... ;)

bgrozev (approved):

bgrozev:
Thanks!

PR #1674, opened 2021-06-21
Fix forwarding of VP9 screenshare

JonathanLennox:
Don't treat layers with unknown frame rates as being low-framerate.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/cc/allocation/SingleSourceAllocation.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/cc/allocation/SingleSourceAllocation.java
@@ -154,8 +154,10 @@ class SingleSourceAllocation
                         .addField("remote_endpoint_id", endpoint.getId());
             for (LayerSnapshot layerSnapshot : ratesList)
             {
+                RtpLayerDesc l = layerSnapshot.layer;
                 ratesTimeSeriesPoint.addField(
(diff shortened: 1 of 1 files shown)


JonathanLennox:
For VP9, Chrome is encoding screenshare streams with an odd structure.  I.e..:

```
It indeed uses a somewhat strange scalability structure. There are no temporal layers, only upto 3 spatial layers, Then, these layers have the same resolution but different framerate!
 The lowest layer has a lowest framerate, then it increases. So, for a given frame only layers i..2 are present. Each frame references the previous of the same spatial layer and the lower one, if it exists.
The structure looks like that:
o----o----o----o----o----o----o----o----
|         |              |
o---------o--------------o--------------
|                        |
o------------------------o--------------
Thus, by choosing how many layers to forward, SFU could control the frame-rate. Resolution for screenshare is always the same.
This structure is thought to be better than the temporal layers because there shouldn't be visible quality flickering. Different temporal layers usually have different qualitiy. 
So if the user is shown tl0, tl1, tl0, tl2 frames, then every other frame is of a visibly worse quality. However, in the spatial layer structure, all the shown frames reference each other and there's way less quality flickering.
```

Since there are no temporal layers, in our `RtpLayerDesc` object the frame rate is reported as -1.  Our current code causes it to be treated as a low-frame-rate stream; this PR fixes that, which seems to fix the problem.

I probably should add some test cases for this case if I can figure out the right place to add them.

bgrozev (approved):

PR #1677, opened 2021-06-25
Only forward audio packets of the loudest speakers in a conference.

jqdrqgnq:


jbg:
Is there a way to turn this functionality off completely? (e.g. `numLoudest = -1`)

e.g. some deployments are using JVB with custom frontends and client-side mixing based on some external factors, and rely on all audio being forwarded by JVB all the time.

bgrozev:
Jenkins, add to whitelist

jvb/src/main/resources/reference.conf
```diff
@@ -289,6 +289,21 @@ videobridge {
     }
   }
 
+  loudest {
+      # The number of current loudest speakers to route audio packets for.
+      numLoudest = 3
```
bgrozev, line 294: Can we have a way to disable the new behavior and default to it?

jqdrqgnq: off by default now.

bgrozev (approved):

PR #1679, opened 2021-07-01
Make trace context properly-formatted JSON.

JonathanLennox:


JonathanLennox:
Without this, the conference name shows up in the trace JSON output as a bare string, not a quoted string.

bgrozev (approved):

PR #1683, opened 2021-07-14
Added option to start IceAgent with ephemeral port

nils-ohlmeier:
This should allow us to create endpoints which will use ephemeral ports for outgoing ICE connections to other JVBs.


jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -347,8 +349,11 @@ class IceTransport @JvmOverloads constructor(
     }
 
     companion object {
-        fun appendHarvesters(iceAgent: Agent) {
+        fun appendHarvesters(iceAgent: Agent, isJvbClient: Boolean) {
             Harvesters.initializeStaticConfiguration()
+            if (isJvbClient) {
+                Harvesters.singlePortHarvesters = SinglePortUdpHarvester.createHarvesters(0)
```
bgrozev, line 355: This replaces the static list of harvesters that normally bind on 10000!

I think we need something like:
```
if (isJvbClient) {
  iceAgent.addCandidateHarvester(SinglePortUdpHarvester(0))
} else {
  Harvesters.singlePortHarvesters?.forEach(iceAgent::addCandidateHarvester)
}
```

bgrozev:
I think it might be best to delay this until we know more about how we'll implement the bridge-to-bridge endpoints (e.g. if we use a new `Endpoint` class).

bgrozev:
Jenkins, add to whitelist

nils-ohlmeier:
> I think it might be best to delay this until we know more about how we'll implement the bridge-to-bridge endpoints (e.g. if we use a new `Endpoint` class).

Fair enough. I'll leave this here in case it will be useful.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
nils-ohlmeier: So I just pushed another version which uses addCandidateHarvester(). It also works in that the bridge advertises an ephemeral port. But since initializeStaticConfiguration() always sets up the harvester on port 10000, we now have two host harvesters running. I don't think that is desirable. That is why I decided to overwrite the static port 10000 harvester. Although I do admit that my previous choice is not a very nice one. I think we still would want a solution to remove the static 10000 harvester in case an ephemeral harvester is added.

bgrozev:
Closing as the feature ended up implemented in other PRs

PR #1684, opened 2021-07-14
Format IceTransport in Unix EOL format to match other files

nils-ohlmeier:
IceTransport.kt is the only file in the whole repo that is formatted with DOS style end of line. This converts it over to the Unix EOL style used by all other files in the repo.


JonathanLennox:
Jenkins, add to whitelist

JonathanLennox (approved):

PR #1691, opened 2021-07-19
Preserve VP9 sequence numbers, picId, and tl0picIdx across suspensions

JonathanLennox:
Otherwise Chrome's decoder can get confused, and decode the video incorrectly.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionContext.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionContext.kt
@@ -106,7 +106,7 @@ class Vp9AdaptiveSourceProjectionContext(
                 val projection: Vp9FrameProjection
                 try {
                     projection = createProjection(
-                        frame = frame, initialPacket = packet,
+                        frame = frame, initialPacket = packet, isResumption = acceptResult.isResumption,
                         isReset = result.isReset, mark = acceptResult.mark, receivedMs = receivedMs
                     )
                 } catch (e: Exception) {
@@ -231,6 +231,7 @@ class Vp9AdaptiveSourceProjectionContext(
         frame: Vp9Frame,
         initialPacket: Vp9Packet,
         mark: Boolean,
+        isResumption: Boolean,
         isReset: Boolean,
         receivedMs: Long
     ): Vp9FrameProjection {
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionContext.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionContext.kt
@@ -101,7 +101,7 @@ class Vp9AdaptiveSourceProjectionContext(
             val receivedMs = packetInfo.receivedTime
             val acceptResult = vp9QualityFilter
                 .acceptFrame(frame, incomingIndex, targetIndex, receivedMs)
-            frame.isAccepted = acceptResult.accept && checkDecodability(frame)
+            frame.isAccepted = acceptResult.accept
             if (frame.isAccepted) {
                 val projection: Vp9FrameProjection
                 try {
@@ -223,16 +223,6 @@ class Vp9AdaptiveSourceProjectionContext(
      */
     private fun findNextBaseTl0(frame: Vp9Frame) =
         vp9PictureMaps.get(frame.ssrc)?.findNextBaseTl0(frame)
-
-    /**
-     * For a frame that's been accepted by the quality filter, verify that
-     * it's decodable given the projection decisions about previous frames
-     * (in case the targetIndex has changed).
-     */
-    private fun checkDecodability(frame: Vp9Frame): Boolean {
-        /* TODO - use SS or flexible mode reference list */
-        return true
-    }
 
     /**
      * Create a projection for this frame.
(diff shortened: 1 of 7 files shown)


jvb/src/test/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionTest.kt
```diff
@@ -1010,13 +1010,172 @@ class Vp9AdaptiveSourceProjectionTest {
...
+            } else {
+                Assert.assertFalse(accepted)
+            }
+            if (endOfPicture) {
+                expectedTs = RtpUtils.applyTimestampDelta(expectedTs, 3000)
+                expectedPicId = applyExtendedPictureIdDelta(expectedPicId, 1)
+            }
+        }
+        var suspended = 64
+        while (suspended < 65536) {
```
bgrozev, line 1071: This loop is equivalent to `for(suspended = 64; suspended < 65536; suspended *= 2)`, right? I find the `for` version easier to read.

jvb/src/test/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionTest.kt
```diff
@@ -1010,13 +1010,172 @@ class Vp9AdaptiveSourceProjectionTest {
...
+                }
+            }
+            suspended *= 2
+        }
+    }
+
+    @Test
+    fun sourceSuspensionTest() {
+        val generator = ScalableVp9PacketGenerator(1)
+        runSourceSuspensionTest(generator, 2)
```
bgrozev, line 1177: nit: I think `getIndex(0, 0, 2)` would be easier to understand (or `getIndex(tid=2)` if you were to add default values to `getIndex`).

jvb/src/main/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionContext.kt
```diff
@@ -334,6 +335,49 @@ class Vp9AdaptiveSourceProjectionContext(
...
+            lastVp9FrameProjection.timestamp,
+            lastProjectedFrame.timestamp
+        )
+        val projectedTs = applyTimestampDelta(frame.timestamp, tsDelta)
+
+        /* Increment picId and tl0picidx by 1 from the last projected frame. */
+        val projectedPicId = applyExtendedPictureIdDelta(lastVp9FrameProjection.pictureId, 1)
+        val projectedTl0PicIdx = applyTl0PicIdxDelta(lastVp9FrameProjection.tl0PICIDX, 1)
+
+        /* Increment sequence numbers based on the last projected frame, but leave a gap
```
fippo, line 366: how does that work? You can see if the current packet is not the beginning of the frame (i.e. B is not set), but you can't know how many packets of the new frame are missing?

I'd expect more than one missing/reordered packet to be quite rare though.

jvb/src/main/kotlin/org/jitsi/videobridge/cc/vp9/Vp9QualityFilter.kt
```diff
@@ -110,7 +111,9 @@ internal class Vp9QualityFilter(parentLogger: Logger) {
                I think this shouldn't be a problem? */
             getSidFromIndex(incomingIndex) == getSidFromIndex(externalTargetIndex)
         }
-        return AcceptResult(accept, mark)
+        val isResumption = (prevIndex == SUSPENDED_INDEX && currentIndex != SUSPENDED_INDEX)
+        if (isResumption) assert(accept) // Pretty sure this has to be true?
```
fippo, line 115: nit: remove the comment
I don't see how you can have a resumption and then don't accept the frame either.

fippo (approved):
LGTM (modulo my missing kotlin skills) with one question and a nit

jvb/src/main/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionContext.kt
JonathanLennox: I'm keeping track of all received frames, not just those in frames that I've forwarded.  Thus I can calculate the sequence number gap between the most recent packet previously received and this packet, and so leave the equivalent gap in the outgoing sequence number space.

bgrozev:
LGTM too (except for the nits).

jvb/src/test/kotlin/org/jitsi/videobridge/cc/vp9/Vp9AdaptiveSourceProjectionTest.kt
JonathanLennox: Kotlin doesn't actually have C-style `for` loops, only `foreach`-style loops.

I can use a helper function of the type seen at https://stackoverflow.com/questions/66377331/how-do-you-iterate-a-for-loop-with-multiplication-or-division-kotlin .

bgrozev: That's annoying. You could also do something like
```
for (i in 5..16) {
  val suspended = 2.0.pow(i).toInt()
}
```

I'm fine with either version, including the original.

bgrozev (dismissed):

bgrozev (approved):

PR #1711, opened 2021-08-24
Fix removing ICE listeners.

bgrozev:


JonathanLennox (approved):

JonathanLennox:
It should be possible to leave the methods as they are, and just call them in the listeners, yes? I feel like that might be cleaner.

PR #1718, opened 2021-08-31
feat: Support owner encoded as a domain-bare JID.

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/xmpp/MediaSourceFactory.java
```diff
@@ -491,24 +486,24 @@ public static String getOwner(SourcePacketExtension source)
...
-     * Given a string containing a JID, return its StringPrep'd Resource part,
-     * or {@link Resourcepart#EMPTY}, as a string.
+     * Given a string, if it is a valid JID with a Resource part return its StringPrep'd Resource part. Otherwise,
+     * return the string itself.
      *
-     * Helper for {@link #getOwner}
+     * Helper for {@link #getOwner}.
      */
-    private static String getResourceFromJid(String jid)
+    private static String getResourceOrDomainFromJid(String jid)
```
JonathanLennox, line 501: I think this function name should change with your updated "whole-string" approach.

PR #1724, opened 2021-09-14
When dumping, run jstack and jmap as root, not the jvb user.

JonathanLennox:
This seems to be necessary with JVM 11, at least in our environment?


bgrozev:
Unfortunately with JVM 8 it's the other way round -- it needs to run as the user, and running as root won't work. Is it worth checking the JVM version, or should we just ignore 8 here?

bgrozev (approved):

PR #1726, opened 2021-09-15
Forward VideoType messages over Octo.

JonathanLennox:


jvb/src/main/java/org/jitsi/videobridge/octo/ConfOctoTransport.java
```diff
@@ -234,6 +238,8 @@ else if (remoteBridges.isEmpty())
                 bridgeOctoTransport.addHandler(conferenceId, this);
             }
             remoteBridges = Collections.unmodifiableMap(newRelays);
+
+            newRelaysAdded(newBridgeAddresses);
```
bgrozev, line 242: I think there's a race between this bridge learning about the new bridge, and the new bridge being ready to accept the message. Jicofo [sends the relay update to existing bridges before it creates a conference on the new bridge](https://github.com/jitsi/jicofo/blob/master/src/main/java/org/jitsi/jicofo/JitsiMeetConferenceImpl.java#L757).

bgrozev: Since this is not meant to be a permanent solution maybe we can just delay the message by e.g. 1s?

bgrozev (approved):

PR #1727, opened 2021-09-16
Fix "flickering" issues (use BWE threshold when BWE increases).

bgrozev:


jvb/src/test/kotlin/org/jitsi/videobridge/cc/allocation/BitrateControllerTest2.kt
```diff
@@ -0,0 +1,141 @@
...
+    private val B = Endpoint("B")
+    private val C = Endpoint("C")
+    private val D = Endpoint("D")
+    private val E = Endpoint("E")
+    private val F = Endpoint("F")
+    private val bc = BitrateControllerWrapper(listOf(A, B, C, D, E, F), clock = clock)
+
+    init {
+        bc.bc.endpointOrderingChanged()
+        val parsedLines = File("${System.getProperty("user.dir")}/src/test/resources/bwe-events.csv")
```
JonathanLennox, line 53: Is this guaranteed to be right both for mvn tests and IntelliJ?

bgrozev: > Is this guaranteed to be right both for mvn tests and IntelliJ?

In my environment it works in both

JonathanLennox (approved):

jvb/src/test/kotlin/org/jitsi/videobridge/cc/allocation/BitrateControllerTest2.kt
ibauersachs: BitrateControllerTest2::class.java.getResourceAsStream("/bwe-events.csv") would be the proper way to retrieve a resource file.

ibauersachs:
@bgrozev BitrateControllerTest2 is not picked up by Maven because it doesn't match the usual test patterns (see https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html) and it fails - sometimes - when running from within IntelliJ: it's successful if the Kotest plugin is installed and ran standalone, but it fails if it's run as part of the testsuite (both the entire jvb suite, as well as just the allocation folder).

Is there some state cleanup that needs to be done?

bgrozev:
@ibauersachs Thanks, fixed in #1766

PR #1729, opened 2021-09-17
fix: Signaling recent speakers fixes.

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -413,42 +414,36 @@ private void lastNEndpointsChanged()
     /**
      * Notifies this instance that {@link #speechActivity} has identified a speaker switch event and there is now a new
      * dominant speaker.
+     * @param recentSpeakers the list of recent speakers (including the dominant speaker at index 0).
      */
-    private void dominantSpeakerChanged()
+    private void dominantSpeakerChanged(List<AbstractEndpoint> recentSpeakers)
```
JonathanLennox, line 419: Do you want to rename this method as well?

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -413,42 +414,36 @@ private void lastNEndpointsChanged()
...
         {
-            logger.info("ds_change ds_id=" + dominantSpeakerId);
+            logger.info("Recent speakers changed: " + recentSpeakersIds);
             getVideobridge().getStatistics().totalDominantSpeakerChanges.increment();
         }
+        broadcastMessage(new DominantSpeakerMessage(recentSpeakersIds));
 
-        if (dominantSpeaker != null)
+        AbstractEndpoint dominantSpeaker = recentSpeakers.get(0);
+        if (getEndpointCount() > 2 && dominantSpeaker != null)
```
JonathanLennox, line 431: Is it possible for the recentSpeakers list to change without the dominant speaker changing?  (E.g. if endpoints further down the list left?) If so, won't this send unnecessary keyframe requests?

bgrozev: Yes, and yes. What would you think about removing the preemptive keyframe requests altogether?

JonathanLennox: Ideally we'd only send preemptive keyframes when someone's actually likely to be moved into on-stage...with tile view being the default now, that's a lot less likely than it used to be.  But I'd be a bit reluctant to degrade the experience for that case.  I wonder how hard it is to distinguish...

bgrozev: I pushed a fix to the regression you found.

 I'll follow up in another PR with sending this only when some local endpoint has someone "on-stage".

jvb/src/test/kotlin/org/jitsi/videobridge/SpeechActivityTest.kt
```diff
@@ -22,6 +22,7 @@ import io.kotest.matchers.collections.shouldContainInOrder
 import io.kotest.matchers.shouldBe
 import io.mockk.every
 import io.mockk.mockk
+import org.jitsi.videobridge.message.DominantSpeakerMessage
```
JonathanLennox: Accidental import?

bgrozev: Yep, fixed

JonathanLennox (approved):

PR #1736, opened 2021-09-23
doc: Add config to enable the REST API to rest.md (reported by @tpo)

bgrozev:


JonathanLennox (approved):

tpo:
Thanks a lot @bgrozev :+1: !

tpo:
@bgrozev Unfortunately that doesn't work in my current jitsi-videobridge2 (2.1-551-g2ad6eb0b-1):

It seems like the system is using "old style config" aka `sip-communicator.properties`. I do have a `jvb.conf` and if I put there:

```
/etc/jitsi/videobridge # cat jvb.conf 
videobridge {
                apis {
                    rest {
                        enabled = true
                    }
                }
            }
```
it won't make the api start unless I put `JVB_OPTS="--apis=rest"` into the systemd `/etc/jitsi/videobridge/config`

bgrozev:
If I'm reading the code correctly, the only way this could happen is if another value is passed as `--apis`. Can you please check what parameters are passed to java if you remove `JVB_OPTS="--apis=rest"`?

tpo:
@bgrozev if I comment out `JVB_OPTS` in `/etc/jitsi/videobridge/config`. Then the videobridge will not start. `/var/log/jitsi/jvb.log` will only contain repeatedly:

```
Usage:
/usr/share/jitsi-videobridge/jvb.sh [OPTIONS], where options can be:
        --apis=APIS where APIS is a comma separated list of APIs to enable. Currently the only supported API is 'rest'. The default is none.
```

as systemd tries to re-re-restart it (that's how I'm interpreting what I'm seeing).

If I reinsert into `/etc/jitsi/videobridge/config` the original `JVB_OPTS="--apis=,"` (which figured in the config file as shipped by the jitsi-vdeobridge package or by its postinst scripts or similar). Then setting:

```
 # cat /etc/jitsi/videobridge/jvb.conf 
videobridge {
                apis {
                    rest {
                        enabled = true
                    }
                }
            }
```

will apparently **not** override the `JVB_OPTS="--apis=,"` setting and the videobridge will not expose the REST API.

bgrozev:
Oh, I see. The shell script and possibly debian packages need an update to not include "--apis=" as a param (which overrides jvb.conf and disables the REST API). I'll work on it when I have some time, but I'm not sure when. Thanks for helping uncover this!

tpo:
@bgrozev care will have to be taken when manipulating the config files or respectively removing settings from config files: the user might have modified the config file. I'm not sure if you are aware, but if the config are be declared as such in Debian/Ubuntu (listed in /var/lib/dpkg/info/*.conffiles), then the package manager will ask the user for feedback before replacing config files (provided that the user is upgrading the package in an interactive session...).

Another option is the /usr/share/doc/$package/NEWS.Debian(.gz) file which should be shown to the user on new entries there. That way the user could be made aware of new/obsolete/deprecated/ineffective etc. settings.

You might notice that I do not really know about the jitsimeet config mechanisms, so please ignore my comment here if it's irrelevant and please excuse the noise.

bgrozev:
@tpo Thanks for your comments. I opened #1738 as a first step -- new installations will not place "--apis" in /etc/jitsi/videobridge/config, but it is still supported (with a deprecation warning logged in the appication log file). I think (@damencho  correct me if I'm wrong) that we take rersponsibility for not overriding /etc/jitsi/videobridge{jvb.conf,config} in the postinstall script, but the files are not declared as conffiles in dpkg. 

We could add some logic to postinstall to read move the "enable REST" option from `config` to `jvb.conf`. Then we can competely remove handing of `--apis` in java. But I'll leave that for a later stage.

damencho:
> I think (@damencho correct me if I'm wrong) that we take rersponsibility for not overriding /etc/jitsi/videobridge{jvb.conf,config} in the postinstall script, but the files are not declared as conffiles in dpkg.

Yep, we try not to touch does files as people may had already done some customizations and we do not want to overwrite those.

PR #1738, opened 2021-09-24
ref: Do not use --apis in new debian installations * Do not require it in jvb.sh * Log a warning if present

bgrozev:


damencho (approved):

bgrozev:
jenkins please test this again please

PR #1752, opened 2021-10-27
feat: implement SourceVideoTypeMessage

paweldomas:
Adds support for the new SourceVideoTypeMessage which allows the client to signal video type on a per media source basis.

The feature is enabled only when videobridge.multi-stream.enabled = true


jvb/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -16,6 +16,7 @@
 package org.jitsi.videobridge;
 
 import org.jetbrains.annotations.*;
+import org.jitsi.nlj.VideoType;
```
bgrozev, line 19: nit: package imports

jvb/src/main/kotlin/org/jitsi/videobridge/SourceNameSignalingConfig.kt
```diff
@@ -0,0 +1,15 @@
+package org.jitsi.videobridge
```
bgrozev, line 1: License header

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -430,6 +433,37 @@ class ReceiverVideoConstraintsMessage(
...
+    endpointId: String? = null
+) : BridgeChannelMessage(TYPE) {
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    var endpointId: String? = endpointId
+        set(value) {
+            field = value
+            resetJsonCache()
+        }
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
bgrozev, line 455: This is always non-null

jvb/src/main/resources/reference.conf
```diff
@@ -294,6 +294,11 @@ videobridge {
     }
   }
 
+  # The experimental multiple streams per endpoint support
+  source-name-signaling {
```
bgrozev, line 298: Should we add a more generic "multi-stream" config context?

paweldomas: can we keep it as is for now with a note here that it's experimental and subject to be removed without warning?
I can't come up with a good name, because can't tell what properties will be there yet. I imagine maybe something around default behavior when picking streams, but the "multi-stream" may not be the best name at that point.

paweldomas: or I will change it, even if this may change it's better to start with something more generic

paweldomas:
jenkins test this please

paweldomas:
jenkins test this please

paweldomas:
There's 0 test failures after a rerun. The one that failed was unrelated (AV moderation).

bgrozev (approved):

PR #1756, opened 2021-10-28
fix: Delay shutdown in graceful mode.

bgrozev:


bgrozev:
@JonathanLennox can you review again please?

jvb/src/main/resources/reference.conf
```diff
@@ -321,8 +321,11 @@ videobridge {
   }
 
   version {
-    // Wheather to announe the jitsi-videobridge version to clients in the ServerHello message.
+    # Wheather to announe the jitsi-videobridge version to clients in the ServerHello message.
```
JonathanLennox, line 324: Typos: "Whether", "announce".  This is old code but might as well fix it when you're editing the line?

JonathanLennox (dismissed):

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -255,18 +264,23 @@ public Videobridge(
...
      */
     private void enableGracefulShutdownMode()
     {
-        if (!shutdownInProgress)
+        if (shutdownRequestedTime == null)
         {
             logger.info("Entered graceful shutdown mode");
         }
-        this.shutdownInProgress = true;
+        this.shutdownRequestedTime = clock.instant();
```
JonathanLennox, line 275: Hm, just noticed this - if enableGracefulShutdownMode is called more than once, this reset the shutdown time.  Is that what we want?

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -598,7 +630,7 @@ public void stop()
     public OrderedJsonObject getDebugState(String conferenceId, String endpointId, boolean full)
     {
         OrderedJsonObject debugState = new OrderedJsonObject();
-        debugState.put("shutdownInProgress", shutdownInProgress);
+        debugState.put("shutdownInProgress", (shutdownRequestedTime != null));
```
JonathanLennox, line 633: Do we want to have the actual requested time in the debug state?

bgrozev: I think having it in the logs is sufficient.

JonathanLennox (approved):

PR #1758, opened 2021-11-11
feat: Add an enable-rtcstats field to ColibriConferenceIQ.

gpolitis:


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1174,6 +1181,12 @@ public JSONObject getDebugState(boolean full, String endpointId)
     {
         JSONObject debugState = new JSONObject();
         debugState.put("id", id);
+
+        if (!isRtcStatsEnabled)
```
bgrozev, line 1185: This would also prevent us from manually debugging such conferences. How about including the `isRtcStatsEnabled` flag in the output and having the rtcstats pusher skip conferences with the flag set?

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -231,7 +232,7 @@ public Videobridge(
      */
     public @NotNull Conference createConference(EntityBareJid name)
     {
-        return createConference(name, Conference.GID_NOT_SET, null);
+        return createConference(name, Conference.GID_NOT_SET, null, false);
```
bgrozev, line 235: Looks like this API is only exposed for testing. Can you add a comment to the effect of "we default to rtcstatsEnabled=false because this is only used for testing"? I'm wary of us later introducing a bug where rtcstatsEnabled=false because we didn't pay attention to the default.

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -430,7 +432,8 @@ private void handleColibriRequest(XmppConnection.ColibriRequest request)
             return createConference(
                     conferenceIq.getName(),
                     ColibriUtil.parseGid(conferenceIq.getGID()),
-                    conferenceIq.getMeetingId());
+                    conferenceIq.getMeetingId(),
+                    conferenceIq.isRtcStatsEnabled());
```
bgrozev, line 436: This will result in rtcstats being disabled when the bridge is used with older jicofo versions or other colibri implementations. Can we default to `true` (just like we default to `true` for the flag in jicofo)

bgrozev (changes requested):

jvb/src/main/java/org/jitsi/videobridge/Conference.java
gpolitis: Will do!

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
gpolitis: Yes, I will make the appropriate change in jitsi-xmpp-extensions.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
gpolitis: https://github.com/jitsi/jvb-rtcstats-push/pull/8

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1174,6 +1193,8 @@ public JSONObject getDebugState(boolean full, String endpointId)
     {
         JSONObject debugState = new JSONObject();
         debugState.put("id", id);
+        debugState.put("rtcstatsEnabled", isRtcStatsEnabled);
```
bgrozev, line 1196: Should we include `isCallstatsEnabled`, too?

bgrozev (approved):

jvb/src/main/java/org/jitsi/videobridge/Conference.java
gpolitis: Might be useful, I'll add it

PR #1759, opened 2021-11-11
fix: use dummy values for raddr and rport instead of dropping them.

nils-ohlmeier:
Firefox can't handle ICE candidates where raddr or rport are missing. So we just use raddr=0.0.0.0 and rport=0 instead of dropping these values from the ICE candidate.


bgrozev (approved):

JonathanLennox:
Technically 8839 says the raddr should be set to "::" rather than "0.0.0.0" for IPv6 addresses, but it doesn't really matter.

PR #1760, opened 2021-11-14
Update dependencies, cleanup deprecations

ibauersachs:
This depends on:
- jitsi/jitsi-xmpp-extensions#55
- jitsi/jicoco#157

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -14,21 +14,25 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        java: [ 8, 11, 17 ]
+
+    name: Java ${{ matrix.java }}
+
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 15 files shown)


bgrozev:
I think the tests require a PR updating jicofo (jicofo's unit tests depend on jitsi-videobridge and probably end up running with the wrong version of jitsi-xmpp-extensions).

ibauersachs:
Thanks. I forgot about jicofo and will create a PR there as well soon. I'd appreciate it if you could look at the linked PRs so I can incorporate any necessary changes here and in jicofo.

ibauersachs:
Okay, so the torture tests of JVB (and Jicofo too) finally passed. The previous runs failed because a wrong parser registration for an extension just blocked within ColibriConferenceIqProvider. Regardless of the registration mistake, this shouldn't happen and I'll try to do more cleanup of ColibriConferenceIqProvider in a distinct PR.

bgrozev (approved):

PR #1763, opened 2021-11-17
colibri2 and secure octo

JonathanLennox:


jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -446,6 +567,47 @@ private void handleColibriRequest(XmppConnection.ColibriRequest request)
...
+                logger.warn("Will not create conference, conference already exists for meetingId=" + meetingId);
+                throw new ConferenceAlreadyExistsException();
+            }
+            if (isShutdownInProgress())
+            {
+                logger.warn("Will not create conference in shutdown mode.");
+                throw new InGracefulShutdownException();
+            }
+
+            /* TODO: race condition if something else created a meeting with this meetingId since the lookup above? */
```
bgrozev, line 591: It looks safe to just encapsulate all of this in `synchronized(conferencesById)`

jvb/src/main/kotlin/org/jitsi/videobridge/EndpointConnectionStatusMonitor.kt
```diff
@@ -129,8 +129,13 @@ class EndpointConnectionStatusMonitor @JvmOverloads constructor(
...
             conference.broadcastMessage(msg, true)
         } else {
-            val ep = conference.getEndpoint(receiverEpId)
-            conference.sendMessage(msg, listOf(ep), false)
+            val ep = conference.getLocalEndpoint(receiverEpId)
+            if (ep != null) {
+                conference.sendMessage(msg, listOf(ep), false)
+            } else {
+                /* TODO: send only to the relevant relays */
+                conference.broadcastMessage(msg, true)
```
bgrozev, line 137: I don't think this needs to be broadcast. The case where `receiverEpId` is not null is when its associated endpoint connected and needs to be notified of other endpoints' status. With just a bit of changes we can pass the `Endpoint` object itself instead its ID.

JonathanLennox: Ah, there's never a case where we need to direct this to a remote endpoint, ok.

The main issue with changing this to take an `Endpoint` would be adjusting `EndpointConnectionStatusMonitorTest`, which uses a lot of mockk-code which I don't understand very well.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayedEndpoint.kt
```diff
@@ -0,0 +1,78 @@
...
+
+class RelayedEndpoint(
+    conference: Conference,
+    val relay: Relay,
+    id: String,
+    parentLogger: Logger
+) : AbstractEndpoint(conference, id, parentLogger) {
+    var audioSources: Array<AudioSourceDesc> = arrayOf()
+
+    override fun receivesSsrc(ssrc: Long): Boolean = audioSources.any { ssrc == it.ssrc } ||
```
bgrozev, line 35: This could be `relay.endpointsBySsrc[ssrc] == this`?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -0,0 +1,549 @@
...
+        }
+        updateTransceiverSources()
+    }
+
+    fun removeRemoteEndpoint(id: String) {
+        val ep: RelayedEndpoint?
+        synchronized(endpointsLock) {
+            ep = relayedEndpoints.remove(id)
+            if (ep != null) {
+                endpointsBySsrc.keys.removeAll(ep.ssrcs)
```
bgrozev, line 436: I don't think it's possible for the endpoint's SSRC and the map to get out of sync, but maybe cleaning by the endpoint would make it more obvious? `endpointsBySsrc.entries.removeIf { it.value == ep }`

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -0,0 +1,426 @@
...
+     * {@inheritDoc}
+     */
+    override fun notifyTransportChannelConnected() {
+        relay.relayMessageTransportConnected()
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    override fun clientHello(message: ClientHelloMessage): BridgeChannelMessage? {
```
bgrozev, line 119: I don't think we need the ClientHello/ServerHello exchange. The websocket events should be sufficient

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -954,6 +1035,17 @@ void endpointExpired(AbstractEndpoint endpoint)
...
     }
 
+    /**
+     * Notifies this conference that one of its relays has expired.
+     *
+     * @param relay the <tt>Relay</tt> which expired.
+     */
+    public void relayExpired(Relay relay)
+    {
+        String id = relay.getId();
```
bgrozev, line 1045: I think we need to call `removeReceiver` for all local endpoints here (like in `endpointExpired`).

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -0,0 +1,426 @@
...
+     * @return
+     */
+    override fun addReceiver(message: AddReceiverMessage): BridgeChannelMessage? {
+        val epId = message.endpointId
+        val ep = relay.conference.getEndpoint(epId) ?: run {
+            logger.warn("Received AddReceiverMessage for unknown epId $epId")
+            return null
+        }
+
+        ep.addReceiver(message.bridgeId, message.videoConstraints)
```
bgrozev, line 147: I think we should use `relay.id` here instead of relying on the value in the message (the value in the message is redundant for relays). This way we're sure we can remove the correct receiver when the relay is expired.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -0,0 +1,549 @@
...
+    fun describeTransport(): IceUdpTransportPacketExtension {
+        val iceUdpTransportPacketExtension = IceUdpTransportPacketExtension()
+        iceTransport.describe(iceUdpTransportPacketExtension)
+        dtlsTransport.describe(iceUdpTransportPacketExtension)
+        val wsPacketExtension = WebSocketPacketExtension()
+
+        /* TODO: this should be dependent on videobridge.websockets.enabled, if we support that being
+         *  disabled for relay.
+         */
+        if (_messageTransport.isActive) {
```
bgrozev, line 262: I think we should stop adding the "active" flag in signaling, as it's not actually used and the semantics are getting confusing. I think `_messageTransport.isActive` means "I have been told to connect to remote relay". We describe the transport only when responding to the "create" message (I just pushed that change), anyway, so we can simply always describe the WS with the URL.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -0,0 +1,426 @@
...
+     */
+    override fun webSocketClosed(ws: ColibriWebSocket, statusCode: Int, reason: String) {
+        synchronized(webSocketSyncRoot) {
+            if (ws == webSocket) {
+                webSocket = null
+                logger.debug { "Web socket closed, statusCode $statusCode ( $reason)." }
+            }
+        }
+        if (outgoingWebsocket != null) {
+            // Try to reconnect.  TODO: how to handle failures?
```
bgrozev, line 294: Is this going to race with `relay` expiration and re-connect only to be closed as soon as the colibri message arrives?

jvb/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -573,15 +581,16 @@ public BridgeChannelMessage endpointMessage(EndpointMessage message)
...
-                targets = Collections.emptyList();
-                sendToOcto = true;
+                targets = Collections.singletonList((Endpoint)targetEndpoint);
+                sendToOcto = false;
             }
             else if (targetEndpoint != null)
             {
-                targets = Collections.singletonList(targetEndpoint);
-                sendToOcto = false;
+                /* TODO: for relayed endpoints, send only to the relevant relay. */
```
bgrozev, line 591: I think, if we don't insist of going through `Conference`, we can do this with just `((RelayedEndpoint) targetEndpoint).getRelay().sendMessage(message);`

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -0,0 +1,426 @@
...
+            return null
+        }
+        val targets = conference.localEndpoints.stream()
+            .filter { ep: Endpoint ->
+                ep.wantsStatsFrom(
+                    from
+                )
+            }
+            .collect(Collectors.toList())
+        conference.sendMessage(message, targets, false)
```
bgrozev, line 413: Do you think it's necessary to send the message via `Conference`? I think I did that at some point, but I don't remember if there was a good reason anymore. It would be simpler to just do `localEndpoints.filter { ...}.forEach { it.sendMessage(message) }`

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
JonathanLennox: That iterates over all the elements in the `endpointsBySsrc` map, though, so it'll be slower for large conferences.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
JonathanLennox: The one thing it provides is the peer JVB version, which would let us know if the servers are out of sync for some reason.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
JonathanLennox: `Conference.endpointExpire` does call `removeReceiver`, unless I"m missing something?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
JonathanLennox: I think the case this is meant to handle is if the initial create message actually included the peer's transport (i.e. you're doing an offer/answer-style negotiation rather than the dual-create style), then there's no need for us to send the URL.  But maybe that's unnecessary since you're not planning on using that mode in jicofo.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
JonathanLennox: If we haven't called `close` here yet, yeah.  Do you think that's a problem?  Do you think we should listen for the status code and treat certain codes as a deliberate close?

jvb/src/main/java/org/jitsi/videobridge/Conference.java
JonathanLennox: Oh, right, the receiver in this case is the relay itself, not the relayed endpoint.  Added.

JonathanLennox: Oh, right, the receiver in this case is the relay itself, not the relayed endpoint.  Added.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1044,6 +1044,8 @@ public void relayExpired(Relay relay)
     {
         String id = relay.getId();
         relaysById.remove(id);
+
+        endpointsById.forEach((i, senderEndpoint) -> senderEndpoint.removeReceiver(id));
```
bgrozev, line 1048: With a full-mesh we should only ever be doing `Endpoint.addReceiver` for local endpoins. Should we enforce it and change this to iterate over localEndpoints?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
bgrozev: Should we log a warning if that's the case?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
bgrozev: I think so, looking at [the docs](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/websocket/api/StatusCode.html) it looks like NORMAL and maybe SHUTDOWN can be considered deliberate.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
JonathanLennox: Not a bad plan.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1022,7 +1022,7 @@ public void endpointExpired(AbstractEndpoint endpoint)
             updateEndpointsCache();
         }
 
-        endpointsById.forEach((i, senderEndpoint) -> senderEndpoint.removeReceiver(id));
+        getLocalEndpoints().forEach(senderEndpoint -> senderEndpoint.removeReceiver(id));
```
bgrozev, line 1025: I think local endpoint can be subscribed to remote endpoints.

bgrozev (approved):

jvb/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -561,40 +570,37 @@ public BridgeChannelMessage endpointMessage(EndpointMessage message)
...
-            if (targetEndpoint instanceof OctoEndpoint)
+            if (targetEndpoint instanceof Endpoint)
+            {
+                ((Endpoint)targetEndpoint).sendMessage(message);
+            }
+            else if (targetEndpoint instanceof RelayedEndpoint)
             {
-                targets = Collections.emptyList();
-                sendToOcto = true;
+                ((RelayedEndpoint)targetEndpoint).getRelay().sendMessage(message);
```
bbaldino, line 592: Why not expose `sendMessage` on `RelayedEndpoint` and the implementation of it just calls `getRelay` first?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -0,0 +1,629 @@
...
+import java.time.Clock
+import java.time.Instant
+import java.util.concurrent.TimeUnit
+import java.util.function.Supplier
+
+/**
+ * Models a relay (remote videobridge) in a [Conference].
+ */
+/* TODO: figure out how best to share code between this and [Endpoint], without multiple inheritance. */
+class Relay @JvmOverloads constructor(
```
bbaldino, line 74: `Relay` models a collection of remote endpoints, is that right?  It seems like it having a `conference` member kinda inverts the hierarchy a bit: ideally the methods it needs to call there would be passed here by some kind of interface rather than making this depend directly on `Conference`.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -0,0 +1,629 @@
...
+            override fun onRttUpdate(newRttMs: Double) {
+                if (newRttMs > 0) {
+                    transceiver.removeEndpointConnectionStatsListener(this)
+                    iceTransport.updateStatsOnInitialRtt(newRttMs)
+                }
+            }
+        }
+
+    /* TODO: we eventually want a smarter Transceiver implementation, that splits processing by
+     *  source endpoint or something similar, but for the initial implementation this should work.
```
bbaldino, line 132: So this loses the concept of demuxing remote incoming packets to individual receive pipelines per remote endpoint?  Before I think I had it set up where there were independent rx pipelines leveraging a shared transport--but maybe I'm misremembering.

bbaldino (commented):
A few random comments.  Overall I'm a bit confused by the hierarchy with the Relay/RelayedEndpoint abstractions: at least the way I had always hoped to model this would be that RemoteEndpoints would share an interface with regular ones and just have a different sort of `transport` situation (i.e. shared network and message transports); generally keeping them as similar as possible to hide away the fact that they are remote from most of the code.  This feels like it "special cases" them a bit, but it's entirely possible I just haven't grokked everything or aren't aware of other limitations.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
JonathanLennox: Yeah - I have this as a TODO in case it ends up being useful, but it'd be pretty complicated, because (unlike old Octo) this has NACK support, so we'd need to wire up retransmission across receive pipelines (because you can't tell which remote source a NACK is for until the SRTCP has been decrypted). We'd also need to propagate SRTP keys across transceivers, etc.  None of this is impossible, but I thought it wasn't needed for MVP.

We do have separate receive and send pipelines per remote relay, so once you get more than two bridges in a conference, we get some degree of parallelization.

I haven't done the performance testing of this yet; whether there's any bottleneck here is one of the things I'll be looking for.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
JonathanLennox: It's a collection of remote endpoints, but also the transport connection to it.

I don't really see that this is that different from `AbstractEndpoint` having a `conference` member.  It might be good to abstract all of that, but I don't think it'd be sensible to abstract it just for relays.

jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceShim.kt
```diff
@@ -0,0 +1,340 @@
...
+import org.jitsi.xmpp.extensions.colibri2.Sources
+import org.jitsi.xmpp.extensions.colibri2.Transport
+import org.jitsi.xmpp.extensions.jingle.ParameterPacketExtension
+import org.jitsi.xmpp.extensions.jingle.RTPHdrExtPacketExtension
+import org.jitsi.xmpp.extensions.jingle.SourceGroupPacketExtension
+import org.jitsi.xmpp.util.IQUtils
+import org.jivesoftware.smack.packet.IQ
+import org.jivesoftware.smack.packet.StanzaError.Condition
+
+class Colibri2ConferenceShim(
```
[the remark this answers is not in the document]
JonathanLennox: Maybe calling it a shim in this case is a misnomer, but it does need code to process the signaling messages and act upon that.  Keeping that isolated seems reasonable to me.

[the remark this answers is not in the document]
bbaldino: Ah ok, yeah maybe just the name then.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
bbaldino: Ah, I forgot `AbstractEndpoint` had `conference` member, yeah I agree would be nice if that was gone too, but not so bad that this doesn't do anything too different, I guess.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
bbaldino: I remember refactoring the Octo stuff we had to move _away_ from the single collection of endpoints and model them more like normal endpoints, just with the shared transport, etc. stuff.  Performance I do think was one benefit, but I think what I liked most is that it felt like it fit more cleanly into the design without these being special-cased.  I think for the most part that worked?

I think I'm missing something about the nack scenario you're talking about: are you talking about the JVB nacking a packet that wasn't received from a remote endpoint?  Or a nack sent to the JVB from a local endpoint that corresponds to a remote endpoint?  I guess not the former, because that seems straightforward.  Maybe not the latter either, because we already have to decrypt all RTCP coming in from local endpoints anyways, no?

JonathanLennox:
> A few random comments. Overall I'm a bit confused by the hierarchy with the Relay/RelayedEndpoint abstractions: at least the way I had always hoped to model this would be that RemoteEndpoints would share an interface with regular ones and just have a different sort of transport situation (i.e. shared network and message transports); generally keeping them as similar as possible to hide away the fact that they are remote from most of the code. This feels like it "special cases" them a bit, but it's entirely possible I just haven't grokked everything or aren't aware of other limitations.

The main difference between relayed endpoints and local ones is that you need to send broadcast messages to every relayed endpoint, whereas you send them only once per relay.  Other than that, things are pretty shared, for the most part, I think.

jvb/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
JonathanLennox: Because I don't want to accidentally iterate over all the `RelayedEndpoint`s and call `sendMessage` on each of them - forcing me to explicitly acknowledge I'm sending through the relay seems more likely to avoid that.

bbaldino:
> > A few random comments. Overall I'm a bit confused by the hierarchy with the Relay/RelayedEndpoint abstractions: at least the way I had always hoped to model this would be that RemoteEndpoints would share an interface with regular ones and just have a different sort of transport situation (i.e. shared network and message transports); generally keeping them as similar as possible to hide away the fact that they are remote from most of the code. This feels like it "special cases" them a bit, but it's entirely possible I just haven't grokked everything or aren't aware of other limitations.
> 
> The main difference between relayed endpoints and local ones is that you need to send broadcast messages to every relayed endpoint, whereas you send them only once per relay. Other than that, things are pretty shared, for the most part, I think.

Good point.  

> > A few random comments. Overall I'm a bit confused by the hierarchy with the Relay/RelayedEndpoint abstractions: at least the way I had always hoped to model this would be that RemoteEndpoints would share an interface with regular ones and just have a different sort of transport situation (i.e. shared network and message transports); generally keeping them as similar as possible to hide away the fact that they are remote from most of the code. This feels like it "special cases" them a bit, but it's entirely possible I just haven't grokked everything or aren't aware of other limitations.
> 
> The main difference between relayed endpoints and local ones is that you need to send broadcast messages to every relayed endpoint, whereas you send them only once per relay. Other than that, things are pretty shared, for the most part, I think.

Good point, that is a bit of an awkward thing to handle.

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
JonathanLennox: I'm talking about a JVB NACKing a packet that wasn't received from the remote JVB - the purpose of this project was to allow Octo connections to go over the public Internet, and one of the things that means is that this connection might be losing packets.

If you have a `Transceiver` object per remote endpoint, then when the upstream JVB receives the NACK, it doesn't know which `Transceiver` it should feed the NACK to.  We'd have to wire up some sort of cross-`Transceiver` NACK handling in order to get this to work.

bbaldino: Ah ha, I see now.  It's a common set of SRTP keys for all the end points, right?  Maybe this would involve moving the SRTP decrypt into the 'common' logic shared across all the receivers, and the receive pipelines pick up after that.  This would also mean they'd have separate entry points for RTP and RTCP (since that demuxing would've already been done).  Given the SRTP is probably on the heavier side of the work, it does mean the performance gain of parallelization would be lost, but maybe that's just how it would have to be.  Would keep the relayed endpoints less "special" potentially, though.

bgrozev (approved):

PR #1767, opened 2021-11-23
feat: allocate multiple streams per endpoint

paweldomas:
Extract Layers data structures into a separate file.

Creates SingleSourceAllocation2.kt which is to be used
by the multi-stream support flow. Eventually this will
be merged later once all changes are identified and
should we still agree it makes sense to merge (as opposed
to keeping separate and deleting old classes once deprecated).

Branches off the multi stream feature flag for the allocation
algorithm in the BandwidthAllocator.java.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
@@ -22,6 +22,7 @@ import org.jitsi.nlj.*;
 import org.jitsi.utils.event.*;
 import org.jitsi.utils.logging.*;
 import org.jitsi.utils.logging2.Logger;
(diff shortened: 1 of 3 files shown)


jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
```diff
@@ -22,6 +22,7 @@
 import org.jitsi.utils.event.*;
 import org.jitsi.utils.logging.*;
 import org.jitsi.utils.logging2.Logger;
+import org.jitsi.videobridge.MultiStreamConfig;
```
bgrozev, line 25: Package imports, let's fix it in the next PR

bgrozev (dismissed):

bgrozev (dismissed):

bgrozev (approved):

Issue #1773, opened 2021-12-10
log4j vulnerability CVE-2021-44228.


daenney:
There's a high severity vulnerability out for log4j, CVE-2021-44228. The videobridge appears to use log4j.

I'd like to understand if:

- The bridge is vulnerable?
- If so, what's the timeline for a fix?

jbg:
A fast mitigation in advance of updates being available is to add:

```
-Dlog4j2.formatMsgNoLookups=true
```

to the JVM command line. This is also relevant to other Jitsi components.

dkasak:
According to https://news.ycombinator.com/item?id=29507263, that mitigation is only available in versions >= 2.10.0, but Jitsi appears to be using 2.3.

jbg:
Sigh!

The comment also mentions this:
- Modify every logging pattern layout to say %m{nolookups} instead of %m in your logging config files, see details at https://issues.apache.org/jira/browse/LOG4J2-2109

... which is interesting because our logging patterns don't have `%m` in them anyway.

jbg:
So far as I can see, `log4j` is not being used directly, only as a transitive dependency from callstats:

```
jbg@Jaspers-MacBook-Air:[~/dev/avstack/jitsi-videobridge]: mvn -pl org.jitsi:jitsi-videobridge dependency:tree -Dincludes=org.apache.logging.log4j:log4j-api
[INFO] org.jitsi:jitsi-videobridge:jar:2.1-SNAPSHOT
[INFO] \- org.jitsi:jitsi-stats:jar:1.0-7-g2a9b765:compile
[INFO]    \- io.callstats:callstats-java-sdk:jar:5.2.1:compile
[INFO]       \- org.apache.logging.log4j:log4j-api:jar:2.3:compile
```

saghul:
Hey folks. The team has been looking into it since this morning. Thanks for being vigilant!

daenney:
> Sigh!
> 
> The comment also mentions this:
> 
>     * Modify every logging pattern layout to say %m{nolookups} instead of %m in your logging config files, see details at https://issues.apache.org/jira/browse/LOG4J2-2109
> 
> 
> ... which is interesting because our logging patterns don't have `%m` in them anyway.

I'm not sure who "our" is referring too, but at least for the video bridge they are in use: https://github.com/jitsi/jitsi-videobridge/blob/56e0dae00c503470cf38d44650ddf20cdf228529/config/log4j2.xml#L9

[...]

PR #1774, opened 2021-12-10
log4j2.xml: Mitigate CVE-2021-44228

daenney:
Relates to #1773.

This is one of a few possible mitigations for the log4j2 RCE. This avoids the need to immediately update the dependency, though that should happen afterwards.

If we can get a rebuild if the packages and the container with this fix merged folks can get to deploying the update in case they don't have an easy way of directly managing `log4j2.xml` themselves right now.

### DIFF
--- a/config/log4j2.xml
+++ b/config/log4j2.xml
@@ -6,7 +6,7 @@
     <Appenders>
         <RollingFile name="RollingFile" fileName="${log-path}/cs.log"
                  filePattern="${log-path}/$${date:yyyy-MM}/cs-%d{MM-dd-yyyy}-%i.log.gz">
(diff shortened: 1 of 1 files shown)


daenney:
This should be mergable without a CLA. It's a format marker change in a config file. By the time I get my employer to sign this CLA there'll hopefully be a release out with the updated dependency.

bgrozev:
Thanks for the contribution!

We're investigating what our expose to the issue is, and we'll provide a fix ASAP (log4j is just a transitive dependency and a fix has already been [pushed upstream](https://github.com/callstats-io/callstats.java/pull/50)).

bgrozev:
#1776 updates the log4j dependency mitigating the issue

PR #1780, opened 2021-12-16
Feat: source name based receiver constraints

paweldomas:


jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/Prioritize.kt
```diff
@@ -39,6 +41,23 @@ fun <T : MediaSourceContainer> prioritize(
     return conferenceEndpoints
 }
 
+fun prioritize2(
+    conferenceSources: MutableList<MediaSourceDesc>,
+    selectedSourceNames: List<String> = emptyList()
+): List<MediaSourceDesc> {
+    // Bump selected sources to the top of the list.
+    selectedSourceNames.asReversed().forEach { selectedSourceName ->
+        // Note the usage of sourceName!! which is expected to be always defined in the multi-stream mode
```
bgrozev, line 50: I wonder if we should fill in missing source names in the bridge? Or make them mandatory in colibri?

bgrozev: I think until we get there you should throw an exception earlier (e.g. at signaling time) if multi stream is enabled and a source doesn't have a name.

bgrozev: With multi stream off, we can enforce:
1. At least one video source per endpoint
2. Set the source's name to the endpoint ID if missing

bgrozev: I wonder if, once multi stream is finished and we're happy with how it works, we can substitude the duplicated code with source-name based code with sourceName=endpointId

jvb/src/test/kotlin/org/jitsi/videobridge/MediaSourceFactoryTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.core.spec.IsolationMode
+import io.kotest.core.spec.Spec
+import io.kotest.core.spec.style.ShouldSpec
+import io.kotest.matchers.shouldBe
+import org.jitsi.config.setNewConfig
+import org.jitsi.videobridge.xmpp.MediaSourceFactory
+import org.jitsi.xmpp.extensions.colibri.SourcePacketExtension
+
+fun createSource(ssrc: Long): SourcePacketExtension {
```
bgrozev, line 28: Just a suggestion: this could be
```
fun createSource(ssrc: Long) = SourcePacketExtension().apply { this.ssrc = ssrc }
```
(unless the name shadowing doesn't work as I expect)

jvb/src/test/kotlin/org/jitsi/videobridge/MediaSourceFactoryTest.kt
```diff
@@ -0,0 +1,72 @@
...
+    val spe = SourcePacketExtension()
+
+    spe.ssrc = ssrc
+
+    return spe
+}
+
+// TODO port MediaSourceFactoryTest to kotlin and unify with this class
```
bgrozev, line 36: I think this has been done?

jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/Prioritize.kt
paweldomas: I will add a check to enforce source name on the Colibri level, but the `!!` should stay here regardless, right?

paweldomas: Enforcing one video source per endpoint may not be a good idea for JVB usage outside of jitsi-meet? Especially not when the multi-stream is off as it would break backwards compatibility?

paweldomas: I like the idea of substituting the duplicated code with source-name based code. Note that there's no backwards compatibility mode implemented in the multi-stream to support older mobile react-native clients yet, but this can be one way of solving it.

jvb/src/test/kotlin/org/jitsi/videobridge/MediaSourceFactoryTest.kt
paweldomas: uhm no, there's still the java version here: https://github.com/jitsi/jitsi-videobridge/blob/9cbc86449e7b0bfe81b23f835fc6eeb8d1db75bc/jvb/src/test/java/org/jitsi/videobridge/xmpp/MediaSourceFactoryTest.java#L28

jvb/src/main/java/org/jitsi/videobridge/shim/ChannelShim.java
```diff
@@ -267,7 +273,19 @@ public void describe(ColibriConferenceIQ.ChannelCommon commonIq)
      * Sets the list of sources signaled for this channel.
      */
     public void setSources(@NotNull List<SourcePacketExtension> sources)
+        throws IqProcessingException
     {
+        if (multiStreamConfig.getEnabled()) {
+            for (SourcePacketExtension s: sources) {
+                // TODO use Kotlin "isNullOrEmpty" once ported to Kotlin
+                if (s.getName() == null || s.getName().trim().isEmpty()) {
```
paweldomas: @bgrozev what's the recommended way of doing "is null or empty check" now? I saw StringUtils were deprecated

jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/Prioritize.kt
bgrozev: I meant enforcing at *most* one video source per endpoint (not at least). Then we can make `MediaSourceDesc.sourceName` non-nullable (and polyfill with the endpoint ID in case of non-multi-stream), so you don't need the `!!`. But I don't want to block because of it.

bgrozev (approved):

PR #1781, opened 2021-12-16
Feat: source name based forwarded sources

paweldomas:
The next step after https://github.com/jitsi/jitsi-videobridge/pull/1780

PR #1782, opened 2021-12-16
Feat: source name based sender constraints

paweldomas:
A follow up on for https://github.com/jitsi/jitsi-videobridge/pull/1781


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -763,6 +763,21 @@ public Relay getRelay(@NotNull String id)
         return relaysById.get(id);
     }
 
+    @Nullable
+    public AbstractEndpoint findSourceOwner(@NotNull String sourceName)
+    {
+        // TODO does it need to also handle OctoEndpoints?
```
bgrozev, line 769: Yes, this needs to work with all endpoints. With Octo/Relayed endpoints we calculate the local max and then request that from the remote bridge.

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
```diff
@@ -78,6 +78,7 @@ public BridgeChannelMessage addReceiver(@NotNull AddReceiverMessage message)
         // only those that reference an endpoint local to this bridge.
         if (endpoint instanceof Endpoint)
         {
+            // TODO Implement source name based receiver constraints for Octo
```
bgrozev, line 81: This is actually not necessary. Remote endpoints can only subscribe to local endpoint. I know this is very confusing, I'll try and sketch a diagram

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
```diff
@@ -94,6 +95,7 @@ public BridgeChannelMessage removeReceiver(@NotNull RemoveReceiverMessage messag
...
```
bgrozev, line 98: Same as above

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -488,6 +505,18 @@ class Endpoint @JvmOverloads constructor(
     fun effectiveVideoConstraintsChanged(
         oldEffectiveConstraints: Map<String, VideoConstraints>,
         newEffectiveConstraints: Map<String, VideoConstraints>
+    ) {
+        if (MultiStreamConfig.config.enabled) {
+            effectiveVideoConstraintsChangedV2(oldEffectiveConstraints, newEffectiveConstraints)
```
bgrozev, line 510: Oh, this is subtle. The java type of the constraints hasn't changed, but the semantics are diffrent (the keys are source names instead of endpoint IDs). Can you add a comment please?

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -440,7 +529,28 @@ public void removeReceiver(String receiverId)
...
+    public void removeReceiverV2(String receiverId, String sourceName)
+    {
+        ReceiverConstraintsMap sourceConstraints = receiverVideoConstraintsMapV2.get(sourceName);
+
+        if (sourceConstraints != null)
+        {
+            if (sourceConstraints.remove(receiverId) != null)
+            {
+                logger.debug(() -> "Removed receiver " + receiverId + " for " + sourceName);
+                receiverVideoConstraintsChanged(sourceConstraints.getMaxHeight());
```
bgrozev, line 548: I think this needs to call `receiverVideoConstraintsChangedV2` with the source name

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -398,6 +402,27 @@ class SenderVideoConstraintsMessage(val videoConstraints: VideoConstraints) : Br
     }
 }
 
+/**
+ * A message sent from the bridge to a client (sender), indicating constraints for the sender's video streams.
```
bgrozev, line 406: This indicates constraint for one specific stream/source, right?

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -398,6 +402,27 @@ class SenderVideoConstraintsMessage(val videoConstraints: VideoConstraints) : Br
...
+    /**
+     * Serialize manually because it's faster than Jackson.
+     *
+     * We use the "idealHeight" format that the jitsi-meet client expects.
+     */
+    override fun createJson(): String =
+        """{"colibriClass":"$TYPE", "sourceName":"$sourceName", "idealHeight":$idealHeight}"""
+
+    companion object {
+        const val TYPE = "SenderVideoConstraintsV2"
```
bgrozev, line 422: I wonder if we can choose a better name. SourceVideoConstraints? SenderSourceConstraints? SourceConstraints?

Since this needs client changes anyway, we should rename `idealHeight` as well. Maybe `maxHeight`?

jvb/src/main/kotlin/org/jitsi/videobridge/octo/OctoEndpoint.kt
```diff
@@ -102,6 +102,10 @@ class OctoEndpoint(
         )
     }
 
+    override fun sendVideoConstraintsV2(sourceName: String, maxVideoConstraints: VideoConstraints) {
+        throw NotImplementedError("sendVideoConstraintsV2 is not implemented for Octo yet")
```
bgrozev, line 106: Let me know if I can help you figure this out

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -81,9 +86,15 @@
...
      * in the conference. The client needs to send _at least_ this to satisfy
      * all receivers.
+     *
+     * @deprecated Use maxReceiverVideoConstraintsMap.
      */
+    @Deprecated
     protected VideoConstraints maxReceiverVideoConstraints = new VideoConstraints(0, 0.0);
 
+    // TODO docs
+    protected Map<String, VideoConstraints> maxReceiverVideoConstraintsMap = new HashMap<>();
```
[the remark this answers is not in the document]
paweldomas: Yes, it does the same as `maxReceiverVideoConstraints`, but on a per source name basis.

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
paweldomas: yes, good catch

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
paweldomas: ok, I'll go with `SenderSourceConstraints` and `maxHeight`

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -474,8 +479,20 @@ class Endpoint @JvmOverloads constructor(
...
+            broadcastSenderConstraints()
+        } else {
+            sendVideoConstraints(maxReceiverVideoConstraints)
+        }
+    }
+
+    private fun broadcastSenderConstraints() {
+        maxReceiverVideoConstraintsMap.forEach {
+            (sourceName, constraints) ->
+            sendVideoConstraintsV2(sourceName, constraints)
```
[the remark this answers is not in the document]
paweldomas: I'm missing something in you suggestion. `sendVideoConstraintsV2` also includes a source name argument, how should the new version work without a source name
Screen Shot 2022-01-20 at 7 57 26 PM
?

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
paweldomas: yeah, I wish there was a way to declare a type for String and call it a SourceName..

paweldomas: actually Kotlin has a typealias feature, I'm curious if that will work: https://kotlinlang.org/docs/type-aliases.html

paweldomas: hm the first obstacle is not being able to use an alias from java files

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
paweldomas: The way I understand this code is that it registers an Octo JVB as a receiver of a local endpoint with specific constraints. I don't understand why that would not be necessary in the source name world - a jvb will register for a given source, no?

paweldomas: well it wouldn't be exactly here, because needs a new message type?

jvb/src/main/java/org/jitsi/videobridge/Conference.java
paweldomas: I've adjusted it to use `endpointsById` which should include Octo endpoints as well

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
bgrozev: I think typealiases are just aliases, they'd be interchangeable. You could extend `String`, but that's probably overkill for something we'll only have temporarily.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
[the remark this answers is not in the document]
bgrozev: Ah, I thought the only use of `sendVideoConstraintsV2` was when sending all sources, but I'm wrong. Can you just rename `broadcast` to something like `sendAllVideoConstraints()`?

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
```diff
@@ -90,13 +104,26 @@ public BridgeChannelMessage addReceiver(@NotNull AddReceiverMessage message)
...
+            // Since we currently broadcast everything in Octo, we may receive messages intended for another bridge.
+            // Handle only those that reference an endpoint local to this bridge.
+            if (endpoint instanceof Endpoint)
+            {
+                endpoint.removeReceiverV2(message.getBridgeId(), sourceName);
+            }
+        }
+        else
+        {
+            AbstractEndpoint endpoint = conference.getEndpoint(message.getEndpointId());
```
paweldomas, line 120: I've been wondering why `conference.getLocalEndpoint` is not used here even though the comment talks about "endpoint local to this bridge"?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -147,7 +147,16 @@ class RelayMessageTransport(
             return null
         }
 
-        ep.addReceiver(relay.id, message.videoConstraints)
+        if (MultiStreamConfig.config.enabled) {
+            val sourceName = message.sourceName
+            if (ep.hasMediaSource(sourceName)) {
```
paweldomas, line 152: Technically, the endpoint ID is not needed with source name in `AddReceiverMessage`, but it makes the lookup a little easier here. I'm not sure whether to keep both endpoint ID or to use only source name in the new mode? Keeping them both may create a false impression of an endpoint being another level of indirection, but source names are conference global (I think).

paweldomas:
The last commit is an attempt of adding source name to receiver Octo messages, but this is not tested yet and there's one part not finished in the `expire()` method. I need to create a setup for testing Octo. What is required to test the relays? Will it be used with just 2 bridges?

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
bgrozev: No good reason as far as I can tell. The code might have been written before `getLocalEndpoint` was available

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
```diff
@@ -73,12 +73,27 @@ public void unhandledMessage(BridgeChannelMessage message)
...
         Conference conference = octoEndpoints.getConference();
-        AbstractEndpoint endpoint = conference.getEndpoint(message.getEndpointId());
-        // Since we currently broadcast everything in Octo, we may receive messages intended for another bridge. Handle
-        // only those that reference an endpoint local to this bridge.
-        if (endpoint instanceof Endpoint)
+
+        if (MultiStreamConfig.config.getEnabled())
+        {
+            String sourceName = message.getSourceName();
+            AbstractEndpoint endpoint = conference.findSourceOwner(sourceName);
```
bgrozev, line 80: You could lookup the endpoint via `message.getEndpointId()` instead of source name (if you decide to keep using it)

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
bgrozev: I think it's cleaner to use only `sourceName`, but since you're reusing the message and have the endpoint ID you might as well use it. We can clean that up (remove endpointId) when we phase out the old code.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -763,6 +763,20 @@ public Relay getRelay(@NotNull String id)
         return relaysById.get(id);
     }
 
+    @Nullable
+    public AbstractEndpoint findSourceOwner(@NotNull String sourceName)
```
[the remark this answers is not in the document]
paweldomas: yes, I'm worried about that too

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
paweldomas: yeah I think it's cleaner to use only `sourceName`

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
paweldomas: I've skipped OctoV1 for multi-stream, because this gives one more combination to test and maintain. Let me know if that's ok.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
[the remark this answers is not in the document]
paweldomas: done

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -66,6 +66,11 @@
      */
     private final ReceiverConstraintsMap receiverVideoConstraintsMap = new ReceiverConstraintsMap();
 
+    /**
+     * The map of source name -> receiver endpoint id -> video constraints.
```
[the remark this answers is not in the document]
paweldomas: I have expanded the `ReceiverConstraintsMap` into the `receiver endpoint id -> video constraints` part, but removed it now since it appears to be confusing.

jvb/src/main/java/org/jitsi/videobridge/octo/OctoEndpointMessageTransport.java
paweldomas: I removed the OctoV1 part and will try to only use the source name. Eventually the lookup by source name will have to be optimized anyway.

paweldomas:
@bgrozev please take a look. I will do another round of testing with the last few commits

paweldomas:
Ok it seem to be working. Octo is working too, but only V2 with multi-stream (see comment above).

jvb/src/main/java/org/jitsi/videobridge/xmpp/MediaSourceFactory.java
```diff
@@ -774,13 +801,17 @@ private static MediaSourceDesc createSource(
         // not optional (there's no good reason for that). Then the error will be thrown automatically by Kotlin.
         if (MultiStreamConfig.config.getEnabled())
         {
-            if (primarySsrcs.name == null)
+            if (owner == null)
             {
                 throw new IllegalArgumentException("The 'name' is missing in the source description");
```
bgrozev, line 806: "owner" and "name" are the wrong way around in the log messages

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -495,6 +537,18 @@ class Endpoint @JvmOverloads constructor(
         }
     }
 
+    override fun sendVideoConstraintsV2(sourceName: String, maxVideoConstraints: VideoConstraints) {
+        // Note that it's up to the client to respect these constraints.
+        if (findMediaSourceDesc(sourceName) == null) {
+            logger.cdebug { "Suppressing sending a SenderVideoConstraints message, endpoint has no such source." }
```
bgrozev, line 543: This should probably be a warning/error

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -398,20 +402,40 @@ class SenderVideoConstraintsMessage(val videoConstraints: VideoConstraints) : Br
...
+    val endpointId: String?, // Used in single stream per endpoint mode and wil be removed
+    val sourceName: String?, // Used in the multi-stream mode
     val videoConstraints: VideoConstraints
 ) : BridgeChannelMessage(TYPE) {
     /**
      * Serialize manually because it's faster than Jackson.
      */
     override fun createJson(): String =
-        """{"colibriClass":"$TYPE","bridgeId":"$bridgeId","endpointId":"$endpointId",""" +
+        """{"colibriClass":"$TYPE","bridgeId":"$bridgeId","endpointId":"$endpointId","sourceName":"$sourceName",""" +
```
bgrozev, line 438: The serialization is broken now that some of the values are nullable

paweldomas: broken in what way? that it puts `null` there? I can make it skip that. I'm asking because I was testing Octo with 3 JVBs and saw no exception about serialization

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
paweldomas: 🤦‍♂️

jvb/src/main/java/org/jitsi/videobridge/xmpp/MediaSourceFactory.java
paweldomas: 🤦‍♂️

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
bgrozev: You can play with the perf test in `BridgeChannelMessageTest`, but I'm not too worried about the performance -- it's not one of the messages that are sent often

jvb/src/main/java/org/jitsi/videobridge/AbstractEndpoint.java
```diff
@@ -424,6 +487,37 @@ public void addReceiver(String receiverId, VideoConstraints newVideoConstraints)
...
+     * Notifies this instance that a specified received wants to receive the specified video constraints from the media
+     * source with the given source name.
+     *
+     * The receiver can be either another endpoint, or a remote bridge.
+     *
+     * @param receiverId the id that specifies the receiver endpoint.
+     * @param sourceName the name of the media source for which the constraints are to be applied.
+     * @param newVideoConstraints the video constraints that the receiver wishes to receive.
+     */
+    public void addReceiverV2(String receiverId, String sourceName, VideoConstraints newVideoConstraints)
```
bgrozev, line 500: Can you please annotate `sourceName` with `@NotNull` or `@Nullable` as appropriate? I think it's expected to be `@NotNull` but you use it as nullable in `RelayMessageTransport`

paweldomas: I've added the annotation, but didn't get any error about using nullable. Is this an IDE warning or something?

paweldomas: oh I was using `!!` which would throw

bgrozev (approved):

PR #1785, opened 2021-12-17
chore: Remove the sentry and agafua-syslog runtime dependencies.

bgrozev:
This is to stop shipping non-essential jars and reduce potential exposure to vulnerabilities. Setting up sentry requires setting an environment variable (and potentially more changes?), so it should be easy to adapt to also include the sentry jar in the classpath.

cc @saghul @sapkra


bgrozev:
@sapkra you're the only one that we're aware of actively using this. Let us know what you think, we can delay merging this if you need time to update your system.

PR #1788, opened 2021-12-21
Refactor colibri shims

bgrozev:
Move more of the quirky colibri1 logic to the shim layer.

- ref: Move Endpoint.channelShims to EndpointShim.
- ref: Move local SSRC generation from the shim to Conference.
- ref: Inline recreateMediaSources (move logic to the shim).
- ref: Move the colibri queue from ConferenceShim to Conference.
- ref: Do not use the shim to count video channels.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/shim/ColibriUtil.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/shim/ColibriUtil.java
@@ -19,10 +19,13 @@ import org.jitsi.utils.*;
 import org.jitsi.utils.logging2.*;
 import org.jitsi.videobridge.*;
 import org.jitsi.videobridge.sctp.*;
+import org.jitsi.videobridge.xmpp.*;
 import org.jitsi.xmpp.extensions.colibri.*;
+import org.jitsi.xmpp.extensions.jingle.*;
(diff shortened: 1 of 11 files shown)


jvb/src/main/java/org/jitsi/videobridge/shim/ColibriUtil.java
```diff
@@ -146,7 +149,33 @@ static String getOctoChannelId(MediaType mediaType)
             // video channel is signaled.
             if (MediaType.VIDEO.equals(contentShim.getMediaType()) && !channelSources.isEmpty())
             {
-                channelShim.getEndpoint().recreateMediaSources();
+                List<ChannelShim> videoChannels = contentShim.getChannelShims().stream()
+                        .filter(c -> c.getMediaType() == MediaType.VIDEO)
+                        .filter(c -> c.getEndpoint() == channelShim.getEndpoint())
```
JonathanLennox, line 154: I'm a little concerned (for performance reasons) that this is iterating over all the channels in the conference.  Since you added `EndpointShim`, would it make sense to have `recreateMediaSources` reference the `ChannelShim`s from there?

JonathanLennox (approved):

PR #1794, opened 2022-01-12
Add http API for drain state.

jqdrqgnq:
Add drain state to bridge statistics.


bgrozev:
LGTM! Can you please also add a config property for the initial value of `drain`? We want to be able to spin up a machine in drain mode for testing (if we try to do it via the REST API right after it spins up we risk some conferences being scheduled on it before we change the state).

bgrozev (approved):

PR #1800, opened 2022-01-19
Disable config caching for all tests, use static fields for config objects.

bgrozev:
- chore: Remove stale logging config.
- ref: Use withNewConfig in JvbLoadManagerTest.
- ref: Disable Metaconfig caching for all tests.
- ref: Use a single static instance of WebsocketServiceConfig.
- ref: Use only the static octo config field.
- ref: Use only the static health config field.
- ref: Use only the static field for VideobridgeExpireThreadConfig.
- ref: Use only the static field for XmppClientConnectionConfig.
- ref: Move the BandwidthProbingConfig instance.
- ref: Use only the static config instance for BitrateControllerConfig.
- ref: Use only the static field for EndpointConnectionStatusConfig.
- ref: Move the StatsManagerConfig instance.
- ref: Use only the static instance of CallstatsConfig.
- fix: Fix CallstatsConfig getters.
- ref: Make constructors of config classes private.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
@@ -26,6 +26,7 @@ import org.jitsi.utils.version.Version
 import org.jitsi.videobridge.Videobridge
 import org.jitsi.videobridge.stats.StatsCollector
 import org.jitsi.videobridge.stats.StatsTransport
+import org.jitsi.videobridge.stats.config.StatsManagerConfig
 import org.jitsi.videobridge.stats.config.StatsTransportConfig
 import java.time.Duration
 
@@ -181,7 +182,7 @@ class CallstatsConfig {
      *
      * For backwards compatibility, we read it from the stats manager "callstatsio" transport, if present.
      */
-    val interval: Duration = StatsCollector.config.transportConfigs.stream()
+    val interval: Duration = StatsManagerConfig.config.transportConfigs.stream()
         .filter { tc -> tc is StatsTransportConfig.CallStatsIoStatsTransportConfig }
         .map(StatsTransportConfig::interval)
         .findFirst()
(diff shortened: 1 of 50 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
```diff
@@ -58,6 +57,8 @@ class CallstatsService(
     val statsTransport: StatsTransport?
         get() = callstatsTransport
 
+    val config = CallstatsConfig.config
```
JonathanLennox, line 60: Is this needed?

jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/XmppConnection.kt
```diff
@@ -54,6 +54,8 @@ class XmppConnection : IQListener {
 
     var eventHandler: EventHandler? = null
 
+    val config = XmppClientConnectionConfig.config
```
JonathanLennox, line 57: Or this?

jvb/src/main/kotlin/org/jitsi/videobridge/stats/callstats/CallstatsService.kt
bgrozev: I didn't think an import of something defined within the same file would work. But it seems to work, so I replaced it.

jvb/src/main/kotlin/org/jitsi/videobridge/xmpp/XmppConnection.kt
bgrozev: I thought it made sense conceptually to expose the config *of this XmppConnection*, but looking again it's not necessary. I removed it.

damencho:
Jenkins, test this please.

JonathanLennox (approved):

bbaldino:
Sorry for not commenting on this sooner, but kotest does have a hook to run some code before the project which would also work for disabling the cache (see `BeforeProjectListener` [here](https://kotest.io/docs/framework/extensions/extensions-introduction.html).  In case you want to give it a try.

bgrozev:
Thanks, I didn't know about that. I think in our case it's equivalent to `beforeAll` in the project config.

PR #1814, opened 2022-02-15
Send release ID in bridge statistics.

jqdrqgnq:
Sending dummy value for release ID for now.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -144,6 +144,7 @@ public class Videobridge
     private final ScheduledFuture<?> loadSamplerTask;
 
     private final Version version;
+    private final String releaseId;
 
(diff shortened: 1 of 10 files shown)


jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -200,6 +201,7 @@ public Videobridge(
             xmppConnection.setEventHandler(new XmppConnectionEventHandler());
         }
         this.version = version;
+        this.releaseId = "346"; /* $ Get real value from somewhere. */
```
bgrozev, line 204: This should read the value from the config file (see here for an example of reading a nullable value https://github.com/jitsi/jitsi-videobridge/blob/master/jvb/src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt#L57)

bgrozev (approved):

PR #1821, opened 2022-02-25
Fix issue where Relays' SrtpTransformers were closed while still in use.  Bump JMT.

JonathanLennox:


JonathanLennox:
Depends on https://github.com/jitsi/jitsi-media-transform/pull/403

bgrozev (approved):

PR #1822, opened 2022-03-01
Properly include relay endpoints and senders in relay bitrate/packet rate stats.

JonathanLennox:


jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -763,6 +764,26 @@ class Relay @JvmOverloads constructor(
...
+            }
+
+    val incomingPacketRate: Long
+        get() = transceiver.getTransceiverStats().rtpReceiverStats.packetStreamStats.packetRate +
+            synchronized(endpointsLock) {
+                relayedEndpoints.values.sumOf { it.getIncomingStats().packetRate }
+            }
+
+    val outgoingBitrateBps: Double
+        get() = transceiver.getTransceiverStats().rtpReceiverStats.packetStreamStats.getBitrateBps() +
```
bgrozev: I think this should be `getTransceiverStats().outgoingPacketStreamStats`, same below for packet rate

bgrozev (approved):

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
JonathanLennox: Yep, I noticed that just after the initial PR

PR #1835, opened 2022-03-17
feat(multi-stream): endpoint ID compatibility

paweldomas:
Jicofo will forward source name capability for
each Colibri2 Endpoint. In Colibri V1 "use source
names" is always false for now.

Based on the info on whether the client uses source
names or not, will use different signaling messages
for talking to this client. All operations under
the hood are happening on source names.


paweldomas:
Opening this as a draft, because it's not well tested yet, wanted to get early feedback.

paweldomas:
Requires https://github.com/jitsi/jicofo/pull/894 and https://github.com/jitsi/jitsi-xmpp-extensions/pull/68

paweldomas:
https://github.com/jitsi/jitsi-videobridge/pull/1842

PR #1845, opened 2022-03-28
Ssrc rewriting

jqdrqgnq:


jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -815,6 +824,46 @@ class Endpoint @JvmOverloads constructor(
     fun setBandwidthAllocationSettings(message: ReceiverVideoConstraintsMessage) =
         bitrateController.setBandwidthAllocationSettings(message)
 
+    private fun logCurrentSsrcs() {
+        logger.debug {
+            var s = String()
```
JonathanLennox, line 841: Should use `buildString`

jqdrqgnq: i tried improving this with joinToString.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1239,4 +1359,379 @@ class Endpoint @JvmOverloads constructor(
...
+        val ssrc = ssrc_ ?: getNextSsrc()
+        private val state = RtpState()
+        private var sequenceNumberDelta = 0
+        private var timestampDelta = 0L
+        private var tl0IndexDelta = 0
+
+        companion object {
+            private val useRandom = true; /* switch off to ease debugging */
+            private var nextSsrc = 777000001L /* use serial number for debugging */
+            private val random = Random(System.currentTimeMillis())
```
JonathanLennox, line 1446: It might be better if this were cryptographically random - if not, you may want to use ThreadLocalRandom.

You may also want to check to avoid colliding with existing SSRCs?  I'm not sure whether you need to worry about that.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -837,17 +887,71 @@ class Endpoint @JvmOverloads constructor(
...
+                        }
+                    }
                     transceiver.sendPacket(packetInfo)
                 } else {
                     logger.warn("Dropping a packet which was supposed to be accepted:$packet")
                 }
                 return
             }
+            is AudioRtpPacket -> if (doSsrcRewriting) audioSsrcs.rewriteRtp(packet)
             is RtcpSrPacket -> {
```
JonathanLennox, line 955: You're not doing any rewriting of RTCP packets - would this be needed? It seems like the receivers will get RTCP for all the original sources, and none for the re-written ones.  I don't know if this matters in practice for the browsers, but it somewhat concerns me in theory?

JonathanLennox (commented):
I can't say I've fully gotten my head around this PR yet, but here are some initial comments...

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -837,17 +887,71 @@ class Endpoint @JvmOverloads constructor(
...
+     */
+    private fun findVideoSourceProps(ssrc: Long): MediaSourceDesc? {
+        conference.getEndpointBySsrc(ssrc)?.let { ep ->
+            ep.mediaSources.forEach { s ->
+                if (s.findRtpEncodingDesc(ssrc) != null) {
+                    return s
+                }
+            }
+        }
+        logger.error { "\tNo properties found for SSRC $ssrc." }
```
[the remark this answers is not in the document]
jqdrqgnq: oops, these are leftover formatting from when i was logging a lot more debug info.
fixed.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1239,4 +1359,379 @@ class Endpoint @JvmOverloads constructor(
...
+                    logger.debug { "added $mediaType receive SSRC: ${packet.ssrc}" }
+                }
+
+                val ss = getSendSource(rs.props.ssrc1, rs.props, mediaType == MediaType.AUDIO, null)
+                if (ss != null) {
+                    send = ss.rewriteRtp(packet, start, rs)
+                    logger.debug { this.toString() }
+                    logger.debug {
+                        if (send) {
+                            "output packet: ${debugInfo(packet)} source=${rs.props.name} start=$start send=$send"
```
JonathanLennox, line 1710: Log messages should generally be capitalized

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1239,4 +1359,379 @@ class Endpoint @JvmOverloads constructor(
...
+        override fun toString(): String = "$send1, $send2" + if (started) "" else " (not started)"
+    }
+
+    /**
+     * Limit the number of local SSRCs used for the given media type to the number specified.
+     * If there are more sources in the conference than the limit, then the least recently used
+     * SSRCs are remapped. RTP packets have their header fields rewritten so the stream appears
+     * to be a continuation of an already advertised SSRC.
+     */
+    private inner class SsrcCache(val mediaType: MediaType, val size: Int) {
```
JonathanLennox, line 1560: Rather than having SsrcCache take a mediaType argument and then have if(mediaType) switches in its code, it seems like it would be cleaner to have an abstract SsrcCache with AudioSsrcCache and VideoSsrcCache subclasses.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1239,4 +1359,379 @@ class Endpoint @JvmOverloads constructor(
 
         override fun trace(f: () -> Unit) = f.invoke()
     }
+
+    /**
+     * Align common fields from different source types.
+     * Perhaps this could become a base class of those types.
+     */
+    private class SourceDesc private constructor(
```
JonathanLennox, line 1367: Is it possible to move the source remapping code out of Endpoint.kt?  It seems like it would make it easier to comprehend, and also (if you abstract its connection to Endpoint, or mock something) it could make it easier to write unit tests for it.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
jqdrqgnq: Added checks for collisions. 
Switched to ThreadLocalRandom.
Can discuss further whether to use a better random.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
jqdrqgnq: fixed.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
jqdrqgnq: Yes, seems much better now. thanks for the suggestion.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
jqdrqgnq: done. yes, this should let me overcome the issues with my unit test. i will update the test and submit it.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1239,4 +1359,379 @@ class Endpoint @JvmOverloads constructor(
...
+                    recv.hasDeltas = true
+                }
+
+                recv.state.update(packet)
+
+                packet.ssrc = ssrc
+                packet.sequenceNumber = RtpUtils.applySequenceNumberDelta(packet.sequenceNumber, sequenceNumberDelta)
+                packet.timestamp = RtpUtils.applyTimestampDelta(packet.timestamp, timestampDelta)
+                if (packet is Vp9Packet) {
+                    packet.TL0PICIDX = (packet.TL0PICIDX + tl0IndexDelta) % 256
```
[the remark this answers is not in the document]
jqdrqgnq: Didn't want to touch VP8 since it appears to be working. If you think it's more consistent and robust to do it anyway, I can make the change.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1082,6 +1083,8 @@ public void endpointMessageTransportConnected(@NotNull AbstractEndpoint abstract
             {
                 endpoint.sendMessage(new DominantSpeakerMessage(recentSpeakers, speechActivity.isInSilence()));
             }
+
+            abstractEndpoint.onMessageTransportConnect();
```
bgrozev, line 1087: We already have `Endpoint.endpointMessageTransportConnected`. I don't see you implement this in `Relay`, so I think you can just merge your code into it.

jqdrqgnq: thanks, fixed.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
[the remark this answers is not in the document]
jqdrqgnq: ok, did it. haven't seen issues with the video thus far.

jvb/src/main/kotlin/org/jitsi/videobridge/SsrcCache.kt
```diff
@@ -0,0 +1,530 @@
...
+
+    /**
+     * Assign send SSRCs to the given sources. Any remapped SSRCs will be notified to the client.
+     */
+    fun activate(sources: List<MediaSourceDesc>) {
+
+        val remappings = mutableListOf<SendSource>()
+
+        synchronized(sendSources) {
+            sources.forEach { source ->
```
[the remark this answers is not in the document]
jqdrqgnq: Good catch. The loop you suggest could still bump a source that should remain active out of the cache. Perhaps I can partition the input source list into active and new, and then process the active first.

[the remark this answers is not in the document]
jqdrqgnq: thanks, good catch. fixed it as you suggested, except also suppressing any sources that stay active from the second pass.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
jqdrqgnq: completed.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
jqdrqgnq: Switched to SecureRandom.

jvb/src/main/kotlin/org/jitsi/videobridge/SsrcCache.kt
```diff
@@ -0,0 +1,616 @@
...
+}
+
+/**
+ * Truncate to 8 unsigned bits.
+ */
+private fun Int.fixByte(): Int {
+    val s = this.rem(256)
+    /* Adjust if dividend was negative. */
+    return if (s < 0) (s + 256) else s
+}
```
JonathanLennox, line 70: Shouldn't `this and 0xff` do this?  Bitwise operations are defined on signed integers in Java/Kotlin.

jqdrqgnq: done, thanks.

JonathanLennox (approved):

bgrozev (approved):

PR #1853, opened 2022-04-04
fix: Log and return an error if a Relay is used without websockets co…

bgrozev:
…nfigured.


damencho (approved):

bgrozev:
Tests failed because jicofo's logs had too many lines. The threshold has been updated (to account for the change to colibri2).

PR #1854, opened 2022-04-04
Add REST API for Colibri2

JonathanLennox:


JonathanLennox:
So far this just has the serializers/deserializers.

JonathanLennox:
I haven't added anything for the GET to the REST URLs, but I don't currently think those are useful; at the very least, I don't know what the use case is for them.

JonathanLennox:
Jenkins, test this please.

jvb/src/main/java/org/jitsi/videobridge/rest/root/colibri/v2/conferences/Conferences.java
```diff
@@ -0,0 +1,162 @@
...
+                throw new BadRequestException();
+            }
+        }
+        catch (ParseException pe)
+        {
+            throw new BadRequestExceptionWithMessage(
+                    "Failed to create conference, could not parse JSON: " + pe.getMessage());
+        }
+
+        ConferenceModifyIQ conferenceModifyIQ
```
bgrozev, line 102: Maybe try/catch any deserialization exceptions? They should be equivalent to bad JSON

bgrozev (approved):

PR #1856, opened 2022-04-05
1. Use String for id in RtpHeaderExtension.

kaiduanx:
2. Use IceCandidatePacketExtension instead of CandidatePacketExtension.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
@@ -118,7 +118,7 @@ public final class JSONDeserializer
             {
                 deserializeCandidate(
                         (JSONObject) candidate,
-                        CandidatePacketExtension.class,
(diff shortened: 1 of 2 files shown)


jvb/src/main/java/org/jitsi/videobridge/rest/JSONDeserializer.java
```diff
@@ -657,7 +657,7 @@ public static RTPHdrExtPacketExtension deserializeHeaderExtension(
         }
         else
         {
-            Long id = (Long)headerExtension.get(RTPHdrExtPacketExtension.ID_ATTR_NAME);
+            String id = (String)headerExtension.get(RTPHdrExtPacketExtension.ID_ATTR_NAME);
```
bgrozev, line 660: This might break existing software that encodes it as a number. Can we change JSONSerializer instead, since the code was introduced just recently?

bgrozev (approved):

JonathanLennox:
Jenkins, test this please

PR #1857, opened 2022-04-06
Shutting down

bgrozev:
Add a time limit to graceful-shutdown. An hour (by default) after entering graceful shutdown, the bridge will shut down. If during this time the number of endpoints drops below 10 (by default) the bridge will shutdown.

Before shutting down the bridge will enter the SHUTTING_DOWN state for 1 minute, which is advertised in presence and should serve as a trigger for jicofo to actively move participants away.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -683,6 +683,11 @@ public class Videobridge
         return shutdownManager.getState() == ShutdownState.GRACEFUL_SHUTDOWN;
     }
 
+    public ShutdownState getShutdownState()
+    {
+        return shutdownManager.getState();
+    }
+
     /**
      * Initiate a shutdown if there are no conferences left.
      */
@@ -707,7 +712,7 @@ public class Videobridge
                 // Make sure that enough time passes for the "graceful shutdown" mode to be announced in presence.
                 // Otherwise, other components may detect this as a bridge going down non-gracefully.
                 Duration delay
-                        = VideobridgeConfig.Companion.getGracefulShutdownDelay().minus(timeSinceShutdownRequested);
+                        = ShutdownConfig.config.getGracefulShutdownDelay().minus(timeSinceShutdownRequested);
                 if (delay.isNegative() || delay.isZero())
                 {
                     shutdownManager.doShutdown();
--- a/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -440,47 +440,6 @@ public class Videobridge
     }
 
     /**
-     * Handles a COLIBRI2 request synchronously. Exposed only for testing (jicofo).
-     * @param conferenceModifyIQ The COLIBRI request.
-     * @return The response in the form of an {@link IQ}. It is either an error or a {@link ColibriConferenceIQ}.
-     */
-    public IQ handleConferenceModifyIq(ConferenceModifyIQ conferenceModifyIQ)
-    {
-        Conference conference;
-        try
-        {
-            conference = getOrCreateConference(conferenceModifyIQ);
-        }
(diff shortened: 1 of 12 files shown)


jvb/src/main/resources/reference.conf
```diff
@@ -355,8 +355,18 @@ videobridge {
     #release = "123"
   }
 
-  # The minimum amount of time to stay in graceful-shutdown mode before terminating.
-  graceful-shutdown-delay = 1 minute
+  shutdown {
+    # The maximum amount of time to stay in GRACEFUL_SHUTDOWN before going into SHUTTING_DOWN. See ShutdownManager.
+    graceful-shutdown-max-duration = 24 hours
+
+    # The minimum numnber of participants required to stay in GRACEFUL_SHUTDOWN. If the number of participants falls
```
JonathanLennox, line 362: Typo, "number"

jvb/src/test/kotlin/org/jitsi/videobridge/VideobridgeTest.kt
```diff
@@ -57,12 +59,17 @@ class VideobridgeTest : ShouldSpec() {
...
+                        repeat(10) { videobridge.endpointExpired() }
+                        videobridge.shutdownState shouldBe ShutdownState.SHUTTING_DOWN
+                        // TODO: how can we verify that the shutdownService has been called when the task is scheduled
+                        // on TaskPools.SCHEDULED_POOL
+                    }
+                    context("When the graceful shutdown period expires") {
+                        clock.elapse(60.mins)
+                        // TODO: how can we verify that the state has transitioned to SHUTTING_DOWN when the task was
+                        // scheduled on TaskPools.SCHEDULED_POOL?
+                        // videobridge.shutdownState shouldBe ShutdownState.SHUTTING_DOWN
```
JonathanLennox, line 72: Do you have a plan to resolve this TODO?  It seems like this isn't currently verifying that the shutdown actually runs.

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -439,47 +439,6 @@ public IQ handleColibriConferenceIQ(ColibriConferenceIQ conferenceIq)
         return conference.getShim().handleColibriConferenceIQ(conferenceIq);
     }
 
-    /**
```
JonathanLennox, line 461: I think it'd be better to cherry-pick this removal to a separate PR, it makes the overall PR diff a bit confusing otherwise.

jvb/src/main/kotlin/org/jitsi/videobridge/shutdown/ShutdownManager.kt
```diff
@@ -0,0 +1,106 @@
...
+                        }
+                    },
+                    config.gracefulShutdownMaxDuration.toMillis(),
+                    TimeUnit.MILLISECONDS
+                )
+            } else {
+                logger.info("Graceful shutdown already initiated mode.")
+            }
+        } else {
+            changeToShuttingDown()
```
JonathanLennox, line 73: If I'm reading this correctly, the non-graceful shutdown time has now gone from 1 second to 1 minute?  That's a pretty big operational difference.

bgrozev: That's correct.

Graceful shutdown was already delayed by 1 minute (even when empty) to assure presence makes it to jicofo. I thought the same shoud apply to non-graceful shutdown and the code was simpler that way. I don't actually know why non-graceful shutfdown exists, I haven't seet it used anywhere.

jvb/src/test/kotlin/org/jitsi/videobridge/VideobridgeTest.kt
bgrozev: Not really. I don't know if there's any good way to mock scheduled tasks.

I tested manually, but of course that's no substitute.

JonathanLennox (approved):

PR #1858, opened 2022-04-06
Update jicofo, smack to 4.4.5.

bgrozev:
Fixes the deadlock previously observed in jibri, now also observed with jvb connected to multiple xmpp servers.


JonathanLennox (approved):
Commit description should say "jicoco" not "jicofo".  I've edited the PR in github but I'm not sure that carries over when doing a squash and merge.

PR #1862, opened 2022-04-14
fix: Catch exceptions parsing RED, log details (JMT)

bgrozev:


bgrozev:
The test failure is unrelated

PR #1864, opened 2022-04-15
Support complex topologies of bridge relays

JonathanLennox:
This allows bridges to be arranged in a tree of full meshes, rather than in a single full-mesh, as currently.  Media packets received from one relay mesh are forwarded to other ones.

Each connection to another relay is identified by a `mesh-id` value, indicating the mesh of relays of which that connection is a part.  It is responsibility of the entity setting up connections between relays (e.g., jicofo) to ensure the topologies are set up correctly.  (This code is not yet written.)

Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/70.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -419,18 +419,22 @@ public class Conference
 
     /**
      * Sends a message that originated from a relay, forwarding it to local endpoints
-     * and to those relays with different mesh-id values than the source relay
+     * (if {@param sendToEndpoints} is true) and to those relays with different mesh-id values than the source relay
      *
      * @param msg the message to be sent
      * @param meshId the ID of the mesh from which the message was received.
      */
     public void sendMessageFromRelay(
         BridgeChannelMessage msg,
+        boolean sendToEndpoints,
         @Nullable String meshId)
     {
-        for (Endpoint endpoint : getLocalEndpoints())
+        if (sendToEndpoints)
         {
-            endpoint.sendMessage(msg);
+            for (Endpoint endpoint : getLocalEndpoints())
+            {
+                endpoint.sendMessage(msg);
+            }
         }
 
         for (Relay relay: relaysById.values())
@@ -441,10 +445,7 @@ public class Conference
             }
         }
 
-        if (tentacle != null && meshId != null)
-        {
-            tentacle.sendMessage(msg);
-        }
+        /* Never need to send to tentacle, meshId != null cannot be true for Colibri1. */
     }
 
     /**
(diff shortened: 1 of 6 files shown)


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -417,6 +417,36 @@ public void sendMessage(
...
+
+        for (Relay relay: relaysById.values())
+        {
+            if (!Objects.equals(meshId, relay.getMeshId()))
+            {
+                relay.sendMessage(msg);
+            }
+        }
+
+        if (tentacle != null && meshId != null)
```
bgrozev, line 444: This condition is `true` only if we use old octo with mesh IDs?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -108,6 +108,7 @@ class Relay @JvmOverloads constructor(
     /**
      * True if the ICE agent for this [Relay] will be initialized to serve as a controlling ICE agent, false otherwise.
      */
+    val meshId: String?,
```
bgrozev, line 111: The `meshId` param is between `iceControlling` and its docs. 

Maybe add a note on the semantics of mesh ID somewhere (that it's associated with a bridge-to-bridge link and not a particular bridge)?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -195,6 +195,8 @@ class RelayMessageTransport(
 
         ep.setVideoType(message.videoType)
 
+        relay.conference.sendMessageFromRelay(message, relay.meshId)
```
bgrozev, line 198: This broadcasts the received video type to all local endpoints. Is this the intention?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -218,6 +220,8 @@ class RelayMessageTransport(
 
         ep.setVideoType(message.sourceName, message.videoType)
 
+        relay.conference.sendMessageFromRelay(message, relay.meshId)
```
bgrozev, line 223: Same as above

jvb/src/main/java/org/jitsi/videobridge/Conference.java
JonathanLennox: Yeah, which is impossible, but I was trying to be complete...do you think I should just have a comment saying it cannot happen for old octo?

jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
JonathanLennox: No, will fix.

bgrozev (approved):

PR #1874, opened 2022-04-26
feat: Updates dependencies(jitsi-srtp) and support libssl3, drops java8.

damencho:


bgrozev:
Can you remove the java 8 handling from `jvb.sh` too? And the GH action that now fails :)

damencho:
I was just doing that :) Let's see now.

bgrozev (approved):

PR #1876, opened 2022-04-26
log: Log when establishing a colibri websocket fails.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/relay/RelayMessageTransport.kt
```diff
@@ -323,6 +323,9 @@ class RelayMessageTransport(
         }
     }
 
+    override fun webSocketError(cause: Throwable) =
+        logger.error("Failed to establish colibri websocket: ${cause.message}")
```
JonathanLennox, line 327: I don't think this happens only on failure to establish - it looks like other things can trigger it (like invalid websocket framing).  Which probably means we want to log for errors on endpoint websockets, too - and maybe do something to tear down already-established ones?

jvb/src/main/java/org/jitsi/videobridge/websocket/ColibriWebSocket.java
```diff
@@ -103,5 +109,6 @@ public interface EventHandler {
          * @param ws the {@link ColibriWebSocket} from which a message was received.
          */
         void webSocketTextReceived(ColibriWebSocket ws, String message);
+        default void webSocketError(Throwable cause) {}
```
JonathanLennox, line 112: Should have javadoc.  I'd also say it should take the ws as an argument, to match the other methods.

JonathanLennox (approved):

PR #1888, opened 2022-05-13
Bump JMT and RTP.

JonathanLennox:
Fix: Don't generate 65535 NACKs when re-receiving the highest sequence number.

Fix: Handle sequence number wrapping in NACK packets.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,12 +102,12 @@
             <dependency>
                 <groupId>${project.groupId}</groupId>
                 <artifactId>rtp</artifactId>
-                <version>1.0-68-g0eb94bf</version>
+                <version>1.0-69-gd8300a6</version>
             </dependency>
             <dependency>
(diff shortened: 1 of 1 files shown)


bgrozev (approved):

bbaldino:
@JonathanLennox inline classes are stable now, wonder if it's worth starting to use [RtpSequenceNumber](https://github.com/jitsi/jitsi-rtp/blob/master/src/main/kotlin/org/jitsi/rtp/rtp/RtpSequenceNumber.kt) in more places.

bbaldino:
Ah, I do see @bgrozev had found an issue there (noted in a comment in that class).  Wonder if that's still a problem or if there's a proper way it should be done.

bbaldino:
I tried the scenario in that comment and couldn't reproduce. [link](https://pl.kotl.in/CRHdvTKhJ")

PR #1895, opened 2022-05-25
Fix reading relay

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
```diff
@@ -62,15 +62,15 @@ class OctoConfig private constructor() {
     /**
      * This is deprecated and only used in octo v1.
      */
-    val bindAddress: String by config {
+    val bindAddress: String? by optionalconfig {
         "bind address from legacy config" { legacyBindAddress!! }
```
JonathanLennox, line 66: Should this still have the `!!` here?

jvb/src/main/kotlin/org/jitsi/videobridge/octo/OctoRelayService.kt
```diff
@@ -48,8 +48,8 @@ class OctoRelayService {
             throw IllegalStateException("Octo relay service is not enabled")
         }
 
-        val address = config.bindAddress
-        val port = config.bindPort
+        val address = config.bindAddress ?: throw IllegalStateException("bind-address not configured.")
```
JonathanLennox, line 51: In the case where `bindAddress` isn't set but `enabled` is true (i.e. you only want to use OctoV2) this will throw `IllegalStateException`, which will cause the `Error initializing OctoRelayService` to be logged, which might be confusing?

jvb/src/main/kotlin/org/jitsi/videobridge/octo/config/OctoConfig.kt
bgrozev: Yes, the function requires a `String`. It should catch the NPE then rethrow a config exception, causing it skip to the next one. But I'll run some tests to verify.

JonathanLennox (approved):

PR #1915, opened 2022-07-07
Add metric wrappers and MetricsContainer for Prometheus endpoint

alexcsf:
This is a PR for the GSoC project [JVB Prometheus Stats](https://summerofcode.withgoogle.com/programs/2022/projects/CtgpJGaV). This is a project idea [originally proposed by Jitsi](https://github.com/jitsi/gsoc-ideas/blob/master/2022/prometheus-stats.md).

Large commit, so here's the commit message:

* Add MetricsContainer which defines and collects metrics.

MetricsContainer provides a central interface to define and collect metrics.
Classes can define and register metrics with MetricsContainer.
Metrics can be exported in JSON and in both formats supported by Prometheus: OpenMetrics and text/plain 0.0.4.
Currently, only a small sample of existing statistics are defined and kept as metrics.

* Add metric wrappers for Prometheus metric types.

Wrappers such as GaugeMetric implement the interface Metric, simplifying the exporting of metrics in MetricsContainer.
These wrappers also provide atomic operations, such as decrementAndGet().
Wrappers such as BooleanMetric and LongGaugeMetric enable stronger type distinction
since Prometheus exposes all samples as 64-bit floats, unlike JSON.

* Update pom.xml with Prometheus JVM lib dependencies.

* Replace some Atomic instances with Metric instances.

* doc: Add copyright notice to Prometheus class

* ref: Rename /prometheus endpoint path to /metrics

* feat: Support for content-type negotiation for /metrics

* fix: Set region stat once in constructor

Signed-off-by: Alexandre Fonseca <4717855+alexcsf@users.noreply.github.com>


jvb/src/main/java/org/jitsi/videobridge/metrics/LongGaugeMetric.java
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return (long) get();
+    }
+
+    /**
+     * Atomically increments the value of this gauge by one, returning the updated value.
+     *
+     * @return the updated value
+     */
+    public long incrementAndGetLong()
```
bgrozev, line 52: Idea: maybe this class is superfluus and these methods should go to the default (double) `GaugeMetric` class.

jvb/src/main/java/org/jitsi/videobridge/metrics/MetricsContainer.java
```diff
@@ -0,0 +1,171 @@
...
+     */
+    public BooleanMetric registerBooleanMetric(String name, String help)
+    {
+        BooleanMetric booleanGauge = null;
+        if (!metrics.containsKey(name))
+        {
+            booleanGauge = new BooleanMetric(name, help, METRICS_NAMESPACE);
+            metrics.put(name, booleanGauge);
+        }
+        return booleanGauge;
```
bgrozev, line 113: Did you mean to return the existing metric if a metric with that name was already registered? Or just return `null` as a failure indication?

I suggest logging a warning or even throwing an exception, because I don't anticipate this happening legitimately.

The same applies to the other `register` methods.

jvb/src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
```diff
@@ -116,12 +119,16 @@
     private final @NotNull Videobridge videobridge;
     private final @NotNull XmppConnection xmppConnection;
 
+    private final BooleanMetric healthy = MetricsContainer.getInstance().registerBooleanMetric(
```
bgrozev, line 122: I think it would be nice to have an optional initial value, so we can have this start off as `true`.

bgrozev:
This looks great overall! I left some comments, but there's nothing major.

Some unit tests are failing as you can see from the gihtub action and the tests run by jenkins (the ones named "default"). This is probably reproducible locally by running the whole test suite with `mvn test`. It's good practice to run `mvn install` locally which will include tests and linting.

There's two more general things to discuss before we can merge:
1. Moving the new code to kotlin. This is not required, but in this case (as in most cases, IMO) it will improve the code. Mostly because of number of lines, default parameter values, and null checks (if you insist on keeping it in java I'll ask for some @NotNull/@Nullable annotations). I'm available to help with this.

2. Unit tests. In this case it makes sense to delay writing them until we settle on the architecture to avoid having to refactor them multiple times. But before merging there should be tests for: the basic metrics classes, MetricsContainer (e.g. what it does if you register the same metric name twice), and the HTTP interface.

jvb/pom.xml
```diff
@@ -281,6 +281,16 @@
...
     </dependency>
+    <dependency>
+      <groupId>io.prometheus</groupId>
+      <artifactId>simpleclient</artifactId>
+      <version>0.16.0</version>
+    </dependency>
+    <dependency>
+      <groupId>io.prometheus</groupId>
+      <artifactId>simpleclient_common</artifactId>
+      <version>0.16.0</version>
```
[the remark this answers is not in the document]
alexcsf: I believe they should match, I'll add a version variable.

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -120,6 +121,11 @@
     @NotNull
     private final Clock clock;
 
+    /**
+     * A (singleton) class that defines and exports metrics.
+     */
+    private static final MetricsContainer METRICS_CONTAINER = MetricsContainer.getInstance();
```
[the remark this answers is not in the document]
alexcsf: No need, just an oversight!

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1165,7 +1177,18 @@ public IQ healthCheckIqReceived(@NotNull HealthCheckIQ iq)
         /**
          * Number of local endpoints that exist currently.
          */
-        public AtomicLong currentLocalEndpoints = new AtomicLong();
+        public LongGaugeMetric currentLocalEndpoints = METRICS_CONTAINER.registerLongGauge(
+                "current_local_endpoints",
```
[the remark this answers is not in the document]
alexcsf: I think we should, that matches the metric name for the current number of conferences: `"conferences"`.
Should the _variable_ name keep the prefix? For `"conferences"` I named the variable `currentConferences` to avoid ambiguity with the Conference class.

jvb/src/main/java/org/jitsi/videobridge/metrics/Metric.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+@FunctionalInterface
+public interface Metric
+{
+    /**
+     * Supplies the value of a metric.
+     *
+     * @return the current value of a metric
+     */
+    Object getMetricValue();
```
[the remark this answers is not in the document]
alexcsf: It crossed my mind; I agree with `Metric<T>`.
 I would no longer extend `GaugeMetric` for both `BooleanMetric` and `LongGaugeMetric`.
I would have to implement them concretely, since we'd have `GaugeMetric implements Metric<Double>`.
I could perhaps make `GaugeMetric` a `DoubleGaugeMetric`? Those would be used for metrics such as bitrate.

Does it make sense to skip the getters (such as `getLong()`) and just use the `Metric<T>` interface method? (less clutter)
```java
public final class LongGaugeMetric implements Metric<Long>
{
    private final Gauge gauge;
    // ...

    @Override
    public Long getMetricValue()
    {
        return (long) gauge.get();
    }
    // ...
}
```

Other operations _could_ also return the parameterized type, such as `Long incrementAndGet()`.
The only issue I see here is returning boxed types but I don't believe that's a problem.

jvb/src/main/java/org/jitsi/videobridge/metrics/LongCounterMetric.java
```diff
@@ -0,0 +1,84 @@
...
+
+import io.prometheus.client.*;
+
+/**
+ * Wraps around a (monotonically increasing) Prometheus {@code Counter} and provides atomic operations
+ * such as {@link #incrementAndGetLong()}.
+ *
+ * @see <a href="https://prometheus.io/docs/concepts/metric_types/#counter">Prometheus Counter</a>
+ */
+public class LongCounterMetric implements Metric
```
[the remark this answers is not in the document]
alexcsf: Sounds good, I wasn't sure if we'd have an unlikely double counter.

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -325,6 +333,8 @@ public void expireConference(Conference conference)
             if (conference.equals(conferencesById.get(id)))
             {
                 conferencesById.remove(id);
+                statistics.currentConferences.decrementAndGetLong();
```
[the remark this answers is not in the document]
alexcsf: Agreed, for simplicity's sake I only mimicked the previous `AtomicLong` behavior

jvb/src/main/java/org/jitsi/videobridge/metrics/Metric.java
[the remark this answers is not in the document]
bgrozev: "Other operations could also return the parameterized type, such as Long incrementAndGet()."

I like that. You could implement LongGaugeMetric and BooleanGaugeMetric by delegating to a DoubleGaugeMetric, but I'm not sure it's worth the effort (compared to re-implementing the few methods they have).

jvb/src/main/java/org/jitsi/videobridge/metrics/MetricsContainer.java
alexcsf: I wasn't sure whether I should throw an exception or return an existing one so I left it as it is for now.
I believe the test failures are due to this.
As an example: in [ConferenceTest.kt](https://github.com/jitsi/jitsi-videobridge/blob/master/jvb/src/test/kotlin/org/jitsi/videobridge/ConferenceTest.kt#L31), initializing new `Videobridge.Statistics()` results in null metrics. This leads to `NullPointerException` when calling methods such as `totalConferencesCreated.incrementAndGetLong()`.

Should we just return the metric if it exists (and perhaps log a warning)?

jvb/src/main/java/org/jitsi/videobridge/stats/VideobridgeStatistics.java
alexcsf: Agreed!

jvb/src/main/kotlin/org/jitsi/videobridge/metrics/MetricsContainer.kt
```diff
@@ -0,0 +1,153 @@
...
+import java.io.StringWriter
+
+/**
+ * `MetricsContainer` gathers and exports metrics from a [Videobridge][org.jitsi.videobridge.Videobridge] instance.
+ */
+class MetricsContainer private constructor() {
+    /**
+     * Map metric names to wrapped Prometheus metric types using the [Metric] interface.
+     */
+    private val metrics: MutableMap<String, Metric<*>> = HashMap()
```
bgrozev, line 31: nit: You can use the kotlin `mutableMapOf()` here

jvb/src/main/kotlin/org/jitsi/videobridge/metrics/MetricsContainer.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Returns the metrics in this instance as a JSON string.
+     *
+     * @return a JSON string of the metrics in this instance
+     */
+    val jsonString: String
+        get() {
+            val map: MutableMap<String, Any?> = HashMap(metrics.size)
```
bgrozev, line 40: You can use `metrics.mapValues { it.value.get() }` here

jvb/src/main/kotlin/org/jitsi/videobridge/metrics/InfoMetric.kt
```diff
@@ -0,0 +1,42 @@
...
+    namespace: String,
+    /** the value of this info metric */
+    value: String?
+) : Metric<String> {
+    private val info = Info.build(name, help).namespace(namespace).register().also { it.info(name, value) }
+
+    /**
+     * Returns the value of this info metric.
+     */
+    override fun get() = info.get()[name]!!
```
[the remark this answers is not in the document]
alexcsf: `null` values are not supported, I'll include the change

jvb/pom.xml
```diff
@@ -281,6 +285,16 @@
       <version>2.3.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
bgrozev, line 288: Can it just inherit this from `jicoco-metrics`?

alexcsf: The `/metrics` endpoint uses media type constants such as `TextFormat.CONTENT_TYPE_004`.
Only `simpleclient_common` is needed for that, but we can have those constants in `MetricsContainer` or write them ourselves in the endpoint.

bgrozev (approved):

PR #1926, opened 2022-07-25
feat: Silence detection

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/ConferenceSpeechActivity.java
```diff
@@ -36,6 +37,11 @@
  */
 public class ConferenceSpeechActivity
 {
+    /**
+     * The special endpoint ID to use for silence (when there is no dominant speaker).
+     */
+    private static final String SILENCE = "SILENCE";
```
JonathanLennox, line 43: Would it be better to use an ID that syntactically cannot be used in colibri?  (Is there one?)

JonathanLennox (approved):

PR #1929, opened 2022-08-02
Bandwidth re-allocation timer

bgrozev:


paweldomas:
Does it affect re-allocation when participants join or leave? Is there a chance this could add 5 second delay before the video of a new participant appears?

bgrozev:
> Does it affect re-allocation when participants join or leave? Is there a chance this could add 5 second delay before the video of a new participant appears?

Nope. That's just changing the trigger from RTCP packets to a timer. When an endpoint joins or re-allocation is needed for another reason it executes independently (like it did before).

jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
```diff
@@ -548,17 +563,48 @@ public boolean hasNonZeroEffectiveConstraints(String endpointId)
...
+            delayMs = period.toMillis();
         }
+        else
+        {
+            delayMs = period.minus(timeSinceLastUpdate).toMillis();
+        }
+
+        // Add 5ms to avoid having to re-schedule right away. This increases the average period at which we
+        // re-allocate by an insignificant amount.
+        TaskPools.SCHEDULED_POOL.schedule(
```
JonathanLennox, line 604: I'm wondering if we should hold on to the ScheduledFuture that this returns, so that `BandwidthAllocator.expire()` can cancel it.  Otherwise the bandwidth allocator object is retained by a reference from the scheduled executor service, which can be problematic if you're doing the "set it to something huge to disable it" hack (e.g. in your testing case where you set it to an hour).

jvb/src/test/kotlin/org/jitsi/videobridge/cc/allocation/BitrateControllerTest.kt
```diff
@@ -64,8 +64,15 @@ class BitrateControllerTest : ShouldSpec() {
      */
     override suspend fun beforeSpec(spec: Spec) = super.beforeSpec(spec).also {
         setNewConfig(
-            "videobridge.cc.bwe-change-threshold=0" +
-                "\n" + configWithMultiStreamDisabled, // Also disable multi stream support,
+            """
```
JonathanLennox, line 67: Presumably the tests that use the BitrateController should expire them when they're done, both to avoid leaking the object and to test the expiration behavior?  (Maybe an explicit test for the expiration behavior?)

JonathanLennox, line 67: And I'm wondering if we should have a FakeExecutorService-based test of the periodic update logic.

jvb/src/main/java/org/jitsi/videobridge/cc/allocation/BandwidthAllocator.java
```diff
@@ -568,6 +573,11 @@ public boolean hasNonZeroEffectiveConstraints(String endpointId)
      */
     void expire()
     {
+        ScheduledFuture<?> updateTask = this.updateTask;
+        if (updateTask != null)
+        {
+            updateTask.cancel(false);
+        }
         expired = true;
```
JonathanLennox, line 581: I'd put the expired = true before the cancel, I think it handles races better?

jvb/src/test/kotlin/org/jitsi/videobridge/cc/allocation/BitrateControllerNewTest.kt
```diff
@@ -53,14 +61,48 @@ class BitrateControllerNewTest : ShouldSpec() {
...
+              // Effectively disable periodic updates.
+              max-time-between-calculations = 1 hour 
+            }
+            """.trimIndent(),
+            true
+        )
     }
 
     override suspend fun afterSpec(spec: Spec) = super.afterSpec(spec).also {
+        bc.bc.expire()
```
JonathanLennox, line 77: Looks like you added this to BitrateControllerNewTest but not BitrateControllerTest?

JonathanLennox (approved):

PR #1932, opened 2022-08-09
ref: Port ICE stats to Metrics.

alexcsf:
Also fixes a test failure in BitrateControllerTraceTest and removes a trailing comma ktlint wasn't happy about.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -1004,24 +1004,36 @@ public class Videobridge
          * The total number of times an ICE Agent failed to establish
          * connectivity.
          */
-        public AtomicInteger totalIceFailed = new AtomicInteger();
+        public CounterMetric iceFailed = VideobridgeMetricsContainer.getInstance().registerCounter(
+                "ice_failed",
+                "Number of times an ICE Agent failed to establish connectivity."
+        );
 
         /**
          * The total number of times an ICE Agent succeeded.
          */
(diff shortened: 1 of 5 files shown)


jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1004,24 +1004,36 @@ public IQ healthCheckIqReceived(@NotNull HealthCheckIQ iq)
          * The total number of times an ICE Agent failed to establish
          * connectivity.
          */
-        public AtomicInteger totalIceFailed = new AtomicInteger();
+        public CounterMetric iceFailed = VideobridgeMetricsContainer.getInstance().registerCounter(
```
bgrozev, line 1007: Can you move these from Videobridge.Statistics to IceTransport.Companion?

I see that iceSucceededTcp and iceSucceededRelayed which explains why I've had "fix ice tcp and relayed stats" ticket on my backlog 🤦 . I'll fix that later.

bgrozev: The increment should also be internal to IceTransport

bgrozev: As an example currently IceTransport calls `connected()`, then `Endpoint`/`Relay` fire `iceSucceeded` and `Conference` increments the `iceSucceeded` metrics. Instead `IceTransport` should increment the metric when it calls `connected()`. 
Basically all the logic for ICE metrics (except for should be in IceTransport

bgrozev (approved):

PR #1939, opened 2022-08-19
fix: Treat maxHeight=-1 as "unlimited".

bgrozev:


JonathanLennox (commented):
I'm presuming you found all the places this is used, I haven't verified that part.

JonathanLennox (dismissed):

bgrozev:
> I'm presuming you found all the places this is used, I haven't verified that part.

I went through all the usage I found in the main code, but not the tests as there's too many.

JonathanLennox (approved):

PR #1942, opened 2022-08-23
Move a few metrics to MetricsContainer.

alexcsf:
I have a couple of concerns regarding metric (and variable) naming.

1. (nitpicking) In this case, I think keeping the "total" prefix reduces ambiguity for Counters with similar Gauge counterparts (e.g., `totalEndpoints` vs. `currentLocalEndpoints`). For standalone Counters we can omit "total".

2. More importantly, the Prometheus library adds a "total" suffix to Counters automatically. However, the JSON (at `/metrics`) , doesn't get this suffix. Worse yet, this behavior can break the `MetricsContainer` (though unlikely):

- Register a CounterMetric named `endpoints` which ends up as `endpoints_total` in the Prometheus registry;
- Register a CounterMetric named `endpoints_total`. This works for `MetricsContainer`, but can't be registered in the Prometheus registry since the library [handles "_total" suffixes gracefully](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Counter.java#L101) (i.e., no duplicate suffixes).

If we want consistency, I believe it is better to ensure every Counter ends with "total". I see 3 options:

1. Leave it as is.
2. Rename every ported CounterMetric, adding a "_total" suffix. New CounterMetrics should have this suffix. Contributors would have to be aware of this naming standard.
3. Change `MetricsContainer` so that registering a CounterMetric adds a suffix when necessary. We could log whenever a metric is renamed (and perhaps suggest renaming the metric itself).

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -1100,24 +1100,32 @@ public class Videobridge
         /**
          * The total number of endpoints created.
          */
-        public AtomicInteger totalEndpoints = new AtomicInteger();
+        public CounterMetric totalEndpoints = VideobridgeMetricsContainer.getInstance().registerCounter(
+                "endpoints",
+                "The total number of endpoints created.");
 
         /**
          * The number of endpoints which had not established an endpoint
          * message transport even after some delay.
          */
-        public AtomicInteger numEndpointsNoMessageTransportAfterDelay = new AtomicInteger();
+        public CounterMetric numEndpointsNoMessageTransportAfterDelay = VideobridgeMetricsContainer.getInstance()
+                .registerCounter("endpoints_no_message_transport_after_delay",
+                "Number of endpoints which had not established a relay message transport even after some delay.");
 
         /**
          * The total number of relays created.
          */
-        public AtomicInteger totalRelays = new AtomicInteger();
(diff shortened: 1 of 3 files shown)


bgrozev:
> 2\. Rename every ported CounterMetric, adding a "_total" suffix. New CounterMetrics should have this suffix. Contributors would have to be aware of this naming standard.

How the the prometheus lib behave in this case -- will it add a suffix resulting in `endpoints_total_total`?

alexcsf:
> > 2. Rename every ported CounterMetric, adding a "_total" suffix. New CounterMetrics should have this suffix. Contributors would have to be aware of this naming standard.
> 
> How the the prometheus lib behave in this case -- will it add a suffix resulting in `endpoints_total_total`?

It removes the suffix (`endpoints`) and adds the suffix only when exposing the metrics. [Source](https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Counter.java#L76).

bgrozev:
And if we register both an `endpoints_total` counter and an `endpoints` gauge things will work as expected? If so, I like your second approach best

alexcsf:
> And if we register both an `endpoints_total` counter and an `endpoints` gauge things will work as expected?

The library will throw an exception since it stores the counter as `endpoints` in the registry. It only adds the suffix when writing the metrics.

> If so, I like your second approach best

Renaming every metric (without touching jicoco), correct?

bgrozev:
Talked offline and agreed oon option 3

PR #1944, opened 2022-08-29
Port a few more stats to MetricsContainer.

alexcsf:
- I assume `dtlsFailedEndpoints` is cumulative, can you confirm?
- `totalDominantSpeakerChanges` is/was a `LongAdder`. Is this for better performance?

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
@@ -1131,13 +1131,17 @@ public class Videobridge
          * The total number of times the dominant speaker in any conference
          * changed.
          */
-        public LongAdder totalDominantSpeakerChanges = new LongAdder();
+        public CounterMetric dominantSpeakerChanges = VideobridgeMetricsContainer.getInstance().registerCounter(
+                "dominant_speaker_changes",
+                "Number of times the dominant speaker in any conference changed.");
 
         /**
          * Number of endpoints whose ICE connection was established, but DTLS
          * wasn't (at the time of expiration).
          */
-        public AtomicInteger dtlsFailedEndpoints = new AtomicInteger();
+        public CounterMetric dtlsFailedEndpoints = VideobridgeMetricsContainer.getInstance().registerCounter(
+                "dtls_failed_endpoints",
+                "Number of endpoints whose ICE connection was established, but DTLS wasn't (at time of expiration).");
 
         /**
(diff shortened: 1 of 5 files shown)


jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1131,13 +1131,17 @@ public IQ healthCheckIqReceived(@NotNull HealthCheckIQ iq)
...
+                "dominant_speaker_changes",
+                "Number of times the dominant speaker in any conference changed.");
 
         /**
          * Number of endpoints whose ICE connection was established, but DTLS
          * wasn't (at the time of expiration).
          */
-        public AtomicInteger dtlsFailedEndpoints = new AtomicInteger();
+        public CounterMetric dtlsFailedEndpoints = VideobridgeMetricsContainer.getInstance().registerCounter(
+                "dtls_failed_endpoints",
```
bgrozev, line 1143: Can you rename this to endpoints_dtls_failed for consistency (we have other endpoints_ metrics, right?)

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -1169,7 +1173,9 @@ public IQ healthCheckIqReceived(@NotNull HealthCheckIQ iq)
         /**
          * The total number of times the layering of an incoming video stream changed (updated on endpoint expiration).
          */
-        public AtomicInteger totalLayeringChangesReceived = new AtomicInteger();
+        public CounterMetric layeringChangesReceived = VideobridgeMetricsContainer.getInstance().registerCounter(
+                "layering_changes_received",
+                "Number of times the layering of an incoming video stream changed (updated on endpoint expiration).");
```
bgrozev, line 1178: FYI: I would like to move to a model where we increment these live insted of on expire(), but that's probably outside the scope of your project

bgrozev:
Yes, dtlsFailedEndpoints is cumulative. I'm not sure about the original reason for using a long adder, but I'm not worried about performance -- the speaker doesn't change nearly often enough to have a significant impact

jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
alexcsf: Will do (and yes, we do)

bgrozev (approved):

PR #1951, opened 2022-09-16
Do not use msid in feedback sources.

bgrozev:


jallamsetty1:
The client uses the `mixedmslabel` streamId to determine that these are jvb feedback sources skips remote track creation :(

bgrozev:
> The client uses the `mixedmslabel` streamId to determine that these are jvb feedback sources to skip remote track creation.

Can we change it to use the "owner" field?

JonathanLennox:
There's a ktlint failure, but otherwise looks good to me pending the fix for the issue Jaya mentioned.

jallamsetty1:
> > The client uses the `mixedmslabel` streamId to determine that these are jvb feedback sources to skip remote track creation.
> 
> Can we change it to use the "owner" field?

It is not very straightforward in the client.  The remote track is created when the browser fires the `track` event after remoteSdp is set on the peerconnection. The only information available during the track creation is the stream id (msid) and transceiver the track is attached to.  The client will either have to generate this msid based on the owner field  or assume that the first 2 transceivers are for the jvb feedback sources.  We need this change in the client before this PR can be merged.

bgrozev:
Thanks, @jallamsetty1!
I changed jicofo (jitsi/jicofo#967) to include the msid, so the bridge doesn't have to. So don't be alarmed when this gets merged, the client will still receive the same thing :)

PR #1966, opened 2022-10-31
Remove colibri1

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/Videobridge.java
```diff
@@ -81,8 +81,10 @@
     public static final Random RANDOM = new Random();
 
     /**
-     * The <tt>Conference</tt>s of this <tt>Videobridge</tt> mapped by their
-     * local IDs.
+     * The <tt>Conference</tt>s of this <tt>Videobridge</tt> mapped by their local IDs.
+     *
+     * TODO: The only remaining use of this ID is for the colibri WebSocket conference identifier. This should be
+     * replaced with meetingId.
```
JonathanLennox, line 87: I believe it's also used for the debug API, where it should also be replaced.

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -235,24 +226,9 @@ public Conference(Videobridge videobridge,
...
-                            response = p.getFirst();
-                            expire = p.getSecond();
-                        }
-                        else
-                        {
-                            throw new IllegalStateException("Bad IQ " + request.getClass() + " passed to colibriIQ");
-                        }
+                        Pair<IQ, Boolean> p = colibri2Handler.handleConferenceModifyIQ(request.getRequest());
+                        IQ response = p.getFirst();
+                        boolean expire = p.getSecond();
```
JonathanLennox, line 231: Is a correct exception still thrown, and thus an XMPP error returned, for an unexpected IQ type?

JonathanLennox (approved):

PR #1969, opened 2022-11-06
Support cryptex.

JonathanLennox:
Supports draft-ietf-avtcore-cryptex.  Part of a package with https://github.com/jitsi/jitsi-srtp/pull/29, https://github.com/jitsi/lib-jitsi-meet/pull/2150, and https://github.com/jitsi/jitsi-xmpp-extensions/pull/87.


JonathanLennox:
I marked this as "ready for review" but we still need the AudioLevelReader split.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -771,6 +773,9 @@ class Endpoint @JvmOverloads constructor(
             } else {
                 logger.info("Ignoring empty DtlsFingerprint extension: ${transportInfo.toXML()}")
             }
+            if (CryptexConfig.endpoint) {
+                cryptex = cryptex || fingerprintExtension.cryptex
```
bgrozev, line 777: I think this should be `&&` now? Otherwise it's just equivalent to the config value. The semantics of `Endpoint.cryptex` is "whether we should use cryptex when encrypting", right?

bgrozev (approved):

PR #1983, opened 2023-01-19
Cleanup DTLS 1.0 leftovers

nils-ohlmeier:
This removes remaining bits from supporting DTLS 1.0, which is no longer needed.

Additionally I extended the list of ciphers based on what Chrome and Firefox support today, to avoid running with a single cipher only.


JonathanLennox (dismissed):

bgrozev (approved):

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/dtls/DtlsConfig.kt
```diff
@@ -23,4 +24,16 @@ class DtlsConfig {
...
+            list.map { it.toBcCipherSuite() }
+        }
+    }
+
+    companion object {
+        val config = DtlsConfig()
+    }
 }
+
+private fun String.toBcCipherSuite(): Int = CipherSuite::class.java.getDeclaredField(this).getInt(null)
```
JonathanLennox, line 39: This will throw `NoSuchFieldException` if the field name is spelled wrong - is that good enough, or do we want a cleaner error message?

jitsi-media-transform/src/main/resources/reference.conf
```diff
@@ -30,6 +30,19 @@ jmt {
...
+    // CipherSuite class.
+    cipher-suites = [
+      TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+      TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+      TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+      TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+      TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
+      TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+      TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,
+      TLS_DHE_RSA_WITH_AES_256_GCM_SHA384
```
JonathanLennox, line 44: This should presumably be updated to the list that @nils-ohlmeier came up with

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/dtls/DtlsConfig.kt
```diff
@@ -23,4 +24,16 @@ class DtlsConfig {
     val handshakeTimeout: Duration by config {
         "jmt.dtls.handshake-timeout".from(JitsiConfig.newConfig)
     }
+
+    val ciphersSuites: List<Int> by config {
```
JonathanLennox, line 28: This should probably be named `cipherSuites` not `ciphersSuites`

JonathanLennox: If the BC code wants this as an IntArray, can we represent it as one here, rather than calling `toIntArray` every time?  Or does the config code not handle `IntArray`?

jitsi-media-transform/src/main/resources/reference.conf
bgrozev: I took this from Nils' latest commit. Do you notice a difference?

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/dtls/DtlsConfig.kt
bgrozev: I don't think our config code supports arrays, so we'd lose the usual behavior (e.g. being able to reset for tests) unless we do something like this:
```
private val cipherSuitesProp: List<Int> by config { ... }
fun cipherSuites(): IntArray = cipherSuitesProp.toIntArray()
```

jitsi-media-transform/src/main/resources/reference.conf
JonathanLennox: I missed his "updated cipher list based on Mozilla SSL configurator" commit, so I was remembering the previous one.  This is ok.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/dtls/DtlsConfig.kt
JonathanLennox: That might be a bit cleaner API-wise?  Not big deal tho.

bgrozev (dismissed):

bgrozev (approved):

Issue #1962, opened 2022-10-25
Videobridge considers endpoints with muted video when building lastN list


bukharin:
## Description

JVB takes endpoints with muted video track when signals clients about LastN changes

## Steps to reproduce

Create conference with `A`, `B` and `C` endpoints with active video tracks. Make `B` as dominant speaker. 

- Endpoint `A` ask only 1 video:

```
{"colibriClass":"ReceiverVideoConstraints","lastN":1,"defaultConstraints":{"maxHeight":180}}
```
JVB answer (correct)

```
{"colibriClass":"LastNEndpointsChangeEvent","lastNEndpoints":["B"]}
```

- **Mute video track** for endpoint `C`. Make `C` as dominant speaker. JVB sends:

```
{"colibriClass":"LastNEndpointsChangeEvent","lastNEndpoints":[]}
```

## Expected Behavior

I expect that JVB will not change LastN list and last event will not sent to client

## Possible Solution

Exclude endpoints with muted video track when forming LastN list.

# Environment details

Version: `2.2.46-gf19f7758`

bgrozev:
I agree with your expectation, but wasn't able to reproduce the problem in a unit test (and I don't think it's easy to simulate with jitsi-meet either). It would be helpful if you can give a way to reproduce with jitsi-meet, or in a unit test. See my attempt here:
https://github.com/bgrozev/jitsi-videobridge/tree/audio-only-test

bukharin:
@bgrozev Ok I will try to reproduce soon. 

I would like to clarify the case: it reproduces only if endpoint `C` joins with video track and mutes them later. If it joins in audio only mode - JVB works as expected

bukharin:
@bgrozev in your attempt you create audio only participany with empty array of `MediaSourceDesc`. 

I guess that in my case `Endpoint` has muted `MediaSourceDesc` in mediaSources list and [getEffectiveContraints](https://github.com/jitsi/jitsi-videobridge/blob/620dfa5dd33a391dd321812efd22f6bd1ba349b0/jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/Prioritize.kt#L42) ignores this to determine which sources sould be forwarded

bukharin:
@bgrozev  I found a way to reproduce the problem on meet.jit.si:

[...]

PR #1989, opened 2023-02-06
fix(#1962): Fix lastn when an endpoint has an enabled and a disabled source.

bgrozev:


JonathanLennox:
With this patch all disabled sources are now prioritized below all enabled sources, even if they're selected.  I think that's ok but we want to make sure we've thought about that explicitly.

JonathanLennox (dismissed):

JonathanLennox (approved):

PR #1999, opened 2023-02-23
ref(colibri2 session participant re-invite): specific colibri2 unknown endpoint error

mykhailo-ua:
Added functionality to send back UNKNOWN_ENDPOINT Colibri2Error when there is no such endpoint. This addition reason will help jicofo to recognise this colibri2 error and re-invite participant instead of invalidating whole bridge and re-inviting all participants.
Added this change as part of https://github.com/jitsi/jicofo/issues/1047

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
@@ -63,6 +63,8 @@ class Colibri2ConferenceHandler(
      */
     fun handleConferenceModifyIQ(conferenceModifyIQ: ConferenceModifyIQ): Pair<IQ, Boolean> = try {
         validateRequest(conferenceModifyIQ)
+
+        val ignoreUnknownEndpoints = shouldIgnoreUnknownEndpoints(conferenceModifyIQ)
         val responseBuilder =
             ConferenceModifiedIQ.builder(ConferenceModifiedIQ.Builder.createResponse(conferenceModifyIQ))
         var expire = conferenceModifyIQ.expire.also {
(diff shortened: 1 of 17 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
```diff
@@ -71,7 +73,7 @@ class Colibri2ConferenceHandler(
 
         /* TODO: is there any reason we might need to handle Endpoints and Relays in in-message order? */
```
JonathanLennox, line 74: Your validation I think now means this TODO can go away, because we don't support it.

jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
```diff
@@ -460,3 +481,18 @@ class Colibri2ConferenceHandler(
...
+/**
+ * Return true if "unknown endpoint" and "unknown relay" errors should be ignored when handling [iq]. We ignore errors
+ * if the request simply expires endpoints or a relay (since presumably the entity is already expired), or if it is a
+ * batch update of force-mute state (for the same reason, in order to simplify error handling).
+ */
+private fun shouldIgnoreUnknownEndpoints(iq: ConferenceModifyIQ): Boolean {
+    if (iq.endpoints.any { it.create || it.media.size > 0 || it.transport != null || it.sources != null }) {
+        return false
+    }
+    if (iq.relays.any { !it.expire }) {
```
JonathanLennox, line 494: What about relays' included endpoint lists?  It looks like right now unknown endpoints there are silently ignored - is that what we want?

bgrozev: I think this is okay, because we only expire them when told to do so in signaling?

bgrozev:
Jenkins add to whitelist

JonathanLennox (approved):

PR #2003, opened 2023-03-15
feat: Support setting the initial value of last-n through colibri2.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1137,6 +1137,15 @@ class Endpoint @JvmOverloads constructor(
         logger.info("Expired.")
     }
 
+    fun setInitialLastN(initialLastN: Int) {
+        if (!_messageTransport.isConnected) {
+            logger.info("Setting initialLastN = $initialLastN")
+            bitrateController.lastN = initialLastN
+        } else {
+            logger.info("Ignoring initialLastN, message transport already connected.")
```
JonathanLennox, line 1145: Seems like there's still a race here - the threshold for ignoring the initialLastN should be that receiver constraints have been received, not that the message transport is connected.  (Data will come after the connect for both websocket and datachannel.)

JonathanLennox (approved):

PR #2009, opened 2023-04-06
Add meshId to debug state.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -301,6 +301,7 @@ class Relay @JvmOverloads constructor(
             put("iceTransport", iceTransport.getDebugState())
             put("dtlsTransport", dtlsTransport.getDebugState())
             put("transceiver", transceiver.getNodeStats().toJson())
+            put("meshId", meshId.toString())
```
JonathanLennox, line 304: Doesn't this do "meshId": "null" if there's no meshId?  If this is json unquoted null is probably ok, but quoted null would be confusing (since that's technically a valid meshId value).

bgrozev: I did this on purpose because I have a vague memory of something down the line breaking if we have null values. But I think I may be getting confused with `OrderedJsonObject` which requires non-nulls but isn't used here.

JonathanLennox (approved):

PR #2010, opened 2023-04-07
Call conference.addEndpoints() a single time when adding RelayedEndpoints.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/relay/Relay.kt
```diff
@@ -733,18 +733,19 @@ class Relay @JvmOverloads constructor(
     fun dtlsAppPacketReceived(data: ByteArray, off: Int, len: Int) =
         sctpHandler.processPacket(PacketInfo(UnparsedPacket(data, off, len)))
 
+    /** Return the newly created endpoint, or null if an endpoint with that ID already existed. */
```
JonathanLennox, line 736: Maybe add a note here that this does not add it to the conference?

JonathanLennox (approved):

PR #2012, opened 2023-04-10
Don't expire a conference that's in multiple meshes.

JonathanLennox:
Even if it has no participants - it might be bridging to a visitor relay.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -1076,12 +1076,33 @@ public class Conference
     }
 
     /**
+     * @return {@code true} if this {@link Conference} is in more than one relay mesh.
+     */
+    private boolean inMultipleMeshes()
+    {
+        String meshId = null;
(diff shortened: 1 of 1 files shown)


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1075,13 +1075,34 @@ public Logger getLogger()
...
+    private boolean inMultipleMeshes()
+    {
+        String meshId = null;
+        for (Relay r : relaysById.values())
+        {
+            if (meshId == null)
+            {
+                meshId = r.getMeshId();
+            }
+            else if (!meshId.equals(r.getMeshId()))
```
bgrozev, line 1090: I'm not sure how `null` should be handled here, but I think it's inconsistent: relay IDs of `[null, "A"]` and `["A", null]` will have different results. Maybe something like `relaysById.values().stream().map(r -> r.getMeshId()).collect(Collectors.toSet()).size() > 1;`?

bgrozev (approved):

PR #2017, opened 2023-04-24
fix: Fix sending preemptive keyframe requests with source names, do not send for screensharing.

bgrozev:


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -493,10 +493,23 @@ private void maybeSendKeyframeRequest(AbstractEndpoint dominantSpeaker)
...
+                for (String onStageSourceName : onStageSourcesNames)
+                {
+                    AbstractEndpoint owner = findSourceOwner(onStageSourceName);
+                    if (owner != null)
+                    {
+                        MediaSourceDesc onStageSource = owner.findMediaSourceDesc(onStageSourceName);
+                        if (onStageSource != null && onStageSource.getVideoType() == VideoType.CAMERA)
+                        {
+                            anyEndpointInStageView = true;
+                            break;
```
JonathanLennox, line 509: We want this to break out of the outer loop, not just the inner one.  I think Java has a labeled loop syntax for this?

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -493,10 +493,23 @@ private void maybeSendKeyframeRequest(AbstractEndpoint dominantSpeaker)
...
-            if (otherEndpoint != dominantSpeaker && otherEndpoint.isInStageView())
+            if (otherEndpoint != dominantSpeaker)
             {
-                anyEndpointInStageView = true;
-                break;
+                // Do not anticipate a switch if all on-stage sources are DESKTOP
+                List<String> onStageSourcesNames = otherEndpoint.getOnStageSources();
+                for (String onStageSourceName : onStageSourcesNames)
+                {
+                    AbstractEndpoint owner = findSourceOwner(onStageSourceName);
```
JonathanLennox, line 502: `findSourceOwner` also iterates over all the endpoints, so this is now quadratic in the number of endpoints, which I don't love.  Is there a better way to structure this?

PR #2025, opened 2023-05-19
fix: Do not update RTT when the calculated value is invalid

bgrozev:
or suspiciously high, or the sender's "delay since last SR" values is
too high.

### DIFF
--- a/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
+++ b/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
@@ -144,20 +144,30 @@ class EndpointConnectionStats(
             // The delaySinceLastSr value is given in 1/65536ths of a second, so divide it by .000065536 to get it
             // in nanoseconds
             val remoteProcessingDelay = Duration.ofNanos((reportBlock.delaySinceLastSr / .000065536).toLong())
-            rtt = (Duration.between(srSentTime, receivedTime) - remoteProcessingDelay).toDoubleMillis()
-            if (rtt > 7.secs.toMillis()) {
-                logger.warn(
-                    "Suspiciously high rtt value: $rtt ms, remote processing delay was " +
-                        "$remoteProcessingDelay (${reportBlock.delaySinceLastSr}), srSentTime was $srSentTime, " +
-                        "received time was $receivedTime"
-                )
-            } else if (rtt < 0) {
-                logger.warn(
-                    "Negative rtt value: $rtt ms, remote processing delay was " +
-                        "$remoteProcessingDelay (${reportBlock.delaySinceLastSr}), srSentTime was $srSentTime, " +
-                        "received time was $receivedTime"
-                )
+            if (remoteProcessingDelay > Duration.ofMinutes(5)) {
+                logger.warn("Ignoring report block with suspiciously long DLSR: $remoteProcessingDelay")
+                return@let
             }
+
+            val newRtt = (Duration.between(srSentTime, receivedTime) - remoteProcessingDelay).toDoubleMillis()
+            if (newRtt > 7.secs.toMillis()) {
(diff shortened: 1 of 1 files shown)


jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
```diff
@@ -144,20 +144,30 @@ class EndpointConnectionStats(
             // The delaySinceLastSr value is given in 1/65536ths of a second, so divide it by .000065536 to get it
             // in nanoseconds
             val remoteProcessingDelay = Duration.ofNanos((reportBlock.delaySinceLastSr / .000065536).toLong())
-            rtt = (Duration.between(srSentTime, receivedTime) - remoteProcessingDelay).toDoubleMillis()
-            if (rtt > 7.secs.toMillis()) {
+            if (remoteProcessingDelay > Duration.ofMinutes(5)) {
```
JonathanLennox, line 147: How did you pick 5 minutes for this value?  It seems high.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
```diff
@@ -144,20 +144,30 @@ class EndpointConnectionStats(
...
+                logger.warn("Ignoring report block with suspiciously long DLSR: $remoteProcessingDelay")
+                return@let
+            }
+
+            val newRtt = (Duration.between(srSentTime, receivedTime) - remoteProcessingDelay).toDoubleMillis()
+            if (newRtt > 7.secs.toMillis()) {
                 logger.warn(
-                    "Suspiciously high rtt value: $rtt ms, remote processing delay was " +
+                    "Ignoring suspiciously high rtt value: $newRtt ms, remote processing delay was " +
```
JonathanLennox, line 155: 7 seconds RTTs are actually possible, usually when someone is in a cellular network that's just marginally connected.  Maybe we should keep a stat for this separate from the RTTs?  (Not sure how best to handle repeated high RTTs from the same endpoint though.)

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
bgrozev: I didn't think much about it. I saw 2h35m in logs and it was causing problems. What's a more reasonable value?

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
bgrozev: What's a better cutoff?

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
JonathanLennox: Did you actually see some large values here (as opposed to the calculated RTT)?

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/stats/EndpointConnectionStats.kt
JonathanLennox: I think it's as reasonable a cutoff as any, it's certainly indicative of a very unhappy endpoint (or a very unhappy cloud instance, sometimes) -- I just think it'd be good to have the stat, too.  Doesn't need to be this PR though.

JonathanLennox (approved):
On reflection I think this is fine as-is...the five minute cutoff is large but it's a bit more than twice the TCP maximum segment lifetime (2 minutes) so that's probably a reasonable number to pick.

PR #2038, opened 2023-08-03
Remove maxReceiverVideoConstraints (unused).

bgrozev:


JonathanLennox:
I don't think the change matches the description?  It seems like you just renamed it, rather than removing it.

bgrozev:
I removed the `maxReceiverVideoConstraints` field, then ranamed `maxReceiverVideoConstraintsMap` to `maxReceiverVideoConstraints`

JonathanLennox (approved):

PR #2044, opened 2023-08-23
Increase consent check and inactivity intervals.

bgrozev:
- Bump consent check and inactivity intervals.
- Add missing whitespace in reference.conf.

### DIFF
--- a/jvb/src/main/resources/reference.conf
+++ b/jvb/src/main/resources/reference.conf
@@ -29,7 +29,7 @@ videobridge {
 
     # How long an endpoint can be 'inactive' before it will
     # be considered disconnected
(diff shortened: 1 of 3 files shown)


jvb/src/main/resources/reference.conf
```diff
@@ -1,54 +1,54 @@
 videobridge {
   entity-expiration {
     # If an entity has no activity after this timeout, it is expired
-    timeout=1 minute
+    timeout =1 minute
```
JonathanLennox, line 4: If you're adding whitespace this should have whitespace on both sides of the equal sign

JonathanLennox (approved):

PR #2046, opened 2023-09-11
Bump Bouncycastle to version 1.76.

JonathanLennox:


bgrozev:
1.76 is not good because of bcgit/bc-java#1489, let's wait for the fix.

PR #2072, opened 2023-11-13
Colibri msg initiated pcap capture

dkirov-dev:
This PR introduces new functionality to start pcap capture per endpoint from a colibri message.
If the colibri message "conference-modify"/"endpoint" has new parameters "context-id" and "pcap-rec-mode" set to enable capture, then jvb will produce pcap files for this specific endpoint.
```
`<endpoint xmlns='jitsi:colibri2' id='32bbe701' create='true' stats-id='Sallie-XSA' context-id='919633fd-93f7-43eb-802e-f285a0248065' pcap-rec-mode='audio'>`
```
The new functionality is configured / enabled by adding a new section named "metadata-pcap-recording" in
jitsi-media-transform/src/main/resources/reference.conf
```
  metadata-pcap-recording {
    enabled=true
    // Base path for storing the pcap and metadata files
    base-path="/tmp/jvb-rec"
  }
```
For each recorded endpoint / context pair, JVB produces 2 pcap and 2 json metadata files describing the pcap files.
The file name is generated as a function of 
```
<base-path>/<context-id>_<endpoint-id>_<capture-end>.
```
For example:
919633fd-93f7-43eb-802e-f285a0248065_5d9bdf3b_fe.json
919633fd-93f7-43eb-802e-f285a0248065_5d9bdf3b_fe.pcap
919633fd-93f7-43eb-802e-f285a0248065_5d9bdf3b_ne.json
919633fd-93f7-43eb-802e-f285a0248065_5d9bdf3b_ne.pcap

The JSON metadata file includes the endpoint_id, context_id, capture_end and payload_map describing a specific pcap file with the same name.
```
{
  "endpoint_id": "5d9bdf3b",
   "context_id": "919633fd-93f7-43eb-802e-f285a0248065",
   "capture_end": "fe",
   "capture_mode": "audio-video",
   "payload_map": [
    {
      "payload_type": 101,
      "media_type": "video",
      "clock_rate": 90000,
      "encoding": "vp9"
    },
    {
      "payload_type": 126,
      "media_type": "audio",
      "clock_rate": 8000,
      "encoding": "telephone-event"
    },
    {
      "payload_type": 111,
      "media_type": "audio",
      "clock_rate": 48000,
      "encoding": "opus"
    }
  ]
}
```
The pcap file will contain RTP packets for audio if the new parameter pcap-rec-mode = "audio" or "audio-video" and it will contain video packets if pcap-rec-mode = "video" or "audio-video".
The "context-id" identifies a specific instance of a jitsi room and allows a jitsi customer to select whether jvb should produce pcap files for a specific endpoint on the fly and to correlate a specific instance of a video conferencing room with a specific participant / endpoint pcap captures. 
This new functionality should be considered as an extension or replacement of TRANSCEIVER_PCAP_DUMP debug feature that is driven from colibri message instead of config settings.
A system integrator can simply start pcap with metadata recordings from prosody plugin for a specific participant of a video conferencing room based on a logic specific to this system integrator and without any other changes in the jitsi code.


dkirov-dev:
@damencho , @emcho , @bgrozev , @JonathanLennox please review.

bgrozev:
Jenkins, add to whitelist

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/CompliancePcapWriter.kt
```diff
@@ -0,0 +1,263 @@
...
+            }
+        }
+
+        meta.put("payload_map", payloadsMap)
+
+        // json-simple produces one long flat line
+        // File(jsonFilename()).writeText(meta.toJSONString())
+
+        // gson produces elements on separate lines
+        val gson = GsonBuilder().setPrettyPrinting().create()
```
bgrozev, line 113: Can we not introduce a new dependency just for this? I think the cleanest way would be to define a kotlin class for your structure and use jackson to serialize to JSON, something like 
```
        data class Metadata(
val endpointId: String,
val contextId: String,
val captureEnd: String,
val payloadType: List<...>
) {
            
        
        fun toJsonString(): String = writer.writeValueAsString(this)
        companon object {
            val writer = jacksonObjectMapper().writerWithDefaultPrettyPrinter()
        }
```

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/CompliancePcapWriter.kt
```diff
@@ -0,0 +1,263 @@
...
+            return
+        }
+
+        if (newMode !in listOf(CAP_MODE_AUDIO, CAP_MODE_VIDEO, CAP_MODE_AUDIO_VIDEO)) {
+            logger.error("Compliance recording:${capId()} invalid mode:$newMode")
+            return
+        }
+
+        if (newContextId.isNullOrEmpty()) {
+            logger.error("Compliance recording:${capId()} invalid context-id:$newContextId")
```
bgrozev, line 152: The logs should already contain CompliancePcapWriter, so no need to include "Complience recording"

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/CompliancePcapWriter.kt
dkirov-dev: Sure we can change this to whatever you guys think is the best, but please note that using jacksonObjectMapper would also introduce new dependency since currently jitsi-media-transform does not include jackson.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/CompliancePcapWriter.kt
dkirov-dev: Done.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/CompliancePcapWriter.kt
bgrozev: That's fine, it's already a dependency in jvb/pom.xml

bgrozev:
Are you also planning to contribute changes to jicofo for this feature?

dkirov-dev:
@bgrozev I am not sure if changes in jicofo are needed for this and we need your team advice here. 
So far I was able to test it with a simple prosody plugin which does not require changing the colibri message structure:
```
function on_iq_received(event)
   if event and event.stanza and event.stanza.name == "iq" then
      local colibri = event.stanza:get_child("conference-modify", "jitsi:colibri2");
      if colibri ~= nil then
         local endpoint = colibri:get_child("endpoint");
         if endpoint ~= nil then
            endpoint.attr["com-rec"] = "audio-video"
            endpoint.attr["context-id"] = "919633fd-93f7-43eb-802e-f285a0248065" 
            module:log("info", "comrec enabled for endpoint: %s", endpoint.attr.id);
         end
      end
   end
end
module:hook("pre-iq/full", on_iq_received);
module:log("info", "Loaded mod_room_comrec");
```
But technically speaking, the "context-id" is more appropriately placed in the conference-modify section and not in the endpoint section, but that would require changing the classes that define the message structures.
Your input here is appreciated. Thanks!

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/CompliancePcapWriter.kt
dkirov-dev: Done. Latest commit uses jackson instead of gson.

PR #2075, opened 2023-12-19
task: graceful shutdown only via REST

aaronkvanmeerten:


damencho (approved):

DanielMcAssey:
Is graceful shutdown not supported by JVB anymore?

aaronkvanmeerten:
> Is graceful shutdown not supported by JVB anymore?

Graceful shutdown is still supported, simply does not require a PID, and no longer needs to wait and kill the JVB.  Instead, after being triggered for graceful shutdown via REST, the JVB will watch itself for a minimum number of participants or a timeout, and then exit gracefully itself either way.

This PR is meant to reflect this new behavior, but maintain the functionality of the graceful-shutdown.sh script.

PR #2081, opened 2024-01-19
ref: Move Harvesters to kotlin.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/Main.kt
```diff
@@ -76,7 +76,9 @@ fun main() {
     }
 
     startIce4j()
-    Harvesters.initializeStaticConfiguration()
+
+    // Initialize, binding on the main ICE port.
+    Harvesters.INSTANCE
```
JonathanLennox, line 81: I understand this constructs the object and thus does the static initialization, but it still feels weird idiomatically to just reference an object and not use it.  Is there any way to make this look like a function call without making other things less idiomatic?

JonathanLennox (approved):

PR #2084, opened 2024-01-23
Send pings periodically on idle websocket connections.

JonathanLennox:


jvb/src/main/resources/reference.conf
```diff
@@ -225,6 +225,12 @@ videobridge {
     server-id = "default-id"
     // Whether to negotiate WebSocket compression (permessage-deflate)
     enable-compression = true
+    // Whether to send keepalive pings on idle websockets
+    send-keepalive-pings = true
+    // The keepalive ping timeout
+    keepalive-ping-timeout = 15 seconds
```
bgrozev, line 231: Should this be named with "interval" instead of "timeout"?

JonathanLennox: Makes sense.

bgrozev (approved):

PR #2087, opened 2024-01-24
Remove backward compat for source names.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/BitrateController.kt
```diff
@@ -130,7 +122,7 @@ class BitrateController<T : MediaSourceContainer> @JvmOverloads constructor(
     /**
      * Return the number of endpoints whose streams are currently being forwarded.
```
JonathanLennox, line 123: Comment needs updating.

JonathanLennox (commented):
I think `sendVideoConstraints` (V1) can also go away?

In general search for any remaining @Deprecated markers and make sure they're not source-names related?

bgrozev:
There are 3 remaining deprecated functions: VideoTypeMessage, onStageEndpoints and onStage sources. Unfortunately I don't know if they are still used by endpoints or not (even though endpoints claim to support source names). I'll see if I can find some data.

Issue #2083, opened 2024-01-23
Does initial-last-n actually works?


bukharin:
## Current behavior

Hi, 

I am trying to use channelLastN option to limit lastN video before datachannel connection is established:

```
channelLastN: 1
```

I see initial-last-n option in the session-accept response from the client:

```
<initial-last-n value="1" xmlns="jitsi:colibri2"/>
```

But when datachannel is established, client receives `ForwardedSources` message which indicates that JVB forwards all existing endpoints (6 in this example)

![image](https://github.com/jitsi/jitsi-videobridge/assets/3126568/5946bde3-50aa-4775-a39a-8d8a1008e1cb)

After that JVB signals about 1 forwarding endpoint as expected (may be client `ReceiverVideoConstraints` change params?)

## Expected Behavior

JVB forwards only `initial-last-n` endpoints before datachannel established. Client does not receive `ForwardedSources` message with all endpoints after datachannel connection is established

# Environment details

Reproducable on latest release of JVB + Jicofo

bgrozev:
Do you see any jvb log messages containing "initialLastN"?

bukharin:
@bgrozev yes, it has

```
Setting initialLastN = 1
```

Also if I try to disable sending `ReceiverVideoConstraints` message in lib-jitsi-meet. In this case I see the same behaivour. It seems that JVB always sends `ForwardedSources` message with all endpoints.

![image](https://github.com/jitsi/jitsi-videobridge/assets/3126568/efd39fc7-9557-448b-a6e6-7584149b2f39)

bgrozev:
That's strange and I can't reproduce it. Do you use SCTP or WS for the data channel?

bukharin:
WebSockets, the screenshots above were taken from Chrome DevTools

bgrozev:
I think I see how this happens. The WS must have been established before the bridge received the initial-last-n from jicofo. In this case the bridge will send the initial forwarded sources message (with lastN=-1), then send the updated on (with lastN=1) once it receives initial-last-n.

I'm surprised it happens often, because the WS requires at least an extra RTT for the TCP handshake, while the initial-last-n is just a packet sent to jicofo which immediately updates the bridge via an existing connection. 

For your use case, would it work if we add a config option to the bridge setting a default initial lastN value?

bukharin:
@bgrozev yes, you are right. It's a race condition. 

[...]

PR #2095, opened 2024-02-13
feat: Add a config option for default initial last n.

bgrozev:
See discussion in #2083


jvb/src/main/resources/reference.conf
```diff
@@ -67,6 +67,10 @@ videobridge {
     # no last-n limit)
     jvb-last-n = -1
 
+    // The initial value for the client-configured last-n setting. This applies before it is overriden by colibri or
+    // data channel signaling.
```
JonathanLennox, line 71: Suggest repeating the "(-1 implies no last-n limit)" text

JonathanLennox (approved):

PR #2101, opened 2024-02-20
Fix bug that would cause VP9/AV1 simulcast not to be routed properly after a switch.

JonathanLennox:


jvb/src/main/kotlin/org/jitsi/videobridge/cc/av1/Av1DDAdaptiveSourceProjectionContext.kt
```diff
@@ -228,6 +228,9 @@ class Av1DDAdaptiveSourceProjectionContext(
 
     private fun frameIsNewSsrc(frame: Av1DDFrame): Boolean = lastAv1FrameProjection.av1Frame?.matchesSSRC(frame) != true
 
+    private fun frameIsProjectable(frame: Av1DDFrame): Boolean =
+        frameIsProjectable(frame) || frame.index >= lastFrameNumberIndexResumption
```
bgrozev, line 232: This doesn't seem right. Does it need to call `frameIsNewSsrc` like vp9?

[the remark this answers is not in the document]
JonathanLennox: You are right, thank you!

JonathanLennox: Damian also caught this...autocomplete error.

bgrozev (approved):

PR #2117, opened 2024-04-04
fix: Catch RTP parsing exceptions.

bgrozev:


jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/RtpParser.kt
```diff
@@ -43,12 +43,25 @@ class RtpParser(
...
-            MediaType.AUDIO -> when (payloadType.encoding) {
-                RED -> packet.toOtherType(::RedAudioRtpPacket)
-                else -> packet.toOtherType(::AudioRtpPacket)
+            MediaType.AUDIO -> try {
+                when (payloadType.encoding) {
+                    RED -> packet.toOtherType(::RedAudioRtpPacket)
+                    else -> packet.toOtherType(::AudioRtpPacket)
+                }
+            } catch (e: Exception) {
+                logger.info("Dropping audio packet due to parse failure: ${e.message}")
```
paweldomas, line 52: Just a note that if someone floods the jvb with invalid packets, catching so many exceptions will have a big performance hit. So as soon as you know what the problem is, it'd be good to try to make it really be an exception rather than the norm. If the parser sees some assumptions don't add up it'd be good it returned null and maybe logged an error every x packets instead of throwing.

bgrozev: Thanks, @paweldomas ! Do you think the performance of catching the exceptions here would be worse compared to letting them bubble up and be caught at the queue? If not, I'd like to keep this change as it cleans up the logs and returns the buffer to the pool.

I agree that it would be better if the parser didn't throw an exception, but it's a non-trivial change because it's [in the `RtpPacket` constructor](https://github.com/jitsi/jitsi-videobridge/blob/master/rtp/src/main/kotlin/org/jitsi/rtp/rtp/RtpPacket.kt#L141).

JonathanLennox (approved):

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/RtpParser.kt
paweldomas: The biggest cost of throwing exceptions is the part that creates the stacktrace, so the place where it's caught does not matter.

PR #2124, opened 2024-04-16
Add an sctp-pcap-dump feature.

JonathanLennox:


jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/transform/node/ToggleablePcapWriter.kt
```diff
@@ -51,11 +51,15 @@ class ToggleablePcapWriter(
 
     fun isEnabled(): Boolean = pcapWriter != null
 
-    fun newObserverNode(): Node = PcapWriterNode("Toggleable pcap writer: $prefix")
+    fun newObserverNode(outbound: Boolean): Node = PcapWriterNode("Toggleable pcap writer: $prefix", outbound)
```
bgrozev, line 54: If it's not private anymore why not change the return type to PcapWriterNode?

PR #2127, opened 2024-04-23
Use JitsiXmppStringprep, update smack

bgrozev:
- **ref: Do not register xmpp extensions for RAW UDP, we only support ICE.**
- **ref: Remove unused code.**
- **feat: Use JitsiXmppStringprep**
- **doc: Fix javadocs.**
- **ref: Move queue stats out of Videobridge.java.**
- **ref: Move ulimit check out of Videobridge.java.**
- **ref: Move videobridgeExpireThread out of Videobridge.java.**

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/stats/QueueStats.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/stats/QueueStats.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright @ 2024 - present 8x8, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/stats/QueueStats.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.jitsi.videobridge.Endpoint
+import org.jitsi.videobridge.relay.Relay
+import org.jitsi.videobridge.relay.RelayEndpointSender
+import org.json.simple.JSONObject
+
+object QueueStats {
+    /**
+     * Gets statistics for the different `PacketQueue`s that this bridge
+     * uses.
+     * TODO: is there a better place for this?
```
JonathanLennox, line 33: I think you've answered this TODO?

JonathanLennox:
Do you have any reason to think we might need to add a flag for the new stringprep here, like you added in jicofo?

bgrozev:
The bridge only sees the JIDs from the brewery and jicofo, I don't see any risk or a reason to disable it.

JonathanLennox (approved):

PR #2131, opened 2024-05-16
Convert to use dcsctp4j for SCTP.

JonathanLennox:
This will depend on https://github.com/JonathanLennox/dcsctp4j becoming a "real" project (i.e. with its packages in a Maven repo somewhere).


jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1172,6 +1088,79 @@ class Endpoint @JvmOverloads constructor(
...
+                        0
+                    } else {
+                        logger.error("Error sending to SCTP: $status")
+                        -1
+                    }
+                },
+                logger
+            )
+            // This handles if the remote side will be opening the data channel
+            dataChannelStack!!.onDataChannelStackEvents { dataChannel ->
```
bbaldino, line 1136: One thing I've done in situations like this is assign the newly-created thing to a local variable, so then the compiler knows it's non-null.  You can assign it to the member somewhere (like a `this.dataChannelStack = dataChannelStack` at the end or right after it's created if there are early-return scenarios), but use it via the local reference then you don't need all the `!!` in the calls.

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1172,6 +1209,80 @@ class Endpoint @JvmOverloads constructor(
...
+                logger
+            )
+            this@Endpoint.dataChannelStack = dataChannelStack
+            // This handles if the remote side will be opening the data channel
+            dataChannelStack.onDataChannelStackEvents { dataChannel ->
+                logger.info("Remote side opened a data channel.")
+                messageTransport.setDataChannel(dataChannel)
+            }
+            dataChannelHandler.setDataChannelStack(dataChannelStack)
+            if (OPEN_DATA_CHANNEL_LOCALLY) {
```
bgrozev, line 1263: OPEN_DATA_CHANNEL_LOCALLY is just hard-coded to `false`. If this mode hasn't been tested should we just remove it?

JonathanLennox: Probably.  We have real code to open the datachannel locally in Relay, so we don't need this as a model for the future.

jvb/src/main/kotlin/org/jitsi/videobridge/dcsctp/DcSctpTransport.kt
```diff
@@ -0,0 +1,146 @@
...
+            this.timeoutId = timeoutId
+            future = TaskPools.SCHEDULED_POOL.schedule(this, duration, TimeUnit.MILLISECONDS)
+        }
+
+        override fun stop() {
+            future?.cancel(false)
+        }
+
+        override fun run() {
+            transport.socket.handleTimeout(timeoutId)
```
bgrozev, line 143: Is this safe to run in the single-threaded SCHEDULED_POOL? Perhaps keep the timer there but move the work to an IO thread?

bgrozev:
We make sure that when videobridge.sctp.enabled=false the usrsctp library is never loaded. Is the same true for dcsctp?

bgrozev (approved):

PR #2136, opened 2024-05-23
fix(AV1 DD): Separate AV1 DD template's decodeTargetLayers and decodeTargetProtectedBy fields.

JonathanLennox:
The former can be present when the latter is not.

### DIFF
--- a/rtp/src/main/kotlin/org/jitsi/rtp/rtp/header_extensions/Av1DependencyDescriptorHeaderExtension.kt
+++ b/rtp/src/main/kotlin/org/jitsi/rtp/rtp/header_extensions/Av1DependencyDescriptorHeaderExtension.kt
@@ -627,6 +627,7 @@ class Av1DependencyDescriptorReader(
          */
         templateCnt = 0
         templateInfo.clear()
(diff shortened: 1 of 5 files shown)


rtp/src/main/kotlin/org/jitsi/rtp/rtp/header_extensions/Av1DependencyDescriptorHeaderExtension.kt
```diff
@@ -623,7 +627,7 @@ class Av1DependencyDescriptorReader(
          */
         templateCnt = 0
         templateInfo.clear()
-        decodeTargetInfo.clear()
+        decodeTargetLayers.clear()
```
bgrozev, line 630: Does `decodeTargetProtectedBy` also need to be cleared?

JonathanLennox: Ah, you're right, yes, thank you!

bgrozev (approved):

PR #2163, opened 2024-06-04
feat(config): Make the packet timeline configurable.

bgrozev:


JonathanLennox:
I'm a little concerned what happens to packets that are in flight at the moment the timeline gets turned on.

Maybe restructure it so that `PacketInfo.timeline` is an `EventTimeline?` which only gets instantiated if `enableTimeline` is set, and then everything else checks whether the specific PacketInfo's timeline val is non-null?

bgrozev:
The timeline is configurable in the config file but once initialized doesn't change (payload verification does, via an http api). I just renamed `ENABLE_TIMELINE` to `enableTimeline` because I think that's the usual convention

JonathanLennox (dismissed):

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
```diff
@@ -24,10 +24,17 @@ import java.time.Clock
...
 @SuppressFBWarnings("CN_IMPLEMENTS_CLONE_BUT_NOT_CLONEABLE")
 class EventTimeline(
-    private val timeline: MutableList<Pair<String, Duration>> = mutableListOf(),
+    /**
+     * We want this thread safe, because while PacketInfo objects are only handled by a single thread at a time,
+     * [StatsKeepingNode] may add an "exit" event after the packets has been added to another queue potentially handled
+     * by a different thread. This is not critical as it only affects the timeline and the result is just some "exit"
+     * events missing from the trace logs.
+     */
+    private val timeline: ConcurrentLinkedQueue<Pair<String, Duration>> = ConcurrentLinkedQueue(),
```
JonathanLennox, line 37: I think `ConcurrentLinkedQueue` is probably more heavyweight than we want for this - it's an actual reader-writer queue, with poll() and the like.

I think Collections.synchronizedList(mutableListOf()), and synchronizing on `timeline` around the `joinToString` (which uses an iterator internally), would be better.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
```diff
@@ -56,7 +58,7 @@ class EventTimeline(
     }
 
     fun clone(): EventTimeline {
-        val clone = EventTimeline(ConcurrentLinkedQueue(timeline))
+        val clone = EventTimeline(timeline.toMutableList())
```
JonathanLennox: Not a new bug, but while you're in there, `clone()` should pass `clock` rather than letting it default to `Clock.systemUTC()`.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
```diff
@@ -34,9 +33,12 @@ class EventTimeline(
      * by a different thread. This is not critical as it only affects the timeline and the result is just some "exit"
      * events missing from the trace logs.
      */
-    private val timeline: ConcurrentLinkedQueue<Pair<String, Duration>> = ConcurrentLinkedQueue(),
+    timelineArg: MutableList<Pair<String, Duration>> = mutableListOf(),
     private val clock: Clock = Clock.systemUTC()
 ) : Iterable<Pair<String, Duration>> {
+
+    private val timeline: MutableList<Pair<String, Duration>> = Collections.synchronizedList(timelineArg)
```
JonathanLennox, line 40: This shouldn't need the type

JonathanLennox (approved):

PR #2182, opened 2024-06-21
Fix(dcsctp): Fix memory leak.

JonathanLennox:
There were reference cycles through JNI, which the garbage collector can't detect.


aaronkvanmeerten (approved):
Approved without full comprehension

bgrozev (approved):

PR #2194, opened 2024-07-18
feat: Remove support for ICE/TCP.

bgrozev:
* TURN/TLS provides a better solution for the same problem.
* The implementation has been known to have issues with threads getting blocked.
* ICE/TCP gets in the way of improvements we're currently working on.
* We (jitsi team) haven't used it in any of our deployments since at least 2018.
* See also: https://github.com/jitsi/docker-jitsi-meet/commit/7a939785a7f018601afe2d49d8f66078812a3dd0


JonathanLennox (approved):

oktay454:
Due to the UDP/10000 port on a JMS we host on our own servers, we cannot communicate with a person in a different institution.
They need to allow our server's UDP/10000 port.
For this problem, I was disabling the HTTP 80 port and assigning the TCP/80 port to the JVB.

What solution do you have for UDP/10000 in corporate networks? (Unless an additional permission is defined in the organization's firewall)

bgrozev:
You can use a TURN server accessible via TLS/443

PR #2198, opened 2024-07-22
fix vp9 temporal layers count in flexibale mode

splinehip:
In flexible mode (chrome version 125+) if we have only one temporal layer with id=0, received client rendering video stream in 8 fps only.

### DIFF
--- a/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/rtp/codec/vp9/Vp9Packet.kt
+++ b/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/rtp/codec/vp9/Vp9Packet.kt
@@ -260,6 +260,13 @@ class Vp9Packet private constructor(
             if (off == -1) {
                 return null
             }
+
+            val isSwitchingUpPoint = DePacketizer.VP9PayloadDescriptor
+                .isSwitchingUpPoint(buffer, payloadOffset, payloadLength)
+
+            val isFlexibleMode = DePacketizer.VP9PayloadDescriptor
+                .isFlexibleMode(buffer, payloadOffset, payloadLength)
+
             val ssHeader = buffer[off].toInt()
 
             val numSpatial = ((ssHeader and 0xE0) shr 5) + 1
@@ -300,6 +307,14 @@ class Vp9Packet private constructor(
                     }
                 }
                 groupSize
+            } else if (isFlexibleMode && isSwitchingUpPoint) {
+                // In flexible mode (chrome version 125+) if we have only one temporal layer with id=0,
+                // received client rendering video stream in 8 fps only.
+                val numTemporal = 3
+                for (tid in 0 until numTemporal) {
+                    tlCounts[tid]++
+                }
+                numTemporal
             } else {
                 tlCounts[0] = 1
                 1


splinehip:
Signed https://www.8x8.com/products/docusign-thank-you?env=517e312c-babc-4fd6-ba23-79b5aadf8c5e&pf=2fbd46b1-ed23-46a3-a57e-c7c0005544ca&r=b2d32bda-5a5e-4611-8604-e491e50345c0

JonathanLennox:
Can you describe how the browser is set to generate this encoding mode, so I can test?  Thanks!

splinehip:
Before join to conference set this config in browser debug console: config.videoQuality = ['VP9']

jallamsetty1:
> Before join to conference set this config in browser debug console: config.videoQuality = ['VP9']

@splinehip How are you forcing the browser to send only one temporal layer? We always configure 3 temporal layers for VP9 both in simulcast and full SVC mode.

jallamsetty1:
Even when I force the client to encode a single temporal layer by setting `L1T1` as the scalability mode, I see 30 fps being sent and received for a JVB call.

Screenshot 2024-07-22 at 3 37 54 PM
Screenshot 2024-07-22 at 3 38 27 PM

sunduev:
@jallamsetty1 @JonathanLennox Hi! Please try Chromium 125+ (which [enabled flexible mode for encoding by default](https://groups.google.com/g/discuss-webrtc/c/RpePKpwGIds/m/botDdva6AwAJ))
And you have to use vp9 for encoding, no more configuration should be required (if L3T3 is used by default, L1T1 probably might work as it is), then you should watch for low fps (around 7-9) on viewers side.

splinehip:
> > Before join to conference set this config in browser debug console: config.videoQuality = ['VP9']
> 
> @splinehip How are you forcing the browser to send only one temporal layer? We always configure 3 temporal layers for VP9 both in simulcast and full SVC mode.

Just use chrome 125+ and set VP9 codec by client side:
Снимок экрана 2024-07-23 в 10 24 49

splinehip:
Снимок экрана 2024-07-23 в 11 47 56

jallamsetty1:
@splinehip Thanks, I am seeing it on meet.jit.si now but not on my lonely.

splinehip:
> @splinehip Thanks, I am seeing it on meet.jit.si now but not on my lonely.

Could you try on yours site use 3 participants in conference, when only 2 participants - clients work through p2p not bridge?

jallamsetty1:
> > @splinehip Thanks, I am seeing it on meet.jit.si now but not on my lonely.
> 
> Could you try on yours site use 3 participants in conference, when only 2 participants - clients work through p2p not bridge? Or set boths clients: config.p2p = false

Yes, I do understand that. My lonely was running a bridge version which is slightly older than the current meet.jit.si version.

Toumassa:
Hello, i can confirm this issue exists with VP9 falling to 7-8 fps.
Can AV1 be concerned with flexible mode ?
I thought setting videoQuality.vp9.useKSVC and / or videoQuality.vp9.scalabilityModeEnabled to false should correct this issue ? But actually it doesn't ...

jallamsetty1:
> Hello, i can confirm this issue exists with VP9 falling to 7-8 fps. Can AV1 be concerned with flexible mode ? I thought setting videoQuality.vp9.useKSVC and / or videoQuality.vp9.scalabilityModeEnabled to false should correct this issue ? But actually it doesn't ...

This shouldn't affect AV1. Setting `videoQuality.vp9.scalabilityModeEnabled=false` will force the client to use the old K-SVC mode for VP9 but it is still VP9.

Toumassa:
> This shouldn't affect AV1. Setting `videoQuality.vp9.scalabilityModeEnabled=false` will force the client to use the old K-SVC mode for VP9 but it is still VP9.

I've set disableSimulcast to true and i get 30fps. However no scalability between filmstrip and large video. Can be a workaround waiting this to be merged.

JonathanLennox:
I have a more general solution (that doesn't hard-code the assumption that there will be three temporal layers) at https://github.com/jitsi/jitsi-videobridge/pull/2199.  Would you be able to test this?

splinehip:
> I have a more general solution (that doesn't hard-code the assumption that there will be three temporal layers) at #2199. Would you be able to test this?

@JonathanLennox Hi! I tested your fix - work fine, thank you!

JonathanLennox:
@splinehip - I pushed some more changes to https://github.com/jitsi/jitsi-videobridge/pull/2199 which should make it more efficient on the server-side - can you test again so we can be sure?  Thanks!

splinehip:
@JonathanLennox HI! Tested, work fine. Thanks!

PR #2220, opened 2024-09-22
feat(WIP): Export audio to a recorder over a WS.

bgrozev:


JonathanLennox (approved):

JonathanLennox:
Jenkins, test this please

JonathanLennox:
Jenkins, test this please

wnikolaichuk:
Hello everyone, when will the merge request be closed? What is needed for this, maybe you need to help in some way? Multi-streaming audio recording Jitsi recording is very important for my project

bgrozev:
I'm resuming active work on it now, should be pretty close.

bgrozev:
@savely-krasovsky Give this new version a go

savely-krasovsky:
@bgrozev it works! Could you maybe suggest what is the proper way to start and stop those recordings? From what I understand it uses COLIBRI2 modify conference request, so the client side probably should be implemented as part of `lib-jitsi-meet`?

savely-krasovsky:
@bgrozev thanks! The puzzle is starting to come together... I also find that for now the only metadata which we can recieve at JMR side is endpoint ID (and we can obtain it using `ffprobe` for example), but it's hard to match it with real conference participant (or not?). From what I understand exporter thing has only access to RTC packet stream and it doesn't hold this kind of metadata at all. For example Jibri collects that information using Selenium context, but what about JMR?

bgrozev:
We use this together with another service which collects metadata about the conference. You should be able to implement the same with this community-maintained prosody module:
https://github.com/jitsi-contrib/prosody-plugins/tree/main/event_sync

JonathanLennox (approved):

PR #2221, opened 2024-09-24
Don't feed SCTP packets to the SCTP stack after it's been closed.

JonathanLennox:
(It results in harmless-but-scary warnings in the logs about verification tags being wrong.)

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/dcsctp/DcSctpTransport.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/dcsctp/DcSctpTransport.kt
@@ -15,11 +15,13 @@
  */
 package org.jitsi.videobridge.dcsctp
 
+import org.jitsi.dcsctp4j.DcSctpMessage
 import org.jitsi.dcsctp4j.DcSctpOptions
 import org.jitsi.dcsctp4j.DcSctpSocketCallbacks
 import org.jitsi.dcsctp4j.DcSctpSocketFactory
 import org.jitsi.dcsctp4j.DcSctpSocketInterface
 import org.jitsi.dcsctp4j.SendOptions
+import org.jitsi.dcsctp4j.SendStatus
 import org.jitsi.dcsctp4j.Timeout
 import org.jitsi.nlj.PacketInfo
 import org.jitsi.utils.OrderedJsonObject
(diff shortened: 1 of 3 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/dcsctp/DcSctpTransport.kt
```diff
@@ -40,19 +42,51 @@ class DcSctpTransport(
...
+    fun stop() {
+        synchronized(lock) {
+            socket?.close()
+            socket = null
+        }
+    }
+
+    fun connect() {
+        synchronized(lock) {
+            socket?.connect()
```
bgrozev, line 70: Should this throw an error when `socket` is `null`?

jvb/src/main/kotlin/org/jitsi/videobridge/dcsctp/DcSctpTransport.kt
```diff
@@ -40,19 +42,51 @@ class DcSctpTransport(
...
+        synchronized(lock) {
+            socket = factory.create(name, callbacks, null, options)
+        }
     }
 
     fun handleIncomingSctp(packetInfo: PacketInfo) {
         val packet = packetInfo.packet
-        socket.receivePacket(packet.getBuffer(), packet.getOffset(), packet.getLength())
+        synchronized(lock) {
+            socket?.receivePacket(packet.getBuffer(), packet.getOffset(), packet.getLength())
```
bgrozev, line 57: Do we care about keeping track of how often this happens?

bgrozev (approved):

jvb/src/main/kotlin/org/jitsi/videobridge/dcsctp/DcSctpTransport.kt
JonathanLennox: I don't think so - it just means that the socket was torn down before it could be connected, which should be fine.

jvb/src/main/kotlin/org/jitsi/videobridge/dcsctp/DcSctpTransport.kt
JonathanLennox: I don't think so - it's normal in the case where a sctp packet arrives after we shut down the sctp stack but before we shut down the dtls stack.  This is currently cluttering up the logs and is harmless, thus this change.

PR #2263, opened 2024-12-12
Add support for the VLA RTP header extension

bgrozev:
- **feat: Add BitReader utilities and tests.**
- **feat: Add a parser for the VLA RTP header extension.**
- **ref: Remove unused function.**
- **feat: Update layers with info found in VLA.**
- **feat: Add an option to use targetBitrate instead of measured bitrate for allocation**


JonathanLennox (commented):
You'll want to make sure the VLA is forwarded by HeaderExtStripper on Octo links, but probably it's not needed to be sent down to endpoints?

JonathanLennox (approved):

PR #2265, opened 2024-12-20
Catch TERM, HUP, and INT signals, and do a clean shutdown.

JonathanLennox:


jvb/src/main/kotlin/org/jitsi/videobridge/Main.kt
```diff
@@ -163,6 +164,21 @@ fun main() {
         null
     }
 
+    var exitStatus = 0
+
+    /* Catch signals and cause them to trigger a clean shutdown. */
+    listOf("TERM", "HUP", "INT").forEach { signalName ->
+        try {
+            Signal.handle(Signal(signalName)) { signal ->
+                shutdownService.beginShutdown()
```
bgrozev, line 173: Suggest logging

bgrozev: And set exitStatus before beginShutdown?

bgrozev (approved):

PR #2277, opened 2025-01-14
Add probing stats, try to fix probing rate

bgrozev:
- **feat: Add stats for bandwidth probing to debug (rtcstats).**
- **fix: Use observed bitrate instead of encoder target when calculating probing needed.**

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/cc/BandwidthProbing.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/cc/BandwidthProbing.kt
@@ -92,7 +92,7 @@ class BandwidthProbing(
 
         // How much padding do we need?
         val totalNeededBps = bitrateControllerStatus.currentIdealBps - bitrateControllerStatus.currentTargetBps
-        if (totalNeededBps < 1) {
+        if (totalNeededBps < 1 || !bitrateControllerStatus.hasNonIdealLayer) {
             // Don't need to send any probing.
             bytesLeftOver = 0
             zeroStats()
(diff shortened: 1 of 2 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/cc/BandwidthProbing.kt
```diff
@@ -36,18 +40,28 @@ class BandwidthProbing(
     /**
      * The sequence number to use if probing with the JVB's SSRC.
      */
-    private var seqNum = Random.nextInt(0xFFFF)
+    private val seqNum = Random.nextInt(0xFFFF)
```
JonathanLennox, line 43: It looks like this (and ts below) isn't used anywhere other than debug state output, so better to delete it than to make it a `val` (which doesn't make any sense for a sequence number)

jvb/src/main/kotlin/org/jitsi/videobridge/cc/allocation/BitrateController.kt
```diff
@@ -200,11 +200,7 @@ class BitrateController<T : MediaSourceContainer> @JvmOverloads constructor(
         val nowMs = clock.instant().toEpochMilli()
         val allocation = bandwidthAllocator.allocation
         allocation.allocations.forEach { singleAllocation ->
-            val allocationTargetBitrate: Bandwidth? = if (config.useVlaTargetBitrate) {
-                singleAllocation.targetLayer?.targetBitrate ?: singleAllocation.targetLayer?.getBitrate(nowMs)
-            } else {
-                singleAllocation.targetLayer?.getBitrate(nowMs)
-            }
+            val allocationTargetBitrate: Bandwidth? = singleAllocation.targetLayer?.getBitrate(nowMs)
```
JonathanLennox, line 203: This will cause probing to happen, even when we're fully sending ideal, if the VLA bitrate is greater than the measured bitrate, which seems bad.
Conversely if the VLA bitrate is less than the measured bitrate we won't probe up, though I'm confused if that will matter.

JonathanLennox (approved):

PR #2283, opened 2025-01-21
VideobridgeExpireThread is created but not started

dsmeytis:
Hello @bgrozev! I'm not really sure whether it is a typo or done on purpose but I noticed that during `VideobridgeExpireThread` move from `Videobridge` to `Main` in commit https://github.com/jitsi/jitsi-videobridge/commit/b623d6af714a17922750aa9d1c8060406ecb6e0a call to `VideobridgeExpireThread.start()` was lost.
Again, maybe it is done on purpose but in my deployment it leads to a weird situation that peers who lost connection are getting stuck in "reconnecting" state without being expired.


JonathanLennox:
Jenkins, test this please

JonathanLennox (approved):

PR #2304, opened 2025-03-07
feat(deb): Moves to using jvb.conf only.

damencho:
The empty jvb.conf has been added from the deb package since late 2020. Now we can assume when this file is missing, this is a new install and will create one with default values.


debian/postinst
```diff
@@ -82,25 +70,8 @@ case "$1" in
...
-            echo "org.jitsi.videobridge.xmpp.user.shard.DOMAIN=auth.$JVB_HOSTNAME" >> $NEW_JITSI_CONFIG
-            echo "org.jitsi.videobridge.xmpp.user.shard.USERNAME=jvb" >> $NEW_JITSI_CONFIG
-            echo "org.jitsi.videobridge.xmpp.user.shard.PASSWORD=$JVB_SECRET" >> $NEW_JITSI_CONFIG
-            echo "org.jitsi.videobridge.xmpp.user.shard.MUC_JIDS=JvbBrewery@internal.auth.$JVB_HOSTNAME" >> $NEW_JITSI_CONFIG
-            echo "org.jitsi.videobridge.xmpp.user.shard.MUC_NICKNAME=$(uuidgen)" >> $NEW_JITSI_CONFIG
-        fi
-
         if [ ! -f $HOCON_CONFIG ]; then
-            echo "Generating an empty hocon config"
+            echo "Generating an default hocon config"
```
JonathanLennox, line 74: Should be "a default"

damencho: Fixed, thank you.

JonathanLennox (approved):

PR #2311, opened 2025-04-04
Periodically send clients the current estimated downlink bandwidth.

JonathanLennox:


jvb/src/main/resources/reference.conf
```diff
@@ -382,6 +382,14 @@ videobridge {
   # Whether to redact remote endpoint IP addresses from logs
   redact-remote-addresses = true
 
+
+  connection-stats {
+    # Whether to periodically send connection stats (currently, estimated downlink bandwidth) to endpoints
+    enabled = true
+    # How often to send these stats
+    interval = 5 seconds
```
damencho, line 390: The clients stats we send every 10 seconds by default ... maybe we can do the same and for these ...

bgrozev (approved):

JonathanLennox:
Jenkins, test this please

PR #2321, opened 2025-04-25
Various enhacements to timeseries debugging

JonathanLennox:
Added during new-BWE experimentation, but independent of it.

### DIFF
--- a/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
+++ b/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
@@ -19,6 +19,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 import org.jitsi.config.JitsiConfig
 import org.jitsi.metaconfig.config
 import org.jitsi.rtp.Packet
+import org.jitsi.rtp.rtcp.RtcpPacket
 import org.jitsi.utils.logging2.createLogger
 import java.time.Clock
 import java.time.Duration
@@ -90,6 +91,16 @@ class EventTimeline(
     }
 }
 
+enum class PacketType {
+    Routed,
+    RTX,
+    Padding,
+    Synthesized,
+    RTCP,
+    Misc
+    /* TODO: handle SCTP and datachannel if we need them */
+}
+
 /**
  * [PacketInfo] is a wrapper around a [Packet] instance to be passed through
  * a pipeline.  Since the [Packet] can change as it moves through the pipeline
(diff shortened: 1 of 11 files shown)


jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
```diff
@@ -90,6 +91,16 @@ class EventTimeline(
     }
 }
 
+enum class PacketType {
+    Routed,
+    RTX,
+    Padding,
+    Synthesized,
+    RTCP,
```
bgrozev, line 99: Not sure what "RTCP" means here. RTCP packets can be Routed, or Synthesized, right?

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
```diff
@@ -140,6 +151,11 @@ open class PacketInfo @JvmOverloads constructor(
      */
     var payloadVerification = if (enablePayloadVerification) packet.payloadVerification else null
 
+    /**
+     * The type of the packet, used for tracking the sources of media being routed.
+     */
+    var packetType: PacketType = if (packet is RtcpPacket) PacketType.RTCP else PacketType.Misc
```
bgrozev, line 157: This is initialized when the PacketInfo is created. I don't think the RTCP parser creates a new instance, so RTCP is only set for packets we generate?

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
```diff
@@ -90,6 +91,16 @@ class EventTimeline(
     }
 }
 
+enum class PacketType {
+    Routed,
+    RTX,
```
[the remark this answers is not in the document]
JonathanLennox: In the new Google CC (which motivated this work), there's separate probing technique (send a short burst of data very fast and measure how quickly it's received) which is different than our padding probing, so I'd like to keep the name "Padding" here.  Ok on the RTX.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/PacketInfo.kt
JonathanLennox: The rate trackers I'm measuring currently are only measuring VideoType packets, so RTCP isn't really handled at all.  Possibly I should just leave RTCP as TODO for now.

bgrozev (approved):

PR #2322, opened 2025-04-25
Add activeDecodeTargets to AV1 DD JSON description

JonathanLennox:
(This is only used for debug output.)


rtp/src/main/kotlin/org/jitsi/rtp/rtp/header_extensions/Av1DependencyDescriptorHeaderExtension.kt
```diff
@@ -262,6 +262,13 @@ class Av1DependencyDescriptorHeaderExtension(
             customDtis?.let { put("customDTIs", it) }
             customFdiffs?.let { put("customFdiffs", it) }
             customChains?.let { put("customChains", it) }
+            activeDecodeTargetsBitmask?.let {
+                if (newTemplateDependencyStructure == null ||
+                    activeDecodeTargetsBitmask != ((1 shl newTemplateDependencyStructure.decodeTargetCount) - 1)
```
bgrozev, line 267: Not sure what the check does, but it should use the `it` reference

bgrozev (approved):

PR #2325, opened 2025-04-30
Compute and store extended sequence numbers as Long rather than Int.

JonathanLennox:
In a long-lived session, I'm pretty sure that extended sequence numbers can exceed an int, especially for transport sequence numbers (the most quickly-incrementing sequence number in our system).

In most cases the rollover will be fine, but there are some cases (e.g. `TccGeneratorNode`) where we compare extended sequence numbers numerically.

### DIFF
--- a/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/util/ArrayCache.kt
+++ b/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/util/ArrayCache.kt
@@ -200,7 +200,7 @@ open class ArrayCache<T>(
         if (head == -1 || index > cache[head].index) {
             return
         }
-        val diff = cache[head].index - index
+        val diff = index - cache[head].index
         val position = position(diff)
         if (cache[position].index == index) {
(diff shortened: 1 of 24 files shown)


jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/util/ArrayCache.kt
```diff
@@ -178,27 +181,27 @@ open class ArrayCache<T>(
...
     }
 
-    private fun doUpdateTimeAdded(index: Int, timeAdded: Long) {
+    private fun doUpdateTimeAdded(index: Long, timeAdded: Long) {
         if (head == -1 || index > cache[head].index) {
             return
         }
         val diff = cache[head].index - index
-        val position = (head - diff) floorMod size
+        val position = position(diff)
```
bgrozev, line 204: `position(diff)` does `head + diff`, but the original code here does `head - diff`

bgrozev (approved):

PR #2329, opened 2025-05-13
feat: Detect CPU steal.

bgrozev:


jvb/src/main/kotlin/org/jitsi/videobridge/load_management/JvbLoadMeasurement.kt
```diff
@@ -24,6 +24,9 @@ interface JvbLoadMeasurement {
     fun getLoad(): Double
 
     operator fun div(other: JvbLoadMeasurement): Double
+    operator fun plus(other: JvbLoadMeasurement): Double {
```
JonathanLennox, line 27: From an operator overloading point of view it's weird to have `plus` not return the type being added.

jvb/src/main/kotlin/org/jitsi/videobridge/load_management/StealDetection.kt
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.videobridge.load_management
+
+import org.jitsi.metrics.DoubleGaugeMetric
+import org.jitsi.videobridge.metrics.VideobridgeMetricsContainer.Companion.instance as metricsContainer
+
+object StealDetection {
```
JonathanLennox, line 21: It'd be nice to have unit tests for this.  Probably by making it a class rather than an object, and having some way to feed it dummy contents of /proc/stat.

JonathanLennox (commented):
Are we confident that CPU steal should be treated the same way as other system load?

JonathanLennox (approved):

PR #2332, opened 2025-05-15
Merge index trackers (rtp seq, picture ID), add timestamp index tracker.

bgrozev:
- **feat: Abstract the rfc3711 index tracker, add RtpTimestampIndexTracker.**
- **ref: Rename to IndexTraker.kt**
- **test: Add index tracker tests.**
- **ref: Rename Rfc3711IndexTracker to RtpSequenceIndexTracker.**
- **ref: Implement PictureIdIndexTracker as an IndexTracker.**
- **ref: Further abstract the index tracker.**


jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/util/IndexTracker.kt
```diff
@@ -0,0 +1,137 @@
...
+        val diff = a - b
+        return when {
+            diff < -(1L shl (bits - 1)) -> diff + (1L shl bits)
+            diff > (1L shl (bits - 1)) -> diff - (1L shl bits)
+            else -> diff
+        }
+    }
+}
+
+sealed class IndexTracker<T> {
```
JonathanLennox, line 65: Why did you make this sealed?

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/util/IndexTracker.kt
```diff
@@ -0,0 +1,137 @@
...
+ * limitations under the License.
+ */
+
+package org.jitsi.nlj.util
+
+/**
+ * Index tracker inspired by the RFC3711 RTP sequence number index tracker.
+ */
+
+class PictureIdIndexTracker : IntIndexTracker(15)
```
JonathanLennox, line 23: Add a doc that this is specifically for VP8/VP9 picture IDs, with a rollover of 0x8000?

jitsi-media-transform/src/test/kotlin/org/jitsi/nlj/util/Rfc3711IndexTrackerTest.kt
```diff
@@ -24,7 +24,7 @@ import io.kotest.matchers.shouldBe
 internal class Rfc3711IndexTrackerTest : ShouldSpec() {
```
JonathanLennox, line 24: Rename this test?  We should probably add tests for the timestamp tracker too.

jitsi-media-transform/src/test/kotlin/org/jitsi/nlj/codec/vpx/PictureIdIndexTrackerTest.kt
```diff
@@ -20,6 +20,7 @@ import io.kotest.core.spec.IsolationMode
 import io.kotest.core.spec.style.ShouldSpec
 import io.kotest.matchers.comparables.shouldBeGreaterThan
 import io.kotest.matchers.shouldBe
+import org.jitsi.nlj.util.PictureIdIndexTracker
 
 internal class PictureIdIndexTrackerTest : ShouldSpec() {
```
JonathanLennox, line 25: This test should probably get moved to the namespace that PictureIdIndexTracker is in now.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/util/IndexTracker.kt
bgrozev: It has to be public because it defines the API, but I want to keep the helper functions internal. I can technically make it just `abstract`, but then you can't define a superclass outside the package because you can't access the internal functions need to be overriden. So it would be effectively sealed except you get a somewhat cryptic error when you try to define a super class: `MyIndexTracker inherits invisible abstract members: internal abstract fun addRollover(seqNum: Int, roc: Long): Long defined in org.jitsi.nlj.util.IndexTracker`

JonathanLennox (approved):

PR #2338, opened 2025-06-06
WIP: add handler for ReceiverAudioSubscriptionMessage

kota-yata:
This is going to be one of PRs for the [Audio Switchboard API](https://summerofcode.withgoogle.com/myprojects/details/aBy319rB).

ReceiverAudioSubscriptionMessage is new bridge message to notify selected audio sources from the receiver. where to store the subscription and how to filter packets are further discussed and implemented later.

Corresponding change to lib-jitsi-meet: https://github.com/jitsi/lib-jitsi-meet/pull/2813

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
@@ -157,6 +157,13 @@ public class EndpointMessageTransport
         return null;
     }
 
+    @Override public BridgeChannelMessage receiverAudioSubscription(ReceiverAudioSubscriptionMessage receiverAudioSubscriptionMessage)
+    {
+        getLogger().info("Received audio subscription: " + receiverAudioSubscriptionMessage);
+        // TODO: store the subscription somewhere
+        return null;
+    }
+
     @Override
     public void unhandledMessage(@NotNull BridgeChannelMessage message)
     {
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
@@ -59,7 +59,8 @@ import java.util.concurrent.atomic.AtomicLong
     JsonSubTypes.Type(value = ReceiverVideoConstraintsMessage::class, name = ReceiverVideoConstraintsMessage.TYPE),
     JsonSubTypes.Type(value = SourceVideoTypeMessage::class, name = SourceVideoTypeMessage.TYPE),
     JsonSubTypes.Type(value = ConnectionStats::class, name = ConnectionStats.TYPE),
-    JsonSubTypes.Type(value = VideoTypeMessage::class, name = VideoTypeMessage.TYPE)
+    JsonSubTypes.Type(value = VideoTypeMessage::class, name = VideoTypeMessage.TYPE),
+    JsonSubTypes.Type(value = ReceiverAudioSubscriptionMessage::class, name = ReceiverAudioSubscriptionMessage.TYPE)
 )
 sealed class BridgeChannelMessage {
     private val jsonCacheDelegate = ResettableLazy { createJson() }
@@ -124,6 +125,7 @@ open class MessageHandler {
             is SourceVideoTypeMessage -> sourceVideoType(message)
             is ConnectionStats -> connectionStats(message)
             is VideoTypeMessage -> videoType(message)
+            is ReceiverAudioSubscriptionMessage -> receiverAudioSubscription(message)
         }
     }
 
@@ -150,6 +152,7 @@ open class MessageHandler {
     open fun sourceVideoType(message: SourceVideoTypeMessage) = unhandledMessageReturnNull(message)
     open fun connectionStats(message: ConnectionStats) = unhandledMessageReturnNull(message)
     open fun videoType(message: VideoTypeMessage) = unhandledMessageReturnNull(message)
+    open fun receiverAudioSubscription(message: ReceiverAudioSubscriptionMessage) = unhandledMessageReturnNull(message)
 
     fun getReceivedCounts() = receivedCounts.mapValues { it.value.get() }
 }
@@ -475,6 +478,14 @@ class SourceVideoTypeMessage(
 
     companion object {
         const val TYPE = "SourceVideoTypeMessage"
+    }
+}
+
+class ReceiverAudioSubscriptionMessage(
+    val sourceNames: List<String>
+) : BridgeChannelMessage() {
+    companion object {
+        const val TYPE = "ReceiverAudioSubscriptionMessage"
     }
 }
 


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1028,6 +1031,21 @@ public void addEndpointSsrc(@NotNull AbstractEndpoint endpoint, long ssrc)
         }
     }
 
+    public List<Long> getAllRemoteAudioSsrcs(String localEndpointId) {
+        List<Long> ssrcs = new ArrayList<>();
+        for (AbstractEndpoint endpoint : getEndpoints()) {
```
bgrozev, line 1036: If you use getLocalEndpoints instead you can remove the `instanceof` check

kota-yata: Used getLocalEndpoints and removed type check

jvb/src/main/java/org/jitsi/videobridge/EndpointMessageTransport.java
```diff
@@ -157,6 +157,18 @@ public BridgeChannelMessage sourceVideoType(SourceVideoTypeMessage sourceVideoTy
         return null;
     }
 
+    @Override public BridgeChannelMessage receiverAudioSubscription(ReceiverAudioSubscriptionMessage receiverAudioSubscriptionMessage)
```
bgrozev, line 160: Add a `@NotNull` annotation to the parameter to match the kotlin interface. Fix the line length (max 120 characters)

jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1028,6 +1031,19 @@ public void addEndpointSsrc(@NotNull AbstractEndpoint endpoint, long ssrc)
         }
     }
 
+    public List<Long> getAllRemoteAudioSsrcs(String callerId) {
```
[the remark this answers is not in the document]
bgrozev: You should be able to run `mvn install` locally and get a clean result. It's currently failing with some linting errors.

jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
```diff
@@ -296,6 +296,12 @@ class Colibri2ConferenceHandler(
             sources.mediaSources.filter { it.type == MediaType.AUDIO }.forEach {
                 it.sources.forEach { s ->
                     audioSources.add(AudioSourceDesc(s.ssrc, c2endpoint.id, it.id))
+                    conference.getLocalEndpoints().forEach {e ->
```
bgrozev, line 299: The purpose of Colibri2ConferenceHandler is to read the signaling message and update the state accordingly. I would prefer that any side effects happen in the native objects themselves. In this case, this should happen in the setter for Endpoint.audioSources, where you also have an opportunity to check for removed sources. You can use something like this:
```
    var audioSources: ArrayList<AudioSourceDesc> = ArrayList()
        set(newValue) {
            val oldValue = field
            val removedSsrcs = oldValue.map { it.ssrc }.toSet() - newValue.map { it.ssrc }.toSet()
            val addedSsrcs = newValue.map { it.ssrc }.toSet() - oldValue.map { it.ssrc }.toSet()
            // Update state as necessary, call into conference. 
            field = newValue
        }
```

And I would actually prefer to use the AudioSource for these callbacks. You can save the SSRC only in the place where you match against a packet.

jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -478,6 +481,15 @@ class SourceVideoTypeMessage(
     }
 }
 
+class ReceiverAudioSubscriptionMessage(
```
bgrozev, line 484: Add brief docs on the semantics please

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -535,6 +541,75 @@ class Endpoint @JvmOverloads constructor(
         }
     }
 
+    private class AudioSubscriptionEntry {
+        private var excludeWildcard = false
+        private var includeWildcard = true
+        private var wantedSsrcs = mutableListOf<Long>()
```
bgrozev, line 547: You can use an unmutable List or Set instead, reducing the risk of issue with concurrent modification. 

It looks like you don't need to keep track of `wantedSsrcs` in the case where `includeWildcard || excludeWildcard` is true, because if `includeWildcard` or `excludeWildcard` ever change then `wantedSsrcs` is reset anyway.

bgrozev: I just noticed that you're matching the SSRCs against the signaled list. The ReceiverAudioSubscriptionMessage should use source names in its lists, as SSRCs are:
1. Not always known to the receivers
2. May be changed for some receivers with "SSRC rewriting"

So we should:
1. Document the "include" and "exclude" lists' semantics, these are source names (that's why they're strings, not numbers)
2. Change the "getAllRemoteAudioSsrcs" method in Conference to return a list of AudioSourceDesc. I also suggest removing the parameter to this method, and filtering it inside Endpoint.AudioSubscriptionEntry

bgrozev:
Please merge latest master

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
kota-yata: This was my mistake. It now retrieves audioSourceDescs from the Conference and update subscription accordingly.

jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
bgrozev: Yes, exactly 👍

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -411,6 +417,18 @@ class Endpoint @JvmOverloads constructor(
      *  Keep track of this endpoint's audio sources.
      */
     var audioSources: ArrayList<AudioSourceDesc> = ArrayList()
+        set(newValue) {
+            val oldValue = field
+            val removedSsrcs = oldValue.map { it.ssrc }.toSet() - newValue.map { it.ssrc }.toSet()
+            val addedSsrcs = newValue.map { it.ssrc }.toSet() - oldValue.map { it.ssrc }.toSet()
+            conference.getLocalEndpoints().forEach { e ->
```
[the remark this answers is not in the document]
kota-yata: fixed with [be8c927](https://github.com/jitsi/jitsi-videobridge/pull/2338/commits/be8c927bc8558b2390bb09617cfd9c1c29c31947)

jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
```diff
@@ -1216,4 +1250,50 @@ class Endpoint @JvmOverloads constructor(
...
+            return wantedSsrcs.contains(ssrc)
+        }
+
+        fun onConferenceSourceAdded(descs: Set<AudioSourceDesc>) {
+            // If the subscription has a wildcard include or exclude, we don't filter the sources.
+            // Getting out of wildcard state is only done by new subscription, which re-initialize watedSsrcs.
+            if (latestSubscription.include.contains("*") || latestSubscription.exclude.contains("*")) {
+                return
+            }
+            wantedSsrcs = wantedSsrcs.union(descs.filter { isSsrcWanted(it.ssrc) }.map(AudioSourceDesc::ssrc).toSet())
```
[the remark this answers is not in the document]
kota-yata: AudioSubscriptionEntry separated at [0c834a0](https://github.com/jitsi/jitsi-videobridge/pull/2338/commits/0c834a01198cd7ba0c2c8037b4b0b902c5ea9891), corresponding test added at [b338756](https://github.com/jitsi/jitsi-videobridge/pull/2338/commits/b338756177dc5120c9ddaff76222882f1777b00c)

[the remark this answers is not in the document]
kota-yata: Now the class checks subscription or sources consistently by sourceNames, never with SSRCs.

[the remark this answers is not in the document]
kota-yata: I've added some tests to see if the class is tolerant of SSRC changes

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscriptionEntry.kt
```diff
@@ -0,0 +1,69 @@
...
+) {
+    private var latestSubscription: ReceiverAudioSubscriptionMessage = ReceiverAudioSubscriptionMessage(
+        listOf("*"),
+        emptyList()
+    )
+    private var wantedSsrcs: Set<Long> = emptySet()
+
+    fun updateSubscription(subscription: ReceiverAudioSubscriptionMessage, sources: List<AudioSourceDesc>) {
+        latestSubscription = subscription
+        if (subscription.exclude.contains("*") || subscription.include.contains("*")) {
```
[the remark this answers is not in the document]
kota-yata: Removed this linear searches by having mode property in `ReceiverAudioSubscriptionMessage`

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscriptionEntry.kt
```diff
@@ -0,0 +1,69 @@
...
+        wantedSsrcs = descs.filter { desc ->
+            subscription.include.contains(desc.sourceName) && !subscription.exclude.contains(desc.sourceName)
+        }.map(AudioSourceDesc::ssrc).toSet()
+    }
+
+    fun isSsrcWanted(ssrc: Long): Boolean {
+        if (latestSubscription.exclude.contains("*")) {
+            return false
+        }
+        if (latestSubscription.include.contains("*")) {
```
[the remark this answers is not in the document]
kota-yata: This is now fixed by having mode property at `ReceiverAudioSubscriptionMessage`

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscriptionEntry.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    fun updateSubscription(subscription: ReceiverAudioSubscriptionMessage, sources: List<AudioSourceDesc>) {
+        latestSubscription = subscription
+        if (subscription.exclude.contains("*") || subscription.include.contains("*")) {
+            return
+        }
+        val descs = sources.filter { desc -> desc.owner != endpointId }
+        wantedSsrcs = descs.filter { desc ->
+            subscription.include.contains(desc.sourceName) && !subscription.exclude.contains(desc.sourceName)
+        }.map(AudioSourceDesc::ssrc).toSet()
```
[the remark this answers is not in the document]
kota-yata: >In particular this means that it's impossible for a receiver to hear a new participant until they've gotten the notification about them, and subscribed, which may lead to clipping.

A receiver can hear a new participant if it has wildcard include subscription. Otherwise it is intended that the new sources aren't forwarded.

Now that the subscription has a mode property, the AudioSubscription class looks at the subscription's mode first. If the mode is `All`, forward everything including new sources. If it is `None`, forward none. If it is `Custom`, forward those contained in the include list and not in the exclude list.

jvb/src/test/kotlin/org/jitsi/videobridge/AudioSubscriptionEntryTest.kt
```diff
@@ -0,0 +1,199 @@
...
+
+class AudioSubscriptionEntryTest : ShouldSpec() {
+    override fun isolationMode(): IsolationMode? = IsolationMode.InstancePerLeaf
+
+    private val endpointId = "test-endpoint"
+    private val source1 = AudioSourceDesc(100L, "endpoint1", "source1")
+    private val source2 = AudioSourceDesc(200L, "endpoint2", "source2")
+    private val source3 = AudioSourceDesc(300L, "endpoint3", "source3")
+    private val ownSource = AudioSourceDesc(400L, endpointId, "own-source")
+    private val allSources = listOf(source1, source2, source3, ownSource)
```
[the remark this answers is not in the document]
kota-yata: Test added at [59c77fe](https://github.com/jitsi/jitsi-videobridge/pull/2338/commits/59c77fe3c3cddb8753ee49d5d0b04fa864bf3bdb)

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscription.kt
```diff
@@ -0,0 +1,88 @@
...
+            }
+            is ReceiverAudioSubscriptionMessage.Exclude -> {
+                wantedSsrcs = sources.filterNot { desc ->
+                    subscription.list.contains(desc.sourceName)
+                }.map(AudioSourceDesc::ssrc).toSet()
+            }
+        }
+    }
+
+    fun isSsrcWanted(ssrc: Long): Boolean {
```
bgrozev, line 47: You can simplify this to:
```
fun isSsrcWanted(ssrc: Long): Boolean = when (latestSubscription) {
        is ReceiverAudioSubscriptionMessage.All -> true
        is ReceiverAudioSubscriptionMessage.None -> false
        else -> wantedSsrcs.contains(ssrc)
    }
```

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscription.kt
```diff
@@ -0,0 +1,88 @@
...
+                return wantedSsrcs.contains(ssrc)
+            }
+            is ReceiverAudioSubscriptionMessage.Exclude -> {
+                return wantedSsrcs.contains(ssrc)
+            }
+        }
+    }
+
+    fun onConferenceSourceAdded(descs: Set<AudioSourceDesc>) {
+        when (latestSubscription) {
```
bgrozev, line 61: You can avoid the explicit cast with something like:
```
latestSubscription.let { subscription ->
    if (subscription is Include) {
        subscription.list.... // subscription is automatically casted to Include
    }
}
```

bgrozev (approved):

PR #2352, opened 2025-07-22
Getting audio subscription and route-loudest-only work together

kota-yata:
This PR fixes the conflict between audio subscription management and the route-loudest-only configuration when handling incoming packets.

When `route-loudest-only=true` is set, the bridge drops a packet before checking endpoints `wants()` it if it's not among the top `numLoudest` loudest sources. However, if a participant explicitly subscribes to a source (using "include" mode), the expectation is that they want to receive that source regardless of its loudness in most use cases. This implies the subscription should always takes precedence over `route-loudest-only` in incoming packet handling.

To handle this conflict, this PR adds new set called `subscribedLocalAudioSources` in `Conference`, which holds local source names that are subscribed by either local or remote endpoints. The set is updated in following situations:
- When a local endpoint has new "Include" subscription
  - the endpoint adds sources in the subscription to `subscribedLocalAudioSources`
- When a source is removed from the conference
  - the endpoint simply removes the source from `subscribedLocalAudioSources`
- When the bridge receives remote subscription notification from other bridges
  - the conference adds the notified source to `subscribedLocalAudioSources`

To implement the last one, this PR will also add a messaging format or something to notify new remote subscription from a bridge to other bridge.

### DIFF
--- a/jvb/src/main/java/org/jitsi/videobridge/Conference.java
+++ b/jvb/src/main/java/org/jitsi/videobridge/Conference.java
@@ -170,6 +170,15 @@ public class Conference
     private final boolean routeLoudestOnly = LoudestConfig.getRouteLoudestOnly();
 
     /**
+     * A list of local audio source names that at least one endpoint subscribes to.
+     */
+    private Set<String> subscribedLocalAudioSources = new HashSet<>();
+
+    public Set<String> getSubscribedLocalAudioSources() {
+        return subscribedLocalAudioSources;
+    }
+
+    /**
      * The task of updating the ordered list of endpoints in the conference. It runs periodically in order to adapt to
      * endpoints stopping or starting to their video streams (which affects the order).
      */
@@ -1262,6 +1271,13 @@ public class Conference
         SpeakerRanking ranking = speechActivity.levelChanged(endpoint, level);
         if (ranking == null || !routeLoudestOnly)
             return false;
+        // return false if the source is explicitly subscribed by any other endpoint.
+        List<AudioSourceDesc> sources = endpoint.getAudioSources();
+        for (AudioSourceDesc source : sources) {
+            if (subscribedLocalAudioSources.contains(source.getSourceName())) {
+                return false;
+            }
+        }
         if (ranking.isDominant && LoudestConfig.Companion.getAlwaysRouteDominant())
             return false;
         if (ranking.energyRanking < LoudestConfig.Companion.getNumLoudest())
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscription.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscription.kt
@@ -19,7 +19,7 @@ package org.jitsi.videobridge
 import org.jitsi.videobridge.message.ReceiverAudioSubscriptionMessage
 import org.jitsi.videobridge.relay.AudioSourceDesc
 
-class AudioSubscription() {
+class AudioSubscription(private val conference: Conference) {
     private var latestSubscription: ReceiverAudioSubscriptionMessage = ReceiverAudioSubscriptionMessage.All
 
     // wantedSsrcs is a set of SSRCs that the endpoint wants to receive audio for.
@@ -32,9 +32,12 @@ class AudioSubscription() {
             is ReceiverAudioSubscriptionMessage.All -> return
             is ReceiverAudioSubscriptionMessage.None -> return
             is ReceiverAudioSubscriptionMessage.Include -> {
-                wantedSsrcs = sources.filter { desc ->
+                val (local, remote) = sources.partition { desc ->
                     subscription.list.contains(desc.sourceName)
-                }.map(AudioSourceDesc::ssrc).toSet()
+                }
+                wantedSsrcs = local.map(AudioSourceDesc::ssrc).toSet()
+                conference.subscribedLocalAudioSources.addAll(local.map(AudioSourceDesc::sourceName))
+                // TODO: Notify remote sources to other relays
             }
             is ReceiverAudioSubscriptionMessage.Exclude -> {
                 wantedSsrcs = sources.filterNot { desc ->
@@ -77,5 +80,8 @@ class AudioSubscription() {
 
     fun onConferenceSourceRemoved(descs: Set<AudioSourceDesc>) {
         wantedSsrcs = wantedSsrcs.subtract(descs.map(AudioSourceDesc::ssrc).toSet())
+        descs.forEach {
+            conference.subscribedLocalAudioSources.remove(it.sourceName)
+        }
     }
 }
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/Endpoint.kt
@@ -368,7 +368,7 @@ class Endpoint @JvmOverloads constructor(
     /**
      * Last updated ReceiverAudioSubscription
      */
-    private var audioSubscription: AudioSubscription = AudioSubscription()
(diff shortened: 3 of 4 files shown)


jvb/src/main/java/org/jitsi/videobridge/Conference.java
```diff
@@ -1262,6 +1271,13 @@ public boolean levelChanged(@NotNull AbstractEndpoint endpoint, long level)
         SpeakerRanking ranking = speechActivity.levelChanged(endpoint, level);
         if (ranking == null || !routeLoudestOnly)
             return false;
+        // return false if the source is explicitly subscribed by any other endpoint.
+        List<AudioSourceDesc> sources = endpoint.getAudioSources();
+        for (AudioSourceDesc source : sources) {
```
bgrozev, line 1276: This is one place that we want to optimize, as it runs very often -- 50 times per second per audio source, even when muted. The existing case (no endpoint using the new signaling) should not degrade.

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscription.kt
```diff
@@ -32,9 +32,12 @@ class AudioSubscription() {
             is ReceiverAudioSubscriptionMessage.All -> return
             is ReceiverAudioSubscriptionMessage.None -> return
             is ReceiverAudioSubscriptionMessage.Include -> {
-                wantedSsrcs = sources.filter { desc ->
+                val (local, remote) = sources.partition { desc ->
```
bgrozev, line 35: I'm confused. `local` are the sources from the conference, which the new subscription subscribes to? And `remote` are the sources that it doesn't subscribe to?

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscription.kt
```diff
@@ -77,5 +80,6 @@ class AudioSubscription() {
 
     fun onConferenceSourceRemoved(descs: Set<AudioSourceDesc>) {
         wantedSsrcs = wantedSsrcs.subtract(descs.map(AudioSourceDesc::ssrc).toSet())
+        conference.subscribedLocalAudioSources.removeAll(descs.map(AudioSourceDesc::sourceName))
```
bgrozev, line 83: What if another endpoint is subscribed to the same source? E.g. ep `A` subscribes to source `S`, then `B` subscribes to `S`, then `B` leaves. I think you'll need a richer structure to hold this.

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscription.kt
```diff
@@ -19,7 +19,7 @@ package org.jitsi.videobridge
 import org.jitsi.videobridge.message.ReceiverAudioSubscriptionMessage
 import org.jitsi.videobridge.relay.AudioSourceDesc
 
-class AudioSubscription() {
+class AudioSubscription(private val conference: Conference) {
```
bgrozev, line 22: I'm not a fan of tying the basic `AudioSubscription` class to an entire `Conference`. You can introduce a simple `interface` that `Conference` implements. But we can clean that up at a later stage, as the code might need to change for other reasons.

bgrozev (changes requested):

jvb/src/test/kotlin/org/jitsi/videobridge/AudioSubscriptionManagerTest.kt
```diff
@@ -0,0 +1,461 @@
...
+    private val manager = AudioSubscriptionManager(conference)
+
+    init {
+        context("Basic subscription management") {
+            should("handle single endpoint subscription") {
+                val subscription = ReceiverAudioSubscriptionMessage.Include(listOf("source1", "source2"))
+                manager.setEndpointAudioSubscription("endpoint1", subscription)
+
+                // Endpoint should want included sources
+                manager.isEndpointAudioWanted("endpoint1", 1001L) shouldBe true // source1
```
bgrozev, line 48: Any implications from the fact that an endpoint can subscribe to its own sources? Packets won't be routed back to it, but maybe something more subtle? I'm just wondering

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscriptionManager.kt
```diff
@@ -0,0 +1,124 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+class AudioSubscriptionManager(private val conference: Conference) {
+    /**
+     * A map of endpoint IDs to their audio subscriptions.
+     */
+    private val audioSubscriptions = ConcurrentHashMap<String, AudioSubscription>()
+
+    /**
+     * A map of local audio source names to a set of endpoint IDs that explicitly subscribe to the source.
```
bgrozev, line 30: Define what you mean by "explicitly" (with an "Include" subscription, not All or Except)

bgrozev (changes requested):

bgrozev:
I left some minor comments in-line. Two more things:
1. Some tests failed, can you try to figure out why (it's possible it's an unrelated false positive)? If you click on the "default" link you'll go to a page with reports and logs. I can help with that if it doesn't make sense

2. I'd like to think about the thread model. New endpoints/sources will come from colibri/xmpp and they are handled on a conference-level queue running in the IO pool. New subscriptions will come from EndpointMessageTransport, which runs on a endpoint-level queue in the IO pool. levelChanged runs on a receiver-level queue in the CPU pool. I'm not quite sure where isEndpointAudioWanted runs. We'll probably need to add some explicit synchronization.

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscriptionManager.kt
```diff
@@ -0,0 +1,124 @@
...
+        audioSubscriptions.values.forEach { subscription ->
+            subscription.onConferenceSourceAdded(sources)
+        }
+    }
+
+    /**
+     * Called when an endpoint is removed from the conference.
+     * @param endpoint the endpoint that was removed
+     */
+    fun onEndpointRemoved(endpoint: AbstractEndpoint) {
```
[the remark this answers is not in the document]
kota-yata: Changed to just pass the id and let its caller call removeSources separately

jvb/src/test/kotlin/org/jitsi/videobridge/AudioSubscriptionManagerTest.kt
kota-yata: If it's with Include type, there will be unnecessary element in the subscribedLocalSources. That's pretty much the only effect I could come up with. No big effect I assume

jvb/src/main/kotlin/org/jitsi/videobridge/AudioSubscriptionManager.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Checks if audio from a given SSRC is wanted by a specific endpoint.
+     * @param endpointId the ID of the endpoint
+     * @param ssrc the SSRC to check
+     * @return true if the audio is wanted, false otherwise
+     */
+    fun isEndpointAudioWanted(endpointId: String, ssrc: Long): Boolean {
+        val subscription = audioSubscriptions[endpointId]
+        return subscription?.isSsrcWanted(ssrc) ?: false
```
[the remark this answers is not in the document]
kota-yata: It should be safe enough. Possible erroneous behaviors here are a packet being sent to an endpoint that doesn't subscribe to it because the subscription arrives a bit late which is unavoidable anyway, or a packet being sent to an endpoint that doesn't even exist, which is less likely to happen.

PR #2354, opened 2025-08-13
Remove endpointId from AddReceiverMessage

kota-yata:
This PR removes unused endpointId argument from AddReceiverMessage


jvb/src/main/kotlin/org/jitsi/videobridge/message/BridgeChannelMessage.kt
```diff
@@ -401,8 +401,6 @@ class SenderSourceConstraintsMessage(
  */
 class AddReceiverMessage(
     val bridgeId: String,
-    // Used in single stream per endpoint mode and wil be removed
-    val endpointId: String?,
     // Used in the multi-stream mode
     val sourceName: String?,
```
bgrozev, line 405: Can you make this non-nullable, too? The comment can go away

bgrozev (approved):

Issue #1306, opened 2020-06-11
Read JVB_SECRET from debconf


hermanbanken:
https://github.com/jitsi/jitsi-videobridge/blob/17a3212a7bac54815285a664595a8893cd62f6a5/debian/postinst#L39

Afterwards this is done:
```
db_set jitsi-videobridge/jvbsecret $JVB_SECRET
```

If we would get the JVB_SECRET from db_get in the first place, the [scalable setup](https://jitsi.github.io/handbook/docs/devops-guide/devops-guide-scalable) would be simpler!

sanvila:
This is a duplicate of an issue which I reported here:

https://github.com/jitsi/jitsi-videobridge/issues/1304

(Just so that everybody knows I don't know if github issues allow to be merged)

damencho:
Fixed by https://github.com/jitsi/jitsi-videobridge/pull/2398

PR #2398, opened 2026-03-25
read JVB_SECRET from debconf if already set, issue  #1304 issue #1306

### LINKED ISSUE #1304: Preseeding with debconf has no effect
## Description

I'm trying to write an ansible role to install a secondary videobridge (i.e. on an instance where jitsi-meet is not going to be installed). To avoid questions, I use debconf preseeding, but it does not work as expected.

## Current behavior

The package chooses a random password every time I install it, even if I preseed the password before installing the package.

## Expected behavior

I expect the package to honor debconf variables when they are preseeded.

## Possible solution

Try to get the value of `jitsi-videobridge/jvbsecret` somewhere in postinst. I see a db_set
but not a db_get anywhere.

## Steps to reproduce

On a Debian 10 system with minimal packages installed I do this:

```
echo "jitsi-videobridge2	jitsi-videobridge/jvbsecret	password	MYPASSWORD" | debconf-set-selections
echo "jitsi-videobridge2	jitsi-videobridge/jvb-hostname	string	MYHOSTNAME" | debconf-set-selections
apt-get install jitsi-videobridge2
```
After doing that, files `config` and `sip-communicator.properties` contain a new password, not the one I preseeded with debconf.

This happens with jitsi-videobridge version 2.1-202-g5f9377b9-1 currently in the stable repository.
(Sorry, didn't try a later version).

Thanks.

jananivarthini455:
During first install, if a value for jitsi-videobridge/jvbsecret is preseeded in debconf, postinst now uses it instead of generating a random secret.
Fixes #1304, #1306, #2398

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -36,7 +36,12 @@ case "$1" in
             JVB_HOSTNAME="$JVB_HOSTNAME_IN"
             if [ "$OLDCONFIG" = "false" ]; then
                 # 8-chars random secret, alternative to pwgen 8
-                JVB_SECRET=`head -c 8 /dev/urandom | tr '\0-\377' 'a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9@@@@####'`
+                db_get jitsi-videobridge/jvbsecret
+                if [ -n "$RET" ]; then
+                    JVB_SECRET="$RET"
+                else
+                    JVB_SECRET=`head -c 8 /dev/urandom | tr '\0-\377' 'a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9@@@@####'`
+                fi
             fi
         fi
 


damencho:
> Before this fix, JVB always generated a new random secret on install/update, ignoring any value preseeded in debconf.

This is not correct. The secret is generated and set only the first time when jvb is installed and config files are non existent.

damencho:
Drop the unrelated changes, please.

jananivarthini455:
@damencho  Thanks for the review! I've fixed both issues — dropped the unrelated doc changes and updated the description to correctly reflect that the secret is only generated on first install. The PR now only touches debian/postinst. Please take another look!

damencho:
Just for the record, this can be used like this:
```
debconf-set-selections <<< "jitsi-videobridge2	jitsi-videobridge/jvbsecret	password my_secret_password"
```
I tested both upgrade and new install, all looks good.

damencho (approved):

PR #2401, opened 2026-03-26
Fix SysV init not working on Debian systems issue 1605

### LINKED ISSUE #1605: SysV init ignored by Debian postinst script
After the installation of jitsi-videobridge on a Debian system with SysV init, the service is not started and the init script links under `/etc/rc<runlevel>.d` are missing. 

Suggested solution: Create according `dh_installinit` logic for the Debian `postinst` script.

jananivarthini455:
Fixes #1605

The service was not starting and rc symlinks were missing
when installing on systems using SysV init instead of systemd.

Changes made:
- Added debian/jitsi-videobridge.init SysV init script
- Removed --noscripts from dh_installinit in debian/rules
- Added SysV stop logic in postinst alongside systemd logic

@damencho 
I don't have access to an older Debian system with SysV init
to test this locally. Could someone please test this on a 
Debian system running SysV init to verify the fix works 
correctly? Any feedback is welcome!


damencho:
What is the purpose of this? Many years ago we have dropped the init.d script in favor of the systemd one. Why do you eant to revert that?

jananivarthini455:
Thank you for the clarification! I wasn't aware that the init.d script had already been dropped in favor of systemd. That makes complete sense. I'll close this issue and look into more relevant areas to contribute. Apologies for the noise!

PR #2404, opened 2026-03-31
Allow other JDK implementations

eppesuig:
The requirement on openjdk-* is quite strict and don't allow usage of other Java implementations. I've just added the generic dependencies that would allow other Java. As an example, in Debian bullseye there is not openjdk 17 nor 21, but you may download amazon corretto debian package. This change allow to use any java implementation that provides a well crafted debian package

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Section: net
 Priority: extra
 Maintainer: Jitsi Team <dev@jitsi.org>
 Uploaders: Emil Ivov <emcho@jitsi.org>, Damian Minkov <damencho@jitsi.org>
-Build-Depends: debhelper (>= 9), dh-systemd, openjdk-17-jdk | openjdk-21-jdk, maven
+Build-Depends: debhelper (>= 9), dh-systemd, java17-sdk-headless | openjdk-17-jdk | openjdk-21-jdk, maven
 Standards-Version: 3.9.3
 Homepage: https://jitsi.org/videobridge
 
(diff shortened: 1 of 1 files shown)


damencho:
Debian bullseye is end of life already. The dependencies we have support all latest stable release of Ubuntu and Debian, we just changed them to match them.

eppesuig:
Well, debian bullseye is currently supported by LTS team and I don't know when that support will end. By the way, this patch would allow more freedom to Jitsi users, and will not add a single restriction or problem. So, why not allowing it?

damencho:
Cause it changes default behavior, change the order so it uses those packages as fallback.

eppesuig:
I did not add a specific package, I added the virtual package names. In fact openjdk-* packages provide that virtual package. If you have openjdk-* packages nothing changes. If you have any other package that provide that virtual names, then it works as well.

```
$ apt show openjdk-17-jdk-headless | grep ^Provides
Provides: java-compiler, java-sdk-headless (= 17), java10-sdk-headless, java11-sdk-headless, java12-sdk-headless, java13-sdk-headless, java14-sdk-headless, java15-sdk-headless, java16-sdk-headless, java17-sdk-headless, java2-sdk-headless, java5-sdk-headless, java6-sdk-headless, java7-sdk-headless, java8-sdk-headless, java9-sdk-headless
```

I added the virtual package in front of the other package names since I seem to remember this is the debian way, but I may be wrong. I am going to check it.

eppesuig:
I just found the documentation. If you suggest one specific implementation, than its package name should be listed prior to the virtual package name, otherwise there is no preference over the implementation packages.

The documentation states
```
To specify which of a set of real packages should be the default to satisfy a particular dependency on a virtual package, list the real package as an alternative before the virtual one.
```
source: https://www.debian.org/doc/debian-policy/ch-relationships.html#virtual-packages-provides

So, you may update my patch changing the package order as you prefer, albeit I hope any JDK implementation will gives the same result.

damencho:
We prefer to put there what we have tested. And no, we cannot guarantee same results if you use some other GC as for realtime it depends how the GC is behaving, so we have tested and adjusted the GC for openjdk that gives best results for high load usage.

https://github.com/jitsi/jitsi-videobridge/blob/4f55d380a76e0ea0ea0e0d7b5bcbecdea4769cce/jvb/resources/jvb.sh#L20

eppesuig:
Well, thank you very much for this explanation. I cannot use OpenJDK, so I'll carefully test the system with Corretto.

PR #2415, opened 2026-05-26
Fix Exporter reconnect backoff loop

bgrozev:
Reset `reconnectAttempts` only after a stable connection (open longer than
`stable-connection-threshold`, default 1 minute), not on every WebSocket open.

Previously the counter was reset in `onWebSocketOpen`, so any connection that
opened and immediately closed would always be treated as attempt 1 (0 ms delay),
causing an infinite tight reconnect loop with no backoff.

Also removes the `maxDelay` parameter from `scheduleReconnect` — it was a
workaround for the same bug (forcing max delay for 1011 errors to avoid the
0 ms loop). With the counter reset fixed, normal backoff handles all error codes.

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/export/Exporter.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/export/Exporter.kt
@@ -57,6 +57,7 @@ internal class Exporter(
     private val isShuttingDown = AtomicBoolean(false)
     private val reconnectAttempts = AtomicInteger(0)
     private var reconnectFuture: ScheduledFuture<*>? = null
+
     @Volatile private var connectionOpenedAtMs: Long = 0
 
     // Ping/pong state
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/export/Exporter.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/export/Exporter.kt
@@ -58,7 +58,7 @@ internal class Exporter(
     private val reconnectAttempts = AtomicInteger(0)
     private var reconnectFuture: ScheduledFuture<*>? = null
 
-    @Volatile private var connectionOpenedAtMs: Long = 0
+    private var connectionOpenedAtMs: Long = 0
 
     // Ping/pong state
     private var pingScheduledFuture: ScheduledFuture<*>? = null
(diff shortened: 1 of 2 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/export/Exporter.kt
```diff
@@ -58,6 +58,8 @@ internal class Exporter(
     private val reconnectAttempts = AtomicInteger(0)
     private var reconnectFuture: ScheduledFuture<*>? = null
 
+    @Volatile private var connectionOpenedAtMs: Long = 0
```
JonathanLennox: Why does this need to be volatile?

jvb/src/main/kotlin/org/jitsi/videobridge/export/Exporter.kt
```diff
@@ -90,6 +92,11 @@ internal class Exporter(
             get() = session != null
 
         override fun onWebSocketClose(statusCode: Int, reason: String?) {
+            val openedAt = connectionOpenedAtMs
+            val stableConnection = openedAt > 0 &&
+                System.currentTimeMillis() - openedAt > stableConnectionThreshold.toMillis()
+            if (stableConnection) reconnectAttempts.set(0)
```
JonathanLennox: This should probably reset `connectionOpenedAtMs`.

jvb/src/main/resources/reference.conf
```diff
@@ -428,5 +428,9 @@ videobridge {
     // and doubles at each step maxing at max-delay.
     base-delay = 1 second
     max-delay = 30 seconds
+    // How long a connection must stay open before it is considered stable. Only stable connections reset the
+    // reconnect attempt counter (and thus the backoff). Prevents tight reconnect loops when the remote closes
+    // the connection immediately after it opens.
+    stable-connection-threshold = 1 minute
```
JonathanLennox: A minute feels long to me, did you have any specific justification for this time, or did you just pick it arbitrarily?

bgrozev: Changed to 30s. I figured it shouldn't be less than max-delay, but I don't have a convincing reason.

JonathanLennox (approved):

PR #2418, opened 2026-06-01
Replace json-simple with Jackson

bgrozev:
Replace the `org.json.simple` dependency with Jackson.

- Migrate all `JSONObject`/`JSONArray` usages to `ObjectNode`/`ArrayNode` across `jvb`, `jitsi-media-transform`, and `rtp` modules
- Debug state methods now return `ObjectNode`
- Update `jitsi-utils` to `1.0-SNAPSHOT`, `jicoco` to `1.1-SNAPSHOT`, `jitsi-xmpp-extensions` to `1.0-SNAPSHOT`
- `Conferences.java` still uses `json-simple` transitively via `jitsi-xmpp-extensions` Colibri2 JSON API (will be removed once that dependency is updated)

### DIFF
--- a/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/MediaSources.kt
+++ b/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/MediaSources.kt
@@ -16,6 +16,7 @@
 
 package org.jitsi.nlj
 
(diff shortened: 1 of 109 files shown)


JonathanLennox (approved):
Approved pending dependency updates.

PR #2419, opened 2026-06-11
Live audio translation: colibri2 connect exports/requests, synthetic sources, inbound injection

JonathanLennox:
Implements live audio translation support in the bridge, on top of the colibri2 `<connect>` and `media-source` extensions added in jitsi-xmpp-extensions (`translator` connect type, `<exports>`/`<requests>`, `synthetic`).

### Connect handling
- Support **multiple `<connect>`s** simultaneously (e.g. a recorder and a translator).
- Connects are applied as **id-keyed deltas**: each `<connect>` names an `id` and either `create`s a new exporter, `expire`s a running one, or (with neither flag) updates the running exporter with that id; connects not mentioned are left running. Changes that can't be applied live (url, HTTP headers, ping, type, audio) are rejected with `feature_not_implemented`; a connect can be **replaced** (e.g. to change its URL) by expiring and re-creating its id in a single request (expires apply before creates). Any other repeated mention of an id in one request is rejected.
- Connect application is serialized against conference expiry, so a delta racing expiry can't start an exporter that would never be stopped.
- Each exporter is a `PotentialPacketHandler` folded into `Conference.sendOut`'s clone-minimizing loop (no more double-cloning through a wrapper).
- **Compatibility**: the `id`/`create`/`expire` attributes are mandatory (id-less `<connect>`s are rejected at parse time by jitsi-xmpp-extensions 1.0-114), so a signaler using the previous single-connect syntax (id-less `<connect>`, empty `<connects/>` as stop — e.g. the released jicofo transcriber path) cannot drive this bridge; the matching jicofo changes are required.

### Outbound (bridge -> translator)
- Read `<connect>`'s `exports`/`requests` source-name lists and the connect `type`.
- Send a media-json **`sources`** event declaring the exported/requested source names (on connect and on update).
- **Filter outbound audio by `exports`**: an empty/absent list means "export all" (current recorder/transcriber behavior); a non-empty list forwards only the named sources. In both modes, audio is exported only if its SSRC **resolves to a signaled source** — audio arriving ahead of its colibri signaling is not exported, mirroring receiver routing.
- Tag outbound media-json `start`/`media` events by **source name** (the same identifier used by `exports`/`requests`), so the protocol keys on source names end to end. There is no `epId-ssrc` fallback tag on the wire: a stream's tag is always its source name, resolved once at stream start.

### Synthetic sources
- Carry `MediaSource.synthetic` into `MediaSourceDesc`/`AudioSourceDesc` (via `MediaSourceFactory` / `Colibri2ConferenceHandler`).
- Synthetic audio sources are **not auto-routed** to clients (even under an "All"/"Exclude"/default subscription) -- only to endpoints that explicitly (Include) subscribed to them by name; they're **excluded from loudest-speaker**; and they're still **forwarded to relays**.
- Synthetic (bridge-generated) audio is **never exported to connects**, so translated audio doesn't feed back into the translator that produced it or into an export-all transcriber/recorder — including when it arrives over a relay.
- Audio **subscriptions now resolve relayed sources** too (previously only local), so this works across bridges (and fixes a pre-existing bug where `Exclude` subscribers received no relayed audio).

### Inbound (translator -> bridge)
- Translated audio returned as media-json `media` events is injected as its **synthetic source**: map the event tag to the synthetic `AudioSourceDesc`, build an Opus RTP packet on that source's SSRC (conference-negotiated Opus PT), and inject. The synthetic-source routing then delivers it to explicit subscribers, keeps it out of loudest-speaker, and forwards it to relays.
- `sendOut` no longer echo-suppresses synthetic sources, so an endpoint can subscribe to a synthetic source associated with itself -- on any bridge, whether the audio was injected locally or received over a relay.

### Dependencies
- Builds on jitsi/jicoco#235 (adds `SourcesEvent` and inbound `MediaEvent` handling to `jicoco-mediajson`), now merged and released; `jicoco.version` is pinned to `1.1-173-g1e2216b`.
- Builds on the released `jitsi-xmpp-extensions:1.0-114-g087129a` (adds the connect `id`/`create`/`expire` attributes).

### Notes / follow-ups
- The translator worker (opus-transcriber-proxy) needs corresponding changes to be driven by the `sources`/`requests` model (and to tag returned media by the requested source name) rather than the earlier spike's `start-translation` frames + `speaker~language` tags. Workers only ever see source-name tags now, so they don't need to handle the `epId-ssrc` form.
- Packet-level paths in the exporter/inbound code aren't unit-tested (heavy Conference/RTP/websocket coupling); the connect-reconcile and audio-subscription logic are covered.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JonathanLennox:
Depends on jitsi/jicoco#235 (adds `SourcesEvent` and inbound `MediaEvent` handling to `jicoco-mediajson`). This branch builds only against jicoco `1.1-SNAPSHOT` until that PR is merged and released and `jicoco.version` is bumped accordingly.

bgrozev (approved):

PR #2421, opened 2026-06-17
Stamp sdes:mid on rewritten media for mid-based demuxing

JonathanLennox:
## Summary

Under SSRC rewriting, optionally stamp the `sdes:mid` RTP header extension on packets forwarded to an endpoint, and signal the per-slot mid in the `Audio`/`VideoSourcesMap` messages, so the client can demux forwarded media by mid. This lets the client negotiate mid (disabling Chrome's payload-type demuxing fallback — the trigger for the silent audio-demux wedge) while the bridge honors the "negotiated ⇒ sent" assumption browsers make.

Part of the cross-repo **bridge-stamped mid** change. Companion `mid-demux-stamping` branches in **jitsi-xmpp-extensions** (the capability constant), **jicofo** (emits the capability), and **lib-jitsi-meet** (consumes the mid); meant to be tested together.

## Details

- Model the `sdes:mid` extension (`RtpExtensionType.MID`); the extension id is already signaled by jicofo via colibri2.
- `MidStamper` egress node resolves the mid from the packet's rewritten SSRC and adds the extension. It sits where the RTX pipeline rejoins the main pipeline, so primary and RTX packets are stamped with the same mid.
- `SsrcCache` assigns a short, stable mid per slot (`a`/`v` + index), bound to the slot's send SSRCs rather than the source, so it is stable across remaps (a remap is metadata-only on the client).
- Gated on the new `rtp-mid-demux` endpoint capability (in addition to SSRC rewriting); no behavior change without it. (The capability is read by string literal pending a jitsi-xmpp-extensions release containing `Capability.CAP_RTP_MID_DEMUX_SUPPORT`.)

## Tests

`MidStamperTest` (3 cases) and `SsrcCacheTest` per-slot-mid cases (assignment, stability across remap, disabled⇒no mid). `mvn test-compile`, ktlint, and the affected tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jitsi-media-transform/src/test/kotlin/org/jitsi/nlj/transform/node/outgoing/MidStamperTest.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.jitsi.nlj.util.StreamInformationStoreImpl
+import org.jitsi.rtp.rtp.RtpPacket
+import org.jitsi.rtp.rtp.header_extensions.SdesHeaderExtension
+
+class MidStamperTest : ShouldSpec() {
+    override fun isolationMode(): IsolationMode = IsolationMode.InstancePerLeaf
+
+    private val midExtId = 10
+    private val ssrcWithMid = 1234L
+    private val mid = "v3"
```
bgrozev: Can we add tests with a longer string?

bgrozev (approved):

JonathanLennox:
Jenkins, test this please

PR #2428, opened 2026-07-09
ref: Bound per-SSRC tracking maps with an LRU

bgrozev:
Several per-SSRC tracking structures on the receive path grew without bound as new SSRCs were seen, with no cap or eviction. This bounds each with an LRU (1024 entries, access-order):

- `IncomingStatisticsTracker.ssrcStats`
- `DuplicateTermination.replayContexts`
- `RetransmissionRequester.streamPacketRequesters` (evicted requesters are stopped to cancel their scheduled tasks)
- `AudioRedHandler.ssrcRedHandlers` (evicted handlers are stopped to release cached buffers)
- `DiscardableDiscarder.rewriters`
- `RtcpRrGenerator.senderInfos`

The 1024 cap is a backstop: legitimate traffic keeps these maps small, so the limit only bounds pathological growth. It is deliberately high — well above the number of SSRCs a receiver tracks even in large or relayed conferences — so active SSRCs are never evicted under normal operation.

Thread-safety is preserved (`synchronizedMap` where the map was previously concurrent; existing external synchronization elsewhere).

Adds `PerSsrcMapCapTest` covering each bounded map.

### DIFF
--- a/jitsi-media-transform/spotbugs-exclude.xml
+++ b/jitsi-media-transform/spotbugs-exclude.xml
@@ -42,6 +42,9 @@
     <Bug pattern="BC_IMPOSSIBLE_INSTANCEOF"/>
     <!-- False positives with Kotlin inlines -->
     <Bug pattern="RV_RETURN_VALUE_IGNORED_NO_SIDE_EFFECT"/>
+    <!-- False positives for the synthetic bridge methods Kotlin generates on anonymous
+         subclasses of generic collections (e.g. an anonymous LRUCache overriding removeEldestEntry). -->
+    <Bug pattern="UMAC_UNCALLABLE_METHOD_OF_ANONYMOUS_CLASS"/>
 
     <!-- False positives for an unknown reason -->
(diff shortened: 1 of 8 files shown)


JonathanLennox (approved):

jitsi-media-transform/spotbugs-exclude.xml
```diff
@@ -50,6 +50,9 @@
     <Bug pattern="UC_USELESS_OBJECT"/>
     <Bug pattern="AT_STALE_THREAD_WRITE_OF_PRIMITIVE"/>
     <Bug pattern="MS_EXPOSE_REP"/>
+    <!-- False positives for the synthetic bridge methods Kotlin generates on anonymous
```
JonathanLennox: It's pretty trivial but since this bypass has a known reason, it should be above the unknown reasons in spotbugs-exclude.xml

JonathanLennox (approved):

PR #2438, opened 2026-08-06
feat: Support in-place ICE restart with a new Agent

bgrozev:
Implements the bridge side of an in-place ICE restart. When a client's network changes, the current recovery is a full session teardown and re-invite. This adds a lighter path that keeps the same DTLS session, SSRCs and RTP state, so media keeps flowing throughout.

On an `ice-restart` request from jicofo the bridge creates a **new** ICE agent with fresh credentials and returns its transport, while the existing agent keeps carrying media. Only once the new agent connects does the bridge cut over, freeing the old one after a transition window so late checks from the old generation are still answered. If the new agent does not connect, the restart is abandoned and the established agent is kept, so a failed restart never breaks a working transport.

### How it works

- `IceTransport` is restructured around an `AgentBundle` (agent + stream + component + listeners + generation). `currentBundle` is what we send on and describe; `pendingBundle` is a restart in flight.
- `requestIceRestart()` creates the pending bundle but deliberately does **not** start connectivity checks. The bridge is the controlling agent, and the checks it sends are authenticated with the peer's password, so it waits for the peer's new credentials first. Checks arriving from the peer in the meantime are queued by ice4j and replayed when establishment starts.
- `applyRemoteCredentialsToPendingRestart()` runs when a transport update arrives with a pending bundle: it matches the `ice-generation`, applies the credentials, and starts checks.
- On the pending bundle reaching COMPLETED, `cutOver()` swaps it in and frees the old bundle after `videobridge.ice.restart.transition-window`.
- On failure or timeout, `abandonPendingRestart()` frees the pending bundle and keeps the established one.

The colibri WebSocket password is pinned to the initial agent (first commit). The ICE password doubles as the WebSocket auth token baked into the URL the client holds, and the client re-dials that URL on exactly the network change that triggers a restart, so it must not rotate with the ICE credentials.

### Config

- `videobridge.ice.restart.enabled` (default true)
- `videobridge.ice.restart.transition-window` (default 2s)
- `videobridge.ice.restart.timeout` (default 10s)

Metrics: `ice_restarts_started`, `_succeeded`, `_failed`, `_rejected`.

### Notes

No ice4j change is needed: this creates a fresh agent rather than re-arming an existing one.

A rejected restart is soft — it logs, bumps `ice_restarts_rejected` and returns no `<transport>`, rather than failing the whole conference-modify and taking other endpoints' updates with it.

Verified end-to-end against a real client: cut-over 656ms after the request, with outbound video non-zero in every 100ms sample across the cut-over, zero freezes and zero dropped frames.

Part of a multi-repo change; all PRs share the `in-place-ice-restart` branch name so CI tests them together:

- jitsi-xmpp-extensions: https://github.com/jitsi/jitsi-xmpp-extensions/pull/148
- jicofo: https://github.com/jitsi/jicofo/pull/1299
- lib-jitsi-meet: https://github.com/jitsi/lib-jitsi-meet/pull/3083
- jitsi-meet: https://github.com/jitsi/jitsi-meet/pull/17700

### DIFF
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
@@ -315,10 +315,16 @@ class Colibri2ConferenceHandler(
             null
         }
 
-        // A transport is signaled back for a restart that started (the new Agent's rotated credentials) and for
-        // one that kept the existing Agent (its unchanged credentials, so the endpoint keeps the connection it
-        // has). Only IceRestartResult.UNAVAILABLE signals nothing: an answer with no <transport> is how the
-        // endpoint learns that this bridge will not restart ICE and that it needs a full re-invite instead.
+        // How the refusal of a restart is signaled: with the *absence* of a <transport> in the
+        // conference-modified for this endpoint. There is no explicit "refused" flag, and the request is not
+        // failed with an error, because an error would fail the whole conference-modify and take every other
+        // endpoint's updates with it. Jicofo pairs a request it sent with the answer it gets back: a
+        // <transport> means the restart happened and is relayed to the endpoint, no <transport> means it did
+        // not and jicofo falls back to a re-invite.
+        //
+        // So a transport is signaled back for a restart that started (the new Agent's rotated credentials) and
+        // for one that kept the existing Agent (its unchanged credentials, so the endpoint keeps the connection
+        // it has, with no re-invite). Only IceRestartResult.UNAVAILABLE signals nothing.
         if (c2endpoint.create || iceRestartResult == IceRestartResult.STARTED ||
             iceRestartResult == IceRestartResult.KEEP_EXISTING
         ) {
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
@@ -259,40 +259,6 @@ class Colibri2ConferenceHandler(
 
         c2endpoint.transport?.iceUdpTransport?.let { endpoint.setTransportInfo(it) }
 
-        // An explicit ICE restart request. Handled after setTransportInfo, so any credentials included in this
-        // same request still belong to (and are applied to) the pre-restart Agent. The restart rotates our own
-        // ICE credentials, so the endpoint has to be told the new ones — its connectivity checks are addressed
-        // to them — even though this is not a create.
-        val iceRestartResult = if (c2endpoint.transport?.iceRestart == true) {
-            endpoint.requestIceRestart().also {
-                if (it != IceRestartResult.STARTED) {
-                    logger.warn("Did not restart ICE for endpoint ${c2endpoint.id}: $it.")
-                }
-            }
-        } else {
-            null
-        }
-
-        // A transport is signaled back for a restart that started (the new Agent's rotated credentials) and for
-        // one that kept the existing Agent (its unchanged credentials, so the endpoint keeps the connection it
-        // has). Only IceRestartResult.UNAVAILABLE signals nothing: an answer with no <transport> is how the
-        // endpoint learns that this bridge will not restart ICE and that it needs a full re-invite instead.
-        if (c2endpoint.create || iceRestartResult == IceRestartResult.STARTED ||
-            iceRestartResult == IceRestartResult.KEEP_EXISTING
-        ) {
-            val transBuilder = Transport.getBuilder()
-            transBuilder.setIceUdpExtension(endpoint.describeTransport())
-            if (c2endpoint.transport?.sctp != null) {
-                transBuilder.setSctp(
-                    Sctp.Builder()
-                        .setPort(DcSctpTransport.DEFAULT_SCTP_PORT)
-                        .setRole(Sctp.Role.SERVER)
-                        .build()
-                )
-            }
-            respBuilder.setTransport(transBuilder.build())
-        }
-
         c2endpoint.sources?.let { sources ->
             if (endpoint.visitor && sources.mediaSources.isNotEmpty()) {
                 throw IqProcessingException(
@@ -329,6 +295,44 @@ class Colibri2ConferenceHandler(
 
         c2endpoint.forceMute?.let {
             endpoint.updateForceMute(it.audio, it.video)
+        }
+
+        // An explicit ICE restart request. Handled after setTransportInfo, so any credentials included in this
+        // same request still belong to (and are applied to) the pre-restart Agent. The restart rotates our own
+        // ICE credentials, so the endpoint has to be told the new ones — its connectivity checks are addressed
+        // to them — even though this is not a create.
+        //
+        // Handled last, once nothing else in this request can throw: a restart is armed as a side effect (it
+        // rotates the credentials we advertise and starts the timeout), so failing the request after arming it
+        // would leave the bridge waiting out that timeout for an answer the endpoint was never asked for.
+        val iceRestartResult = if (c2endpoint.transport?.iceRestart == true) {
+            endpoint.requestIceRestart().also {
+                if (it != IceRestartResult.STARTED) {
+                    logger.warn("Did not restart ICE for endpoint ${c2endpoint.id}: $it.")
+                }
+            }
+        } else {
+            null
+        }
+
+        // A transport is signaled back for a restart that started (the new Agent's rotated credentials) and for
+        // one that kept the existing Agent (its unchanged credentials, so the endpoint keeps the connection it
+        // has). Only IceRestartResult.UNAVAILABLE signals nothing: an answer with no <transport> is how the
+        // endpoint learns that this bridge will not restart ICE and that it needs a full re-invite instead.
+        if (c2endpoint.create || iceRestartResult == IceRestartResult.STARTED ||
+            iceRestartResult == IceRestartResult.KEEP_EXISTING
+        ) {
+            val transBuilder = Transport.getBuilder()
+            transBuilder.setIceUdpExtension(endpoint.describeTransport())
+            if (c2endpoint.transport?.sctp != null) {
+                transBuilder.setSctp(
+                    Sctp.Builder()
+                        .setPort(DcSctpTransport.DEFAULT_SCTP_PORT)
+                        .setRole(Sctp.Role.SERVER)
+                        .build()
+                )
+            }
+            respBuilder.setTransport(transBuilder.build())
         }
 
         return respBuilder.build()
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
@@ -257,7 +257,22 @@ class Colibri2ConferenceHandler(
         }
 
         c2endpoint.transport?.iceUdpTransport?.let { endpoint.setTransportInfo(it) }
-        if (c2endpoint.create) {
+
+        // An explicit ICE restart request. Handled after setTransportInfo, so any credentials included in this
+        // same request still belong to (and are applied to) the pre-restart Agent. The restart rotates our own
+        // ICE credentials, so the endpoint has to be told the new ones — its connectivity checks are addressed
+        // to them — even though this is not a create.
+        val iceRestarted = if (c2endpoint.transport?.iceRestart == true) {
+            endpoint.requestIceRestart().also {
+                if (!it) {
+                    logger.warn("Rejected an ICE restart request for endpoint ${c2endpoint.id}.")
+                }
+            }
+        } else {
+            false
+        }
+
+        if (c2endpoint.create || iceRestarted) {
             val transBuilder = Transport.getBuilder()
             transBuilder.setIceUdpExtension(endpoint.describeTransport())
             if (c2endpoint.transport?.sctp != null) {
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
@@ -31,6 +31,7 @@ import org.jitsi.videobridge.relay.AudioSourceDesc
 import org.jitsi.videobridge.relay.Relay
 import org.jitsi.videobridge.relay.RelayConfig
 import org.jitsi.videobridge.sctp.SctpConfig
+import org.jitsi.videobridge.transport.ice.IceRestartResult
 import org.jitsi.videobridge.util.PayloadTypeUtil.Companion.create
 import org.jitsi.videobridge.websocket.config.WebsocketServiceConfig
 import org.jitsi.videobridge.xmpp.MediaSourceFactory
@@ -262,17 +263,23 @@ class Colibri2ConferenceHandler(
         // same request still belong to (and are applied to) the pre-restart Agent. The restart rotates our own
         // ICE credentials, so the endpoint has to be told the new ones — its connectivity checks are addressed
         // to them — even though this is not a create.
-        val iceRestarted = if (c2endpoint.transport?.iceRestart == true) {
+        val iceRestartResult = if (c2endpoint.transport?.iceRestart == true) {
             endpoint.requestIceRestart().also {
-                if (!it) {
-                    logger.warn("Rejected an ICE restart request for endpoint ${c2endpoint.id}.")
+                if (it != IceRestartResult.STARTED) {
+                    logger.warn("Did not restart ICE for endpoint ${c2endpoint.id}: $it.")
                 }
             }
         } else {
-            false
+            null
         }
 
-        if (c2endpoint.create || iceRestarted) {
+        // A transport is signaled back for a restart that started (the new Agent's rotated credentials) and for
+        // one that kept the existing Agent (its unchanged credentials, so the endpoint keeps the connection it
+        // has). Only IceRestartResult.UNAVAILABLE signals nothing: an answer with no <transport> is how the
+        // endpoint learns that this bridge will not restart ICE and that it needs a full re-invite instead.
+        if (c2endpoint.create || iceRestartResult == IceRestartResult.STARTED ||
+            iceRestartResult == IceRestartResult.KEEP_EXISTING
+        ) {
             val transBuilder = Transport.getBuilder()
             transBuilder.setIceUdpExtension(endpoint.describeTransport())
             if (c2endpoint.transport?.sctp != null) {
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
@@ -91,7 +91,8 @@ class IceConfig private constructor() {
 
     /**
      * How long to wait for the new [org.ice4j.ice.Agent] of an ICE restart to connect before giving up on the
-     * restart and keeping the existing Agent.
+     * restart and keeping the existing Agent. A value that is not positive disables ICE restarts: the new Agent
+     * would be freed before the endpoint could answer it.
      */
     val restartTimeout: Duration by config(
         "videobridge.ice.restart.timeout".from(JitsiConfig.newConfig)
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
@@ -82,8 +82,9 @@ class IceConfig private constructor() {
 
     /**
      * How long the old [org.ice4j.ice.Agent] is kept alive after an ICE restart has cut over to the new one.
-     * Both Agents accept incoming connectivity checks during this window, so checks still in flight from the
-     * old generation are answered rather than dropped.
+     * Both Agents keep their sockets during this window, and ice4j routes each packet by the address it came
+     * from, so the endpoint's old-generation checks — which come from its old address — are answered by the old
+     * Agent rather than dropped.
      */
     val restartTransitionWindow: Duration by config(
         "videobridge.ice.restart.transition-window".from(JitsiConfig.newConfig)
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
@@ -22,6 +22,7 @@ import org.jitsi.config.JitsiConfig
 import org.jitsi.metaconfig.config
 import org.jitsi.metaconfig.from
 import org.jitsi.metaconfig.optionalconfig
+import java.time.Duration
 
 class IceConfig private constructor() {
     /**
@@ -68,6 +69,32 @@ class IceConfig private constructor() {
      */
     val advertisePrivateCandidates: Boolean by config(
         "videobridge.ice.advertise-private-candidates".from(JitsiConfig.newConfig)
+    )
+
+    /**
+     * Whether ICE restarts are enabled: when an endpoint explicitly requests one (colibri2
+     * `<transport ice-restart="true"/>`), create a second [org.ice4j.ice.Agent] with rotated local credentials
+     * and run it alongside the established one, instead of rejecting the request.
+     */
+    val restartEnabled: Boolean by config(
+        "videobridge.ice.restart.enabled".from(JitsiConfig.newConfig)
+    )
+
+    /**
+     * How long the old [org.ice4j.ice.Agent] is kept alive after an ICE restart has cut over to the new one.
+     * Both Agents accept incoming connectivity checks during this window, so checks still in flight from the
+     * old generation are answered rather than dropped.
+     */
+    val restartTransitionWindow: Duration by config(
+        "videobridge.ice.restart.transition-window".from(JitsiConfig.newConfig)
+    )
+
+    /**
+     * How long to wait for the new [org.ice4j.ice.Agent] of an ICE restart to connect before giving up on the
+     * restart and keeping the existing Agent.
+     */
+    val restartTimeout: Duration by config(
+        "videobridge.ice.restart.timeout".from(JitsiConfig.newConfig)
     )
 
     companion object {
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
@@ -59,6 +59,27 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.LongAdder
 
+/**
+ * The outcome of an ICE restart request, which decides what the caller signals back to the peer.
+ */
+enum class IceRestartResult {
+    /** A new Agent was created. Signal its transport, so the peer moves its checks to it. */
+    STARTED,
+
+    /**
+     * No new Agent was created, but the established one is still usable. Signal its transport unchanged: the
+     * peer keeps the connection it has, instead of escalating to a full re-invite over a transport that is
+     * merely still connecting.
+     */
+    KEEP_EXISTING,
+
+    /**
+     * This bridge can not restart ICE. Signal no transport at all, which is how the peer learns to fall back to
+     * a full re-invite.
+     */
+    UNAVAILABLE
+}
+
 class IceTransport @JvmOverloads constructor(
     id: String,
     /**
@@ -332,33 +353,35 @@ class IceTransport @JvmOverloads constructor(
      * (make-before-break); we only [cutOver] once the new Agent connects, and keep the old one alive for
      * [IceConfig.restartTransitionWindow] after that so old-generation checks still in flight are answered.
      *
-     * @return true if a restart was started, in which case the caller must signal our new transport (rotated
-     * credentials plus the new `ice-generation`) back to the peer. False if the request was rejected, in which
-     * case nothing changed and the established Agent keeps running.
+     * @return [IceRestartResult.STARTED] if a restart was started, in which case the caller must signal our new
+     * transport (rotated credentials plus the new `ice-generation`) back to the peer.
+     * [IceRestartResult.KEEP_EXISTING] if no restart was started but the established transport is still usable,
+     * in which case the caller must signal that transport unchanged. [IceRestartResult.UNAVAILABLE] if this
+     * bridge can not restart ICE at all, in which case the caller must signal no transport so that the peer
+     * falls back to a full re-invite.
      */
-    fun requestIceRestart(): Boolean {
+    fun requestIceRestart(): IceRestartResult {
         if (!running.get()) {
-            logger.warn("Rejecting ICE restart request: the transport is not running.")
+            logger.warn("Can not restart ICE: the transport is not running.")
             iceRestartsRejected.inc()
-            return false
+            return IceRestartResult.UNAVAILABLE
         }
         if (!IceConfig.config.restartEnabled) {
             logger.warn(
-                "Rejecting ICE restart request: ICE restarts are disabled " +
-                    "(videobridge.ice.restart.enabled=false)."
+                "Can not restart ICE: ICE restarts are disabled (videobridge.ice.restart.enabled=false)."
             )
             iceRestartsRejected.inc()
-            return false
+            return IceRestartResult.UNAVAILABLE
         }
         if (!currentBundle.agent.state.isEstablished) {
             // There is no established connectivity to preserve, so there is nothing for a make-before-break
             // restart to do. The initial Agent is still gathering/checking and the peer should keep using it.
             logger.warn(
-                "Rejecting ICE restart request: the transport is not established yet " +
-                    "(state=${currentBundle.agent.state})."
+                "Not restarting ICE: the transport is not established yet " +
+                    "(state=${currentBundle.agent.state}). Keeping the existing Agent."
             )
             iceRestartsRejected.inc()
-            return false
+            return IceRestartResult.KEEP_EXISTING
         }
 
         val newBundle: AgentBundle
@@ -390,7 +413,7 @@ class IceTransport @JvmOverloads constructor(
             TimeUnit.MILLISECONDS
         )
 
-        return true
+        return IceRestartResult.STARTED
     }
 
     /**
@@ -836,12 +859,13 @@ class IceTransport @JvmOverloads constructor(
         )
 
         /**
-         * The total number of ICE restart requests rejected before a new Agent was created: the feature is
-         * disabled, or the transport is not running.
+         * The total number of ICE restart requests for which no new Agent was created: the feature is
+         * disabled, the transport is not running, or it is not established yet.
          */
         val iceRestartsRejected = VideobridgeMetricsContainer.instance.registerCounter(
             "ice_restarts_rejected",
-            "Number of ICE restart requests rejected (the feature is disabled, or the transport is stopped)."
+            "Number of ICE restart requests for which no new Agent was created (the feature is disabled, or " +
+                "the transport is stopped or not established)."
         )
 
         /**
--- a/jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
+++ b/jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
@@ -412,6 +412,17 @@ class IceTransport @JvmOverloads constructor(
         val newBundle: AgentBundle
         val generation: Int
         synchronized(restartLock) {
+            // A repeated request for a restart that has not started its checks yet (the peer has not answered
+            // with its own credentials) is answered with the bundle we already have. Clients do fire duplicate
+            // network-change events, and rolling a new generation here would free a bundle the peer may already
+            // be checking against, discard its progress and restart the timeout.
+            pendingBundle?.let { pending ->
+                if (!pending.checksStarted.get()) {
+                    logger.info("An ICE restart is already pending and has not started checks: $pending.")
+                    return IceRestartResult.STARTED
+                }
+            }
+
             // Create the new Agent before touching any state, so that a failure leaves the restart in flight
             // (if there is one) alone rather than freeing a bundle the peer may already be checking against.
             generation = restartGeneration + 1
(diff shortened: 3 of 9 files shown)


jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
```diff
@@ -257,7 +257,22 @@ class Colibri2ConferenceHandler(
...
-        if (c2endpoint.create) {
+
+        // An explicit ICE restart request. Handled after setTransportInfo, so any credentials included in this
+        // same request still belong to (and are applied to) the pre-restart Agent. The restart rotates our own
+        // ICE credentials, so the endpoint has to be told the new ones — its connectivity checks are addressed
+        // to them — even though this is not a create.
+        val iceRestarted = if (c2endpoint.transport?.iceRestart == true) {
+            endpoint.requestIceRestart().also {
+                if (!it) {
+                    logger.warn("Rejected an ICE restart request for endpoint ${c2endpoint.id}.")
```
JonathanLennox: Document (at least in comments) how the peer knows that the iceRestart was refused.  This is described in jicofo but should be described here too.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -123,48 +126,121 @@ class IceTransport @JvmOverloads constructor(
...
-        performConsentFreshness = true
-        nominationStrategy = IceConfig.config.nominationStrategy
-        addStateChangeListener(iceStateChangeListener)
-    }.also {
-        logger.addContext("local_ufrag", it.localUfrag)
-    }
+    /**
+     * Whether the ice4j [Agent]s created by this transport take the 'controlling' role.
+     */
+    private val controlling = controlling
```
JonathanLennox, line 171: In general it's possible for the controlling role to differ across different ICE restarts.  I think we've structured things so that's not the case but would be a good idea to leave this flexible?  How hard would it be?

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -123,48 +126,121 @@ class IceTransport @JvmOverloads constructor(
...
-     * whenever the WebSocket reconnects, so this value must stay fixed for the lifetime of the transport. Hence
-     * it is pinned here rather than read from the [Agent] on each access.
+     * This is deliberately pinned to the *initial* Agent's password and is NOT rotated by an ICE restart:
+     * the client is handed a WebSocket URL containing this password once (when its transport is described)
+     * and re-dials that same URL when the WebSocket reconnects — which is precisely what happens on the
+     * network change that triggers a restart. Rotating it would invalidate the URL the client already holds.
+     * The ICE-level password (advertised in [describe] and used for connectivity checks) does rotate; these
+     * are separate concerns that happened to share a field before ICE restarts existed.
+     */
+    val webSocketPassword: String = currentBundle.agent.localPassword
```
JonathanLennox: I like splitting off websocketPassword but I'm not sure it should still live in iceTransport

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -214,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+                    "(videobridge.ice.restart.enabled=false)."
+            )
+            iceRestartsRejected.inc()
+            return false
+        }
+        if (!currentBundle.agent.state.isEstablished) {
+            // There is no established connectivity to preserve, so there is nothing for a make-before-break
+            // restart to do. The initial Agent is still gathering/checking and the peer should keep using it.
+            logger.warn(
+                "Rejecting ICE restart request: the transport is not established yet " +
```
JonathanLennox: I think this will then trigger a reconnect in jicofo, which seems like not what we want in this case?

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -214,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+        pending.agent.startConnectivityEstablishment()
+    }
+
+    /**
+     * Switches [currentBundle] over to [newBundle], which has just connected, and schedules the old bundle to be
+     * freed after [IceConfig.restartTransitionWindow]. Both bundles accept incoming connectivity checks during
+     * that window (they have different local ufrags, so ice4j's single-port demultiplexing routes each
+     * generation's checks to the right Agent), which is what keeps the handover from dropping checks that were
+     * already in flight.
+     */
```
JonathanLennox, line 615: Maybe comment that both bundles accept incoming media, too?

jvb/src/main/resources/reference.conf
```diff
@@ -315,6 +315,25 @@ videobridge {
...
+
+        # How long the old Agent is kept alive after we cut over to the new one. During this window both
+        # Agents accept incoming connectivity checks, so checks still in flight from the old generation
+        # are answered rather than dropped. Set to 0 to free the old Agent immediately on cutover.
+        transition-window = 2 seconds
+
+        # How long to wait for a new Agent to connect (counted from when it was created, which includes
+        # waiting for the endpoint's own new credentials to arrive) before giving up on the restart and
+        # keeping the existing Agent.
+        timeout = 10 seconds
```
JonathanLennox: These two numbers feel short to me.  Did you derive them based on something specific?

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+                "$remoteCandidateCount, local ufrag=${pending.agent.localUfrag}), starting connectivity checks."
+        )
+        pending.agent.startConnectivityEstablishment()
+    }
+
+    /**
+     * Switches [currentBundle] over to [newBundle], which has just connected, and schedules the old bundle to be
+     * freed after [IceConfig.restartTransitionWindow]. Both bundles accept incoming connectivity checks during
+     * that window (they have different local ufrags, so ice4j's single-port demultiplexing routes each
+     * generation's checks to the right Agent), which is what keeps the handover from dropping checks that were
```
JonathanLennox: **The stated demultiplexing mechanism isn't what ice4j does, and the restart silently depends on the peer's source address changing.**

`AbstractUdpListener.runInHarvesterThreadPush()` demultiplexes on the **remote address alone**:

```java
destinationSocket = sockets.get(remoteAddress);
if (destinationSocket == null) { /* only now parse USERNAME -> ufrag */ }
```

The ufrag is consulted *only* for packets from an address not already in `sockets`, and the entry is removed only when that `MySocket` closes. `SinglePortUdpHarvester.harvest()` spells it out: "we use the local ufrag from an incoming STUN packet to setup de-multiplexing **based on remote transport address**."

So:

* If the peer's post-restart 5-tuple as seen by the bridge is unchanged, its new-generation checks are delivered to the **old** Agent's socket, which drops them (`ConnectivityCheckServer.processRequest` returns early on a local-ufrag mismatch) rather than answering. The pending Agent has no signalled remote candidates and depends entirely on peer-reflexive discovery from those checks, so it starves and the restart is abandoned. Worse than a clean failure: the peer has already committed to the new generation, so it's left with a dead session rather than the "restart failed, keep the old Agent" outcome the PR promises.
* Since endpoints use `useUniquePort=false`, the new Agent advertises the *same* IP:port as the old one. Nothing in the transport we return asks the peer for a new local port -- the restart works only because `restartIce()` happens to re-gather. That's an undocumented load-bearing assumption, not a property of the design, and it's probably why the end-to-end test passed.
* The transition window does work, but by a different mechanism than described: late old-generation checks arrive from the *old* address, which is still mapped to the old socket.

Minimum fix is to correct this comment and state the assumption, because as written it invites the reader to conclude the same-address case is covered. To make it robust the pending Agent needs the peer's address -- e.g. seeded as a remote candidate from the old Agent's selected pair -- plus something to hand the mapping over between agents.

jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
```diff
@@ -70,6 +71,32 @@ class IceConfig private constructor() {
...
+     * `<transport ice-restart="true"/>`), create a second [org.ice4j.ice.Agent] with rotated local credentials
+     * and run it alongside the established one, instead of rejecting the request.
+     */
+    val restartEnabled: Boolean by config(
+        "videobridge.ice.restart.enabled".from(JitsiConfig.newConfig)
+    )
+
+    /**
+     * How long the old [org.ice4j.ice.Agent] is kept alive after an ICE restart has cut over to the new one.
+     * Both Agents accept incoming connectivity checks during this window, so checks still in flight from the
```
JonathanLennox: Same as the note on `cutOver`: the two Agents don't accept each other's checks by ufrag. ice4j's single-port harvester demultiplexes by remote address, so what actually makes this window useful is that late old-generation checks arrive from the *old* address, which is still mapped to the old Agent's socket. Worth rewording so the mechanism isn't misdescribed in two places.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+            generation != pending.generation
+        ) {
+            logger.warn(
+                "Ignoring remote credentials for ICE generation $generation, the pending ICE restart is " +
+                    "generation ${pending.generation}."
+            )
+            iceRestartsRejected.inc()
+            return
+        }
+        if (generation == IceUdpTransportPacketExtension.GENERATION_UNSPECIFIED) {
```
JonathanLennox: **Guessing here can poison the restart round irrecoverably.**

Combined with the one-shot `checksStarted` CAS just below, a single `<transport ufrag=... pwd=...>` carrying no `ice-generation` -- an old-generation trickle `transport-info` still in flight, an older client, jigasi/jibri -- will:

1. apply the *old* remote password to the pending Agent,
2. start its checks, which then all fail MESSAGE-INTEGRITY at the peer,
3. and set `checksStarted`, so when the peer's real generation-tagged credentials arrive they are dropped with "Already started connectivity checks...".

The round then cannot recover; it burns the full timeout. This looks reachable: jicofo routes the peer's credentials through the existing `onTransportInfo` -> `updateTransport` path, which is the same path ordinary trickle candidates take, and lib-jitsi-meet only stamps `ice-generation` on the restart answer itself.

Two independent fixes, and I'd suggest both: don't guess on `GENERATION_UNSPECIFIED` while a restart is pending (drop it, or route it to the current bundle), and set `checksStarted` only *after* the credentials have actually been applied, so a bad update can't consume the round.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+        val password = transportPacketExtension.password
+        if (ufrag == null || password == null) {
+            logger.warn(
+                "Ignoring a transport update with no ufrag/pwd while ICE restart generation " +
+                    "${pending.generation} is pending."
+            )
+            return
+        }
+
+        if (!pending.checksStarted.compareAndSet(false, true)) {
```
JonathanLennox: Because `startConnectivityEstablishment` returns unconditionally once it has diverted to the pending bundle, this gate also means every transport update is withheld from the current bundle for the whole duration of a restart, and trickled candidates arriving after the answer are silently discarded. Harmless for the bridge (it discovers peer-reflexive addresses anyway), but worth a comment saying so deliberately -- and see the note above about setting the flag before the credentials are known to be good.

jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
```diff
@@ -257,7 +257,22 @@ class Colibri2ConferenceHandler(
         }
 
         c2endpoint.transport?.iceUdpTransport?.let { endpoint.setTransportInfo(it) }
-        if (c2endpoint.create) {
+
+        // An explicit ICE restart request. Handled after setTransportInfo, so any credentials included in this
+        // same request still belong to (and are applied to) the pre-restart Agent. The restart rotates our own
+        // ICE credentials, so the endpoint has to be told the new ones — its connectivity checks are addressed
+        // to them — even though this is not a create.
+        val iceRestarted = if (c2endpoint.transport?.iceRestart == true) {
```
JonathanLennox: **The restart is armed as a side effect, before the response is known to be usable.**

`requestIceRestart()` rotates the advertised credentials and starts the timeout here, and `describe()` then unconditionally prefers the pending bundle. If anything downstream fails -- a later `IqProcessingException` in this same function (e.g. the visitor-sources check), a lost `conference-modified`, or jicofo's own latest-wins guard dropping the transport -- the peer never learns the new credentials, but the bridge still spends the full `restart.timeout` advertising them and diverting every transport update to a bundle that can never connect.

The lost-response case does degrade gracefully (the peer hasn't rotated yet, so media keeps flowing and it can retry), so this is robustness rather than a live bug. Arming the restart last, or abandoning it if the response isn't produced, would close it.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+
+        logger.info(
+            "ICE restart requested (generation=$generation): created a pending Agent with local ufrag=" +
+                "${newBundle.agent.localUfrag} (current local ufrag=${currentBundle.agent.localUfrag}). " +
+                "Waiting for the peer's new remote credentials before starting connectivity checks."
+        )
+        iceRestartsStarted.inc()
+
+        val timeout = IceConfig.config.restartTimeout
+        TaskPools.SCHEDULED_POOL.schedule(
```
JonathanLennox: **This timeout covers the entire signalling round trip.** bridge -> jicofo -> client -> jicofo -> bridge, plus the client's re-gathering, plus ICE itself -- on a client whose network just changed and whose own XMPP connection may be reconnecting. The 656 ms measured in a clean test doesn't say much about a mobile handover.

Consider either raising it substantially or splitting it in two: a generous window for the peer's credentials to arrive, and a separate shorter one for ICE once checks have actually started.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+        synchronized(restartLock) {
+            // A newer restart supersedes one that has not connected yet.
+            pendingBundle?.let { superseded ->
+                logger.info("Superseding an in-flight ICE restart: $superseded")
+                pendingBundle = null
+                TaskPools.IO_POOL.submit { superseded.free() }
+            }
+
+            generation = ++restartGeneration
+            newBundle = AgentBundle(generation)
```
JonathanLennox: **Agent construction here is unguarded, and it's inside the lock.**

`agent.createComponent(...)` declares `throws IOException, BindException`. Kotlin won't make you handle it, so a harvest/bind failure propagates out of `requestIceRestart()` -> `Endpoint.requestIceRestart()` -> `handleColibri2Endpoint`, where none of the `catch` clauses match -- failing the whole conference-modify and taking every other endpoint's updates with it, which is exactly what the soft-rejection design was chosen to avoid.

It also leaks: if the `component` initializer throws, the already-constructed `agent` and `stream` are never freed, so the ufrag stays in `SinglePortUdpHarvester.candidates` for the life of the process.

Suggest wrapping it, freeing the partial bundle, and returning false. Building the bundle outside `restartLock` and only swapping it in under the lock would also keep candidate harvesting off the lock.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+            )
+            iceRestartsRejected.inc()
+            return false
+        }
+
+        val newBundle: AgentBundle
+        val generation: Int
+        synchronized(restartLock) {
+            // A newer restart supersedes one that has not connected yet.
+            pendingBundle?.let { superseded ->
```
JonathanLennox, line 496: **Restart requests aren't idempotent.** A repeated request for a still-pending, not-yet-started restart rolls a new generation and frees the bundle the peer may already be checking against, restarting the clock and discarding progress. Re-describing the existing pending bundle would be strictly better here -- clients do fire duplicate network-change events.

(The supersede logic itself is fine, and freeing immediately rather than after a transition window is right, since the superseded bundle never carried media.)

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +300,249 @@ class IceTransport @JvmOverloads constructor(
...
+        if (transitionWindow.isZero || transitionWindow.isNegative) {
+            TaskPools.IO_POOL.submit { oldBundle.free() }
+        } else {
+            TaskPools.SCHEDULED_POOL.schedule(
+                {
+                    logger.info(
+                        "ICE restart (generation=${newBundle.generation}) transition window elapsed, freeing " +
+                            "the old Agent with local ufrag ${oldBundle.agent.localUfrag}."
+                    )
+                    oldBundle.free()
```
JonathanLennox: **`Agent.free()` shouldn't run inline on this pool.** `TaskPools.SCHEDULED_POOL` is `Executors.newSingleThreadScheduledExecutor` and is shared bridge-wide, while `Agent.free()` shuts down the StunStack, closes sockets and joins threads. The other two free sites deliberately hop to `IO_POOL` -- `abandonPendingRestart` even carries the comment explaining why. Suggest `TaskPools.IO_POOL.submit { oldBundle.free() }` here too.

JonathanLennox (commented):
**Note: this review was generated by AI** (Claude Code), from reading this PR against the pinned ice4j `3.2-15-g6da2b08` sources and the jicofo / lib-jitsi-meet sides of the change. It has not been validated by running the code, so please treat each point as something to check rather than as established fact.

The overall design holds up well. The generation counter really does handle overlapping restarts and reordered colibri2 responses; the `pendingBundle !== bundle` identity guards make a double `free()` impossible on every path I traced (timeout vs. FAILED vs. supersede vs. cutover); and removing the listeners before `agent.free()` correctly suppresses the spurious `RUNNING -> TERMINATED` "failure". The comments below are about things around that core.

The two I'd consider blocking are the inline notes on `cutOver`'s demultiplexing rationale (ice4j routes by remote address, not by ufrag, which makes the restart quietly dependent on the peer's 5-tuple changing) and on `applyRemoteCredentialsToPendingRestart`, where one untagged `<transport>` can poison a restart round irrecoverably.

Two further points that don't attach to a specific line:

**No prompt recovery from an abandoned restart.** An established ice4j Agent never transitions to `FAILED` -- `ConnectivityCheckClient.updateCheckListAndTimerStates` bails out on `getState().isEstablished()` -- so consent-freshness failure only stops refreshing `lastIceActivityInstant`. After `abandonPendingRestart` keeps the old Agent, `iceTransport.hasFailed()` stays false and `Endpoint.shouldExpire()` falls through to the inactivity path: `entity-expiration.timeout` is 1 minute with a 1-minute `check-interval`, i.e. up to ~2 minutes of a live endpoint carrying no media. Real recovery therefore rests entirely on the client noticing its own ICE failure and falling back to `session-terminate restart="true"`. Worth stating explicitly in the PR description, and it's an argument for surfacing the abandonment to jicofo rather than only bumping a counter.

**No tests.** `jvb/src/test` has nothing covering `IceTransport`, and this adds a real state machine (supersede, abandon-vs-cutover, generation mismatch, late credentials). `clock` is already injected; the ice4j `Agent` is the obstacle. It would be worth covering at least the pure transition logic if it can be prised apart from Agent construction.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -280,9 +602,14 @@ class IceTransport @JvmOverloads constructor(
     fun stop() {
         if (running.compareAndSet(true, false)) {
             logger.info("Stopping")
-            iceAgent.removeStateChangeListener(iceStateChangeListener)
-            iceStream.removePairStateChangeListener(iceStreamPairChangedListener)
-            iceAgent.free()
+            val bundles = synchronized(restartLock) {
+                buildList {
+                    add(currentBundle)
```
JonathanLennox, line 782: **`stop()` misses a bundle inside its transition window.** It collects only `currentBundle` and `pendingBundle`; a retiring bundle is neither -- it's reachable only from the scheduled closure -- so it isn't freed until that task fires. Self-healing at the default 2 s, but it means agents can outlive the transport, and it scales with a larger configured `transition-window`. `stop()` also doesn't cancel either scheduled task. Holding a reference to the retiring bundle (and to the futures) would let `stop()` handle both.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -453,6 +830,41 @@ class IceTransport @JvmOverloads constructor(
...
+            "ice_restarts_started",
+            "Number of ICE restarts started."
+        )
+
+        /**
+         * The total number of ICE restart requests rejected before a new Agent was created: the feature is
+         * disabled, the transport was not established (or not running), or the peer's credentials arrived
+         * tagged with a generation that does not match the pending restart.
+         */
+        val iceRestartsRejected = VideobridgeMetricsContainer.instance.registerCounter(
```
JonathanLennox, line 1032: **Metric accounting.** Two small things:

* A superseded restart increments `started` but never `succeeded` or `failed`, so the three never reconcile. A `_superseded` counter, or folding it into `failed`, would make the series readable.
* This counter does double duty: restart *requests* rejected up front, and stale-generation credential drops in `applyRemoteCredentialsToPendingRestart`. Those are very different events, and the latter isn't a rejected request at all -- it's the normal outcome of a superseded round.

jvb/src/main/kotlin/org/jitsi/videobridge/ice/IceConfig.kt
```diff
@@ -70,6 +71,32 @@ class IceConfig private constructor() {
...
+     */
+    val restartTransitionWindow: Duration by config(
+        "videobridge.ice.restart.transition-window".from(JitsiConfig.newConfig)
+    )
+
+    /**
+     * How long to wait for the new [org.ice4j.ice.Agent] of an ICE restart to connect before giving up on the
+     * restart and keeping the existing Agent.
+     */
+    val restartTimeout: Duration by config(
```
JonathanLennox, line 98: No validation on this one. Zero or negative fires the abandon task immediately -- after `describe()` has already handed the pending credentials to the peer. `transition-window` handles zero/negative explicitly in `cutOver`; this doesn't.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -294,17 +621,38 @@ class IceTransport @JvmOverloads constructor(
...
+        // checking against an Agent we are about to retire. With no restart in flight this is the established
+        // bundle, so the initial allocation path is unchanged.
+        val bundle = pendingBundle ?: currentBundle
         with(pe) {
-            password = iceAgent.localPassword
-            ufrag = iceAgent.localUfrag
-            iceComponent.localCandidates?.forEach { cand ->
+            password = bundle.agent.localPassword
+            ufrag = bundle.agent.localUfrag
+            if (bundle.generation != IceUdpTransportPacketExtension.GENERATION_UNSPECIFIED) {
```
JonathanLennox: Latent trap rather than a bug today: after a successful cutover `currentBundle.generation` is >= 1, so any later `describe()` stamps a generation the client's strictly-newer check would reject. Safe right now because the only caller is the create/restart path in `Colibri2ConferenceHandler`, but it will bite if `describeTransport()` ever gains another caller. Might be worth only stamping the generation when the bundle is actually pending.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bgrozev: Leaving this as it is for now. The role is fixed for the lifetime of the transport: every `AgentBundle` reads the same field, and nothing in colibri2 signals a per-restart role today, so a settable role would be unreachable. It stays easy to change if we ever need it -- make the field a `var` and set it before the bundle is created.

jvb/src/main/resources/reference.conf
bgrozev: Not derived from anything specific, so raised in bb664854f: `timeout` 10s -> 20s, `transition-window` 2s -> 3s. The timeout covers the whole signaling round trip plus the endpoint's re-gathering plus ICE, which the comment now says. Kept as one timeout rather than splitting it into a credentials-wait and an ICE-connect window; that can follow if 20s turns out to be the wrong shape.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
bgrozev: Same as the reply on `reference.conf`: raised to 20s in bb664854f, and the comment there now spells out what the timeout covers.

bgrozev:
(Also partially AI generated)
Thanks -- all of the review points are addressed, one commit each (`0afc79e5..ed6aa44f`). The two you flagged as blocking:

- **Demultiplexing.** Confirmed against the pinned ice4j: `AbstractUdpListener` looks the remote address up in its socket map first and parses the ufrag only for an address it has not seen, and `ConnectivityCheckServer` drops on a local-ufrag mismatch. The comments claimed the opposite, so they are corrected in `ba946889`, along with the assumption this leaves on the restart path: the new Agent is reachable only because the endpoint's address changes. Making it work for an unchanged address needs the socket mapping to be handed between Agents, which is an ice4j change, so it is not in this PR. An endpoint that keeps its address gets the ordinary failed-restart outcome: the restart is abandoned and the established Agent is kept.
- **An untagged `<transport>` poisoning a restart round.** Real and reachable: lib-jitsi-meet stamps `ice-generation` on the restart answer only, so an untagged update is an ordinary one. Fixed in `0afc79e5`: untagged updates go to the established bundle, and the round is claimed together with the credentials rather than before them.

On the two points that did not attach to a line:

- **No prompt recovery from an abandoned restart.** Correct. Surfacing the abandonment to jicofo needs a bridge-initiated colibri2 message, which we do not have at all today, so that is future work. Recovery stays with the endpoint for now.
- **No tests.** Added in `ed6aa44f`. The ice4j Agent is created through a factory now, so the tests drive the state machine with mock Agents: request outcomes, a repeated request, supersede, generation matching, untagged updates, cutover, abandonment, `stop()`, and a failure to create the Agent. The four fixes that changed behaviour silently were each checked by reverting the fix and watching the test fail.

One behaviour change worth calling out, from the comment about rejections triggering a reconnect: a restart requested before ICE is established no longer answers with no `<transport>`. It answers with the established transport unchanged, so the endpoint keeps the connection it has instead of escalating to a re-invite. A bridge that cannot restart at all (disabled, stopped, or a non-positive `restart.timeout`) still answers with no transport and still escalates.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +386,329 @@ class IceTransport @JvmOverloads constructor(
...
+        synchronized(restartLock) {
+            // Re-check under the lock: the restart may have been superseded or abandoned since we read it.
+            if (pendingBundle !== pending) {
+                logger.info(
+                    "ICE restart generation ${pending.generation} is no longer pending, dropping the " +
+                        "remote credentials that arrived for it."
+                )
+                return
+            }
+            pending.stream.remoteUfrag = ufrag
```
JonathanLennox: **A duplicate update is applied, then reported as ignored.**

```kotlin
pending.stream.remoteUfrag = ufrag
pending.stream.remotePassword = password
if (!pending.checksStarted.compareAndSet(false, true)) { ...ignoring a repeated transport update... ; return }
```

A second tagged update for the same generation overwrites the remote credentials of an Agent whose checks are already running -- ice4j reads `stream.getRemotePassword()` live in `Agent.getRemoteKey`, so this changes the credentials used for both outgoing checks and validating incoming ones -- while the log line says it was ignored.

Everything that can make an update unusable is already checked before this point (null ufrag/pwd above, wrong generation above, no longer pending inside the lock), so the concern in the comment above is satisfied with the CAS moved *before* the two assignments. That also makes the behaviour match the log message.

jvb/src/main/kotlin/org/jitsi/videobridge/colibri2/Colibri2ConferenceHandler.kt
```diff
@@ -309,6 +297,50 @@ class Colibri2ConferenceHandler(
...
+        // failed with an error, because an error would fail the whole conference-modify and take every other
+        // endpoint's updates with it. Jicofo pairs a request it sent with the answer it gets back: a
+        // <transport> means the restart happened and is relayed to the endpoint, no <transport> means it did
+        // not and jicofo falls back to a re-invite.
+        //
+        // So a transport is signaled back for a restart that started (the new Agent's rotated credentials) and
+        // for one that kept the existing Agent (its unchanged credentials, so the endpoint keeps the connection
+        // it has, with no re-invite). Only IceRestartResult.UNAVAILABLE signals nothing.
+        if (c2endpoint.create || iceRestartResult == IceRestartResult.STARTED ||
+            iceRestartResult == IceRestartResult.KEEP_EXISTING
```
JonathanLennox, line 329: **`KEEP_EXISTING` is signalled as a transport the client discards -- worth confirming the far end agrees.**

The transport described here for `KEEP_EXISTING` carries no `ice-generation` (correct -- `describe()` only stamps a pending bundle). Following it through jicofo's `ColibriV2SessionManager.endpointIceRestarted`:

* It passes the staleness guard, then hits `participantInfo.lastRelayedIceGeneration = generation` unconditionally, which **resets the high-water mark to `GENERATION_UNSPECIFIED`** and loses staleness protection for the next reordered response.
* jicofo counts it as a relayed restart, so `IceRestartMetrics` will over-count in-place restarts that never happened.
* It is relayed to the client, whose generation guard then drops it -- which is what actually implements "keep what you have". The lib-jitsi-meet tests cover a non-numeric generation and `0`; worth confirming an *absent* attribute takes the same path.

The reasoning behind `KEEP_EXISTING` itself looks right for the not-established case: the client has not rotated anything, so the initial Agent can still pick up its new address peer-reflexively. It's just that "keep existing" is currently expressed as "send something the peer ignores" rather than signalled explicitly.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -206,23 +386,329 @@ class IceTransport @JvmOverloads constructor(
...
+            }
+
+            // Create the new Agent before touching any state, so that a failure leaves the restart in flight
+            // (if there is one) alone rather than freeing a bundle the peer may already be checking against.
+            generation = restartGeneration + 1
+            newBundle = createAgentBundle(generation) ?: run {
+                // The established Agent is untouched and still carrying media, so keep using it.
+                logger.warn("Not restarting ICE: failed to create the new Agent. Keeping the existing one.")
+                iceRestartsRejected.inc()
+                return IceRestartResult.KEEP_EXISTING
```
JonathanLennox: This is the other caller of `KEEP_EXISTING`, and it seems less clear-cut than the not-established one. Failing to create an Agent is a real resource problem (a bind failure, exhausted ports), and answering with the established transport deliberately bypasses jicofo's new "re-invite the participant if the bridge declines" fallback -- the endpoint's only remaining recovery is its own ICE failure detection. `UNAVAILABLE` would escalate immediately, which for a resource failure may be what you want. Genuinely a judgement call, but worth being deliberate about, since the two paths returning the same value have quite different causes.

JonathanLennox (commented):
**Note: this review was generated by AI** (Claude Code), like the previous one. I re-read all 15 commits and ran the new test class in a worktree: 22 tests, all passing, `BUILD SUCCESS`. As before, please treat each point as something to check rather than as established fact.

Every point from the earlier review looks addressed, and several of the fixes go further than what was raised:

* **Demux** -- corrected in `cutOver`, `requestIceRestart`, `IceConfig` and `reference.conf`, and `requestIceRestart`'s KDoc now states the consequence outright (a peer whose address does not change cannot connect the new Agent, and the restart is abandoned). Documenting rather than fixing seems like the right call here.
* **Untagged updates** -- both halves fixed: only generation-tagged updates route to the pending bundle, and the `checksStarted` claim moved inside the lock.
* **Arming order** -- the restart block moved past everything in `handleColibri2Endpoint` that can throw, with a comment saying why.
* **Timeouts, `free()` on the scheduled pool, `stop()` missing the retiring bundle, Agent-creation failure, idempotency, metrics, timeout validation, the `describe()` generation stamp** -- all fixed. The new task bookkeeping (`restartTimeoutTask` / `transitionWindowTask`, each with an "unless it's already over" re-check after scheduling) holds up under every interleaving I could trace, and making `free()` idempotent covers the new `stop()`-vs-transition-task overlap that `retiringBundle` introduces.

The colibri WebSocket password rework goes beyond what was raised -- decoupling it from ICE entirely rather than just pinning it, and dropping the expected secret from the failure log, which was a small real leak.

A few things left, inline below. The only one I'd call a defect is the credential/CAS ordering in `applyRemoteCredentialsToPendingRestart`; the rest are questions and test coverage.

One further note that doesn't attach to a line: **`createAgentBundle` still harvests candidates while holding `restartLock`**, which `stop()` and `cutOver` contend for. The comment explains the choice (a failure then leaves an in-flight restart alone), and it looks like a fair trade -- just flagging that binding a port now happens under a lock on the cutover path.

jvb/src/main/kotlin/org/jitsi/videobridge/transport/ice/IceTransport.kt
```diff
@@ -453,6 +1010,51 @@ class IceTransport @JvmOverloads constructor(
...
+        val iceRestartsStarted = VideobridgeMetricsContainer.instance.registerCounter(
+            "ice_restarts_started",
+            "Number of ICE restarts started."
+        )
+
+        /**
+         * The total number of ICE restart requests for which no new Agent was created: the feature is
+         * disabled, the transport is not running, or it is not established yet.
+         */
+        val iceRestartsRejected = VideobridgeMetricsContainer.instance.registerCounter(
```
JonathanLennox, line 1032: Now that `KEEP_EXISTING` exists, this counter merges two quite different events: the routine "not established yet" case (normal traffic, the endpoint keeps a working session) and a failure to create the Agent (a resource problem you would want to alert on). Splitting them, or at least noting the difference in the help text, would make the metric actionable.

jvb/src/test/kotlin/org/jitsi/videobridge/transport/ice/IceTransportRestartTest.kt
```diff
@@ -0,0 +1,329 @@
...
+                // The generation marks a transport to restart against. The established one is not that.
+                transport.requestIceRestart()
+                agents.created[1].fireState(IceProcessingState.COMPLETED)
+
+                with(transport.describe()) {
+                    ufrag shouldBe agents.created[1].ufrag
+                    iceGeneration shouldBe IceUdpTransportPacketExtension.GENERATION_UNSPECIFIED
+                }
+            }
+            should("free the old Agent only after the transition window") {
```
JonathanLennox, line 194: This asserts only the "not yet" half -- that the old Agent survives the cutover. Neither timer path is exercised anywhere in the class: the transition window elapsing (which frees the bundle, clears `retiringBundle` and hops to the IO pool) and the restart timeout firing (`abandonPendingRestart` via the scheduled task, as opposed to the `FAILED` route that *is* covered) both go untested.

Both are reachable: `TaskPools.SCHEDULED_POOL` is a public mutable field and `VideobridgeTest` already substitutes a fake executor for exactly this. Given that this is where the new task-cancellation bookkeeping lives, it seems like the part most worth pinning down.

jvb/src/test/kotlin/org/jitsi/videobridge/transport/ice/IceTransportRestartTest.kt
```diff
@@ -0,0 +1,329 @@
...
+        FakeAgent(created.size).also { created.add(it) }.agent
+    }
+}
+
+private class FakeAgent(index: Int) {
+    val ufrag = "ufrag-$index"
+    val password = "password-$index"
+
+    var state: IceProcessingState = IceProcessingState.WAITING
+    var freed = false
```
JonathanLennox, line 370: `freed` is written from the IO pool (`free()` is submitted there) and spun on from the test thread in `awaitTrue`, with no synchronisation and no happens-before. It will almost always be observed promptly on x86, but it can in principle spin out the full 5 s and fail spuriously in CI. `@Volatile` here (and on `created`'s usage, if an Agent is ever created off the test thread) would make it sound.

jvb/src/test/kotlin/org/jitsi/videobridge/transport/ice/IceTransportRestartTest.kt
```diff
@@ -0,0 +1,329 @@
...
+                transport.requestIceRestart() shouldBe IceRestartResult.STARTED
+                transport.requestIceRestart() shouldBe IceRestartResult.STARTED
+
+                agents.created.size shouldBe 2
+                agents.created[1].freed shouldBe false
+                transport.describe().iceGeneration shouldBe 1
+            }
+            should("supersede a restart whose checks have started") {
+                transport.requestIceRestart()
+                transport.startConnectivityEstablishment(remoteTransport(generation = 1))
```
JonathanLennox, line 120: Nothing in the class covers the ordinary, non-restart path through `startConnectivityEstablishment` -- every call is inside `With ICE established`, where the current bundle short-circuits on "Connection already established". Since the routing at the top of that method was restructured for this PR, a leaf under `Before ICE is established` asserting that an untagged update still reaches the initial Agent (sets its remote credentials, starts its checks) would guard the main negotiation path against a future change to the pending-bundle condition.

jvb/src/test/kotlin/org/jitsi/videobridge/transport/ice/IceTransportRestartTest.kt
```diff
@@ -0,0 +1,329 @@
...
+        context("When the new Agent fails") {
+            should("keep the established one") {
+                transport.requestIceRestart()
+                agents.created[1].fireState(IceProcessingState.FAILED)
+
+                transport.hasFailed() shouldBe false
+                transport.describe().ufrag shouldBe initial.ufrag
+                awaitTrue { agents.created[1].freed }
+            }
+            should("not restart the transport's own failure handling") {
```
JonathanLennox: Reads like it is missing a word -- "not trigger the transport's own failure handling"?

JonathanLennox (approved):

PR #2440, opened 2026-08-19
feat: expose unmatched transport-cc feedback in statistics

savely-krasovsky:
We are investigating egress traffic storms on our videobridge deployment: bridges serving 15–18 participant conferences would suddenly send 2–4 Gbit/s for minutes at a time, with outgoing loss of 60%+ and no corresponding increase on ingress, then recover on their own.

The mechanism we reproduced locally: once the egress path saturates, a growing share of transport-cc feedback reports fails to match the send-time history and is silently discarded (the "Received feedback before packet" path in TransportFeedbackAdapter). The estimator then loses both its loss and delay signals at exactly the moment they matter, so the estimate freezes above the real path capacity and the saturation sustains itself.

The unmatched-report counter added here turned out to be the single most indicative signal of this state -- near zero in healthy operation, in the hundreds of thousands during a storm window -- so we use it for detection
and alerting. We also have an experimental pacer (placed upstream of transport-cc tagging, following libwebrtc's RtpSenderEgress ordering) that eliminates the problem in our reproduction. We plan to propose it separately once it has soaked in production. The metric is useful on it own regardless of that outcome.

### DIFF
--- a/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/rtp/TransportCcEngine.kt
+++ b/jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/rtp/TransportCcEngine.kt
@@ -69,6 +69,12 @@ abstract class TransportCcEngine : RtcpListener {
 
     abstract class StatisticsSnapshot {
         abstract fun toJson(): ObjectNode
+
+        /** Cumulative count of transport-cc feedback reports that could not be matched against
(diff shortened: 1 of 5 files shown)


savely-krasovsky:
cc @bgrozev @JonathanLennox

savely-krasovsky:
I signed it previously.

bgrozev:
jenkins add to whitelist

bgrozev (approved):

PR #2452, opened 2026-09-15
Add metrics to measure the keyframe budget's effect on receivers

bgrozev:
## Problem

The keyframe budget's counters (in #2448) show that it is binding and roughly how much it costs, but not whether that is good or bad for the receivers waiting. Two gaps in particular:

1. Every counter is a single aggregate per source, so a wait caused by a bridge-generated request that unblocks a stalled layer switch (e.g. a BWE-driven downswitch) looks the same as one caused by a redundant PLI echoed from a receiver's own decoder during a storm. Those two cases have very different consequences: delaying the former keeps the bridge oversending a stale layer to a receiver that is trying to shed bandwidth.
2. `total_budget_wait_ms` only times the first receiver dropped for a source until the next request is sent, so it undercounts when several receivers are waiting concurrently (documented as a known limit).

## Changes

- Split the existing keyframe-budget counters in `KeyframeRequester` by request origin: bridge-generated (`requestKeyframe()`, e.g. a layer switch or a dominant-speaker switch) vs. forwarded from a receiver's own PLI/FIR (`transform()`). New `_api` counters sit alongside the existing aggregates.
- Fixed the wait-time accounting to key pending waits by `(source, requester)` instead of just `source`, so one sent request correctly resolves every receiver that was waiting on it, not only the first.
- Added a max wait time per source, alongside the existing total.
- Added timing in `AdaptiveSourceProjection` for how long a projection is stuck forwarding a stale layer while `needsKeyframe()` is true. This is independent of what is actually governing the request's timing (per-receiver limit, source-wide limit, or the budget), so it measures the effect on the receiver directly rather than inferring it from request-limiter state, and it is available identically whether the budget is enabled or not — a real before/after comparison, unlike the budget-specific counters, which read as zero with the budget off.
- Registered Prometheus counters/gauges for both sets of stats in `VideobridgeMetrics`, so they can be watched in production before deciding whether to enable the budget by default:
  - `keyframe_requests_dropped_by_budget[_api]`, `keyframe_budget_waits[_api]`, `keyframe_budget_wait_milliseconds_total[_api]` — aggregated from `KeyframeRequester` at endpoint/relay expiration, matching how `keyframes_received` etc. already work.
  - `layer_switch_keyframe_waits_completed`, `layer_switch_keyframe_wait_milliseconds_total`, `layer_switch_keyframe_waits_in_progress` — updated live from `AdaptiveSourceProjection`, since that class already lives in the `jvb` module.

## Known limits

- On a relaying bridge, only the relay's own transceiver is aggregated into the `KeyframeRequester`-derived metrics; each `RelayEndpointSender`'s own `KeyframeRequester` (for PLIs forwarded from local receivers) is not, matching the existing scope of that expiration method for other stats.
- The `_api` origin split is coarse: it does not yet distinguish a layer switch needing a keyframe to go up (a quality request) from one needing it to go down (a bandwidth-relieving one) — both currently count as bridge-generated. Narrowing that further would need the switch direction threaded from `BitrateController` through the `keyframeNeeded` callback chain, which seemed like more plumbing than was justified before seeing what the coarser split shows in practice.

## Testing

- Extended `KeyframeRequesterTest` to cover the origin split (a forwarded PLI and a bridge-generated request both dropped by the budget for the same source, sent counters attributed correctly) and the multi-receiver wait fix (both waiting requesters counted, and attributed by their own origin, when the request that resolves them finally sends).
- Full `jitsi-media-transform` and `jvb` module test suites pass; ktlint and checkstyle clean.


jvb/src/main/java/org/jitsi/videobridge/cc/AdaptiveSourceProjection.java
```diff
@@ -178,6 +198,36 @@ public boolean accept(@NotNull PacketInfo packetInfo)
...
+     * {@code jmt.keyframe.budget}) the keyframe budget lengthening it.
+     */
+    private void updateNeedsKeyframeWaitStats(boolean needsKeyframeNow)
+    {
+        if (needsKeyframeNow)
+        {
+            if (needsKeyframeSinceMs < 0)
+            {
+                needsKeyframeSinceMs = System.currentTimeMillis();
+                VideobridgeMetrics.layerSwitchKeyframeWaitsInProgress.inc();
```
JonathanLennox: **Gauge only drifts upward.** This is incremented when a wait starts and decremented only when a later `accept()` sees the wait end. `PacketHandler` never removes a projection from `adaptiveSourceProjectionMap`, and `AdaptiveSourceProjection` has no teardown hook, so any projection still waiting when its endpoint expires, or whose source simply stops sending, leaves a permanent increment. On a busy bridge `layer_switch_keyframe_waits_in_progress` will climb monotonically within hours and stop meaning anything.

Either decrement from an expire path (e.g. have `BitrateController`/`PacketHandler` tell each projection it is going away, and settle an open wait there), or drop the gauge and keep the two counters, which don't have this problem.

jvb/src/main/java/org/jitsi/videobridge/cc/AdaptiveSourceProjection.java
```diff
@@ -169,7 +187,9 @@ public boolean accept(@NotNull PacketInfo packetInfo)
         // sufficient to only check for needing a key frame if the packet wasn't
         // accepted. But this wouldn't be enough, as we may be accepting packets
         // of low-quality, while we wish to switch to high-quality.
-        if (contextCopy.needsKeyframe()
+        boolean needsKeyframeNow = contextCopy.needsKeyframe();
+        updateNeedsKeyframeWaitStats(needsKeyframeNow);
```
JonathanLennox, line 195: **The timer runs while the projection is suspended.** The request on the next line is gated on `targetIndexCopy > RtpLayerDesc.SUSPENDED_INDEX`, but this call isn't. `GenericAdaptiveSourceProjectionContext.accept()` sets `needsKeyframe = true` on every packet while the target is suspended, so for H.264 and other unparsed codecs every bandwidth suspension gets timed as a "layer switch wait" even though the bridge is requesting nothing. The VP8/VP9/AV1 filters don't set the flag on suspension, but they don't clear it either, so a projection that was already waiting when it got suspended keeps accruing for the whole suspension.

Suggest gating the timing on the same condition as the request, and treating a transition to suspended as ending (or discarding) the current wait. A test with a suspended target on the generic context would pin this.

jvb/src/main/java/org/jitsi/videobridge/cc/AdaptiveSourceProjection.java
```diff
@@ -178,6 +198,36 @@ public boolean accept(@NotNull PacketInfo packetInfo)
         return accept;
     }
 
+    /**
+     * Times how long the context needs a keyframe to make progress on a layer switch for, which is how long the
```
JonathanLennox: **"Layer switch" overclaims a bit.** `Vp9QualityFilter` sets `needsKeyframe` on the first frame of a fresh projection (the current encoding starts suspended, so `externalTargetEncoding != currentEncoding` holds), and the generic context starts with the flag true. So every join and every resume-from-suspended is counted here as a layer switch wait, alongside the actual switches. The description notes the up/down gap but not this one.

Either name these `keyframe_wait_*` and say in the help text that they include initial start, or exclude a wait that begins with the projection's first packet.

jitsi-media-transform/src/main/kotlin/org/jitsi/nlj/rtcp/KeyframeRequester.kt
```diff
@@ -341,9 +383,20 @@ class KeyframeRequester @JvmOverloads constructor(
...
+        addNumber("num_requests_sent_budget_lengthened_api", numRequestsSentBudgetLengthenedApi)
+        addNumber("total_budget_extension_ms_api", totalBudgetExtensionMsApi)
+        addNumber("num_requests_dropped_by_budget_api", numRequestsDroppedByBudgetApi)
+        addNumber("num_budget_waits_api", numBudgetWaitsApi)
+        addNumber("total_budget_wait_ms_api", totalBudgetWaitMsApi)
+        addNumber("max_budget_wait_ms_api", maxBudgetWaitMsApi)
         lastSourceWideLimits.forEach { (ssrc, limit) ->
             addJson("source_wide_limit_$ssrc", limit.toJson())
         }
+        budgetWaitStarts.forEach { (ssrc, waiters) ->
```
JonathanLennox: **Unsynchronized iteration over a map mutated under the lock.** `budgetWaitStarts` is a plain `mutableMapOf` and every write to it is inside `synchronized(keyframeLimiterSyncRoot)`, but this `forEach` is not, so a debug dump that coincides with a request can throw `ConcurrentModificationException` and take down the whole debug response. (`lastSourceWideLimits` two lines up is a `ConcurrentHashMap`, so that loop is fine.)

Snapshot the sizes under the lock, or make this a `ConcurrentHashMap` too.

jvb/src/main/kotlin/org/jitsi/videobridge/metrics/VideobridgeMetrics.kt
```diff
@@ -204,6 +204,70 @@ object VideobridgeMetrics {
...
+        "keyframe_budget_wait_milliseconds_total",
+        "Total milliseconds spent waiting for a keyframe request the budget had delayed, summed across every " +
+            "receiver waiting (updated on endpoint/relay expiration). Divide by keyframe_budget_waits for the mean."
+    )
+
+    @JvmField
+    val keyframeBudgetWaitMillisecondsTotalApi = metricsContainer.registerCounter(
+        // Not "..._total_api": registerCounter appends "_total" only when the name doesn't already end with it,
+        // so "..._total_api" would come out as "..._total_api_total" instead of ending in the unit suffix.
+        "keyframe_budget_wait_milliseconds_api",
```
JonathanLennox: Nit: this comes out as `keyframe_budget_wait_milliseconds_api_total`. The Prometheus convention is unit last, so `keyframe_budget_wait_api_milliseconds` (giving `..._api_milliseconds_total`) would read more consistently next to `keyframe_budget_wait_milliseconds_total`. Not important.

JonathanLennox (commented):
Thanks, the `KeyframeRequester` half of this is good and I'd take it as-is: the `(source, requester)` keying fixes a real undercount I had listed as a known limit, the API/forwarded split is attributed at the one place both paths meet, and the expiration-time aggregation mirrors `keyframes_received` in both `Endpoint` and `Relay`. I also confirmed jicoco's `registerCounter` appends `_total` unless the name already ends in it, so the second commit is right.

I built and ran both module suites on the head (ktlint, checkstyle, SpotBugs, 28 requester cases) and everything passes. The four things below are all in behavior the tests don't exercise: suspension, projection teardown, and a debug dump racing a request.

Of these, I'd want the first two fixed before merging, since they affect the metric this PR exists to add. The other two are small. Details inline.

